PR #44, opened 2019-07-22
new legendLocation argument (and more locations)

Padanian:
…sition groupbox


swharden:
Thanks for your hard work @Padanian !

PR #53, opened 2019-07-26
Commit markers in legend

Padanian:


Padanian:
Fixes for #53 and #54

swharden:
@Padanian this looks great! Thanks so much!

Note that I had to revert to the original tick code: https://github.com/swharden/ScottPlot/pull/53/commits/4f715ee96abda92961b40d0283f45335cf9c4865 because something was causing the cookbook to crash - apparently there was a condition where the infinite loop was never satisfied.

We plan to redo the tick module from scratch (https://github.com/swharden/ScottPlot/issues/47), so we probably don't need to worry about it for now.

PR #60, opened 2019-07-30
support custom line styles

Padanian:
This pull request creates the `LineStyle` enumeration and adds line style as an argument for PlotScatter, PlotVLine, and PlotHLine. Example use:

```cs
plt.PlotScatter(xs, ys, lineStyle: ScottPlot.LineStyle.Dash);
```


swharden:
Thanks @Padanian this looks great!

Padanian:
Happy you liked it. Please note that there is one unused variable in Legend.cs named stubHeight that can be safely removed unless for future features.

PR #72, opened 2019-08-07
add parallel processing to PlotSignal()

StendProg:
Added RenderHighDensityParallel method to SignalPlot class.
Improve rendering speed for large signals.
Using parallel may slow render for small signals,
also may have some multithreading problems then updating signal, need lots of tests before accepted.
For PlotSignalConst gives nothing after fullLoad, i can't load cpu more then 25% with 60M signal.
This pull request more for testing, feel free to reject it.


swharden:
This is excellent! A great performance increase. 

Eventually there will be more parts of ScottPlot which may benefit from parallel processing. I added a `useParallel` bool to the Plottable class. Rather than add "useParallel" as an argument to every plottable, I made it so you can turn parallel processing on (for all plottables) with:

```cs
plt.Parallel(true);
```

What do you think of this solution? I agree this could use more testing, but since parallel processing is off by default I feel good about merging this with the master branch.

StendProg:
Parallel.For works good  on  heavy loops with time computation say starting from seconds and gives nothing for milliseconds loops(Parallel.For overhead). Whole library fast enouth and get milliseconds for computations. One place is big signals then you get long min/max calculations, and its closed by SignalConst class. Log2(n) time insted of linear (n) time computation. It faster in most cases even if you make linear time ( n / (threadcount) ) with Parallel.For. 
I see one  good place for parallel loop  is iterarion over multiple plot then render. Just need get Graphics( its not thread safe) outside loop. This would be not easy task for all plottable classes. And i think its gives nothing in most cases because bottleneck is Grapics.

swharden:
Thank you so much @StendProg!

I added an example of this to the cookbook:
https://github.com/swharden/ScottPlot/tree/master/cookbook#signal-with-parallel-processing

PR #80, opened 2019-08-10
Add ability to update SignalConst

StendProg:
Manualy updating source signal not allowed.
To update signal SignalConst.updateData must be call.
Updates work much slower compare to Signal class.
Update  range work faster then updating by single element in loop.
Added demo button in plot types project for testing updates.


swharden:
This looks good! Thanks @StendProg! I'll continue to work on making all plot types return objects. I'll also create a demo dedicated to advanced SignalConst features.

PR #83, opened 2019-08-11
Changed default bitmap pixel format to PixelFormat.Format32bppPArgb

StendProg:
Simple solution that gives about 1.5x render time reducing.
It's hard to precisely measure with eyes.


swharden:
Thanks @StendProg for noticing this.

If I understand this correctly, the default pixel format is `PixelFormat.Canonical` but this pull request changes it to `PixelFormat.Format32bppPArgb` - a _premultiplied alpha_ format which is the same amount of memory (32 bits per pixel, 8 bits each R, G, B, A) draws and renders faster due to how the alpha channel is managed.

* [Microsoft documentation of pixel formats](https://docs.microsoft.com/en-us/dotnet/api/system.drawing.imaging.pixelformat)
* [premultiplied alpha notes](http://www.shawnhargreaves.com/blog/premultiplied-alpha.html)

swharden:
Wow! this change increased drawing speed by about 30%!

### Format32bppArgb (old)
image

### PixelFormat.Format32bppPArgb (new)
image

StendProg:
Thanks for creating Grapical benchmark. With some durty hacks i make plot render in 2560x1400 (usefull case for me). with signal 10K. Small enough not to run into min/max calculations but enough to render in HighDensityMode.

**Format32bppArgb (old)**
2KWithoutPixelFormat

**Format32bppPArgb (new)** 
2KwithPixelFormat

Thats gives 40% render time reducing for me.

I doesn't investigate much on this solution, just try and it works. It can be easy revert if some negative sides will be found.

[Some tips about GDI+ optimizations](https://www.codeproject.com/Tips/66909/Rendering-fast-with-GDI-What-to-do-and-what-not-to)

Issue #15, opened 2019-05-28
nuget package 


plumforest:
I saw ScottPlot at the nuget and wanted to be the first downloader.
Howerver, I failed since .net framework version 4.5/4.6 lower than that one you build(4.7?).

I do suggest you to make the minimum version reset to be 4.5. It should be work as I successful run the source code v3.

swharden:
Hi @plumforest, I'm happy to hear you found ScottPlot on NuGet!

I just recompiled ScottPlot using the .NET 4.5 Framework and it is available as ScottPlot 3.0.2 on NuGet:
https://www.nuget.org/packages/ScottPlot/

Let me know if this works for you!

plumforest:
@swharden , the 3.0.2 version nuget package  works very well.

PR #85, opened 2019-08-11
Add support for .NET Core

petarpetrovt:
Migrate to dotnet core .csprojs and target both .Net45 and netcoreapp3.0


swharden:
Sorry I made this pull complicated by merging a recent commit.

Do you think it would be better to support .NET Core 2.2?  I do not know much about .NET Core, but I am having some difficulty getting this to compile on my system. I look forward to hearing your suggestion!

StendProg:
Gui app available only for .NET Core 3.0, and it is still in preview. Good improvement for you library. I also do not know much about .NET Core.

petarpetrovt:
You need to install dotnet core 3.0 SDK to be able to build it. 
Download link: [click here](https://dotnet.microsoft.com/download/dotnet-core/3.0)

swharden:
I'm concerned about this switch. It looks like .NET Core 3.0 is still in preview and only supports certain preview versions of Visual Studio preview. I did a couple installs and reboots and still must not have the right version of things installed.

My concern is that making this hard to compile will limit who is able to use and contribute to the project. Maybe we can note in the readme that this code is configured as a .NET 4.5 project but is compatible with .NET Core (tested on 2.2 and 3.0).

I'll continue to search/read about publishing projects for multiple frameworks.

swharden:
In Visual Studio 2019 you have to:

Tools > Options > Environment > Preview Features > Use preview of the .NET Core SDK

image

... I'm still favoring _documenting_ that ScottPlot is comparable with .NET Core 3.0, rather than switching the project to it and requiring users to figure this out.

swharden:
It looks like oxyplot (a similar charting library) keeps separate folders, one for every platform they target: https://github.com/oxyplot/oxyplot/tree/develop/Source

If we supported multiple frameworks I think it means we would have to maintain multiple .csproj projects (and a separate NuGet packages). This seems like a good goal for the future when ScottPlot matures and settles down. 

For now ScottPlot is being actively developed and evolving rapidly so at this time I'll probably keep it .NET 4.5 (which was already toned down far from 4.8 (#15)). Thank you @petarpetrovt for bringing my attention to the fact that this project compiles in .NET Core without modification!

StendProg:
@petarpetrovt maybe you can make standalone script that compiles current .NET Framework project with .NET Core 3.0 and this script to project. .NET Core have a good command line support as i know. That will be a good for everyone solution.

swharden:
I'll give it a try! I'll do it in Python :smiling_imp:

petarpetrovt:
We could define compilation variable `USE_DOTNET_CORE` - disabled by default. And pass it when compiling to netcoreapp 3.0

StendProg:
Other solution make separate branch for this PR. But someone will need to synchronize it periodically.

StendProg:
[Good article about this problem](https://docs.microsoft.com/en-us/dotnet/core/porting/project-structure)
So Visual Studio 2017 is lowest posible version for new type of projects. We can remove target=netcoreapp3.0, to bypass netcore 3.0 installation. And easy add then it need. Cant test this, base PR fully working for me.

swharden:
This branch now has two new files (`ScottPlot-dotNetCore.csproj` and `ScottPlot-dotNetCore.sln`) but  all the original code remains unchanged. This is similar to the [Keep existing projects and create a .NET Core project strategy](https://docs.microsoft.com/en-us/dotnet/core/porting/project-structure#keep-existing-projects-and-create-a-net-core-project) linked to by @StendProg. This gives .NET Core users a solution/project they can work with, but also allows traditional users to open the project as usual (even with older versions of visual studio).

What do you think about this "solution"?

StendProg:
ScottPlot-dotNetCore.sln loaded with erros and targeted .NET Framework 4.5. 
ScottPlot-dotNetCore.csproj works fine.
Good decision, but you need to manage both of them from now ;).
I think more and more people interested in .NET Core, especially after gui applications in 3.0. So this would be good improvement to support it.

swharden:
> I think more and more people are interested in .NET Core, especially after GUI applications in 3.0. So this would be good improvement to support it.

Adding an extra .csproj for .NET Core users may work, but it feels like a hack. 

I'll modify the folder structure of this project to support multiple frameworks and use the [oxyplot library](https://github.com/oxyplot/oxyplot/tree/develop/Source) as a reference for how to do this.

StendProg:
Microsoft thinked a lot about how to make people migrate to .NET Core with no cost. I am also surprised that it turned out to be so simple. To be completely sure that it works, need .NET Core 3.0 gui demo app, that uses compiled library.

swharden:
Okay I finally decided what I want to do. 

The simplest way to support multiple frameworks this from a developer's standpoint is to replace existing NET45 projects with .NET Core ones (described [here](https://docs.microsoft.com/en-us/dotnet/core/porting/project-structure#replace-existing-projects-with-a-multi-targeted-net-core-project)). That's what @petarpetrovt's initial pull request did.

The reason that solution is not good now is because .NET Core 3.0 projects are difficult to compile at this point in time. .NET Core 3.0 is prerelease, and you have to go through menus to even enable it in Visual Studio. A difficult to compile open source project is bad because it discourages community involvement.

However, when .NET Core 3.0 is officially released (in about a month) and is supported by regular versions of Visual Studio, then I'll migrate all the .csproj files in this project to .NET Core. At that time I will make NuGet build scripts for separate ScottPlot.Core and ScottPlot.Net45 NuGet packages. For now, the ScottPlot-dotNetCore.csproj file can live in the main source folder.

StendProg:
>At that time I will make NuGet build scripts for separate ScottPlot.Core and ScottPlot.Net45 NuGet packages. For now, the ScottPlot-dotNetCore.csproj file can live in the main source folder.

I think thats enouth. You dont really need to remake all demos, just add 1-2 than demonstrates .NET Core usage. Main target of this PR as i think is to get ScottPlot.Core on nuget.
When it releases i think they make .NET Standart 3.0 library type that supports both Framework and Core projects

swharden:
Note that a .csproj like this allows .cs files to be sucked in from file paths:

```xml
<Project Sdk="Microsoft.NET.Sdk">

  <PropertyGroup>
    <TargetFramework>netcoreapp3.0</TargetFramework>
    <Version>3.1.0</Version>
    <PackageId>ScottPlot</PackageId>
    <GeneratePackageOnBuild>true</GeneratePackageOnBuild>
    <RootNamespace>ScottPlot</RootNamespace>
  </PropertyGroup>

  <ItemGroup>
    <Compile Include="..\ScottPlot\*.cs" />
    <Compile Include="..\ScottPlot\**\*.cs" />
  </ItemGroup>

</Project>

```

image

swharden:
Following-up, https://www.nuget.org/packages/ScottPlot.WinForms/ now supports .NET Core 3.0 and .NET Framework 4.6.1

PR #89, opened 2019-08-12
plot low quality (anti-aliasing off) while dragging

StendProg:
Mode enabled by set lowQualityWhileDragging=true in ScottPlotUC.
In this mode AA disabled for user interactions with plot like paning or zooming.
But then user stop interacting last render call with settings.antiAliasData.
This makes good speed improvment during user interactions, in cost of image quality.
But plot always looks perfect at static.
AA for figure never turned off, only for data.

### DIFF
--- a/demos/src/plottable-const/Form1.cs
+++ b/demos/src/plottable-const/Form1.cs
@@ -27,6 +27,7 @@ namespace plottable_const
         {
             data = ScottPlot.DataGen.RandomWalk(rand, 10_000_000);
             signal = scottPlotUC1.plt.PlotSignalConst(data);
+            scottPlotUC1.useDynamicAA = true;
             scottPlotUC1.plt.Benchmark();
             scottPlotUC1.Render();
         }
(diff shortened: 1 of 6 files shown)


swharden:
This is a good idea @StendProg!

This would be good to have in the WPF user control too (it should not difficult to add because that user control is so simple).

Want to simplify this by adding an argument to `GetBitmap` instead of overloading it? The argument name could be `bool lowQuality=false` and it temporarily makes antiAliasFigure and antiAliasData false.

The user control property that controls this behavior could be called `lowQualityWhileDragging` instead of `useDynamicAA` to make it more clear what it does.

StendProg:
I want to avoid making antiAliasFigure set to false, it gives nothing at speed but visibly reduses ticks quality.
>This would be good to have in the WPF user control too (it should not be difficult to add because that user control is so simple).

I will try

StendProg:
Done. WPF Control so laggy, and haven't context menu, double click handlers

swharden:
Thanks @StendProg this is really good!

* This is now the default behavior
* I moved the `lowQualityWhileDragging` variable to the MouseTracker class
* This can be toggled through the right-click menu

The right-click menu will get simpler and better soon (#91)

swharden:
Interestingly this implementation crashed .NET Core 3.0 applications. #138 

Adding a null check fixed it (350f5cf7b06db378b38635ae4ebcc9123cfe444e).

Issue #92, opened 2019-08-13
Legend looks bad while dragging


bukkideme:
Hello,
I just found that when you show Legend, and start to pan with keeping Mouse left button, all curves with their marker types in the Legend changes to "ScottPlot.MarkerShape.filledSquare" type from the previously set types. Stop panning and all markers reverts to original.

Edit: actually it looks like this bug happens only, when we create a curve via PlotScatter() method not explicitly specifying marker type. In this case the default marker type is "filledCircle". During mouse panning these default filledCircle markers turn into filledSquare ones.

StendProg:
It is rendered correct. Small filled circle look like filled square with anti aliasing dissabled. After PR #89 antialiasing off during pan by default. If you want to avoid this effect you can set at startup:

```c#
scottPlotUC1.plt.mouseTracker.lowQualityWhileDragging = false;
```

swharden:
You can also toggle this behavior on and off with the right-click menu (which hopefully will be improving soon, #91)

When `lowQualityWhileDragging` is enabled, perhaps we should render the frame and legend in high quality, and only render the data in low quality? I guess then that causes a problem because circles would render well in the legend put poorly on the graph itself...

bukkideme:
Thanks, clear now!

swharden:
To make this clearer in the future should we display a little message in the corner when a frame is rendered low quality? Maybe something that looks kind of like the benchmark? ... or turn this back off by default?

I like how low quality rendering while dragging makes the user interaction faster, but I'm concerned this behavior might be misinterpreted as a glitch...

StendProg:
@swharden it looks like legend render in low resulution and then upscaled with some default method.
Maybe render legend in high resolution initialy help to resolve this problem.

StendProg:
Posible solutions:
- Make new bitmap layer (like figure, and data)  for legend, with AA always on, and just copy it on end render end.
- render each legend element on separate small bitmap with AA disabled, and copy it on dataBitmap.
- disable lowQualityWhileDragging by default

swharden:
Making a standalone legend bitmap is an interesting idea and one I considered before. This would make it easy to render the legend with independent settings from the rest of the graph, but there may be another advantage:

[...]

PR #93, opened 2019-08-13
Delayed HQ render after mousewheel event

StendProg:
Additional speed improvement to PR #89.
Allow zoom with mouse wheel at high speed.
Mouse Wheel event render whith AA off, after configurable delay it render with current setting.
This PR will make legend looks bad on mouse wheel scroll too. #92

### DIFF
--- a/src/ScottPlot/MouseTracker.cs
+++ b/src/ScottPlot/MouseTracker.cs
@@ -13,8 +13,9 @@ namespace ScottPlot
     {
         public Stopwatch mouseDownStopwatch = new Stopwatch();
         private Plottable draggingObject;
-        public bool lowQualityWhileDragging = true;
-        public int mouseWheelHQRenderDelay = 500; // in milliseconds, 0 meens disabled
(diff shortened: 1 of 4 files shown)


swharden:
I like this behavior. 

What do you think about doing the same thing on the MouseUp events? Click-dragging would be low quality, and it would wait a bit after releasing the mouse before rendering in high quality?

Perhaps this would it make it more obvious to the user it's rendering in low quality (and not just appear glitchy)

StendProg:
I can make it, but not today. And i prefer keep old PR #89 behavior as a configurable option. Delayed render for mouse wheel this is a necessary action, but not for mouse up that can RenderHQ  instantly.

StendProg:
HQ render after delay on mouse up. Delay can be set with mouseUpHQRenderDelay = 300.
 With using mouseUpHQRenderDelay = 0 it works like in PR #89.

swharden:
This works great, thanks @StendProg!

Issue #86, opened 2019-08-11
Migrate to .NET Core 3.0


swharden:
This topic was initially discussed in #85. When the time comes consider asking @petarpetrovt for assistance/review of the migration and implementation. Migration has been practiced (#95) and should be straightforward.

**Wait to migrate until Visual Studio fully supports Windows Forms in .NET Core 3.0:** Many demo applications have advanced features (such mouse-draggable axis lines) which are only supported in Windows Forms. Currently, the Visual Studio deisgner does not support Windows Forms in .NET Core 3.0. ScottPlot will conservatively wait to migrate to .NET Core 3.0 until Visual Studio fully supports it.

- [x] consolidate all demos into a single application
- [ ] migrate ScottPlot project to .NET Core 3.0 (using System.Drawing.Common)
- [ ] ensure [target frameworks](https://docs.microsoft.com/en-us/dotnet/standard/frameworks) include `net45` and `netcoreapp3.0`
- [ ] add tests for .NET Framework and .NET Core (to detect issues like #138)
_currently Azure Pipelines does not support running tests in solution files containing mixed .NET Framework and .NET Core projects._
- [ ] publish ScottPlot as a multi-framework package on NuGet

swharden:
After considering whether or not to break this project up into 3 (a plotting class library, a winforms user control, and a WPF user control) I decided to keep it all one project (#87). Briefly, this makes maintaining it and distributing it easier, and I've found that WinForms and WPF can coexist peacefully.

However, I do want to rename `ScottPlotUC` and `ScottPlotWPF` to `FormsPlot` and `WpfPlot`. I know this will be a breaking change which might be a pain to resolve for existing projects (requiring getting into the Designer.cs to fix). I think this change is logical though, and I think it makes sense to do this earlier than later. The transition from .NET Framework to .NET Core next month may be the best time to do this.

Thinking forward, the UserControls folder may eventually contain GUI elements like windows forms (e.g., custom save data dialogues).

WinForms and WPF controls together | The project can stay clean and organized
---|---
![image](https://user-images.githubusercontent.com/4165489/62906718-dc687a00-bd3d-11e9-999e-64290a25bc6d.png) | ![image](https://user-images.githubusercontent.com/4165489/62906721-e68a7880-bd3d-11e9-8849-a7b59ebbed3a.png)

swharden:
* ScottPlot is now .NET Core 3.0 and builds for .NET Framework 4.5 too
[...]

PR #95, opened 2019-08-13
example migration to .NET Core 3.0 and renaming of user controls

swharden:
This is a full migration. Both .NET 4.5 Framework and .NET Core 3.0 are supported by ScottPlot. 

Demos of each platform are included. The cookbook is now .NET 3.0. 

This also renames `ScottPlotUC` and `ScottPlotWPF` to `FormsPlot` and `WpfPlot`

Details are discussed in #86


swharden:
This branch demonstrates a .NET Core migration, but it will not be used. Rather than maintaining this branch, a new migration will occur when .NET Core 3.0 is officially released.

PR #97, opened 2019-08-15
singlePrecision trees for SignalConst

StendProg:
With singlePrecision=true SignalConst allocate and use float[] Trees. That halves memory usage, without visible changes. singlePrecision = false by default.
Additional maked small improvments on SignalConst Class.
This PR makes a lot of code duplicates in SignalConst class. I don't see another way to make it.

### DIFF
--- a/src/ScottPlot/Plot.cs
+++ b/src/ScottPlot/Plot.cs
@@ -369,7 +369,8 @@ namespace ScottPlot
             Color? color = null,
             double lineWidth = 1,
             double markerSize = 5,
-            string label = null
+            string label = null,
+            bool singlePrecision = false
             )
         {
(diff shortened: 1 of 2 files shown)


swharden:
This is really nice!

Supporting doubles and floats nearly doubles the size of the SignalConst class. What do you think about just making all the trees float (and removing support for doubles entirely)? This would probably always be okay since these numbers in the trees are just used to determine pixel positions (which are very low precision relative to the original data).

Looking forward to hearing your thoughts

StendProg:
There is a better and more universal solution using generics in PR #100. Thats allow using trees of source type. Also allow user plot int[] arrays (and many othes) without conversion to double[].

PR #104, opened 2019-08-18
Legend rendered at separate layer

StendProg:
Fix for Issue #92.  Legend rendered at small bitmap with separate AA settings, and draw to figure layer after data drawed.


swharden:
This is looking great! Thanks so much @StendProg!

image

PR #114, opened 2019-08-23
Render with SkiaSharp (and OpengGL)

StendProg:
I partially implemented a `FormsPlotSkia` user control which renders data using [Skia graphic library](https://en.wikipedia.org/wiki/Skia_Graphics_Engine). 
* This is superior to GDI+ for large images (full screen) and images with many lines
* For small plots Skia and GDI+ perform similarly
* Only `PlotSignalConst()` is implemented for now
* Projects that use this control must specify x64 or x86 (untested) mode. ("Any CPU" mode is not supported)
* `FormsPlotSkia` was implemented in a separate project
  * New dependencies are SkiaSharp and OpenGL.NET
  * Main ScottPlot project still has no dependencies
* To enable the Angle backend for Skia you must manually add `libegl.dll` and `libglesv2.dll` to the output folder
  * The simplest way to obtain these files is from a Chrome installation.
  * If you do not do this it uses OpenGL by default

This PR is for tests and to demonstrate the concept. It will be a lot of work to fully support the Skia library, including changing core ScottPlot drawing calls to simplify Skia library injection.


swharden:
This is an impressive improvement @StendProg! GDI+ is slow full-screen (especially with a 4K monitor), and using Skia is an excellent way to improve this limitation.

I created a new branch of ScottPlot for Skia:
https://github.com/swharden/ScottPlot/tree/Skia
Do you want to make a pull request there instead of master?

While Skia has many strengths, I think the master branch of ScottPlot will always be dependency-free. If the Skia branch gets developed fully we can make separate "ScottPlot.Skia" NuGet package for it.

StendProg:
I don't clearly understand the reason of separate branch. In this pr ScotPlot Project stay untouched by dependencies ( just maked some methods protected to allow override). All new dependencies only in other ScotPlotSkia project. You already can make ScottPlot nuget without dependencies. And ScottPlot.Skia nuget from skia project. May be i am wrong, not familar with nuget.
Or you want have copy ScottPlot project on skia branch that make directly  calls skia draws instead gdi+?
I thinked about backend solution. Where main ScottPlot lib call drawings to abstract graphic class not directly to gdi+. ScottPlot main lib implement that class using gdi+.So no dependencies.
Other Skia project in solution implement this abstract class using skia. And use existing code of ScottPlot main lib.

swharden:
> new dependencies are only in the ScotPlotSkia project

After initially inspecting this pull request I overlooked this important point. You created a very good design! I will delete the Skia branch and work toward merging this now.

swharden:
On my system running the Skia demo is faster than GDI+ in fullscreen mode, but _slower_ than GDI+ when using typical window sizes. I find this surprising and interesting.

After thinking about this topic for a long time I decided to move the Skia project to `/dev` where it has its own solution file and its own demos, completely separate from ScottPlot.

***Two of the primary design goals for ScottPlot are that it (1) has no dependencies and (2) is radically simple and easy to use by anybody.***

* ScottPlotSkia requires dependencies, and even though _ScottPlot_ doesn't have dependencies it does imply that the Skia project is a distinctly different from ScottPlot. I don't think Skia code belongs in the ScottPlot solution.

* I have never used Skia before. It seems very good, but it also adds complexity. For example, copying DLLs from a Google Chrome installation. Further, I still can't get my Visual Studio to stop crashing every time I drag/drop the FormsPlotSkia user control onto my designer window (screenshot below). There may be a simple fix to this, but the fact that I have not been able to determine the fix so far means that projects which include ScottPlotSkia are not easy to use. I really like the concept of ScottPlotSkia, but I think it should be cleanly segregated into its own folder and solution.

I think continuing to develop this Skia project has great potential, but I don't think it's consistent enough with the core design goals of ScottPlot to include it the primary `ScottPlotV3.sln` file. 
  * I moved your code to `/dev` and created its own demos and solution file there
  * I am happy to help you further develop your Skia project by modifying ScottPlot (such as making functions `protected` and separating `Action` from `Impl` methods wherever possible). If you make these pull requests, I will act quickly to merge them!

What do you think about this style of organization?

Thank you again for all of your valuable contributions to this project!

image

StendProg:
> After thinking about this topic for a long time I decided to move the Skia project to /dev

Very good solution. I like it.

> For example, copying DLLs from a Google Chrome installation

Native solution is to complile you own from Angle repo. I was disappointed too. There some nuget packages  like `Avalonia.Angle.Natives`, but i get platform not specified error during instalation.

>  I still can't get my Visual Studio to stop crashing every time I drag/drop the FormsPlotSkia user control onto my designer window (screenshot below). 

VS editor try to create and render component in Any CPU mode.  Setting x64 for FormPlotSkia and Demo project fix this crashes. SkiaSharp dont't work with Any CPU. Its only wrapper on Skia native lib (x86 and x64).

> I am happy to help you further develop your Skia project by modifying ScottPlot (such as making functions protected and separating Action from Impl methods wherever possible). If you make these pull requests, I will act quickly to merge them!

This some quick and durty hacks to make it work. Static methods on delegates is worst of that.
No need to make any of this changes if whole library will call draws on absract class like this.
```c#
//settings.gfxData.DrawLine(pen, xPx, 0, xPx, settings.dataSize.Height);
settings.gfxBackend.DrawLine(xPx, 0, xPx, settings.dataSize.Height, lineColor, linewidth, etc. );
...

//settings.gfxFigure.Clear(settings.figureBackgroundColor);
settings.gfxBackend.Clear(settings.figureBackgroundColor);
...

public class GDIBackend:IGraphicBackend
{
    DrawLine(float x, float y, float x1, float y1,Color lineColor,float  lineWidth, etc. )
    {
        var pen = new Pen(lineColor, lineWidth)
        settings.gfxData.DrawLine(settings.gfxData.DrawLine(pen, xPx, 0, xPx, settings.dataSize.Height);
    }

   Clear(Color color)
    {
        settings.gfxData.Clear(color);
    }

// ... implementation of all posible graphic calls from ScottPlot lib
// ...
}
```

And backend can be init in Plot() like this:
```c#
Plot plt = new Plot(new GDIBackend());

Plot(IGraphicBackend backend)
{
//.............
   settings.gfxBackend = backend;   
}
public class Settings
{
//...
    IGraphicBackend gfxBackend;   
}
```

So implementing other graphic libs would be just implementing IGraphicBackend.
There a lots of work to replace all GDI calls in abstract calls that provide enough information for different libs.

StendProg:
Fix designer crashes. For working with designer `Any CPU` mode must be set and rebuild solution. For debuging `x64` mode must be set.

StendProg:
Skia backend changed from offscreen rendering with opengl.net to onscreen rendering to OpenTK.GlControl. 
It works realy fast for now 1-10 ms to render plot at any resolution.
Negative side of  onscreen rendering, it's not posible render legend and benchmark without remaking them to onscreen.

swharden:
Hi @StendProg, I'm still studying this topic. Thank you again for your insightful contributions! I would like to make ScottPlot easier to use for different rendering engines: GDK, Skia, and ImageSharp.

What is the primary reason you are using Skia? Is it because GDK is poorly supported on Linux or mobile? ... or is it because Skia is faster for rendering large images when combined with hardware acceleration? I'm interested in learning more about your use case.

StendProg:
Hi @swharden.
My primary reason is performance. There is only 10 fps on fullscreen plot using GDI+. I'm interested in windows platform only. At fist looked for some DirectX(Direct2D) library, but to my surprise did not find anything worthwhile. SharpDX is most powerfull lib for DirectX, but it's still too complex.
Skia looked very simple to use. It is developed by Google for Chrome. SkiaSharp is developed by Mono (Microsoft for now). So it looked very serious.
After some work on it, i think maybe simple DirectX(Direct2D) wrapper would be enouth. There is no so complex drawing technics used in ScottPlot. SharpDX closed at this year, it is very disappointing.
Anyway i will try to finish work on skia support. It goes slower bacause haven't much free time for now, and spending a lot of time on invistigations (zero expirience on opengl/skia).

swharden:
Thanks for the details! This helps me understand why you chose Skia.

swharden:
Hi @StendProg this pull has been extremely useful! Thanks again for working on it.

After more research, I concluded the complexity I disliked came from OpenGL (not SkiaSharp). It's OpenGL that requires your hardware and system to support it, requires you to compile for the right target, and when OpenGL fails it crashes Visual Studio with non-helpful error messages.

However, SkiaSharp is great! I have used it enough now to really like it. It is similar in speed to GDI+ (when not using OpenGL), but it has no dependencies which is consistent with the goals of this project. It also has great extra features like numerous image encoding options, and its objects (like colors and rectangles) are much more logical to use.

I'd like to design a new plotting library ([QuickPlot](https://github.com/swharden/QuickPlot)) to use Skia everywhere (no `System.Drawing` at all). It will render on a mouse-interactive user control. A _separate_ project (QuickPlotGL) can inherit the QuickPlot user control and replace the drawing area with an OpenGL control (like you do with this pull). This way we keep OpenGL and its complexity out of the primary package, but also fully support hardware-accelerated rendering while developing a plotting library for Skia.

What do you think?

StendProg:
Hi Scott.
I also liked Skia library, simple interface and powerfull implementation. OpenGl just one of the posible backend, it can be setup by few lines of code. It's possible to make fallback to software render if no OpenGL hardware available. And all this backend things good separated from main canvas abstraction, so you can choose design that you like, separate controls in separate projects.

One negative thing about SkiaSharp, it's just a wrapper on native skia library. PInvoke make performance costs. Multiple drawing calls directly to skiasharp kill all performance. This problem can be solved by replacing multiple calls with constracting SKPath and single call to draw it(like i maded in last commits). This solution add some complexity to drawing...

Designer crashes these are mostly problems of designer, many people blame it.
I will plan to make some investigation to fix it  after SkiaBackend will be finished.
OpenTK.GlControl have some interesting solution. They create dummy control if Designer mode detected, and it works stable in designer.

>I'd like to design a new plotting library (QuickPlot) to use Skia everywhere (no System.Drawing at all). It will render on a mouse-interactive user control. A separate project (QuickPlotGL) can inherit the QuickPlot user control and replace the drawing area with an OpenGL control (like you do with this pull). This way we keep OpenGL and its complexity out of the primary package, but also fully support hardware-accelerated rendering while developing a plotting library for Skia.

Sorry haven't time to look carefull at this for now(very busy at main work), but very interested, maybe later. As i think it's this PR way. Make abstract call like DrawRectangles(color, widths, heigths) to backend and resolve it in backend with specific call to lib. It's very flexible with low performance cost.
If you chose skia as a primiry lib, you can make things in different ways. For example you can once draw plot, and multiple zoom and pan without redrawing, just change transformation matrix (don't really tryed, but think it's possible).

swharden:
For what it's worth, I'm finding that System.Drawing (System.Drawing.Common NuGet package) is faster than I expected, especially on .NET Core 3 using WPF...

There's a compiled demo here if you're interested in running it yourself
https://github.com/swharden/QuickPlot/tree/master/demos/compiled

StendProg:
On my machine i get about 50 FPS with you demo. But it not use heavy draw calls. Tried to make some tests and get  build errors swharden/QuickPlot#2.
WPF have DirectX acceleration for UI, but i am not sure you can get it from inmemory rendering to bitmap. Onscreen rendering in wpf control can achive high performance, and it can be done with help of this PR. But ScottPlot WPF user control initialy lagging for no reason.

swharden:
## Observations of WPF performance

> WPF has DirectX acceleration for UI ... onscreen rendering in WPF control can achieve high performance

I've tried many different ways to draw lines on the screen using WPF. Full-screen performance is great for 1-50 lines. At 1000 lines it becomes much ***slower*** than GDI+, even at large resolutions. A simple-case example is: [2019-10-09-WPF-Draw-Lines](https://github.com/swharden/Csharp-Data-Visualization/tree/master/examples/2019-10-09-WPF-Draw-Lines). I know there are slightly more efficient ways to draw lines (such as LineSegment) but it doesn't seem to make much difference. I have noticed that most of the slowness comes from anti-aliasing of long lines.

I think this is because WPF is optimized to place complex objects (images, shapes, etc) and move them around on the screen. It does not seem optimized to display a large number of small lines, especially as they are rapidly created and destroyed.

## Thanks for your work on this PR!

System.Drawing.Common works everywhere, but is slow. SkiaSharp is almost perfect! You are doing a wonderful job with this PR, and although I am exploring alternative rendering methods I highly value the work and time you have invested into this project. 

Is your NuGet package publicly available? I'd like to try it on my system. Ironically, all of my troubles with SkiaSharp have been because it crashes Visual Studio on my computer when I try to interact with user controls that require it. Once compiled, the executable runs great and demonstrate fantastic performance!

swharden:
Microsoft has an interactive charting user control for WPF
https://github.com/microsoft/InteractiveDataDisplay.WPF

Performance is pretty poor in my opinion. I tested it by running the LineGraphSample and resizing it by dragging the corner. It's ~5 Hz update rate. If that's the best Microsoft can do with WPF, maybe I shouldn't feel bad...

StendProg:
>Is your NuGet package publicly available? I'd like to try it on my system. Ironically, all of my troubles with SkiaSharp have been because it crashes Visual Studio on my computer when I try to interact with user controls that require it. Once compiled, the executable runs great and demonstrate fantastic performance!

Started work on fixing crashes, but surprisely spend a hour on tryies to crash designer. As a result designer crashes only on first build of clean project (just downloaded sources), because nuget packages not restored. After restoring works stable. But i totaly did nothing to make it stable.
Can you check last version of this PR and try to crash it? if you get success write steps to reproduce.

There are 2 suggestions why this fixed itself:
- You replaced Plot.Render with stub in designer mode #131.
- I update VS to 16.3.1, maybe they fixed something.

swharden:
> > all of my troubles with SkiaSharp have been because it crashes Visual Studio on my computer when I try to interact with user controls that require it

> Can you check last version of this PR and try to crash it? if you get success write steps to reproduce.

* I can compile the project and the output works great.
* If I open Form1.cs in Visual Studio, I get a designer crash
* If I create a new blank Form2.cs and try to drag/drop the FormsPlotSkia user control onto it, I get another designer crash

This problem is identical to the problem I had with QuickPlot. I have tested this on 2 different computers now with the same result. Note that the program _compiles_ fine! It's just visual studio that crashes.

image

image

StendProg:
what am I doing wrong?
image

swharden:
What! I updated Visual Studio (to 16.3.3) like you suggested and it works now. My Visual Studio was less than a month old! (16.3.1)

This probably means I should move QuickPlot back to Skia 🤕

swharden:
> This probably means I should move QuickPlot back to Skia 🤕

I rewrote QuickPlot to use Skia. It seems to work now, and I even got SkiaSharp to install and run properly on Azure Pipelines when the tests are run. I placed QuickPlot packages on NuGet.

Performance looks great! I am glad you encouraged me to continue working on this.

If you want to check it out, make a new Windows Forms application, use NuGet to install `QuickPlot.WinForms`, add the user control to a form, and place the following code to the startup sequence:

```cs
int pointCount = 1000;
double[] xs = QuickPlot.Generate.Consecutative(pointCount);
double[] ys = QuickPlot.Generate.Random(pointCount);
interactivePlot1.figure.plot.Scatter(xs, ys);
```

It should work even without specifying a build target! 🤞

swharden:
I closed this PR because I don't think it will ever get merged.

There's a lot of valuable code here though! Perhaps this code would be better housed in a `Skia` branch?

StendProg:
Sad but true, its too complex. Its already in branch of my fork, so i don't think it need another.

Issue #136, opened 2019-09-30
WpfPlot user control does not render consistently with different display DPI setting


morningkyle:
Tested with master branch and built-in "quickstart wpf" project.

Expected output:
     <img width="434" src="https://user-images.githubusercontent.com/31173219/65860684-595abc00-e39d-11e9-9749-39ce3e45e7de.png">

Tested output with DPI = 144:
<img width="434" alt="dpi144" src="https://user-images.githubusercontent.com/31173219/65860812-99ba3a00-e39d-11e9-93cf-6dbd9e6f991b.png">

And if display DPI is set to 96, we could get the expected rendering result.

morningkyle:
I'v made a [PR](https://github.com/swharden/ScottPlot/pull/137) to fix this issue.

swharden:
Hi @morningkyle, thank you so much for identifying this issue and suggesting a fix! DPI scaling can be quite a frustration!

The solution you came up with is this:
https://github.com/swharden/ScottPlot/blob/48a623195e102a206994eaf9b104f6611fb7b14a/src/ScottPlot/UserControls/WpfPlot.xaml.cs#L56-L62

However the use of [VisualTreeHelper.GetDpi()](https://docs.microsoft.com/en-us/dotnet/api/system.windows.media.visualtreehelper.getdpi?view=netframework-4.8#System_Windows_Media_VisualTreeHelper_GetDpi_System_Windows_Media_Visual_) requires .NET Framework 4.6.2. Including this call requires ScottPlot's requirements to jump from .NET Framework 4.5 (released in 2012) to 4.6.2 (released in 2016). While this is a trivial difference for new projects, my concern is it may limit this library's use in legacy applications (#15). An ideal solution would be backward-compatible. 

Would a function similar to this (supported by .NET Framework 4.5) be able to detect DPI well enough to come up with a multiplier? I suspect `dpi/expectedDpi` is the same thing as your `sc.DpiScaleX`.

https://github.com/swharden/ScottPlot/blob/908961240726165f1df8764582ec9abca90e07c1/src/ScottPlot/UserControls/FormsPlot.cs#L41-L42

Looking forward to hearing your thoughts!

morningkyle:
@swharden  Thanks for pointing out the Graphics.DpiX/DpiY property.    It works as expected after a quick investigation.   I'll update the PR next if it is not done by someone else.  Yes, no need to update .net framework if no new api is needed.

swharden:
[...]

PR #137, opened 2019-09-30
Fixed issue #136. Make WpfPlot aware of DPI scaling

morningkyle:
(1) Size in Wpf is in device-independent units, while canvas is in pixels;
(2) Upgrade .net framework dependency from 4.5 to 4.6.2 to support [VisualTreeHelper.GetDpi()](https://docs.microsoft.com/en-us/dotnet/api/system.windows.media.visualtreehelper.getdpi?view=netframework-4.8#System_Windows_Media_VisualTreeHelper_GetDpi_System_Windows_Media_Visual_).

### DIFF
--- a/cookbook/src/App.config
+++ b/cookbook/src/App.config
@@ -1,9 +1,9 @@
 <?xml version="1.0" encoding="utf-8"?>
 <configuration>
     <startup> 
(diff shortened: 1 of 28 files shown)


swharden:
Thanks @morningkyle, this looks great! I made a few small improvements to the WPF user control while I was in there. It still lags pretty far behind FormsPlot as far as features go, but support for DPI scaling is a great plus!

PR #148, opened 2019-10-04
ScottPlotTests: add more tests

morningkyle:
Caught an unexpected exception while saving figure without rendering first.   Saving an empty bitmap might be more appropriate.

### DIFF
--- a/tests/ScottPlotTests/UnitTest1.cs
+++ b/tests/ScottPlotTests/UnitTest1.cs
@@ -31,7 +31,7 @@ namespace ScottPlotTests
         {
             {
                 var plt = new ScottPlot.Plot(1, 1);
-                string file = Path.GetTempPath() + "test.bmp";
+                string file = Path.GetRandomFileName() + ".bmp";
                 plt.SaveFig(file);
                 Bitmap bmp = new Bitmap(file);
                 Assert.AreEqual(bmp.Width, 1);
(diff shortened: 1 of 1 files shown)


morningkyle:
>> Assert.Fail failed. Expected no exception, but got: A generic error occurred in GDI+.
"All check have failed" because I added an expected-to-fail test, which might need to be fixed.  All old tests are still passed as before.

swharden:
Interestingly the same code runs fine in a console application. It only fails in the test. 

It seems the crash is only caused within tests when `SaveFig()` is called twice in a row on the same filename. Is this because the tests occur in parallel, so the same file cannot be written to by two tests at the same time?

In any case, a fix (for the test) is to use a unique filename for every output file path.

morningkyle:
yes, confirmed.  Make the saved file name unique would pass the test.  So exception is likely due to the test framework itself.

morningkyle:
@swharden   I tried to avoid using a different/random file name each time in the revised tests, to make things more predictable.

After some debugging, I found the crash was not caused by the unit test framework.  It was due to failed to open the target file in the 2nd plt.SaveFig() call.  The exception could be simply triggered in a console project with following code.  It's surprising that the _bmp_ object kept _file_ open even it is out of scope.

```
            string file = Path.GetTempPath() + "test.bmp";
            {
                var plt = new ScottPlot.Plot(1, 1);
                plt.SaveFig(file);
                Bitmap bmp = new Bitmap(file);  // bmp keeps file open
            }
            // throws an exception here saying the file is in use and can not be open again
            FileStream f = File.Open(Path.GetTempPath() + "test.bmp", FileMode.Open);
```

morningkyle:
I improved the test code a little bit again, to make sure the Bitmap release/dispose the underlying resource(fig file in this case) before running next test.

morningkyle:
Anything else I need to do with the added tests?

swharden:
This looks good! I'm glad we figured out the "multiple saves in a test" issue

PR #192, opened 2019-12-20
option to color Signal plots by density

StendProg:
Currently `PlottableSignal` in `HighDensityMode` draw each column with single color from min to max.
This PR allow extract more information from points containing in column.
Distribution of points in column display with color map.
For example for color map of 10 colors:
Lowest 10% points draw with first color from color map.
Next 10% points draw with next color from color map. 
etc.
Highest 10% points draw with last color from color map.

To enable this mode user need set color map param:
```c#
PlotSignal(ys, colorMap: new Color[]{Color.Green, Color.Red, Color.Green});
``` 
By default `colorMap = null` and Signal render work in `HighDensityMode` as before.

Performance very week, works fast for 100_000, and acceptable for 1M.
It use naive implementation which sort each column points on each render.


swharden:
Hi @StendProg, thanks for this PR - it looks really interesting! I will try to get this reviewed and merged this weekend

swharden:
I added an example to the cookbook: https://github.com/swharden/ScottPlot/tree/master/cookbook#color-by-density

image

StendProg:
Grayscaled looks sad.
To highligh benefits of this mode, its better to use more powerfull noise with stdDev 5 times bigger then signal amplitude like in demo.
For cookbook example may look better:
```c#
double[] noise = ScottPlot.DataGen.RandomNormal(rand, pointCount, 0, 5);
```
It's hard to see any trend without color by density mode.

I prototyped some preprocessing algoritms to make this mode works fast like SignalConst do for Min/Max calculations. But it take a long time to implement this.

citizen3942:
Perhaps use Viridis color scale for this application.

swharden:
Thanks for the suggestions @StendProg and @citizen3942, this new example looks better: 

```cs
// plot the noisy signal using the traditional method
plt.PlotSignal(data, yOffset: -40, color: Color.Red);

// use a color array for displaying data from low to high density
Color[] colors = new Color[]
{
	ColorTranslator.FromHtml("#73D055"),
	ColorTranslator.FromHtml("#1F968B"),
	ColorTranslator.FromHtml("#39568C"),
	ColorTranslator.FromHtml("#440154")
};
plt.PlotSignal(data, colorByDensity: colors);
```

image

**However, I found a bug:** when using the colormap, the `yOffset` argument is ignored. This prevents me from being able to see both graphs when I use:

```cs
plt.PlotSignal(data, color: Color.Red);
plt.PlotSignal(data, yOffset: -40, colorByDensity: colors);
```

PR #212, opened 2020-01-20
Add support for Linux and macOS

Hexxonite:
Add fix for window-less operation, Add GlobalFont class to return a valid Default on multiple systems.

_EDIT: further discussion in #211_

PR #216, opened 2020-01-21
Alternative font select implementation

StendProg:
Initially in PR #212 `GetDefaultFontName()` return installed font with name containing preferred words (not strictly). 
In final implementation, return installed font with name exactly as preferred.
This PR brings initial algoritm back, (with priority order).

Also this PR adds durty `public` `GetDefaultFontName(IEnumerable<string> installedFonts)`. This method needed to substitute  `static` `FontFamily.Families`. That allow  to write some unit tests. Can be collapsed back to single `GetDefaultFontName()`, but with loosing tests.

I actualy not see big benefits of this PR (anyway `GetDefaultFontName()` return existing `SystemFonts.DefaultFont.Name` in worst cases ). All the time i thinked "does it really need". So feel free to reject this PR.

If this PR will be rejected, small improvement can be applied to existing method:
~~`string[] installedFonts = FontFamily.Families.Select(font => font.Name.ToUpper()).ToArray();`~~
```c#
var installedFonts = FontFamily.Families.Select(font => font.Name.ToUpper());
```
No need ToArray() cast (this produce full enumeration). Foreach loop can be applied to `IEnumerable` and call `Select` lambda only for elements before return statement.

### DIFF
--- a/src/ScottPlot/Config/Fonts.cs
+++ b/src/ScottPlot/Config/Fonts.cs
@@ -28,8 +28,13 @@ namespace ScottPlot.Config
         /// </summary>
         public static string GetDefaultFontName()
         {
+            return GetDefaultFontName(FontFamily.Families.Select(font => font.Name.ToUpper()).ToArray());
+        }
+
+        public static string GetDefaultFontName(string[] installedFonts)
+        {
             string[] preferredFonts = { "Segoe UI", "DejaVu Sans" };
-            string[] installedFonts = FontFamily.Families.Select(font => font.Name.ToUpper()).ToArray();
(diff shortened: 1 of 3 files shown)


swharden:
> This PR brings initial algorithm back, (with priority order).

Thanks @StendProg, I really like this!

> I actually don't see big benefits of this PR

I do! Improved readability and improved testability. It will be very easy to make changes to this module in the future, and the tests will confirm future changes never break anything.

> No need ToArray() cast (this produce full enumeration). Foreach loop can be applied to `IEnumerable` and call `Select` lambda only for elements before return statement.

This is excellent. I'm still learning to use LINQ, and PRs like this are very helpful. Thank you again!

Issue #217, opened 2020-01-22
MatchAxis with vertical:false not scaling


ckovamees:
Thanks for a great project, I'm just starting to use it.

I'm using `MultiPlot` (in lieu of QuickPlot's `secondY: true`) and want to have similar X-axis (dates), but let the Y-axis represent different data, with different scale. 

I'm use 'Cookbook.Figure_73_Multiplot()' to reproduce and illustrate here.
If I add `vertical:false`, like this 
`plt.subplots[2].MatchAxis(plt.subplots[3], vertical:false);`
Then the X-axis adapts to the source plot's X-axis, but the Y-axis is left empty.
![keep_multiplot](https://user-images.githubusercontent.com/9557072/72942150-c0332500-3d40-11ea-9994-04dcbf2e7057.png)

StendProg:
Hi @ckovamees.
It looks like you want to call `MatchAxis()` for  `X` axis only, and get `AutoSize()` for `Y` axis only.(Don't clearly understand).
You can add to your code:
```c#
plt.subplots[2].MatchAxis(plt.subplots[3], vertical:false);
plt.subplots[2].AxisAutoY();
```
And some investigation shows what it actualy call `Settings.AxisAuto()` with correct `yExpandOnly = true`

Problem is that using this `xExpandOnly` or `yExpandOnly` params not implemented in `Settings.AxisAuto()`.
https://github.com/swharden/ScottPlot/blob/9a43b0cf77c09fa764a1414c3041e34da14acefd/src/ScottPlot/Settings.cs#L136
So we have correct API with no implementation at the end.
To fix this, `Axes.Expand()` method must be expanded to get additional `xExpandOnly` and `yExpandOnly` params. And then called from `Settings.AxisAuto()` with this `ExpandOnly` params.
https://github.com/swharden/ScottPlot/blob/9a43b0cf77c09fa764a1414c3041e34da14acefd/src/ScottPlot/Config/Axes.cs#L56
https://github.com/swharden/ScottPlot/blob/9a43b0cf77c09fa764a1414c3041e34da14acefd/src/ScottPlot/Settings.cs#L151

ckovamees:
Thanks for your clarification @StendProg. I'm still trying to catch up on the code and don't completely get everything you mention.

So the call `MatchAxis(otherplot, vertical:false)` isn't supposed to make the vertical axis auto size? I didn't understand that and that I had to call `AutoY()` explicitly. 

[...]

PR #218, opened 2020-01-22
Fixed issue with MatchAxis using vertical: false.

ckovamees:
Fixing issue #217. This was due to `Axes.hasBeenSet` uses OR instead of AND to compare, the correct line should be 
`return ((x.hasBeenSet) && (y.hasBeenSet));`

keep_issue_multiplot


ckovamees:
I'm afraid this doesn't solve more complex issues. Will have to noodle on this...

PR #226, opened 2020-01-31
AxSpan Fix

StendProg:
`PlottableAxSpan` produce strange effect, if deeply zoomed in on one of of `AxSpan` border . During paning, border start "dancing" around correct coordinate.

In `Render()` of axis span it try to draw huge rectangle, say from (-1000000, 0) to (574, 800). And this huge rectangle clipped to bitmap size by GDI, which possible produce this effect.

This PR manualy clip huge coordinates outside of `DataBitmap` and fix AxSpan border "dancing".


swharden:
Thanks for reporting and fixing this @StendProg!

PR #231, opened 2020-02-03
DraggableRefactor

StendProg:
Fix lagging while dragging.

Refactored:
1. Convert `IDraggable` `DragEnable()` and `GetDragCursor()` methods to properties. Looks more clear for me.
2.  Add `StartDragMethod` to `IDraggable`. This method called then drag started, and allow `AxSpan` calculate and store dragged edge. Calculation of dragged edge of span in `IsUnderMouse()` check method didn't look obvious.
3. Split `AxSpan` to 2 simplier classes `VSpan` and `HSpan`. Removes lots of if (vertical) checks. It is possible to extract code duplicates to base `AxSpanBase` class, but they already simple.
4. The same split for `AxLine`. `PlottableAxLine` not deleteded because it referenced by https://github.com/swharden/ScottPlot/blob/334f93dec027cd8d950bc63561f0925612d7e3c1/src/ScottPlot/Settings.cs#L237-L238
it can be safely deleted with this method.


swharden:
The return types of `Plot.PlotVLine()` (and others) have now changed. **This is a breaking change** for ScottPlot. This is okay, but it might mean it is a good time to switch from version 4.0 to 4.1

Since this is a breaking change, it may be a good time to add other breaking changes. The primary one I'd like to do is remove "plottable" from class names, and move all plottable objects into the Plottable namespace.

```cs
var sig = new ScottPlot.PlottableSignalConst();

if (sig is ScottPlot.PlottableSignalConst) { /*  */ }
```

would become:

```cs
var sig = new ScottPlot.Plottable.SignalConst();

if (sig is ScottPlot.Plottable.SignalConst) { /*  */ }
```

🤔

swharden:
I created a `dev/4.1` branch we can use while we implement more breaking changes.

I'm going to close this PR since it's for master, but I'll merge your refactor into the dev branch (#233).

Thanks again for all your hard work! This software is getting a lot better because of it.

swharden:
#233 took on a life of its own... this is an excellent PR @StendProg, thanks for working on it!

PR #234, opened 2020-02-04
ScottPlot 4.1.0

swharden:
When merged this PR will upgrade the master branch from ScottPlot version 4.0 to 4.1

This new version introduces minor but user-facing breaking changes.


StendProg:
Hi @swharden.
You done huge amount of work.
I checked this, but can't do it carefully, it realy a lot.
A few comments if you allow:
- I strongly disagree with `IsUnderMouse()` changing internal state, it's not obvious. Everyone reading name like `IsUnderMouse()` think that it make some checks, return bool and nothing more. If this method additionaly will be used in other scenarios, it can break existing code. Additionaly this `IsUnderMouse()` called many many times, on each mouse move and doing unnessesary job(calclulate and store selected edge, it doesn’t matter but ...).  Yes it realy simple, and easy to control it, but it looks like bad design.
- `Plottable` to `IPlottable` is a good solution that has been asking for a long time
- `axSpan` and `axLine` can be converted to `IAxSpan` and `IAxLine`. They actualy don't have much in common code. Leave as is, also not bad.
- Does it really need bind using inheritance `axSpan` and `axLine`? They are already quite simple.
Anyway i see that inheritance from other side `axSpan`:`axLine`. Inherited classes should extend the functionality of base.
- typo in last commit b7112a4fef78ff3ead04fa7645596db1ce4ddf18
```c#
return String.Format("HSpan (X1={0:0.000}, X2={0:0.000})", position1, position2);
```
X2={~0~ 1 :0.000})
with string interpolation it's not possible to make this mistake:
```c#
return $"HSpan (X1={position1:0.000}, X2={position2:0.000})";
```

Do not pay much attention to all these comments, you have your own vision of the project and it really works.

Funny fact:
 With this abstract `axSpan` and `axLine` in initial PR. There is no api breaking changes.
```c#
PlottableAxSpan PlotVSpan(...)
{
    return new PlottableVSpan(...)
}
```

swharden:
@StendProg A lot of this refactoring is experimental still so it's definitely a lot to review, but I greatly appreciate your feedback! I'll refactor this some more tonight and see how it goes...

> It is not obvious `IsUnderMouse()` changes internal state

You're right! I didn't think about it that way before. I'll add methods like `DragStart()` and `DragEnd()` back in to the interface to make it more obvious.

> create interfaces `IAxSpan` and `IAxLine`

I like this idea!

> `axSpan` vs `axLine` inheritance is confusing

I didn't initially intend to have one inherit the other, but then later I observed they were almost identical it seemed like one should inherit the other. The reason I thought axis line should inherit from axis span is because I thought an axis line was a more advanced type axis span: it's an axis span with extra functionality ensure position1 and position2 are always the same.

In either case, inheriting one from the other is still confusing. I'll try making them both interfaces and this issue will probably resolve itself.

> in initial PR. There is no api breaking changes.

Ha! I now realize I could have merged your initial PR all along 🙈 

Thank you for giving me a good excuse to experiment with breaking changes in version 4.1 😈 

I'm still working toward the goal of making it easy to switch between System.Drawing and Skia, and then in the future to whatever drawing system Microsoft invents next.

swharden:
This PR has a lot of excellent discussion! It's too hard to maintain in parallel with all the active work on the main branch. I'll keep these notes/records but work incrementally on the master branch. 

Thanks again @StendProg for all your excellent advice!

PR #243, opened 2020-02-10
Creates a PlottableFunction, for plotting expressions

Benny121221:
This PR lets you pass in a delegate of type `Func<double, double>`, and plot that function.

I used the following code to test it out:

```
using System;
using System.Diagnostics;

namespace ScottPlotTesting
{
	class Program
	{
		{
			string filename = "test.png";
			plt.PlotFunction((x) => 10 * Math.Sin(x), -10, 10, -10, 50);
			plt.PlotFunction((x) => Math.Pow(x, 3), -10, 10, -10, 50);
			plt.SaveFig(filename);

				proc.Start();
			}
		}
	}
}
```

This plots `y=10sinx` and `y=x^3` on the domain [-10,10] and the range [-10,50]

### DIFF
--- a/src/ScottPlot/Plot.cs
+++ b/src/ScottPlot/Plot.cs
@@ -14,1282 +14,1282 @@ using System.Drawing.Imaging;
 
 namespace ScottPlot
 {
-	public class Plot
-	{
-		public PixelFormat pixelFormat = PixelFormat.Format32bppPArgb;
-		private readonly Settings settings;
-
-		public Plot(int width = 800, int height = 600)
-		{
-			if (width <= 0 || height <= 0)
-				throw new ArgumentException("width and height must each be greater than 0");
-			settings = new Settings();
-			Resize(width, height);
-			TightenLayout();
-		}
-
-		public override string ToString()
-		{
-			return string.Format($"ScottPlot ({0:n0} x {1:n0}) with {2:n0} objects ({3:n0} points)",
-				settings.figureSize.Width, settings.figureSize.Height,
-				GetPlottables().Count, GetTotalPoints());
-		}
-
-		#region Bitmaps, Graphics, and Drawing Settings
-
-		public void Resize(int? width = null, int? height = null)
-		{
-			if (width == null)
-				width = settings.figureSize.Width;
-			if (height == null)
-				height = settings.figureSize.Height;
-
-			if (width < 1)
-				width = 1;
-			if (height < 1)
-				height = 1;
-
-			settings.Resize((int)width, (int)height);
-			InitializeBitmaps();
-		}
-
-		private void InitializeLegend(Size size)
-		{
-			settings.bmpLegend = new Bitmap(size.Width, size.Height, pixelFormat);
-			settings.gfxLegend = Graphics.FromImage(settings.bmpLegend);
-		}
-
-		private void InitializeBitmaps()
-		{
-			settings.bmpFigure = null;
-			settings.gfxFigure = null;
-			settings.bmpData = null;
-			settings.gfxData = null;
-			settings.gfxLegend = null;
-			settings.bmpLegend = null;
-
-			if (settings.figureSize.Width > 0 && settings.figureSize.Height > 0)
-			{
-				settings.bmpFigure = new Bitmap(settings.figureSize.Width, settings.figureSize.Height, pixelFormat);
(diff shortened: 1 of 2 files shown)


Benny121221:
If a function is only defined on say, the interval [a,b], the delegate can return null for x ∉ [a,b]

Benny121221:
Worth noting for the cookbook, that certain built-in functions like `Math.Log()` and `Math.Sqrt()` don't throw a domain exception when given input outside of their domain. Instead, they go nutty and cause a stackoverflow. This means the user has to do some domain checking of their own with something like this: 

```
double? y;
if (x < 0)
{
	//Either return null, or throw an exception
	//Returning null is faster, but throwing an exception may be easier
}
else {
	y = Math.Log(x);
}
return y;
```

swharden:
Hi @Benny121221, thanks for your work on this! This is a really interesting topic I hadn't considered before. The ability to plot arbitrary formulas is of obvious usefulness to a plotting library like this one!

I'm always extremely hesitant to add of new plottables since each one is a lot to maintain. Further, the ability to plot functions is nice but this PR limits it to a single plottable (scatter plot). Not only would Signal and SignalConst benefit from this, but how would the user plot all 3 of these using a formula for the X values instead of just the Y values?

This may seem like a disappointing alternative considering all the work you put in here, but what do you think about code like the block below? It lets you use functions, `DataGen.Consecutive()` has arguments to control spacing and offset, and it works as-is with all our existing plot types.

```cs
double func(double x) => Math.Sin(x / 5);
double[] xs = ScottPlot.DataGen.Consecutive(100);
double[] ys = xs.Select(x => func(x)).ToArray();

plt.PlotScatter(xs, ys);
plt.PlotSignal(ys, yOffset: .2);
plt.PlotSignalConst(ys, yOffset: .4);
```

image

This PR makes me think an example like this would be useful to add to the cookbook. 

Perhaps we could also add a helper function to the DataGen module which takes in a _function_ and other arguments similar to your `PlotFunction()` and returns the Y values as a double array ready to plot. 

Looking forward to hearing your thoughts!

swharden:
@Benny121221 after thinking about it a bit more I realized I may have overlooked something fundamental. Using your functional method we could design plots to generate the Y values on the fly at a density identical to horizontal pixel density and just for X values within the range being viewed 🤔 

@StendProg this might be useful for SignalPlot

I'll play with this a bit more and see what I come up with! Thanks again for bringing this up!@Benny121221. 

---

For what it's worth, here's a plot using the function-to-double method similar in style to the initial post:

```cs
double[] xs = ScottPlot.DataGen.Consecutive(200, 0.1, -10);
plt.PlotScatter(xs, xs.Select(x => 10 * Math.Sin(x)).ToArray());
plt.PlotScatter(xs, xs.Select(x => Math.Pow(x, 3)).ToArray());
plt.Axis(-10, 10, -10, 50);
```

image

Benny121221:
@swharden I assumed that optimization would be on scatterplot already? Or does it draw all points, whether or not they are in frame?

swharden:
> Or does it draw all points, whether or not they are in frame?

I think a scatter draws all points even not in the frame. Consider a plot with axis limis: -10, 10, -10, 10. If two points were added at -11, -11 and 11, 11, the scatter plot would have to draw them so your field would show a big diagonal line over it even though it contains no data points.

Because of this the Signal plot works differently and may be better optimized for plotting functions.

(SignalConst probably doesn't have any advantage here)

StendProg:
Hi @Benny121221 you PR have very interesting idea. But implementation can be improved.
Recreating scatter object on each render looks scarry. You can create Scater once and update its data on next render calls. Use Signal instead Scatter another improvement.
Anyway I see this plottable function in more general way. Render calculate points to draw on the fly like @swharden point.
There is possible to implement min/max subpixel calculations like Signal do. And some interesting min max search using gradient descent.

StendProg:
I prototyped `Signal` render from function in https://github.com/StendProg/ScottPlot/commit/384ec23e376f2d63aecff66e1a28ad4bdd50b1ca
Very dirty, but show the concept.

swharden:
Thanks @Benny121221 this looks great!

* I might still refactor a bit and change some argument names over the next few days as I practice using this
* We still need to make a demo program and cookbook example using this new plot type #250

swharden:
I'm going to re-open this because I think I can improve Render() to render exactly 1 point per horizontal pixel.

Also there's some weirdness...

func

Issue #244, opened 2020-02-10
Add support for box plots


Benny121221:
The statistics plot options are currently pretty limited. Box plots could change that.

https://www.isixsigma.com/wp-content/uploads/2019/07/Box-Plot.gif

^Example in case you aren't familiar

swharden:
@Benny121221 this is an excellent idea!

https://en.wikipedia.org/wiki/Box_plot

The PlottableBar could be modified to create a PlottableBox

https://github.com/swharden/ScottPlot/tree/master/cookbook#plot-bar-data

We could also add tools to the statistics module to turn a double arrays into the various box components. One thing to keep in mind is that many people use boxes to show quartiles, but I often use boxes to show standard error, so we should keep it flexible.

Benny121221:
#246

swharden:
### I created `ScottPlot.Statistics.PopulationStats()`
to make it easy to get min, max, mean, median, standard deviation, standard error, etc.

https://github.com/swharden/ScottPlot/blob/ade0446036fb4d814d49d4d4aff8b938b2e9ce81/src/ScottPlot/Statistics/PopulationStats.cs#L8-L58

### I created `ScottPlot.Statistics.BoxAndWhisker`
it is similar to OHLC. You're intended to build them and pass an array of these into Plot functions

https://github.com/swharden/ScottPlot/blob/ade0446036fb4d814d49d4d4aff8b938b2e9ce81/src/ScottPlot/Statistics/BoxAndWhisker.cs#L7-L75

### You can build a `BoxAndWhisker` like this

https://github.com/swharden/ScottPlot/blob/ade0446036fb4d814d49d4d4aff8b938b2e9ce81/demos/WinForms/Demos/FormBoxWhisker.cs#L20-L32

### You can plot `BoxAndWhiskers[]` like this

https://github.com/swharden/ScottPlot/blob/ade0446036fb4d814d49d4d4aff8b938b2e9ce81/demos/WinForms/Demos/FormBoxWhisker.cs#L82-L97

![image](https://user-images.githubusercontent.com/4165489/74205246-88831300-4c44-11ea-9324-7407c031d839.png)

swharden:
@Benny121221 did an awesome job with #246! I modified the interactive demo to showcase both of our plottables. I'm impressed there were no merge conflicts even though we both did the same thing at the same time 😎 

[...]

PR #246, opened 2020-02-11
Adds Support for Box and Whisker Plots

Benny121221:
#244 

I tested it out using this code:

```
using ScottPlot;
using System;
using System.Diagnostics;

namespace ScottPlotTesting
{
	class Program
	{
		{
			Random rand = new Random();
			string filename = "test.png";
			double[] xs = new double[3];
			double[][] ys = new double[3][];
			{
				xs[i] = i;
				ys[i] = new double[10];
				for (int j = 0; j < 10; j++)
				{
					if (j == 0)
					{
						ys[i][j] = 2;//Just to guarantee that there is an outlier
						continue;
					}
					ys[i][j] = rand.NextDouble();
				}
			}

			plt.PlotBoxWhisker(xs, ys);

			plt.SaveFig(filename);

				proc.Start();
			}
		}
	}
}
```


Benny121221:
Sorry, the force push was to rebase

swharden:
oops! It looks like we were working on this at the same time 😝 

Thanks for all your work on this! 

Let me see if I can rig the demo program I made to make it switch between both of our implementations so we can compare...

swharden:
Thanks so much @Benny121221! We can pick up the discussion on #244

PR #247, opened 2020-02-11
BoxAndWhisker improvements

Benny121221:
… individual points

As before, here is the code that I used to test:

```
using ScottPlot;
using System;
using System.Diagnostics;

namespace ScottPlotTesting
{
	class Program
	{
		{
			Random rand = new Random();
			string filename = "test.png";
			double[] xs = new double[3];
			double[][] ys = new double[3][];
			{
				xs[i] = i;
				ys[i] = new double[10];
				for (int j = 0; j < 10; j++)
				{
					if (j == 0)
					{
						ys[i][j] = 2;//Just to guarantee that there is an outlier
						continue;
					}
					ys[i][j] = rand.NextDouble();
				}
			}

			ScottPlot.Statistics.BoxAndWhisker[] boxAndWhisker = new ScottPlot.Statistics.BoxAndWhisker[ys.Length];
			for (int i = 0; i < ys.Length; i++) {
				boxAndWhisker[i] = ScottPlot.Statistics.BoxAndWhiskerFromData.GetBoxAndWhisker_OutliersAndQuartiles(ys[i], i);
			}
			plt.PlotBoxAndWhisker(boxAndWhisker);

			plt.SaveFig(filename);

				proc.Start();
			}
		}
	}
}

```

You can use `Statistics.BoxAndWhiskerFromData.GetBoxAndWhisker_OutliersAndQuartiles` or `Statistics.BoxAndWhiskerFromData.GetBoxAndWhisker_LimitsAndDeviation`, I tested both, they both work fine.

### DIFF
--- a/src/ScottPlot/Statistics/BoxAndWhisker.cs
+++ b/src/ScottPlot/Statistics/BoxAndWhisker.cs
@@ -44,6 +44,7 @@ namespace ScottPlot.Statistics
         public Midline midline = new Midline();
         public double xPosition;
         public string label;
+        public double[] points;
 
         public BoxAndWhisker(double xPosition)
         {
--- a/src/ScottPlot/Statistics/BoxAndWhisker.cs
+++ b/src/ScottPlot/Statistics/BoxAndWhisker.cs
@@ -49,6 +49,7 @@ namespace ScottPlot.Statistics
         public BoxAndWhisker(double xPosition)
         {
             this.xPosition = xPosition;
+            points = new double[] { };
             ResetStyle();
         }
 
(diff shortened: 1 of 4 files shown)


Benny121221:
Made some small oversights in the first commit

StendProg:
> //These could be made faster knowing that we have already sorted the list above

Possible small improvements assuming sorted array
```c#
lowOutliers = sortedValues.TakeWhile(y => y < lowerBoundary).ToArray();
...
minNonOutlier = sortedValues.Reverse().TakeWhile(y => y > lowerBoundary).LastOrDefault();
maxNonOutlier = sortedValues.TakeWhile(y => y < upperBoundary).LastOrDefault();
```

Benny121221:
Did you one better, both the lists of outliers and the min/max non-outliers can be optimized knowing that the list is sorted

StendProg:
Yep that's most performance. But you don't need temponary lists. You can find indexes and then use `Array.Copy` from SortedArray to results with right indexing.
Anyway this improvements cost nothing, because main time consumption is initial sort.
In this case previous linq also have rights to exist. (easy to read).

Benny121221:
You got it chief

swharden:
@Benny121221 thanks for this one too! It looks great!

I might continue to refine it a bit over the next few days. I'll track further work on this in #251

Issue #245, opened 2020-02-10
create tests to assess thread safety


StendProg:
`Settings.GetPixel()` and `Settings.GetLocation()` are not thread safe.
This break `Signal` and `SignalConst` parallel render.
This methods use `xAxisScale` and `yAxisScale`, and they reference to `Bitmap`(bmpData). Bitmap class not thread safe. 
Will it possible to make xAxisScale regular field, and update it then bitmap size actually changed?

Additionaly it will be usefull to have some `GetPixelX()` and `GetPixelY()`. It allow to make a bit optimizations to signal render. In some cases it need to calc only one coordinate.

swharden:
> `xAxisScale` and `yAxisScale` reference to `Bitmap`(bmpData). Bitmap class not thread safe.
> Will it possible to make xAxisScale regular field, and update it then bitmap size actually changed?

Yes, I can make this change!

Do you think you can make a multi-threaded test which fails so I can be confident that my fix is successful? I don't know much about threading so I'm not sure how difficult this is...

> Additionaly it will be usefull to have some `GetPixelX()` and `GetPixelY()`. It allow to make a bit optimizations to signal render. In some cases it need to calc only one coordinate.

Good idea - I'll do this as well.

StendProg:
> Do you think you can make a multi-threaded test which fails so I can be confident that my fix is successful? I don't know much about threading so I'm not sure how difficult this is...

Existing `RenderHighDensityParallel` throws on bitmap access. Good idea to add tests, but I can do it only tomorrow.

swharden:
> > Do you think you can make a multi-threaded test which fails so I can be confident that my fix is successful? I don't know much about threading so I'm not sure how difficult this is...
> 
> Existing `RenderHighDensityParallel` throws on bitmap access. Good idea to add tests, but I can do it only tomorrow.

Thanks for looking into it @StendProg! Take as long as you need.

This is why I disabled `useParallel` (#72) by default. If we get good at writing tests to prove things are thread safe I'd be happy to turn this back on.

I'm working hard on ScottPlot (in the dev branch) to remove `System.Drawing` from ScottPlot. Once that's finished we can be more confident everything is thread safe! It might take some more time to develop though 😅

swharden:
> `Settings.GetPixel()` and `Settings.GetLocation()` are not thread safe.
> This break `Signal` and `SignalConst` parallel render.
[...]

PR #248, opened 2020-02-11
optimize Signal & create parallel processing tests

StendProg:
Add tests for currently implemented parallel renderers.
Additionaly tests for thread safe calls of `Settings.GetPixel` and `Settings.GetLocation` methods.
This PR prevents future Issues like #245.

Replaced `Settings.GetPixel(0, value).Y` with new `Settings.GetPixelY()` in `Signal` and `SignalConst`.


swharden:
I made a benchmark program which gives a little more context to this optimization
image

Issue #290, opened 2020-03-12
Linear regression fails on unevenly-spaced X data


Benny121221:
I used this set: 

```
x,y
1,2
2,2
3,3
4,3
5,3.8
6,4.2
7,4
```

On my (emulated) TI-84 Plus, I got this result: https://photos.app.goo.gl/vmQ47RBT8p6tqNaG6 

And on a separate statistics program (https://github.com/Benny121221/wstat) I got this printout:
![image](https://user-images.githubusercontent.com/8635304/76569061-d70c1300-6477-11ea-9b80-9d207107e9e8.png)

However, using ScottPlot, I got the same offset, but the slope was 0.3200000. The proper slope is 0.4

I cannot see anything in the code that is obviously incorrect, however it is strange to me that the linreg code cares about the x-spacing, the spacing of x values doesn't matter to the LSRL.

I will make a PR to try and fix this, hopefully without causing any breaking changes.

StendProg:
#293 fix bug in spanX calculation.
With this fix ScottPlot result equal to @Benny121221 TI-84 Plus.

swharden:
Thanks @Benny121221 and @StendProg for working on this!

I added rSquared calculation to the fitter too. 

Excel might not be the pinnacle of statistical analysis software, but our values matching theirs gives us some reassurance we are on the right track!

ScottPlot | Excel
---|---
![image](https://user-images.githubusercontent.com/4165489/76683720-abef0400-65dc-11ea-8801-f69b1b326f26.png) | ![image](https://user-images.githubusercontent.com/4165489/76683723-aee9f480-65dc-11ea-89e6-1ba1b87d6f17.png)

swharden:
A problem remains: the linear regression fitter does not work with unevenly-spaced data.

I'm going to merge PR #291 because it is an improvement, but I will leave this issue open until the fitter can properly fit unevenly-spaced data.

@Benny121221 or @StendProg do one of you want to figure this out? I'm not great at statistics 😅 

Evenly-spaced Xs | Random Xs
---|---
![image](https://user-images.githubusercontent.com/4165489/76684028-e5c10a00-65de-11ea-861c-4703ad22dbf4.png) | ![image](https://user-images.githubusercontent.com/4165489/76684033-eeb1db80-65de-11ea-9dbb-6f7492cc2a5e.png)

swharden:
This test is used to generate the randomly-spaced data points shown in the screenshot. It can be used to help develop/test the solution.

[...]

PR #291, opened 2020-03-12
Fixes linreg slope issue and gets rid of xspacing thing

Benny121221:
#290


Benny121221:
This does have **one** breaking change, the signature of `GetResiduals` was changed because there is no reason for it to take a parameter

Benny121221:
Even though the xspacing thing is  no longer used in the code, you can still use that constructor

StendProg:
Hi @Benny121221.
I checked your PR.
Can you calculate `xs.Average()` `ys.Average()` only once, and then use this calculated values.
```c#
meanX = xs.Average();
meanY = ys.Average();
...
sumXYResidual += (xs[i] - meanX) * (ys[i] - meanY);
sumXSquareResidual += (xs[i] - meanX) * (xs[i] - meanX);
...
```

Benny121221:
I got you bud

swharden:
Thanks @Benny121221 this looks great! There's a bit more work to do (#290) but this is an excellent improvement.

PR #293, opened 2020-03-13
Fix wrong span calculation

StendProg:
This PR fix wrong span calculation in LinearRegressionLine #290.
@Benny121221 make more usefull PR #291, which also fix this bug and allow LineRegression for not evently space data.
At least you can grab test from this PR.


swharden:
Thanks for this PR @StendProg!

It looks like @Benny121221 fixed the span error in #291.

To avoid merge conflicts I manually added your new test to master:
https://github.com/swharden/ScottPlot/blob/3a652b4d1bf4f8c2bbece862e949597ae7ee5ce0/tests/Statistics/Regression.cs#L91-L99

Issue #207, opened 2020-01-16
Logarithmic axis


senged:
As far as I can tell this currently isn't supported. Request would be to allow setting any or all of the axis as log mode. Thanks.

swharden:
Hi @senged, thanks for your suggestion. This topic comes up from time to time (#29, #41), but there is currently no way to plot data on log scales.

I will create some helper functions and code demos to demonstrate how to transform log data such that it can be plotted on a linear axis (giving the appearance of a log scale), but true rendering of nonlinear axes is unlikely to be supported.

I'll leave this ticket open to track progress toward this end. Thanks again for the suggestion!

swharden:
I made a little tool to make this task easier and added a cookbook example showing how to do this.

https://github.com/swharden/ScottPlot/tree/master/cookbook#log-axis

I'll publish this on NuGet shortly

```cs
// plot the data using a linear axis
subplot1.PlotScatter(dataXs, dataYs, lineWidth: 0);
subplot1.Title("Data (Linear Scale)");
subplot1.YLabel("Vertical Units");
subplot1.XLabel("Horizontal Units");

// plot the same data using a log axis
subplot2.PlotScatter(dataXs, ScottPlot.Tools.Log10(dataYs), lineWidth: 0);
subplot2.Title("Data (Log Scale)");
subplot2.YLabel("Vertical Units (10^)");
subplot2.XLabel("Horizontal Units");
```

![image](https://user-images.githubusercontent.com/4165489/72669016-cea8d600-39fa-11ea-8f44-1c1ec06f8ce8.png)

at2software:
Hello @swharden and sorry for reaching out in this way. A few months ago we sent a development inquiry via email which is kind of related to this issue. I don't know if you received it (I can resend it of course if you're interested), but we would still be very interested in this feature. Are there any possibilities that this can be re-considered to implement?
Please feel free to delete this post if you should find it inadequate.

Issue #41, opened 2019-07-16
Arbitrary non-orthogonal axis


Padanian:
Like in the example below, it would be great to have additional non-orthogonal axis, for functions z=f(x,y) where x,y are horizontal and vertical axis

![psy](https://user-images.githubusercontent.com/7769348/61295561-95d43e00-a7d8-11e9-92c8-312d4e0c4d5c.jpg)

swharden:
YES! While I don't have a personal use case for non-orthogonal axes (maybe [smith charts](https://en.wikipedia.org/wiki/Smith_chart)?), I have wanted to implement logarithmic axes for a while (#29). I did some preliminary work toward implementing this and it was _not_ easy because so many systems rely on on the current coordinate system (mouse click-dragging, tick spacing, grid line rendering, etc). The fix would require some refactoring, but I'd love to do work on it with you.

A better place to start would be to code a proof-of-concept demo as a standalone application. I created a Cartesian plotting demo easily within 100 lines (https://github.com/swharden/Csharp-Data-Visualization/blob/master/examples/2019-06-03-coordinate-system) and I think something like this could be expanded to support nonlinear / non-orthogonal axes. Once we get a handle on how to do it technically (and provide a good interface for it), we can then figure out how to refactor ScottPlot to accomodate it. What do you think of this idea?

A good place to start would be to create a similar app which:
* has a way to draw _points_ on the graph. This is easiest in my experience. 
* has a way to draw _lines_ on the graph. This is where I keep getting overwhelmed with complexity. Grid lines will be unevenly spaced and may curve.
* allows the user select which coordinate system to use
  * perhaps we would make a `CoordinateSystem` class and provide some pre-made ones like `cartesian` and `logarithmic`, `polar`, but let the user make their own with custom formulas if desired.
  * I'm guessing functions like coordinate-to-pixel, drawing of lines, and generation of ticks would all be handled inside these classes

@Padanian I'm looking forward to hearing your thoughts

Padanian:
A coordinate class is the way to go. Any non orthogonal axis (suppose
a y' axis with some arbitrary clockwise angle t and intercept k with
respect to X-axis) can be resolved with the formula y'=-x sin(t) +y
cos(t) + k.Therefore all standard stuff apply, up until actual
representation on the rotated axis.Imagine the example demo
double[] xs = new double[] {0, 1, 2, 3, 4, 5};
double[] ys = new double[] {0, 1, 4, 9, 16, 25};

[...]

Issue #29, opened 2019-06-25
support log scales


swharden:
Consider a `plt.LogScale()` function like:
```cs
void LogScale(bool xAxis: false, bool yAxis: false)
```
This will have all sorts of side effects, but let's explore them.

A strong start is to modify the coordinate/pixel conversion functions to do a log conversion:
https://github.com/swharden/ScottPlot/blob/4bc377cc10a33ee25b20e8784a4822aee1b8ff91/src/ScottPlot/Settings.cs#L395-L411

swharden:
What a can of worms! Lots of the code currently expects a linear canvas:
* mouse click/drag panning and zooming
* axis tick labels
* rendering functions (data-to-screen pixel lookups)

I'm going to put idea on the back burner until at least the next major release. This will require re-coding a LOT to support, but I'd still like to get there eventually.

swharden:
I added a cookbook example showing how to do this:
https://github.com/swharden/ScottPlot/tree/master/cookbook#log-axis

![image](https://user-images.githubusercontent.com/4165489/72669139-5e02b900-39fc-11ea-90c2-10c336322635.png)

PR #295, opened 2020-03-13
Polar plotting for Scatter Plot

Benny121221:


StendProg:
Hi @Benny121221.
1. It looks like you base you work on unmerged branch. Can you rebase `polar-plot` on `master` branch?
2. I think it's better to make helper method for converting polar coordinates. `DataGen.cs` good place for that. (You may not agree, I do not make decisions)

Benny121221:
Sorry, I accidentally based this branch off of my linreg branch, not master, I'll fix that when I get home.

swharden:
Hey @Benny121221, thanks for this PR!

The topic of plotting data on non-linear coordinates comes up every once and a while. #41 #207 #29

Since ScottPlot renders on Bitmaps using pixel coordinates, all math assumes Cartesian coordinates. The best way to display non-Cartesian data with ScottPlot's Cartesian coordinate system is to convert the data to Cartesian coordinates before plotting it. It is probably best if the user does this conversion.

If you want to write a tool to simplify conversions between Polar and Cartesian coordinates, a function which accepts a polar data and returns Cartesian data would be ideal. You can model your function after `ScottPlot.Tools.Log10()` which performs a similar conversion for plotting log data.

https://github.com/swharden/ScottPlot/blob/3a652b4d1bf4f8c2bbece862e949597ae7ee5ce0/src/ScottPlot/Tools.cs#L221-L227

If you make such a function, it's a good opportunity to make a new PR based on the master branch.

PR #298, opened 2020-03-14
Tools.ConvertPolarCoordinates()

Benny121221:
This time the conversion is done in `src/ScottPlot/Tools.cs`, not in the PlotScatter function. The only thing is this returns a tuple, I would be happy to change the return type if you feel like that would better reflect the direction of the project.


swharden:
Thanks @Benny121221, this is looking great!

Want to create a cookbook example to highlight the best way to use this function?

You can just add another little class to the bottom of this file and model it after the `LogAxis` class. Just by adding a class here (you don't have to do anything else) it will show up in the cookbook 👍 

https://github.com/swharden/ScottPlot/blob/981645942e24ebb8f6f9ad152ec56d96609333f2/src/ScottPlot.Demo/Customize/Axis.cs#L49-L73

Benny121221:
I added a demo to the cookbook, it looks like this:

image

StendProg:
You can give name to toople elements. This would be more readable and intellisense grab this names:
```c#
public static (double[] xs, double[] ys) ConvertPolarCoordinates(double[] rs, double[] thetas) 
```

swharden:
Looks great @Benny121221, thanks! I made the change @StendProg suggested and merged 👍

Issue #302, opened 2020-03-14
PlotFunction rendering can be improved


swharden:
Initially implemented by @Benny121221 in #243, I think there's room to improve this plottable.

There's a cookbook example showing how to use it, but it's obvious it renders a little bit finny.

Render() could be improved so it renders exactly one point per x pixel.

![func](https://user-images.githubusercontent.com/4165489/76690558-93e8a600-6617-11ea-9f29-9e88f9246b3a.gif)

swharden:
I'm currently trying to work 100% on #271, so @Benny121221 if you feel up to giving this a try you might be able to do it faster than me!

swharden:
@Benny121221, awesome job fixing this up! Just in time for the next release. Thanks!!

![func2](https://user-images.githubusercontent.com/4165489/76692491-f056c000-662d-11ea-9874-1e605d3a4862.gif)

swharden:
Actually, are the arguments still necessary in `Plot.PlotFunction()`? If you remove the min and max arguments you can also remove the obsolete tag while you're in there, then I'll merge it in 👍 

https://github.com/swharden/ScottPlot/blob/d0132655b926db529595f8f7df07b1b49ceb8888/src/ScottPlot/Plot.cs#L334-L347

swharden:
... and if you change the arguments, you should also update the cookbook

https://github.com/swharden/ScottPlot/blob/d0132655b926db529595f8f7df07b1b49ceb8888/src/ScottPlot.Demo/Experimental/Function.cs#L9-L19

Benny121221:
Got it. The only thing left is the error handling, if someone tries to plot `Log(x)`, they're gonna have a bad time unless they do this: `if(x<=0) return null; else return Math.Log(x);` which I think could be less clunky.

There are try/catch statements in `PlottableFunction.Render` but they don't actually stop it from bugging out. I think the best option is either fix that or write very publicly in the documentation that for functions with a restricted domain, make sure they don't throw any exceptions but instead return null.

Benny121221:
*I went with the second option, I added a note to the description of the demo

swharden:
Thanks so much @Benny121221, this looks great! I moved the demo out of "experimental" and into "plot types" 😎 

Just published ScottPlot 4.0.21 on NuGet

PR #303, opened 2020-03-14
Optimizations to function plotting allow minimal rendered points

Benny121221:


Benny121221:
#302

Benny121221:
This made two big changes, firstly it now will not calculate (or render) points closer than one pixel together.

Further, it will not calculate points outside the viewport, which gets rid of the issue where zooming in would cause my computer to have a stroke.

All of this means that for a function with domain `[-10,10]`, the number of points drawn will not exceed 1956 (in the WPF demo on a 1920x1080 monitor with the window maximized). This will drop massively if you zoom out. Previously the number of points would be much larger, and blow up if you zoom out. Plus, the graph was very jagged and imprecise if you zoomed in.

swharden:
Great way to address #302 - thanks @Benny121221!

Issue #301, opened 2020-03-14
Set number and date format using a custom culture


SeidChr:
Hi,

i work on machines with different culture settings and sometimes even on vscode remote on a virtual azure box.
All the time i got slightly differntly formatted numbers and dates (i didnt even realize it in the beginning).
To combat this and to set up some custom formattings matching the purpose of the plot, i had to create a complete culture and set the desired formatting there:
https://github.com/SeidChr/COVID-19-Report-Data-Pipeline/blob/f4b2b1cc03ebc105979c7dff7e1359d1fc0da790/src/Pipelines/DailyReportPipeline.cs#L201
It would be great to be able to set these formatting-strings directly without the need to create a new culture.

swharden:
Hi @SeidChr, 

Thanks for noting this issue! A similar question was asked recently #297 so it seems this project may benefit from a better/simpler way to do let the user define how ticks are formatted.

Your solution of creating a custom culture is creative! I wouldn't have thought of that. Perhaps an intermediate solution would be an overload of `plt.SetCulture()` which has several arguments to let the user do something like:

```cs
var plt = new ScottPlot.Plot();
plt.SetCulture(ShortDatePattern = "MM-dd", decimalDigits: 2, decimalSeparator: ",");
```

Then the custom culture creation happens behind the scenes (using default culture values for all arguments not explicitly stated).

What do you think of this possible solution? If you want to give it a go, add an overload here and issue a PR and I can take a look!

https://github.com/swharden/ScottPlot/blob/69ede7b1e6cdb80eaea81567e617db599b9868c1/src/ScottPlot/Plot.cs#L1348-L1351

SeidChr:
Wouldn't have thought of using dotnet cultures at all, when creating a plotting engine, but i assume it makes things easier. Really, im not a fan of .net cultures at all 😅 
Having a variable formatting when executing the same software on another system really makes things indeterministic sometimes.
I would just directly format the numbers or dates with `.ToString(SPECIFIED_FORMAT_STRING)` wherever they are printed. Just as i can print custom tickmarks as i like.

Anyway, I have added a pullrequest with your suggestion.

swharden:
just released in this on NuGet in ScottPlot 4.0.21, thanks again for your help @SeidChr!

swharden:
> Wouldn't have thought of using dotnet cultures at all, when creating a plotting engine, but i assume it makes things easier.

[...]

PR #304, opened 2020-03-15
add SetCulture overload

SeidChr:
As suggested in #301

Build of the project was successfull, but i was unable to build or run the whole solution in vcode online.  
Could you check whether this is working as expected?

Cheers


swharden:
Looks great, thanks @SeidChr! #301

Issue #315, opened 2020-03-20
Refactor Bar and Box Plots


swharden:
The current bar plots work somewhat, but it's not easy to add new features to them. The bar plot module should be refactored to accomodate several new desired features. This issue unifies #24, #260, and #277. After this module is complete, it can probably be modified slightly to support box plots too (#244) to make it easier to display grouped box plots. Perhaps a single class can be used for bar and box plots.

### Example (from matplotlib)

![image](https://user-images.githubusercontent.com/4165489/77127431-cda81b00-6a22-11ea-8a14-c7c6c676f83d.png)

* This graph has 5 ***groups*** (G1, G2, G3, G4, and G5)
* This graph has 2 ***bar series*** (Men and Women)

One can say that if several ***bar series*** objects have the same number of elements in them, it is possible to display them organized by ***group***. Each bar series and each group has its own label.

### Features:
- [x] can display a single bar series
- [x] can display several bar series arranged by groups
- [x] each bar series can be individually customized
- [x] each bar series should appear separately in the legend
- [x] it should be easy to display as vertical or horizontal bars
- [x] it should be possible to display as stacked bars
- [x] optional error-bars should be supported
- [x] add option to customize lines
- [x] add option to display values over bars
- [x] add support for box plots
- [x] display scatter data next to plots
- [x] display data distribution curve

swharden:
Experimental code is in:
https://github.com/swharden/ScottPlot/blob/master/tests/Experimental/BarGraphTests.cs

### Single Bar Series

```cs
double[,] values = new double[,] { { 7, 12, 40, 40, 100, 125, 172, 550, 560, 600, 2496, 2789 } };
string[] rowLabels = new string[] { null };
string[] colLabels = new string[] { "ant", "bird", "mouse", "human", "cat", "dog", "frog", "lion", "elephant", "horse", "shark", "hippo" };

var plt = new ScottPlot.Plot(400, 300);

// create and add an experimental plottable (not in the ScottPlot namespace)
var pltBar = new PlottableBar2(rowLabels, colLabels, values);
plt.Add(pltBar);

plt.Title("Body-to-Brain Mass Ratio");
plt.XTicks(colLabels);
plt.Axis(-1, colLabels.Length, 0, null);
plt.Grid(enableVertical: false);

// customize ticks to display better at high density
plt.Ticks(useMultiplierNotation: false);
plt.Ticks(xTickRotation: 45);
plt.Layout(xLabelHeight: 40);
```

[...]

Issue #313, opened 2020-03-19
plots with no data render as a black square


swharden:
Plots look odd until data is plotted

```cs
var plt = new ScottPlot.Plot(400, 300);
plt.Title("empty plot");
plt.SaveFig("test.png");
```

![image](https://user-images.githubusercontent.com/4165489/77018217-17293500-6953-11ea-9756-fdf981d9b76b.png)

This is jarring for user controls because as soon as you cick/drag, grid lines and tick labels appear.

Interestingly, just setting the axis solves this.
```cs
plt.AxisAuto();
```

![image](https://user-images.githubusercontent.com/4165489/77018255-4344b600-6953-11ea-9118-049ad5fa0757.png)

swharden:
Ahh! I think https://github.com/swharden/ScottPlot/commit/6972d32977abab05a468a769729c4d8b30f956dc may have been there for a reason. The demo doesn't auto-axis properly when it starts now:

![image](https://user-images.githubusercontent.com/4165489/77242325-a5abe980-6bd3-11ea-9fe2-76116963f3ad.png)

PR #316, opened 2020-03-20
Add asymmetric error bars to scatter plots

zrolfs:
Hi Scott!
I wanted to use asymmetric error bars, so I added in some functionality to allow users to do that. To clarify, this pull request should allow users to:
-create asymmetric error bars
-create only a single (positive or negative) error bar

I used the overloading mechanism so that the methods "PlotScatter" and "PlottableScatter" should still be compatible with previous calls.

Thanks for your consideration with this pull request! I'm happy to answer any questions or make changes if you disagree with something.
Zach


StendProg:
Hi @zrolfs I think you need to add overrides for `PlotPoint` and `PlotStep` too.

Using overload works, but quickly inflates code.
It might be better to think through an API to avoid this.
I see some possible solutions:
1. Add optional `errorXarrayNegative` and `errorYarrayNegative` params with default null value.
If they not null, then you interpret existing `errorX` and `errorY` as `errorXarrayPositive` and `errorYarrayPositive`.
2. Add optional `errorXarrayDelta` and `errorXarrayDelta` params with default null value.
`errorXarrayNegative = errorX + errorXarrayDelta`....
3. To use assymmetric error bars user must provide double sized arrays to `errorX` and `errorY`.
There first half contain positive error values, second half contain negative error values.

Do not rush to realize this, I think Scott will choose 4 variant even better.

zrolfs:
Hi @StendProg 
Thanks for your feedback! I agree the overload is not an ideal solution. I appreciate your proposed solutions, but I'm not sure how intuitive those are for the users. 
My preferred solution would be to simply change the original constructors of PlotScatter and PlottableScatter so that they accept the positive/negative arrays (which is what I did with PlotPoint; PlotStep doesn't use error bars). I wrote in the overrides in case we wanted to prevent build errors when users update their versions. I'd be happy to remove them if we don't think that's going to be a big deal.

StendProg:
I looked more closely at the PR, and didn’t cling to anything, each line expands the functionality and is necessary.
I can offer only one improvement, although it does not related to this PR.
You can simplify min/max search in Scatter `GetLimits()` with linq:
```c#
limits[1] = xs.Zip(errorXPositive, (x, e) => x + e).Max();
...
```
Not sure if this is easier to read for everyone, but shorter.

Additionaly you can add `Demo` for this new error bar mode. It's simple task with new demo engine.

swharden:
Hey @zrolfs I haven't taken a look at this yet but it sounds interesting! @StendProg is right, if you add a new demo for this feature it will be a lot easier to assess.

Perhaps you can model your demo after this one:
http://swharden.com/scottplot/cookbooks/4.0.22/#PlotTypes_Scatter_ErrorBars

by mimicking the class:
https://github.com/swharden/ScottPlot/blob/6972d32977abab05a468a769729c4d8b30f956dc/src/ScottPlot.Demo/PlotTypes/Scatter.cs#L109-L143

zrolfs:
Sounds good! I'm busy right now, but I should have time this evening.

swharden:
I think there's some type of rendering error based on the demo...

image

but before going much farther far down this path, I'm wondering if `PlottableErrorbars` should be its own type of Plottable, not included inside `PlottableScatter`? If it is separate from Scatter, it can be added/removed/styled independently, and even used with other plot types (such as bar plots). Also the code would be in its own file, not require modification of any existing plottables, and probably be easier to write and maintain 🤔 

@zrolfs what do you think?

zrolfs:
I like that idea! I'll need to think of a clean way to ensure the color of the error bars matches the color of the plotted series. Would you like me to remove the original _symmetric_ error bar code that is in PlottableScatter and the bar plots?

swharden:
> I like that idea! I'll need to think of a clean way to ensure the color of the error bars matches the color of the plotted series. 

Good point... I'm not really sure of a way to do that automatically other than to define the color of the plotted series when you make it, and also define the color of the errorbars when you make that too. An easy way to get those colors is to use this function which returns a pre-defined series of nice-looking colors based on an integer value:
```cs
var thisColor = new ScottPlot.Config.Colors().GetColor(indexValue);
```

> Would you like me to remove the original _symmetric_ error bar code that is in PlottableScatter and the bar plots?

I think that would be easiest because the "files changed" by this PR as it it is right now is a bit intimidating to review 😅 It would be a lot easier if most of the changes were compartmentalized in a new file.

https://github.com/swharden/ScottPlot/pull/316/files

swharden:
Thanks for woking on this @zrolfs! Let me know when you think it's ready to merge and I'll take a closer look 👍

zrolfs:
Hi Scott! I think it's ready. 
I hit a snag for a while on the error bar offset, but I figured it out. I think this solution is a lot cleaner that the first draft (but there might be even more line differences now)! However, one thing that's kinda hacky is the implementation of the default error bar color. These are lines 63-65 in Settings.cs. The idea is to have the user make their scatter/bar plot and then their error bar plot. We can also scrap this and just make the default error bar color black. Users can just figure out the color of the previous serious by saving the prior plot as a local variable. I don't have strong opinions either way, but I wanted to draw your attention to it. Let me know which solution you prefer.

swharden:
> Hi Scott! I think it's ready .... Let me know which solution you prefer.

Thanks @zrolfs I'll take a look! It might take me a day or so, but I'll take it from here and add some modifications/commits to this branch as I go.

I'll post an update here when things are looking good to merge!

swharden:
This is not @zrolfs's fault but there is a GDI rendering glitch where diagonal lines get drawn inappropriately. I ran into this before but I forgot how I solved it...

err

swharden:
> I hit a snag for a while on the error bar offset, but I figured it out

No kidding - I'm hitting the same snag! This horizontal offset stuff is super confusing. I'm refactoring the bar plots as part of #315 and they'll be a little smarter about error bars... so I'll try removing the xOffset code from `PlottableErrorBars` and see if it makes it easier to work with. 

I understand it may break errorbars some of the multi-bar-plot demos, but those are about to be replaced anyway due to #315 so I'm okay with that.

zrolfs:
> This is not @zrolfs's fault but there is a GDI rendering glitch where diagonal lines get drawn inappropriately. I ran into this before but I forgot how I solved it...
> 
> err
> 
> image

Yeah! I saw that too! It strangely went away if I didn't touch it for a while...

src/ScottPlot/plottables/PlottableErrorBars.cs
```diff
@@ -14,30 +14,24 @@ public class PlottableErrorBars : Plottable
...
             //save properties
             this.xs = xs;
             this.ys = ys;
-            this.xPositiveError = xPositiveError;
-            this.xNegativeError = xNegativeError;
-            this.yPositiveError = yPositiveError;
-            this.yNegativeError = yNegativeError;
-            this.capLength = (float)capLength;
-            this.xOffSet = xOffSet;
+            this.xPositiveError = SanitizeErrors(xPositiveError, xs.Length);
```
zrolfs: Good change!

src/ScottPlot.Demo/PlotTypes/Bar.cs
```diff
@@ -27,11 +27,10 @@ public void Render(Plot plt)
                 }
 
                 // make the bar plot
-                plt.PlotBar(Xs, dataA);
-                plt.PlotErrorBars(Xs, dataA, null, null, errorA, null);
+                plt.PlotBar(Xs, dataA, errorY: errorA);
 
                 // customize the plot to make it look nicer
-                plt.Axis(-10, 110, 0, 110);
+                plt.Axis(null, null, 0, null);
```
zrolfs: I changed these to help me debug. For WPF, the null defaults made the plots super zoomed in on the first bar. I thought I had screwed up the code until I zoomed out. Just a heads up.

swharden: This might be a bug that just got introduced by #313... I'll take a closer look! Other cookbook have messed up auto-axis suddenly

swharden:
@zrolfs this is looking great! Sorry if I was ambiguous earlier when I suggested having a plottable errorbar would be cool for things like scatter plots and bar plots - my thinking wasn't to remove that functionality from scatters and bars, but rather to provide a secondary (more customizable) way to add scatter plots. A lot of my recent comments were undoing changes to those modules which makes sure this merge won't break anybody's code out there who relies on the existing functionality (and it also makes the PR/merge easier to review because it's fewer files!)

I going to keep working on this a bit more, but thought I'd give you an update

zrolfs:
Thanks for keeping me posted! I've really appreciated your support and communication. The code base was pretty intuitive and fun to work on, too! If this doesn't get merged by Monday, I'll probably just make a local nuget release for my project, so no rush.

swharden:
> Thanks for keeping me posted! I've really appreciated your support and communication. The code base was pretty intuitive and fun to work on, too! 

That's great to hear! It's a lot of work to keep this library minimal and light, but it's worth it because it's so much nicer of a library to work in!

> If this doesn't get merged by Monday, I'll probably just make a local nuget release for my project, so no rush.

Whatever you're comfortable with. I suspect this will be done by then, but the auto-axis bug that _just_ got introduced in #313 might wind up in there too so you might do well to wait for the "official" one which will be right on its heels.

Regarding this PR, we're getting there! Still want to make a few tweaks, but I think this plottable stands nicely on its own

image

swharden:
> there is a GDI rendering glitch where diagonal lines get drawn inappropriately. I ran into this before but I forgot how I solved it...
> 
> image

**HOW TO FIX THE GDI RENDERING GLITCH / NOTE TO FUTURE SCOTT:** This GDI rendering glitch only affects small, perfectly level lines drawn with anti-aliasing mode enabled. I fixed this glitch by applying a slight offset to one of the points of the line so it's not perfectly flat. The tilt is invisible to humans, but it fixes the glitch

swharden:
@zrolfs this looks great! I'm gong to merge and close this PR now. I'll release a new NuGet package later today. Thanks for all your hard work!

PR #367, opened 2020-05-02
Simplify creation of grouped series bar graphs

Benny121221:
…the same bar plot more easily

**Purpose:**
To allow plotting multiple bars on the same plot more easily.

**New functionality (code):**
I added a new demo to show it off under "Multiple Bar Graphs (New)"

image


Benny121221:
The screenshot doesn't show the full sourcecode, that's because the demo app requires using spaces for indenting, and my editor is set up to use tabs. I changed to spaces in 6cbcf122f6dbc9c5d28724d87f309d3287a98e44

swharden:
Hey @Benny121221, thanks for creating this example! I'm planning on working on ScottPlot most of tomorrow so I'll merge this then and hopefully close-out all the open issues too 🤞

swharden:
Hey @Benny121221, thanks for this PR! I'm _really_ hesitant to add a new bar type plottable though since we already have two (and it seems like we should just have one!)

However, I agree with your observation that plotting grouped series data is really complicated (http://swharden.com/scottplot/cookbooks/4.0.29/#plottypes-populations-plot-a-population-multi-series). Note that if that example were modified using the 3 lines shown here, the output is almost identical to what you're trying to accomplish:

```cs
var popPlot = plt.PlotPopulations(multiSeries);
popPlot.boxStyle = PlottablePopulations.BoxStyle.BarMeanStdErr;
popPlot.displayItems = PlottablePopulations.DisplayItems.BoxOnly;
```

image

In general I like your idea of using 2D arrays to group data. Maybe we could create a `Plot.MultiBar()` method in the `Plot` module like you made which accepts 2D arrays, but when it's called it really just builds a population plot like the one in the picture?

I should probably work on this some more, because I don't think you can create a population plot just by giving it means and errors... currently it requires individual data values, and it calculates the means and errors.

I'll keep working on this! I'll leave this PR open while I do. Thanks for bringing attention to this issue @Benny121221

Benny121221:
Just putting this out there, it might make sense to transplant my changes to `RenderBarVertical/Horizontal` into the `PlottableBar` and then change `PlotMultiBar` to call `PlottableBar` with a different constructor.

This means we don't add another plottable, and we don't break any compatibility.

Going the population series route might be easier, I'll admit I haven't had a gander at it.

swharden:
> Just putting this out there, it might make sense to transplant my changes to `RenderBarVertical/Horizontal` into the `PlottableBar` and then change `PlotMultiBar` to call `PlottableBar` with a different constructor.

I like this idea! Want to try modifying `PlottableBar` to add this 2D functionality? The function signature of `Plot.PlotMultiBar()` can stay like it is, and the demo program probably doesn't have to change.

Also, `Plot.PlotMultiBar()` probably doesn't need to accept `xs` since it could generate the consecutive numbers itself based on the length of Ys.

What do you think? I like the idea of avoiding the population code. It's quite a lot 😅

swharden:
Actually... probably don't make those changes to PlottableBar quite yet. 

Let me review your code more thoroughly and think about this problem some more first. I don't want you to put work in to something like this if we end up changing it later. The thing that I'm stuck on is the best way to handle labels and legends, which gets complicated when we have 2D groups.

I'll revisit this tomorrow! Thanks again for all your help on this @Benny121221

swharden:
I added a test to develop against. We know we want this code to produce this output. 

This discussion and PR is just about what happens under the hood when you call PlotMultiBar 🤔 

```cs
plt.PlotMultiBar(
    xs: new double[][] { xs, xs, xs }, 
    ys: new double[][] { ys1, ys2, ys3 }, 
    yErr: new double[][] { err1, err2, err3 });
```
image

swharden:
Something that got overlooked is that both groups and series need labels. Group labels are ticks on the horizontal axis, series labels should appear in the legend.

```cs
string[] groupNames = { "one", "two", "three", "four", "five", "six", "seven", "eight", "nine", "ten" };
string[] seriesNames = { "alpha", "beta", "gamma" };
```

@Benny121221 what do you think of the possible solution I committed to this PR? I created an overload for your PlotMultiBar which accepts 2D arrays and creates appropriately labeled and positioned bar graphs. It takes care of the X positions and horizontal ticks too.

```cs
plt.PlotMultiBar(
    groupLabels: groupNames,
    seriesLabels: seriesNames,
    ys: new double[][] { ys1, ys2, ys3 },
    yErr: new double[][] { err1, err2, err3 });
```

image

Benny121221:
Looks good! The only thing is I think it should still be an option to put in non-consecutive x values, or x values that don't start at one

swharden:
> I think it should still be an option to put in non-consecutive x values, or x values that don't start at one

Actually, the user is never prompted for x values. Notice that the function only asks for Y values and group labels. All the x code is handled behind the scenes.

Benny121221:
> Actually, the user is never prompted for x values.

I noticed that, but surely there is a usecase where that becomes necessary? Regular bar plots support this, it would be a shame to cut a feature in their big brother plottable.

swharden:
> > Actually, the user is never prompted for x values.
> 
> I noticed that, but surely there is a use case where that becomes necessary? Regular bar plots support this, it would be a shame to cut a feature in their big brother plottable.

It seems like _grouped_ bar plots will always require the user to forgo numeric X units, otherwise the bars of all series would overlap 🤔 

I'm thinking that use cases which require absolute numerical X positioning should probably avoid "grouped" plots altogether, and create specific bar plots the traditional way that lets users define Xs and Ys.

swharden:
> It seems like _grouped_ bar plots will always require the user to forgo numeric X units, otherwise the bars of all series would overlap 🤔

This is especially true since the new `PlotMultiBar()` destroys the X axis, removing numeric tick labels and replacing them with group names

Benny121221:
> This is especially true since the new PlotMultiBar() destroys the X axis, removing numeric tick labels and replacing them with group names

Ah, I didn't notice that. I suppose group names makes more sense for grouped plots. Plus I don't suppose there's anything stopping someone from passing in group names like `"1", "2", "4", "5"`, which would achieve the same thing as what I was saying.

swharden:
Cool, I'll refine this a bit and merge it tonight. Thanks again for suggesting it @Benny121221!

I'm thinking of changing the name start with Bar so it's easier to find (all the bar methods will cluster alphabetically). Something like `PlotBarSeries()` or `PlotBarGrouped()`...

... or I could just make it an overload of `PlotBar`? It might not be as easy to discover that way though

swharden:
I moved things around a bit but the function is the same. The method is now called `PlotBarGroups`. I think it looks pretty good! 

I'm ready to merge if you think it looks good too @Benny121221 

```cs
string[] groupNames = { "one", "two", "three", "four", "five" };
string[] seriesNames = { "alpha", "beta", "gamma" };

int groupCount = groupNames.Length;
Random rand = new Random(0);
double[] xs = DataGen.Consecutive(groupCount);
double[] ys1 = DataGen.RandomNormal(rand, groupCount, 20, 5);
double[] ys2 = DataGen.RandomNormal(rand, groupCount, 20, 5);
double[] ys3 = DataGen.RandomNormal(rand, groupCount, 20, 5);
double[] err1 = DataGen.RandomNormal(rand, groupCount, 5, 2);
double[] err2 = DataGen.RandomNormal(rand, groupCount, 5, 2);
double[] err3 = DataGen.RandomNormal(rand, groupCount, 5, 2);

plt.PlotBarGroups(
    groupLabels: groupNames,
    seriesLabels: seriesNames,
    ys: new double[][] { ys1, ys2, ys3 },
    yErr: new double[][] { err1, err2, err3 });
```

image

Benny121221:
Yep, looks good

PR #370, opened 2020-05-03
Easier function plotting

Benny121221:

**Purpose:**
This makes plotting functions with restricted domain easier.

**New functionality (code):**
Previously, if I wanted to plot the function `ln x`, I would have to do it like this:

`x => x > 0 ? Math.Log(x) : (double?)null`

Now I can just do `x => Math.Log(x)` and it will be taken care of.


Benny121221:
Sorry about not including this originally when I was working on the function plotting, the realization didn't hit until I was using the library in my own project this morning and I had to replicate this commit. I figured the change would probably be useful on the upstream library.

swharden:
looks great @Benny121221, thanks! I'll push this to NuGet tonight

Issue #151, opened 2019-10-20
WpfPlot (control) does not appear in toolbox


RalphLAtGitHub:
Issue: After installing ScottPlot via PackageManager the UserControl "WpfPlot" does not appear in the Toolbox (VS 2019).

I found two solutions that might be mentioned in the Quickstart document to make life easier for a beginner:

Solution 1: Add Control manually in the XAML editor without adding the UserControl to the Toolbox:
` <Grid>
        <scottplot:WpfPlot Name="wpfPlot1" Margin="10"/>
        ...
  </Grid>`

Solution 2: Right-click the Toolbox, pick "Choose Items...", click "Browse" in any tab, find the ScottPlot.dll, select it and click "Open".

In both cases you have to add in the window tag:
`xmlns:scottplot="clr-namespace:ScottPlot;assembly=ScottPlot"`

swharden:
Hi @RalphLAtGitHub, thanks for mentioning this! Although someone can use the `WpfPlot` control by manually creating it in text, having it appear in the toolbox would be much easier. I hadn't noticed it doesn't appear by default.

After some googling I think this is what I need to do to fix this issue: https://docs.microsoft.com/en-us/nuget/guides/create-ui-controls#add-toolboxassets-pane-support-for-xaml-controls

I've give it a shot and update this issue when I learn more

swharden:
@RalphLAtGitHub thanks again for bringing this to my attention!

I couldn't easily figure out how to do this with the current packaging system. I'm about to use a new packaging system when we migrate to .NET Core 3.0 (#86) so I won't prioritize this for now.

Based on your tips I updated the WPF quickstart to clarify the toolbox issue.
https://github.com/swharden/ScottPlot/blob/master/doc/quickstart/README.md#wpf-application

Fingers-crossed this will fix itself after the migration to .NET Core 🤞

swharden:
I found an open-source NuGet package that adds a custom control to the toolbox. No xmlns registration needed in your xaml files either. I'll see if I can mimic what they do...

https://github.com/dotMorten/UniversalWPF

swharden:
I can't believe I spent 5 hours on this... and it's still not working 💀 So far I:

* made a `VisualStudioToolsManifest.xml` and added a `ToolboxItems` section
* edited the csproj file to:
  * include the xml manifest
  * set OutputType to library
  * set RootNamespace to ScottPlot
[...]

PR #371, opened 2020-05-03
show WpfPlot in Toolbox

swharden:
try to get WpfPlot to show in the WPF toolbox, ideally without requiring a xmlns definition #151


swharden:
accomplished by e3e7dd83dd7df61bd1917115517b88bfae1aa1db

PR #386, opened 2020-05-12
Ability to use custom-drawn markers

henkburgstra:

**Purpose:**
This is a non-evasive change. The existing interface remains unchanged.

I needed the ability to use custom-drawn markers. This change adds this ability. The change adds public field Settings.markerDrawer of type IMarkerDrawer. PlottableScatter.Render uses this field if present to draw markers.

**New functionality (code):**

Implementation of IMarkerDrawer (open triangle in this case) to be used in an Audiogram (https://en.wikipedia.org/wiki/Audiogram)
```cs
    public class LuchtGemaskeerdRechts : IMarkerDrawer
    {
        public void DrawMarker(Graphics gfx, PointF pixelLocation, MarkerShape shape, float size, Color color)
        {
            var c = new MarkerConfig(gfx, pixelLocation, shape, size, color);
            gfx.DrawPolygon(c.Pen, new PointF[] { 
                new PointF(pixelLocation.X, pixelLocation.Y - size / 2),
                new PointF(pixelLocation.X - size / 2, pixelLocation.Y + size / 2),
                new PointF(pixelLocation.X + size / 2, pixelLocation.Y + size / 2)
            });
        }
    }
```

To be used like this:
```cs
            plt.MarkerDrawer(new LuchtGemaskeerdRechts());
```

image


swharden:
Hi @henkburgstra, thank you for this insightful suggestion! I really like how this PR uses interfaces to let the user create a custom marker. This seems like a really useful feature that could benefit many plottable objects.

However, this PR only addresses scatter plots. Rather than merge this PR, I added this topic to the [ScottPlot roadmap](https://github.com/swharden/ScottPlot/blob/master/dev/roadmap.md) because I think the whole library would benefit from refactoring away from the enum-based marker system to a class-based system with interfaces like the one you implemented here.

This is a really good idea, and it pains mt to close this PR, but I'd really like to implement this idea on a much wider scale. I'll ping you when I do so you can take a look and let me know what you think!

henkburgstra:
Hi Scott, thank you very much for your thoughtful response!

Could I maybe help you with this? Thank you for keeping me updated, I would rather use the original ScottPlot library than my fork in our project :)

swharden:
> Could I maybe help you with this? Thank you for keeping me updated, I would rather use the original ScottPlot library than my fork in our project :)

Since the current marker system is enum based, markers are hard-coded into the library. The fastest way could be for me to add your custom markers to the enum-based system so you can use them immediately, then I work toward supporting custom markers in a future release.

To add a custom marker you could just add its name to the list of current markers.

https://github.com/swharden/ScottPlot/blob/ef59756715ac87c6d778fe783c81357a593561b2/src/ScottPlot/Markers.cs#L10-L27

Then add a case for that marker in the `DrawMarker` function (using the arguments that are passed in for location, size, and color)

https://github.com/swharden/ScottPlot/blob/ef59756715ac87c6d778fe783c81357a593561b2/src/ScottPlot/Markers.cs#L44-L46

You could either paste that code in this issue, or change the code and make a PR, and I can incorporate it into the main library pretty quick.

henkburgstra:
Hi Scott,

Many thanks for taking the time to respond. I apologize for my late reply.

I think I will do as you suggested for now. Hopefully I can make a pull
request later today or tomorrow.

Thanks again,
Henk.

PR #413, opened 2020-05-21
Allows enforcing a dedicated GPU from user code.

Benny121221:

**Purpose:**
This allows user code to enforce the higher-performance GPU profile on systems with nVidia GPUs. This can also be accomplished in the windows settings. I cannot yet test what will happen in a computer without an nVidia GPU. I believe it will not change the GPU profile and will continue as normal. I am unsure if this is necessary functionality for the library per se, as the `DedicatedGPU` class could easily be implemented by whomever is using the library. Perhaps this makes more sense in a demo?

Also, the implementation is a tad jank.

**New functionality (code):**
All one has to do is call `ScottPlot.DedicatedGPU.RunWithDedicatedGraphics()` *before* any window is opened. For WPF this means in the constructor before `InitializeComponent()` is invoked.

image
On my laptop GPU1 is the higher-performance graphics accelerator (an nVidia MX150). GPU0 is the battery-saving accelerator (integrated Intel UHD620)


Benny121221:
I have an old PC with a Radeon card in, I can try it on that. It might take some time because currently that computer is bricked -_-

It shouldn't work, but it also shouldn't cause any issues.

Benny121221:
I tested it on Windows 10 machine with Intel HD2500 integrated graphics and a Radeon R7 240. I had no issues. No matter whether I was trying to force dedicated graphics or not, it would not use either GPU. I suspect that both are so old that they would not be used for GPU acceleration. Or my drivers are messed, like everything else on that PC.

For comparison, the behaviour on my laptop is to use the integrated graphics for acceleration unless forced to use the dedicated graphics. I was unable to coerce my laptop into running in software rendering mode.

At the very least we know it works on a non-nvidia machine. It looks like doing the work to enforce a performance mode on AMD is significantly more difficult.

swharden:
Hey @Benny121221, this looks pretty interesting! I have a basic question though - does this actually improve performance of ScottPlot?

I haven't studied this topic in detail yet, but my gut reaction is this tweak controls rendering options for things like DirectX and OpenGL, which ScottPlot doesn't actually use.

It's been on my to-do list to create a GUI benchmark test, so maybe now's a good time for me to do that so we can see how much performance we can gain using this method.

Benny121221:
On my computer, probably not. An MX150 is not much stronger than a UHD620. Plus, I suspect that ScottPlot would benefit more fr dedicated video ram than the actual video card. And an MX150 uses system RAM.

swharden:
I made an automated testing/benchmark program (in /dev/Benchmark) to help with this

image

I tested this PR by editing `Program.cs`:

```cs
/// <summary>
/// The main entry point for the application.
/// </summary>
[STAThread]
static void Main()
{
    ScottPlot.DedicatedGPU.RunWithDedicatedGraphics();
    Application.EnableVisualStyles();
    Application.SetCompatibleTextRenderingDefault(false);
    Application.Run(new FormMain());
}
```

I ran the benchmark full-screen, anti-aliased, with parallel processing.

* FPS without dedicated graphics:  `26.11`, `26.44`, `26.27` (`25.273 ± .165`)
* FPS with dedicated graphics: `25.64`, `25.74`, `25.83` (`25.737 ± .095`)

This implies a very slight (~1%) but real (`p=0.0082`) performance enhancement 🤔

StendProg:
This is a very light load to reveal benefits.
I think `Scatter` with `WhiteNoise` data make havy load on drawings, but not on CPU. (Signal and Signal Const never draw more than 2000-5000 lines, RandomWalk also bad candidate because each line small, GDI+ like it).
To exclude everything else, you need to select a `Scatter` size so that it takes seconds to draw.

Anyway i dislike this PR. It brings platform specific code (Windows only, nvidia only). But in fact, it brings a small helper method that weakly relates directly to plotting.
It would be possible to arrange it like a gist somewhere in the project, because there are those to whom this will simplify the task. But adding everything to the code base (especially platform specific) is especially weakly related to the library, this can quickly become a problem (NVIDIA can change API tomorrow, and you must support this changes in library). If you have hundreds of such decisions accumulated, you will only be able to update them.

swharden:
> This is a very light load to reveal benefits.
> I think `Scatter` with `WhiteNoise` data make havy load on drawings, but not on CPU. (Signal and Signal Const never draw more than 2000-5000 lines, RandomWalk also bad candidate because each line small, GDI+ like it).
> To exclude everything else, you need to select a `Scatter` size so that it takes seconds to draw.

Here's a scatter plot with random Xs and Ys

image

* without this PR: `11.35 FPS`
* with this PR: `11.44 FPS`

> Anyway i dislike this PR. It brings platform specific code (Windows only, nvidia only). But in fact, it brings a small helper method that weakly relates directly to plotting.

I agree with this view. The code works, but it's not related to plotting data. It's an OS-specific and hardware-specific implementation detail

> It would be possible to arrange it like a gist somewhere in the project, because there are those to whom this will simplify the task. But adding everything to the code base (especially platform specific) is especially weakly related to the library, this can quickly become a problem (NVIDIA can change API tomorrow, and you must support this changes in library). If you have hundreds of such decisions accumulated, you will only be able to update them.

I'l close this PR without merging it, but I'm glad I became aware of this topic! Thanks for mentioning it @Benny121221. This PR led to the creation of a useful benchmarking app.

PR #415, opened 2020-05-22
Added higlightable scatter plot

Benny121221:

**Purpose:**
https://github.com/swharden/ScottPlot/issues/414

**New functionality (code):**
There is a demo, but the gist of it is this:

```cs
PlottableScatterHighlight highlightPlot = plt.PlotScatterHighlight(x, rand);
highlightPlot.HighlightPoint(4);
highlightPlot.HighlightPointNearest(8);
highlightPlot.HighlightPointNearest(5, 1);
//highlightPlot.HighlightClear();
```

image


swharden:
haha, wow you're so fast @Benny121221! This is pretty cool. Nice LINQ in there too.

Don't throw these ideas in just because I suggested them - I'm more asking what you think about these potential other features... what about a flag to turn on/off drawing the Y value of the point on the image, similar in appearance (and perhaps with similar options) as the annotation plottable? Also, what do you think about a flag to enable a "crosshair" mode which draws lines (perhaps with similar options as HLine and VLine) to the highlghted point? This could make it feel really interactive when I refactor the code to make it a little easier to show values under the mouse cursor.

Looking forward to hearing your thoughts!

Benny121221:
Do you mean this thing: 
image

I don't know that that would make much sense for this plottable, but I could see value in a method to return the point closest to a specified coordinate pair? And I think its a similar story for the crosshair mode. I think those both make the most sense in a scenario where you want to snap the mouse to the nearest point. It would look quite messy to have 12 annotations and 24 lines drawn because the user decided to highlight points because they are outliers.

I think it could get messy to be calling other plottables from this one. And if you don't create it as a plottable (i.e. just using the drawing library) it becomes less easy to customize, because you can't expose the same API you use for drawing normal VLines to the user.

I think the best option is to add a function to make it easier to get the closest point to a set of coordinates. This also makes it trivial to merge the two scatter plottables

swharden:
> I think the best option is to add a function to make it easier to get the closest point to a set of coordinates. This also makes it trivial to merge the two scatter plottables

I like this conclusion. We can stick with smaller, simpler pieces that snap together. This also makes me think `crosshair` (essentially a VLine and HLine combined) might be a useful plot type 🤔 I'll keep this in mind!

Thanks again for your work on this @Benny121221! I'll work this a bit more over the weekend.

Benny121221:
No problem. I probably can't work on this tomorrow because I have a test, but on the weekend I can help out.

Benny121221:
Also, if there's a crosshair plottable, I think it is necessary that there is an easter egg which lets you quickscope the graph

swharden:
> quickscope the graph

XDXDXD

swharden:
> > quickscope the graph
> 
> XDXDXD

counter-terrorists win

StendProg:
Hi there.
This PR contains interesting idea, and good implementation.
I have a few comments on the code and want to review if you don't mind.

First of all, it's better to make `PlottableScatterHighlight` dirived from `PlottableScatter` to remove a lots of copy/pasted code(`GetLimits()`, `GetCSV()`, Scatter constructor etc).
If you go further, you can put marker rendering code into a separate virtual function in `PlottableScatter`. This will prevent you from copying the `Render()` too.

Good idea to declare `IHighlightable` interface with all this `HighlightPointNearest() ...`  methods.

All lambdas like: 
```c#
points.Select((p, i) => (p.x, p.y, i)).OrderBy(p => pointDistance(p.x, p.y)).First().i
``` 
looks cool, but not most effective way to find minimum value.
Using sort you get result in `O(N*log(N))`.
Using simple minimum search you will get the same result for `O(N)`.
There is no way to find minimum index using lambdas in .NetFramework. (Its possible in MoreLinq, but it's external package).
Finding index using `for` loop, looks less elegant, but its few lines of code and simple too.

src/ScottPlot/plottables/PlottableScatterHighlight.cs
```diff
@@ -0,0 +1,288 @@
...
+        public (double, double) GetPointNearest(double x) { 
+            int index = xs.Select((p, i) => (p, i)).OrderBy(p => Math.Abs(p.p - x)).First().i;
+            return (xs[index], ys[index]);
+        }
+
+        public (double, double) GetPointNearest(double x, double y)
+        {
+            List<(double x, double y)> points = xs.Zip(ys, (first, second) => (first, second)).ToList();
+
+            double pointDistance(double x1, double y1) => Math.Sqrt((x1 - x) * (x1 - x) + (y1 - y) * (y1 - y));
```
StendProg, line 270: Code duplicate, maybe move it to private field or method?

src/ScottPlot/plottables/PlottableScatterHighlight.cs
```diff
@@ -0,0 +1,288 @@
...
+                else
+                {
+                    settings.gfxData.DrawLines(penLine, points.ToArray());
+                }
+            }
+
+            // draw a marker at each point
+            if ((markerSize > 0) && (markerShape != MarkerShape.none))
+                for (int i = 0; i < points.Count; i++)
+                    if (highlightedIndexes.Contains(i))
```
StendProg, line 211: For dozens of highlighted points this `Contains()` work fast. Thousands of `highlilghted` points will significant drop performance because this `Contains()`. 
It's possible to make this check fast if `highlightedindexes` stored sorted. But it's hard to implement and i think don't really need, so:

this is just a note.

src/ScottPlot/plottables/PlottableScatterHighlight.cs
```diff
@@ -0,0 +1,288 @@
...
+        {
+            List<(double x, double y)> points = xs.Zip(ys, (first, second) => (first, second)).ToList();
+
+            double pointDistance(double x1, double y1) => Math.Sqrt((x1 - x) * (x1 - x) + (y1 - y) * (y1 - y));
+
+            int index = points.Select((p, i) => (p.x, p.y, i)).OrderBy(p => pointDistance(p.x, p.y)).First().i; // I feel like I might have gone too Haskel-ey
+            HighlightPoint(index);
+        }
+
+        public (double, double) GetPointNearest(double x) {
```
StendProg, line 261: public `(double xn, double yn)` GetPointNearest looks better and intellisense grab this tuple names.

src/ScottPlot/plottables/PlottableScatterHighlight.cs
```diff
@@ -0,0 +1,288 @@
...
+            int index = points.Select((p, i) => (p.x, p.y, i)).OrderBy(p => pointDistance(p.x, p.y)).First().i; // I feel like I might have gone too Haskel-ey
+            HighlightPoint(index);
+        }
+
+        public (double, double) GetPointNearest(double x) { 
+            int index = xs.Select((p, i) => (p, i)).OrderBy(p => Math.Abs(p.p - x)).First().i;
+            return (xs[index], ys[index]);
+        }
+
+        public (double, double) GetPointNearest(double x, double y)
```
StendProg, line 266: public `(double xn, double xn)` GetPointNearest

Benny121221:
> First of all, it's better to make PlottableScatterHighlight dirived from PlottableScatter to remove a lots of copy/pasted code(GetLimits(), GetCSV(), Scatter constructor etc).
If you go further, you can put marker rendering code into a separate virtual function in PlottableScatter. This will prevent you from copying the Render() too.

@StendProg I agree, although I think it makes even more sense for the two scatterplots to be merged. For now I'll work on your suggestions.

src/ScottPlot/plottables/PlottableScatterHighlight.cs
Benny121221: I decided to sort it, it wasn't difficult

src/ScottPlot/plottables/PlottableScatterHighlight.cs
```diff
@@ -0,0 +1,288 @@
...
+
+        public void HighlightPointNearest(double x)
+        {
+            int index = xs.Select((p, i) => (p, i)).OrderBy(p => Math.Abs(p.p - x)).First().i;
+            HighlightPoint(index);
+        }
+
+        public void HighlightPointNearest(double x, double y)
+        {
+            List<(double x, double y)> points = xs.Zip(ys, (first, second) => (first, second)).ToList();
```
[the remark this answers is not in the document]
Benny121221: Since I changed it to a for loop you do. You cannot index an IEnumerable, you can only enumerate it. Since the index was important I kept the `.ToList()`

Benny121221:
I like the changes, PlottableScatterHighlight.cs is much shorter now. The only thing is the constructor signature is a tad long
https://github.com/Benny121221/ScottPlot/blob/18a56600ecacffa994d50bb5ab8e3b0d8aff7381/src/ScottPlot/plottables/PlottableScatterHighlight.cs#L22

swharden:
> I like the changes, PlottableScatterHighlight.cs is much shorter now. The only thing is the constructor signature is a tad long

They are large by C# standards, but it's consistent with the style of other ScottPlot plottables so I probably wouldn't worry too much about it.

ScottPlot's API uses command-like methods to create plots, and optional arguments to customize them. This style results in methods (and constructors) with lots of arguments. I think methods with lots of arguments is considered a code smell .NET land... but interestingly it is considered Pythonic because that language has some advanced "kwarg" features. Since I try to style the ScottPlot API after matplotlib (a Python library that makes heavy use of kwargs), this tilts toward methods with lots of optional arguments. 

It wasn't until recently that I realized how uncommon this style is in the .NET world. A MVC-style (requiring the user to compose plot objects, style them by setting properties, then pass them into a plot viewer) would be more .NETish, but why I like ScottPlot is that it's much more familiar and easier to pick up by people who haven't used .NET/MVC before, and modeling its API after matplotlib is how I keep it easy to use for newcomers to the language (many data scientists, like me, come from Python backgrounds).

That's a longer response than it needed to be, but I wanted to put it out there for context and perhaps see if @StendProg has any comments, as @StendProg has a lot more .NET experience than I do

Benny121221:
My observation about the constructor signature was more a comment on the `: base(args....)` than it was about the parameter list. I have a Java background, so that syntax for constructor overloading is uncomfortable. In Java it would be part of the constructor body and it would look like `super(args...)`. The base constructor call technically doesn't technically count as part of the method header. It's borrowed from the C++ member initializer list syntax (which until a recent revision was not used for constructor overloading) and it's not considered part of the header.

I think the most .NET-ic solution is to create a settings struct which carries all of the parameters. Something like: `var scatter = plt.PlotScatter(settings);`

Keep in mind most of my .NET experience is from web development, where the above solution is preferred, as the client is sending a JSON object already. Although .NET supports binding JSON to a list of arguments, it's simpler to just bind it to a class or struct, especially as there is no argument list exposed to the client.

I normally don't like the *pythonic* way, but I think here it is more clear, if perhaps more verbose. Of course, the enterprise solution is to use a `PlotSettingsManagerServiceFactory` to instantiate a `PlotSettingsManagerService` which sorts it all out for us.

StendProg:
@Benny121221 .
Binary search do the trick.
Additionally  you can optimize
```c#
public void HighlightPoint(int index)
{
    highlightedIndexes.Add(index);
    highlightedIndexes.Sort();
}
```
if you insert index value each time in right place of List, you keep it sorted without manual `Sort` call.

another microoptimization.
```c#
minDistance = value[0];
for ( int i = 1; ....)
```

You replace spaces with tabs in last commit. And have problem with formating in `PlottableScatterHighlight.cs`. Please fix it.

All others looks perfect, good job.

Benny121221:
I got you. The only thing is this method is written a little unintuitively now: https://github.com/swharden/ScottPlot/blob/f4550c8ed2e21eee961a52187dcc1448f692e6f2/src/ScottPlot/plottables/PlottableScatterHighlight.cs#L49:L60

I added a comment explaining how it works because I would normally assume that `List<T>.BinarySearch()` would return -1 if it doesn't find the value. Plus, most people who aren't systems programmers are made uncomfortable by bitwise operators in general.

StendProg:
@swharden 
I thinked a lot on this library API.
In fact, it is very friendly to beginners. And it also attracted me in due time.
And even having experience working with the library, it’s very convenient to set about 5  parameters and do what you need with one line of code. As a result, the library is very convenient from the point of view of the user (IMHO).

But from the point of view of the developer, supporting this API quickly steps into hell. Methods with dozen of parameters that need to be thrown through several function calls. Besides being annoying, it is also a potential source of various errors.
The standard technique for refactoring functions with a large number of parameters is to combine the parameters into a structure and pass it as a parameter. But from the point of view of the user, you need to create a structure, fill in its fields, for this you need to go into the documentation to read what kind of fields and how to fill them out. Many may choose to go look for another library.

As a result, I think that the upper level of the API(user level) should remain as it is, otherwise it will be a completely different library.
At developer level (not visible to user) you can move away from strict rules, and transmit data through fields, you can even use DI and factories, etc., just don’t rush to redo everything, but constantly think about this problem and apply it only in the new code.

swharden:
> it’s very convenient to set about 5 parameters and do what you need with one line of code

> from the point of view of the developer, supporting this API quickly steps into hell

This is a very good description of the advantage and disadvantage of this design 😅 

> I think that the upper level of the API (user level) should remain as it is

I agree

> constantly think about this problem and apply it only in the new code

What do you think about an intermediate strategy that seeks the best from both worlds? The public API can remain as it is (lots of simple arguments), but inside those methods we developers can start to use a "style object" model to create Plottables, simplifying the constructors and making errors less likely.

After we make a few style objects, we could slowly refactor toward this model one Plottable at a time, and the public API never changes. This may slowly make the existing code base support easier for the developer to support 🤔

StendProg:
@Benny121221 
I can't commit directly to this PR, so I make PR to branch of you fork with Unit tests for this new plottable, please check it.

swharden:
I updated Azure Pipelines to fail the build if the autoformatter wasn't run. 

You can run the autoformatter in visual studio (ALT+K, ALT+D) or by installing a command line tool and running that.

```
dotnet format -w src/ScottPlotV4.sln
```

My hope is this will make it easier to compare "files changed" of PRs, because right now things like tabs vs. spaces and indentation size make it appear like every line in a file has been modified.

_I'm still working to make format errors more descriptive. Right now it shows the build failing on every platform (Linux, MacOS, Windows) but instead I want to have a 4th platform that's just for format checking. This way we can confirm the actual code runs and tests pass, and it's more obvious that the build failed for formatting reasons. Right now this requires digging into the build messages to figure out..._

swharden:
Following-up on the previous comment, I extracted the format checker to its own check now, but it's failing because of a configuration issue. Hopefully I'll have it working properly before much longer...

Benny121221:
I think a .editorconfig file might also be a good idea...

https://docs.microsoft.com/en-us/visualstudio/ide/create-portable-custom-editor-options?view=vs-2019

Most of our massive diffs come from me using tabs and forgetting to click the thingy to switch to spaces. The file is supposed to be able to convert my tabs to 4 (or whatever amount) spaces

Benny121221:
Also your build is failing because you're publishing artifacts from the formatchecker, and it doesn't like having duplicates. https://dev.azure.com/swharden/swharden/_build/results?buildId=1378&view=logs&j=b58c4c1a-41b4-5e9f-fca9-6926815bf697&t=aac76872-f903-5b0b-877a-47c8b79dee7f

swharden:
> I think a .editorconfig file might also be a good idea

I recently added one to /src (and it's loaded in the .sln) (#425) so I hope it helps

> Also your build is failing because...

I'm working to refactor the azure pipelines strategy matrix and pool to try to make this simpler... this YML file is a nightmare 💀 

https://github.com/swharden/ScottPlot/blob/28b0fece44c182367f71d58fb500bee50b1537ba/dev/devops/azure-pipelines.yml#L4-L16

swharden:
Finally got Azure Pipelines to do what I wanted... and it took way longer than I expected 😝 

I still can't figure out how to set the displayName of the jobs created by the strategy matrix in [azure-pipelines.yml](https://github.com/swharden/ScottPlot/blob/master/dev/devops/azure-pipelines.yml), as I'd really like to remove the word "Job" from the labels.

It looks like this PR build is currently failing because:
```
  ScottPlot/Plot.cs(633,42): Fix whitespace formatting.
  ScottPlot/Plot.cs(637,42): Fix whitespace formatting.
  ScottPlot/Plot.cs(641,47): Fix whitespace formatting.

  ScottPlot/plottables/PlottableScatterHighlight.cs(24,73): Fix whitespace formatting.
  ScottPlot/plottables/PlottableScatterHighlight.cs(52,36): Fix whitespace formatting.
  ScottPlot/plottables/PlottableScatterHighlight.cs(53,50): Fix whitespace formatting.
  ScottPlot/plottables/PlottableScatterHighlight.cs(53,86): Fix whitespace formatting.
  ScottPlot/plottables/PlottableScatterHighlight.cs(54,180): Fix whitespace formatting.
  ScottPlot/plottables/PlottableScatterHighlight.cs(55,92): Fix whitespace formatting.
  ScottPlot/plottables/PlottableScatterHighlight.cs(56,129): Fix whitespace formatting.
  ScottPlot/plottables/PlottableScatterHighlight.cs(57,221): Fix whitespace formatting.
  ScottPlot/plottables/PlottableScatterHighlight.cs(62,51): Fix whitespace formatting.
  ScottPlot/plottables/PlottableScatterHighlight.cs(107,62): Fix whitespace formatting.
  ScottPlot/plottables/PlottableScatterHighlight.cs(107,64): Fix whitespace formatting.
```

swharden:
Hey @Benny121221 sorry I made this PR more complicated by editing the azure pipelines file. This was a rookie mistake as I forgot I was still on your branch 🙈. 

I'm going to step away from the computer for a bit, come back refreshed, and review the content of this PR. I see you incorporated @StendProg's suggestions. Do you think this one is ready to merge into master now?

Benny121221:
@swharden I think its ready, at least as soon as I move the demo from experimental

src/ScottPlot/plottables/PlottableScatterHighlight.cs
```diff
@@ -0,0 +1,118 @@
...
+            int insertPosition = highlightedIndexes.BinarySearch(index);
+            if (insertPosition < 0) {
+                insertPosition = ~insertPosition;   //This might look like witchcraft
+                                                    //However, List<T>.BinarySearch returns the ones-compliment (bitwise inverse) of the index of the next-higher value in the list
+                                                    //if the value is not part of the list.
+                                                    //If there is no higher value it returns the ones-compliment of this.Count()
+                                                    //Source: https://docs.microsoft.com/en-us/dotnet/api/system.collections.generic.list-1.binarysearch?view=netcore-3.1#System_Collections_Generic_List_1_BinarySearch__0_
+            }
+            highlightedIndexes.Insert(insertPosition, index);
+        }
```
swharden, line 60: I'm a little confused here. Not about the bit-inverse binary search result (which is cool! I didn't know that), but about what the intent of this function is. 

Currently if you do this:

```cs
highlightPlot.HighlightPointNearest(8);
highlightPlot.HighlightPointNearest(8);
highlightPlot.HighlightPointNearest(8);
highlightPlot.HighlightPointNearest(8);
highlightPlot.HighlightPointNearest(8);
```

It will keep adding the same index to the `highlightedIndexes` list. Is this the intended behavior? 

Perhaps this method could be simplified to:

```cs
public void HighlightPoint(int index)
{
    int searchResult = highlightedIndexes.BinarySearch(index);

    if (searchResult < 0)
        highlightedIndexes.Insert(~searchResult, index);
}
```

or even simpler:

```cs
public void HighlightPoint(int index)
{
    if (highlightedIndexes.Contains(index))
        highlightedIndexes.Add(index);
}
```

src/ScottPlot/plottables/PlottableScatterHighlight.cs
```diff
@@ -0,0 +1,118 @@
...
+
+        private int GetPointNearestIndex(double x) {
```
swharden, line 62: I think this method name is trying to be `GetIndexNearestPoint()`, or `GetIndexNearestX()`. The second option would fix the overload issue that would arise if you create a `GetIndexNearestY()` too (do you think we should have one?)

src/ScottPlot/plottables/PlottableScatterHighlight.cs
swharden: Is the motivation that BinarySearch is faster on ordered lists, and it's important because it's called frequently (in the render loop)?

If performance is the concern, maybe a `bool[] isHighlighted` the same length as `double[] xs` is an alternative way to keep track of this. Then the render loop would require no searching at all. It's a little more memory, but not much.

src/ScottPlot/plottables/PlottableScatterHighlight.cs
Benny121221: I don't know how useful `GetIndexNearestY()` is, but I added it. This is mostly for following the mouse cursor around, and I think most users would prefer to snap to either the closest point, or the closest point to the x span of the cursor. However, it's entirely possible that someone could find a use for it outside of mouse controls

Benny121221:
Ah, now I remember why I was using a list of indices. It makes it very simple to add/remove points

src/ScottPlot/plottables/PlottableScatterHighlight.cs
swharden: Sounds good! A use case might be finding the nearest bar graph to the cursor when horizontal bars are used if later we make bar graphs `IHighlightable`

https://swharden.com/scottplot/cookbooks/4.0.32/#plottypes-bar-horizontal-bar-graph

Now seems like the best time to go ahead and add X and Y methods to `IHighlightable`

swharden:
I made a mouse interaction demo here

image

image

I refactored this module a bit to return tuples, making it easier to get x/y/index information back out of the method that finds the nearest point

```cs
private void formsPlot1_MouseMove(object sender, MouseEventArgs e)
{
    double mouseX = formsPlot1.plt.CoordinateFromPixelX(e.Location.X);
    double mouseY = formsPlot1.plt.CoordinateFromPixelY(e.Location.Y);

    sph.HighlightClear();
    var (x, y, index) = sph.HighlightPointNearest(mouseX, mouseY);
    formsPlot1.Render(skipIfCurrentlyRendering: true);

    label1.Text = $"Closest point to ({mouseX:N2}, {mouseY:N2}) " +
        $"is index {index} ({x:N2}, {y:N2})";
}
```

This is a vastly simpler "show value on hover" solution!

swharden:
@StendProg thank you for making these thorough tests! They revealed an uncommon but fundamental issue with scatter plots: `xs` and `ys` must always be the same length, but this is not enforced. Same for `errorX` and `errorY` if not `null`. The problem is when the length of the array changes. Users can manually modify these public properties to cause the scatter plot to enter a bad state. With ScatterHighlight, this is also true with the `isHighlightable` array.

One solution may be to make all the properties `{get; private set;}` and create methods like `Update(newXs, newYs)` which handle the new array creation and force constraints, but this would be a breaking change.

Another idea I've been considering is something like a `ScatterConst()` which uses `readonly double[]` (valid state checked at initialization), and something like `ScatterMutable()` which uses `List<double>` to make it easier for data we know will resize (valid state checked on every render).

As far as this PR goes, I think we've gotten about as good as we can with the base class we have.

StendProg:
Simpiest way to fix it, is copy xs ys arrays to temp arrays at the begining of Render loop. Check them for length equality and work with temp arrays only in render loop. Copy means only ref copy like tempxs=xs. The same way, you need to do in all methods that works correct only on equal length.
You need a lot more complex solutions if you want to achive true thread safe result. Does it really needed? Maybe make user contract that scatter support only equal arrays and remove all check will be better solution?

Benny121221:
There were some merge conflicts but I rebased. Everything seems dandy now

PR #423, opened 2020-05-23
Pie chart

Benny121221:

**Purpose:**
Adds a pie chart plottable.

**New functionality (code):**

```cs
double[] proportions = new double[5];
double total = 0;
for (int i = 0; i < proportions.Length; i++)
{
	if (total >= 1) {
		break;
	}
	proportions[i] = rand.NextDouble() * (1 - total);   //Make sure we don't add up to more than 1
	total += proportions[i];
}
if (total < 1) {
	proportions[proportions.Length - 1] += 1 - total; //If we are less than 1, add the remnant to the last slice
}

plt.PlotPie(proportions);
plt.Legend();
```

image

### DIFF
--- a/dev/autoformat.bat
+++ b/dev/autoformat.bat
@@ -1,0 +1,3 @@
+cd ..\src
+dotnet format
+pause
--- a/dev/changelog.md
+++ b/dev/changelog.md
@@ -1,6 +1,7 @@
 # ScottPlot Changelog
 
 ## ScottPlot 4.0.35
+* Pie charts can now be made with `plt.PlotPie()`. See cookbook and demo application for examples. (#421, #423) _Thanks @Benny121221_
 * `ScottPlot.FormsPlotViewer(Plot)` no longer resets the new window's plot to the default style (#416)  _Thanks @StendProg_
 
 ## ScottPlot 4.0.34
--- a/src/ScottPlot.Demo/PlotTypes/Pie.cs
+++ b/src/ScottPlot.Demo/PlotTypes/Pie.cs
@@ -1,5 +1,6 @@
 ﻿using System;
 using System.Collections.Generic;
+using System.Linq;
 using System.Text;
 
 namespace ScottPlot.Demo.PlotTypes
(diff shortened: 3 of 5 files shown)


Benny121221:
Not to point the finger, but that build failed because of a file I had no clue existed, so I'm pretty sure it's not because of this PR

swharden:
> Not to point the finger, but that build failed because of a file I had no clue existed, so I'm pretty sure it's not because of this PR

Ha, that was me. I added a new project to the main solution tonight and I guess a path got messed up.  I think it's fixed now so if you merge maser into this branch it should build properly.

This is excellent progress! What do you think about an "exploded" pie chart? gaps between the slices? And also labels and/or values somewhere on the plot? I'm thinking the demo would look best if the grid and axes were disabled too

```cs
plt.Grid(false);
plt.Frame(false);
plt.Ticks(false, false);
```

Benny121221:
I rebased, lets see if it works.

And I can have a look at those suggestions.

swharden:
> I rebased, lets see if it works.

Success!

> And I can have a look at those suggestions.

I recognize the "exploding" effect isn't trivial to implement... but it would look pretty cool, and I look forward to seeing what you come up with!

Benny121221:
How's this for an exploding effect?

image

swharden:
It's looking close! I'm a bit confused why/how some pie slice tips are touching each other while some others are not 🤔

Benny121221:
Because they're small slices, the midpoint of their arc are in approximately the same spot. Which means when you explode the pie they move in the same direction.

Excel avoids this problem by adding an outline to the slices:
image

swharden:
Heh, Excel takes the easy path 😝

I haven't looked it up yet so I don't know the answer, but what does matplotlib do? My go-to is usually to mimic their behavior.

Benny121221:
I'm not sure about matplotlibs behaviour. How does this one look?

image

Benny121221:
And I'm not sure we want to emulate matplotlib in this situation...
image

Benny121221:
matplotlib does this:

image

It doesn't draw an outline around the whole thing, which means if one of the slices is large, their edges don't line up:

image

Otherwise, it seems to act the same as excel. I can turn off the outline, but I think it looks better

swharden:
This is an excellent improvement to this libraries capabilities. Thanks @Benny121221!

Issue #420, opened 2020-05-23
Heatmaps (Categorized and Continuous)


Benny121221:
**What can we do to make ScottPlot better?**
Perhaps a new plottable for heatmaps.

Including continuous heatmaps: 
![image](https://user-images.githubusercontent.com/8635304/82717485-9f7ccc80-9c59-11ea-9058-2228855dfd03.png)
As well as categorical heamaps:
![image](https://user-images.githubusercontent.com/8635304/82717491-ab688e80-9c59-11ea-9dfa-30688aeb89b9.png)

```cs
var heatmap = plt.PlotHeatmap(xs, ys, values, colourmap);
```

My experience with these plottables primarily comes from matplotlib.pyplot, both are implemented with `imshow`. However, in my opinion, categorical heatmaps are just tables with added colours, so perhaps continuous heatmaps ought to be their own plottable, and a new table plottable is created, which can optionally take a colourmap argument?

Check the ScottPlot Roadmap to see if your feature is already planned for an upcoming release:
https://github.com/swharden/ScottPlot/blob/master/dev/roadmap.md

swharden:
@Benny121221 I'm really interested in this idea! I would definitely like to achieve this because right now I have a spectrogram library https://github.com/swharden/Spectrogram that doesn't have axis labels or interactivity, and supporting heatmaps in ScottPlot would make tasks like this achievable. Here are a few thoughts...

* I'm not entirely sure the best way to draw this
  * System.Drawing doesn't always render rectangles perfectly if they touch edge to edge. With anti-aliasing on you can see through them a bit.
  * Maybe the solution here is to render a totally new `Bitmap`, populate pixels based on the colormap, then the plottable could `Graphics.DrawImage()` to render that image onto the plot?
  * Optimal rendering might involve marshaled memory, and you may find some useful code there in my spectrogram demo 
* The scalebar could be a separate plottable, related to #388
* Let's support multiple colormaps with viridis being the default https://www.youtube.com/watch?v=xAoljeRJ3lU (good talk!)

Due to all the stuff I'm working on (#412) it might take me a while to move on this... do you want to give this one a shot? As long as we can keep it in the "experimental" category we could make it a real plottable, include it in the demo program, and incrementally improve upon it over time

Benny121221:
[...]

PR #424, opened 2020-05-23
Beginnings of heatmaps

Benny121221:
**Purpose:**
New heatmap plottable

**New functionality (code):**

```cs
int[] xs = Enumerable.Range(0, 100).ToArray();
int[] ys = Enumerable.Range(0, 100).ToArray();
double[][] intensities = new double[ys.Length][];

for (int i = 0; i < ys.Length; i++) {
	intensities[i] = new double[xs.Length];
	for (int j = 0; j < xs.Length; j++) {
		intensities[i][j] = Math.Sqrt(Math.Pow(xs[j] - 50, 2) + Math.Pow(ys[i] - 50, 2));
	}
}

plt.PlotHeatmap(intensities);
```

image

Currently the only supported colourmap is grayscale, and the scale is worthless. I was going to continue working on it but I've got some stuff to take care of. I should be back to finish it up, but I thought I'd open the PR now in case you want to have a crack at it.


Benny121221:
#420 (Nice)

Benny121221:
I rebased, so theoretically we should have all the .editorconfig goodness

Benny121221:
image

I think it looks pretty good? I think after I add a couple more colourmaps the only thing left to do is sort out axes

Benny121221:
@swharden We now have six colourmaps (although grayscale/grayscale inverted perhaps counts as only one). I haven't looked at any of matplotlibs other colourmaps but some of them may be under a friendly license. I would be a tad interested in adding some (in)famous ones like jet, but I'm again not sure about the licensing.

On the repo you linked you have two colourmaps I don't recognize: https://github.com/swharden/Spectrogram/blob/master/src/Spectrogram/Colormaps/VdBlues.cs
and https://github.com/swharden/Spectrogram/blob/master/src/Spectrogram/Colormaps/VdGreens.cs

Are these your own? Or are they under a generous license? I know you said adding heatmaps to this library would make your spectogram simpler.

swharden:
@Benny121221 this is looking so good!

> I rebased, so theoretically we should have all the .editorconfig goodness

Does Ctrl+K, Ctrl+D autoformat the document for you?

> Google's Turbo is under Apache 2.0 (discussed in #420, nice)

I'm not a lawyer, but since we are distributing a derivative work (not the unmodified original work) it looks like it's okay to include in a MIT project: https://en.wikipedia.org/wiki/Apache_License#Licensing_conditions

This matplotlib discussion https://github.com/matplotlib/matplotlib/issues/15091 had some interesting points about turbo.

> VdBlues and VdGreens

There are simple math-based colormaps I "invented" around 2010 when I made https://github.com/swharden/QRSS-VD (one of my first programming projects, the code is awful). It relates to radio frequency experiments I work on like https://github.com/swharden/QRSS-hardware and https://github.com/swharden/QRSS-hardware/tree/master/spots

TLDR; they're mine, but matplotlib's colormaps are way better

Benny121221:
> Does Ctrl+K, Ctrl+D autoformat the document for you?

Yes but it still uses tabs. If I press tab it gets replaced with spaces, but autoformat doesn't care.

> I'm not a lawyer, but since we are distributing a derivative work (not the unmodified original work) it looks like it's okay to include in a MIT project:

Great, I'll add it

> TLDR; they're mine, but matplotlib's colormaps are way better

Ok, I'll leave them be for now

Benny121221:
# Update
Currently, it looks like this:
image

# Additions
I think it's in a pretty good spot, the only thing that seems lacking is the axes, I feel like it could be useful to show a bit more granularity. The easiest way to do that is just to darken the grid, which looks like this: 


image

Let me know what you think. Just darken the grid, or put some more work into the axes.

swharden:
> > Does Ctrl+K, Ctrl+D autoformat the document for you?
> 
> Yes but it still uses tabs. If I press tab it gets replaced with spaces, but autoformat doesn't care.

Hm, I guess visual studio is ignoring the format file. In a command window you can 

```
cd C:\path\to\ScottPlot\src
dotnet format
```

> Currently, it looks like this

This is looking so good! I really love the colormap scalebar on the right. Don't put too much more time into labeling the corners. I think what should happen here is I should figure out a way to resize the data area to fit the heatmap that way we can use the built-in scalebars. I need to refactor the layout system which is I suggest you don't worry too much about figuring that out because it's a bit thorny and likely to change soon.

In the mean time what do you think about figuring out how to make the heatmap stretch to fill 100% of the viewable area? Maybe this is an interesting intermediate option. If the heatmap describes real data from 0-10 on X and Y, you can `plt.Axis(0, 0, 10, 10);`, then draw your heatmap with something like `stretch: true`, and it will fill the data area. This way the axis ticks will actually line up with the heat map... it seems too easy 😈 

If you want to render the scalebar outside the data area, add a lot of padding to the right (https://swharden.com/scottplot/cookbooks/4.0.32/#customize-figure-figure-padding) and draw on `settings.gfxFigure` instead of `gfxData`. I'll make a more elegant solution when I refactor the layout system as described above...

Benny121221:
> In the mean time what do you think about figuring out how to make the heatmap stretch to fill 100% of the viewable area? Maybe this is an interesting intermediate option. If the heatmap describes real data from 0-10 on X and Y, you can plt.Axis(0, 0, 10, 10);, then draw your heatmap with something like stretch: true, and it will fill the data area. This way the axis ticks will actually line up with the heat map... it seems too easy 😈there may be a catch 🤔

Stretching the heatmap will wreck the aspect ratio unless you resize the plot as well. You could force a plotsize which fits that aspect ratio, but that seems inelegant. As a parameter, it ain't a bad idea, but it won't work everywhere. If someone has a heatmap that's 1000x50 or some similar shape it's going to be comical. You either don't fit the aspect ratio and the whole thing is blurrier than a video of bigfoot. Or you do and the window is 11 feet wide and 6mm tall.

I'll move the scalebar and then I should get some sleep.

Benny121221:
I made the scale fullsized, I think it looks better:

image

swharden:
I'm a little confused by the soft edges. Is this a colormap issue?

https://github.com/swharden/ScottPlot/blob/44e82cc9d2c9fc387f2acd38ec0a271562c09d5b/src/ScottPlot.Demo/Experimental/Heatmap.cs#L18-L21

image

Also I added a new demo with image data. This looks really cool!

image

swharden:
I'm going to put this down for a while, but I got the controls and layout system updated so your custom scalebar isn't broken by mouse interactions anymore 👍

Benny121221:
The soft edges could be because of the interpolation mode? When I made the scalebar it looked really bad at first because it wasn't in nearest neighbour.

Benny121221:
Your fix only seemed to work on WinForms, looks like you got confused between `sourceTextBox` and `sourceCodeLabel` on WPF. I fixed the issue.

I'm wondering though, should we stop the recalculating of layouts for all heatmaps internally? Instead of expecting the user to do it? For example, any call to `PlotHeatmap` disables recalculating the layout?

swharden:
> Your fix only seemed to work on WinForms, looks like you got confused between `sourceTextBox` and `sourceCodeLabel` on WPF. I fixed the issue.

Thanks for noticing and correcting that!

> I'm wondering though, should we stop the recalculating of layouts for all heatmaps internally? Instead of expecting the user to do it? For example, any call to `PlotHeatmap` disables recalculating the layout?

I think the new option added to the controls' `Configure()` method is good because users may want to manually tweak padding for other reasons. 

You're definitely right though, it may be best to not require the user to know to disable this when plotting heatmaps. Perhaps the best way to do this is to automatically disable recalculation OnMouseUp in the user controls if a heatmap is detected in the plottables list.

Benny121221:
I did it. `recalculateLayoutOnMouseUpIf ` is now nullable, and null by default. If it is set then it overrides. Otherwise, it checks if the plot contains a heatmap and acts accordingly.

Benny121221:
Also I have a better solution to me using tabs, I added this to my gitconfig. Now everytime I add changes they should be formatted with spaces. 🤞 

https://gist.github.com/Benny121221/50343cf14d5dbe6b33da285ac2f7ea9e#file-gitconfig-L21

swharden:
> Also I have a better solution to me using tabs, I added this to my gitconfig. Now everytime I add changes they should be formatted with spaces. 🤞
> 
> https://gist.github.com/Benny121221/50343cf14d5dbe6b33da285ac2f7ea9e#file-gitconfig-L21

Thanks for this tip! 

I think I can make Azure Pipelines do this automatically... but (1) I'm afraid to troubleshoot that since it involves automated modification of the whole code base 💀 and (2) I suspect it would cause merge conflicts the next time the user tries to upload their code, so I'm guessing showing a failed build is the best thing to do here 🤔

Benny121221:
Yeah, modifying source files with a pipeline can be very spooky.

Source: babel and the typescript compiler do not play nicely together.

Benny121221:
I'm working on some performance stuff. The slowest part by far was normalizing the data, which didn't seem right.
it cm
The times for the 2-D waveform example:
Previous time elapsed (constructor only): 2.556s
Now: 0.298s

I still have a couple questions, the main one is how many threads it should open, currently it's set at 12 regardless of input size. Multithreading gives a modest performance uplift, but the big discovery was here.

image

If you do this it runs `input.Min()` and `input.Max()` on every iteration...

Benny121221:
I'll work on multithreading a couple of the other pieces, but I don't expect a very big change. I always thought it was strange how slow it was, and I recklessly assumed that the LINQ thingy would be better optimized.

Benny121221:
Ah, good, we can render heatmaps in 4K now
image

It takes a few seconds, but before rendering a 100x100 heatmap took a few seconds, so I'm happy. Most heatmaps are pretty low res anyways.

StendProg:
Hi @Benny121221 , 
Parallel processing in linq easy thing if you already have a querry.
You can simplify you code with just one line:
```c#
this.intensitiesNormalized = intensitiesFlattened.AsParallel().AsOrdered().Select( intens => (intens - min) / (max-min) ).ToArray();
```

Besides the fact that it’s simpler and difficult to make mistakes here, .NET selects the optimal number of threads depending on the configuration of the CPU on which it is running.

This will work even faster if the internal operation (element projection) is long enough. For example you can iterate on lines in parallel, but inside iterate on pixels in line not parallel.
This will look a little more complicated already, and not the fact that it will be easy to write in the form of a querry. But you always can write close to standart `for` or `foreach` loops using `Parallel.For()` or `Parallel.Foreach()`.

Benny121221:
> Parallel processing in linq easy thing if you already have a querry.
You can simplify you code with just one line:
`this.intensitiesNormalized = intensitiesFlattened.AsParallel().AsOrdered().Select( intens => (intens min) / (max-min) ).ToArray();`

Yeesh, that's a lot simpler, I had no idea LINQ could do that

swharden:
I got a question from Olof S. via email asking if we plan to implement filled contour plots...

image

It seems like this is a pretty jump in complexity so my assumption at this time is, "no", but I wanted to put it out there and get your reflection on the idea. It seems like the implementation would be pretty complex. 

Perhaps an official statement somewhere that we support heatmaps, but not labeled contours, would be helpful for others wondering this same thing.

StendProg:
>I got a question from Olof S. via email asking if we plan to implement filled contour plots...
It is very close to HeatMap.
Levels can be obtained using ColorMap in `low resolution`.
Lines can be drawn by coloring  with black each pixel that has a neighbor from a lower-level color map.
But to draw signatures in the right place and at the right angles is really a difficult task.

Benny121221:
I would look at it, but I think it could take a while to get right. If you want to get heatmaps into the next release I don't expect it would be ready (or at least not ready and polished)

swharden:
> I would look at it, but I think it could take a while to get right. If you want to get heatmaps into the next release I don't expect it would be ready (or at least not ready and polished)

If we decide to experiment with contour maps, I like the idea of having a `PlottableContourMap` to experiment with, keeping the `PlottableHeatmap` light (and released soon)

swharden:
@Benny121221 this is developing nicely! I actually want to merge this in ASAP, and we can keep it tagged as experimental, and open a new PR to refine it. 

Merging small PRs frequently and hiding experimental features behind feature flags avoids merge hell and keeps the project nimble.

... but I might be too late 😝 41 files chanaged? 💀 I'm not a git pro... so I'm not sure when it got out of control or if it was you or me that did it. 

If you issue a simple a PR of something I can easily merge, I'm guessing all the future development can just be modifications to the 1 `PlottableHeatmap.cs` file and the PR will be easy from there. Up to you if you want to try to repair this PR or create a new one. Also if you'd rather me give it a shot, let me know (and stop making commits here) and I'll see what I can do.

Benny121221:
~40 real commits, and the rest are merge or rebase related. There should be no conflicts

swharden:
If it helps, these are the only files that should be modified by this PR

image

If you want to divide and conquer a bit, modifications to the user controls and colormaps could probably be their own PRs. That simplifies the heatmap PR greatly.

swharden:
> ~40 real commits, and the rest are merge or rebase related. There should be no conflicts

Well... I guess I can try the merge and roll it back if it gets weird.

In the future I'll consider frequent rebases an indication things are staring to go off the rails somewhere... I like to review the files changed line by line while merging, but this one's gotten pretty overwhelming

Benny121221:
> Well... I guess I can try the merge and roll it back if it gets weird.
> In the future I'll consider frequent rebases an indication things are staring to go off the rails somewhere... I like to review the files changed line by line while merging, but this one's gotten pretty overwhelming

I can cherry-pick the commits, it's not a problem

olofszymczak:
Hi,

Just wondering if there will be a capability to be able to add a background image or to make the heatmap transparent?

heatmap

Another cool feature I was looking at, is demoed by Azure maps
https://azuremapscodesamples.azurewebsites.net/Heat%20Map%20Layer/Heat%20Map%20Layer%20Options.html

In the layer options tab you can play around with:
- Radius size
- Opacity
- intensity
- Weight
image
image

Benny121221:
@olofszymczak I added transparency and background pictures to a new issue #457.

I don't see much use in expanding the radius of points personally, to me it just makes it less readable. If @swharden disagrees then I'm sure it can make it into a PR

Benny121221:
If you have more ideas on heatmaps I would suggest commenting on #457 or making a new issue, just so its easier to find your suggestions and input

swharden:
Hey @olofszymczak, I think @Benny121221 is doing an awesome job with #457 and I agree with their comments from the previous couple of posts.

One thing to keep in mind is that ScottPlot is intended to be a _plotting_ library, with the goal of making it easy to display X/Y data interactively in a 2D coordinate space. In other words, the primary focus of ScottPlot is to create mouse-interactive plots with have labeled X and Y axes and maybe a title. While heatmaps are probably still within this scope, we should temper how many heatmap features we want to support. 

There's probably a point where advanced heatmap functionality features merits its own library. Such a library could be massively simpler that ScottPlot since it would focus on only _image processing_, and not deal with the advanced interactive 2D coordinate and rendering systems that underlies the majority of ScottPlot's code. I'm afraid that diving too deeply into heatmap functionality tilts ScottPlot toward an image processing library, and I'd rather keep it simple and focused and good at one thing: plotting X and Y data.

I'd rather intentionally under-support heatmaps rather than try to do everything heatmap related but end-up doing a bad job at it, or end-up with a heatmap module that's so advanced it's hard to maintain. I don't know what the best balance of features is yet, but I thought I'd share my thought process with the group. Again I'm really excited about how #424 is growing, and looking forward to finding where that ideal balance between features and simplicity will be!

olofszymczak:
Cool, will add any new ideas/suggestions that I have to #457.

Issue #428, opened 2020-05-26
Radar Graphs


Benny121221:
**What can we do to make ScottPlot better?**
A new plottable for Radar Graphs (AKA spider/star chart):

A 7-Point Radar Graph:
![image](https://user-images.githubusercontent.com/8635304/82853151-2633e280-9ec2-11ea-8133-d540757640fd.png)

A 4-Point Radar Graph:
![image](https://user-images.githubusercontent.com/8635304/82853198-4a8fbf00-9ec2-11ea-9d77-a1c7c5480d7c.png)

User code being something like this:

```cs
var radar = plt.PlotRadar(double[] values, string[] labels);
```

swharden:
This looks great @Benny121221, thanks!

PR #430, opened 2020-05-26
Radar graph

Benny121221:

**Purpose:**
#428 

**New functionality (code):**

```cs
double[,] values = { { 78, 83, 84, 76, 43 }, { 100, 50, 70, 60, 90 } };
plt.PlotRadar(values);
```

image

image


swharden:
Wow! I saw this message in my inbox earlier in the week and thought the screenshot was of another plotting library. This is really cool! I'll make a couple tweaks but this is merging in today. Thanks so much @Benny121221!

swharden:
I'm on the fence about the legend color. I'm leaning toward making it the same color as the fill... 

image

Probably the _best_ solution would be to refactor the legend to support a true rectangle shape (not just a fat line) with a definable edge color. If we go there let's do it in another PR

swharden:
@Benny121221 FYI, this is a nightmare topic, and something I'm not going to even to suggest trying to fix because it's broken everywhere in ScottPlot* but I wanted you to be aware of it: Many `System.Drawing` objects implement `IDisposable` and have to be manually deleted, otherwise the garbage collector has to do a ton of work. The proper way to use these objects is to wrap them in `using` statements.

The things you'd expect are `IDisposable`:
* System.Drawing.Image
* System.Drawing.Bitmap
* System.Drawing.Graphics

... but so are these doozies:
* System.Drawing.Pen
* System.Drawing.Brush
* System.Drawing.Font
* System.Drawing.FontFamily
* System.Drawing.StringFormat

_*The reason I suggest not worrying about it in ScottPlot currently is I'm working toward abstracting System.Drawing out of ScottPlot to make it easier to use another rendering system. When I do this refactor I'll respect `IDisposable` everywhere._

Issue #436, opened 2020-05-28
ALT+ScrollWheel causes loss of focus


tomwimmenhove:
**Describe the bug**
After you zoom vertically, using alt+scroll-wheel, zooming horizontally using ctrl+scroll-wheel doesn't work, until the user clicks in, or pans the graph.

**To Reproduce**
Press alt and scroll to zoom in/out vertically. Release alt, then press ctrl and scroll to zoom horizontally. => You'd expect the graph to zoom horizontally, but nothing happens.

```cs
    public partial class Form1 : Form
    {
        public Form1()
        {
            InitializeComponent();

        }

        private void Form1_Load(object sender, EventArgs e)
        {
            double[] dataX = new double[] { 1, 2, 3, 4, 5 };
            double[] dataY = new double[] { 1, 4, 9, 16, 25 };
            formsPlot1.plt.PlotScatter(dataX, dataY);
            formsPlot1.Render();
        }
    }
```

StendProg:
Hi @tomwimmenhove 
This problem is caused by a system event that causes the control to lose focus on `Alt` key press.
I maked PR that fix that, but not sure if he deserves to be accepted into the codebase.
If you want a fast solution you can disable main menu of `Form` in your application by setting `Form.ControlBox` property to `false`.
Also, loss of focus occurs on an odd `Alt` press. Each even `Alt` press returns focus...

swharden:
Hi @tomwimmenhove thanks for reporting this, and @StendProg thanks for figuring out the cause of this issue!

@StendProg's solution (#437) fixes the problem on the WinForms control (but it's still present on the WPF control). 

An alternative solution here may be to stop using the ALT key. Perhaps CTRL could lock the vertical axis and SHIFT could lock the horizontal axis. Currently SHIFT-click-drag is used to zoom-to-rectangle, but we could give that function to ALT, or even CTRL+SHIFT...

Another alternative option may be to have mousewheel events create call a method to focus the control 50 ms after the wheel event. I'm not sure if this is possible yet. Something like:

```cs
Task.Delay(50).ContinueWith(t=> GetFocusAgain());
```

swharden:
I think the best long-term solution here is to avoid using ALT in combination with scroll-wheel actions. ALT _can_ be used with click-drag actions because the control maintains focus. Switching the function of ALT and SHIFT for the WinForms and WPF controls is probably the way to go here. The new list of actions is:

* Left-click-drag: pan
* Right-click-drag: zoom
[...]

PR #437, opened 2020-05-28
Fix loosing control focus on alt key press.

StendProg:
**Purpose:**
Pressing the Alt Key in a WinForms Application Loses the Focus of the Control. #436
This PR filter all `alt` key windows messages, and prevent application to catch that.
Solution taken from the discussion of this problem:
https://stackoverflow.com/questions/18235781/how-do-i-prevent-the-control-box-getting-focus-when-i-press-alt/18235868#18235868
> Beware of the many side effects, it stops a menu behaving normally and shortcut keystrokes that use Alt are not going to work anymore. Do consider using Ctrl or Shift instead.

Additionaly a small fix to remove duplicate code in the Winform user control constructors.

### DIFF
--- a/src/ScottPlot.WinForms/FormsPlot.cs
+++ b/src/ScottPlot.WinForms/FormsPlot.cs
@@ -45,11 +45,8 @@ namespace ScottPlot
             Application.AddMessageFilter(this);
             Reset(plt);
         }
-        public FormsPlot()
+        public FormsPlot() : this(null)
         {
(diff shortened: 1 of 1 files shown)


swharden:
Thanks for this suggestion @StendProg! I might merge this PR if I can't find a better solution, but for now I'm going to close it and seek a solution that is more identical between WinForms and WPF so it will be easier to maintain.

Issue #438, opened 2020-05-28
Vector fields


Benny121221:
**What can we do to make ScottPlot better?**
Add a vector field plottable, e.g.: 
![image](https://user-images.githubusercontent.com/8635304/83201571-04756e00-a103-11ea-9e46-d3ce3452ef81.png)

Does your proposed feature allow users to do new things in source code? If so, show us how you'd like to be able to use the new feature:

```cs
plt.PlotVectorField(Vector2[] vectors, double[] xs, double[] ys)
```

It might be valuable for the arrows to change in colour to indicate magnitude, as vector fields typically have scaled-down arrows so the plot remains readable

Conveniently, Microsoft provides vector types for 2, 3, and 4 dimensions: https://docs.microsoft.com/en-us/dotnet/api/system.numerics.vector2?view=netcore-3.1

swharden:
Haha, consistent with previous suggestions @Benny121221 you surprised me with a really cool idea that seemed to come out of nowhere! I like it.

Use of System.Numerics is an excellent idea. I've been getting familiar with their `Complex` struct/methods lately with [FftSharp](https://github.com/swharden/FftSharp)

You may find the `PlotArrow()` method useful to refer to if you decide to make a demo of this one.

https://github.com/swharden/ScottPlot/blob/641126646d77c3aefad055e9f31d44da333eb03f/src/ScottPlot/Plot.cs#L689-L718

swharden:
We have some experimental code merged in the master branch now (#424)

There may be some utility in expanding the colormap module to let those colormaps also be used to color line plots to produce graphs like this:

![image](https://user-images.githubusercontent.com/4165489/83372829-4f99b600-a37b-11ea-8116-93889cbbbcd1.png)

swharden:
Just merged this in, thanks @Benny121221!

Issue #425, opened 2020-05-23
auto-format pushes and PRs


swharden:
Reviewing PRs would be simpler if formatting changes (e.g., tabs vs. spaces) didn't appear as modified lines of code.

I think with GitHub actions I can auto-format the codebase. This would be nice to run with every push.
https://github.com/marketplace/actions/dotnet-format

swharden:
Actually since we're already using Azure Pipelines, we should probably figure out how to run it within that system.

https://www.meziantou.net/enforce-dotnet-code-style-in-ci-with-dotnet-format.htm

swharden:
Finally figured this one out. This is what I learned.
* Define formatting rules in `.editorconfig`
* Visual Studio auto-formatter (CTRL+K, CTRL+D) respects it
* A command line autoformatter can be installed too
  * install: `dotnet tool install --global dotnet-format`
  * run: `dotnet format`
* On Azure Pipelines this tool requires the latest .NET Core 3, so use `UseDotNet` to get that. Doing this means you also have to run `UseDotNet` to grab .NET Core 2 again.
* When run with `--dry-run --check` the auto-formatter will return with a nonzero exit code if formatting is required
  * this is sufficient to "break the build"
  * `--verbosity diagnostic` details the problem

swharden:
You can modify your `.gitconfig` to autoformat on every commit:

```
addf = !"cd src; dotnet-format; cd ..; git add"	
```

Example (#424): https://gist.github.com/Benny121221/50343cf14d5dbe6b33da285ac2f7ea9e#file-gitconfig-L21

PR #439, opened 2020-05-29
Vector fields

Benny121221:
**Purpose:**
#438 

**New functionality (code):**

```cs
plt.PlotVectorField(vectors, xs, ys);
```

image

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -1,12 +1,11 @@
-New contributors should review CONTRIBUTING.md:
-https://github.com/swharden/ScottPlot/blob/master/CONTRIBUTING.md
+New contributors should review [CONTRIBUTING.md](https://github.com/swharden/ScottPlot/blob/master/CONTRIBUTING.md)
 
 **Purpose:**
 Describe what is accomplished with this code modification. 
 If this pull request relates to an issue, link to it here.
 
 **New functionality (code):**
-Provide a code example demonstrating new functionality achieved with this pull request (if applicable)
+Provide a code example demonstrating new functionality achieved with this pull request (if applicable):
 
 ```cs
 var newThing = new SpecialObject();
@@ -14,3 +13,11 @@ var newThing = new SpecialObject();
 
 **New functionality (image):**
 If this pull request changes how a plot is rendered, provide an image or screenshot demonstrating the new functionality:
+
+**Autoformat your code:**
+The build will fail if your code is not auto-formatted. Auto-format your code in Visual Studio, or from the console using these commands:
+```
+cd ScottPlot/src/
+dotnet tool install --global dotnet-format
+dotnet format
+```
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -21,6 +21,14 @@ ScottPlot is made better by contributions from the open-source community!
 * If you are considering make a large change, create an issue first so the authors may suggest ideal implementation
 * You don't have to create an issue if the bug is sufficiently described in the pull request
 
+## Automatic Code Formatting
+The build will fail if code in pull requests has not been auto-formatted. Auto-format your code in Visual Studio, or from the console using these commands:
+```
+cd ScottPlot/src/
+dotnet tool install --global dotnet-format
+dotnet format
+```
+
 ## Modify an Existing Plottable Object
 
 **For extensive changes,** duplicate existing plottable source code (adding `Experimental` to the end of its name) and follow the steps below to create a new Plottable.
--- a/NOTICE
+++ b/NOTICE
@@ -1,10 +1,9 @@
 ScottPlot
 Copyright 2020 Scott Harden / Harden Technologies, LLC
 
-This product includes the colormap Turbo developed at Google (https://www.google.com/)
-Minor modifications to suit ScottPlot's image handling have been made. The modified file is at /src/ScottPlot/Config/ColorMaps/Turbo.cs
-More detail about modifications is included in the file.
-Turbo is under the Apache 2.0 License:
+ScottPlot includes the Turbo colormap originally developed by Google (https://www.google.com/).
+The Turbo source code has been modified as described in /src/ScottPlot/Config/ColorMaps/Turbo.cs
+Google's Turbo colormap is distributed under the Apache 2.0 License:
 
 Copyright 2019 Google LLC.
 
--- a/dev/changelog.md
+++ b/dev/changelog.md
@@ -1,5 +1,9 @@
 # ScottPlot Changelog
 
+## ScottPlot 4.0.37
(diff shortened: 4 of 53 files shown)


Benny121221:
I think this looks pretty good, the only thing is it doesn't use colour to show magnitude, which I feel would be a nice feature. I don't want to duplicate code that's already in the heatmap PR however.

swharden:
The example in #438 shows the size of the arrowhead scaling with the magnitude of the vector... perhaps this would make the gap at X=0 look less awkward? Maybe something similar for line width? 🤔

swharden:
Matplotlib calls these "quiver plots". Makes sense since a quiver is a collection of arrows 😄 

https://pythonforundergradengineers.com/quiver-plot-with-maatplotlib-and-jupyter-notebooks.html

I think `PlotVectorField()` like you have it reads better...

Benny121221:
> Makes sense since a quiver is a collection of arrows 😄

Oooooohhh... that's why it's called that. I was thinking quiver the verb, and I couldn't think of any way that it made sense.

Benny121221:
I added the pendulum example, because it is the canonical example of vector fields, at least for physics:

image

Benny121221:
I'm curious what your thoughts would be on having LaTeX axis labels? For example, the y axis in the pendulum screenshot is a little bland in comparison to
```tex
\dot{\theta}
```
image

Or if you insist on Leibniz's notation:
```tex
\frac{d\theta}{dt}
```
image

I know you aren't keen on dependencies, and this isn't exactly something that's fun or easy to implement on your own. Especially when it's just to make labels and titles look a little spicier.

swharden:
> I'm curious what your thoughts would be on having LaTeX axis labels?

I like the concept... but I don't see `System.Drawing` being able to keep up. We'd have to bring in a tight dependency to another package, and right now ScottPlot is dependency-free and I'd love to keep it that way.

If a user really wants that level of customization, the way to go is probably for them to use their tex package to render a Bitmap then draw it on top of an existing plot.

hhubschle:
I just committed another quiver plot pull request, Wasn't aware that this exists already.

swharden:
> I just committed another quiver plot pull request (#440), Wasn't aware that this exists already.

no worries! I'll close that PR, merge this one shortly, and we can continue to experiment with alternative implementations in new PRs as ideas pop up

swharden:
@Benny121221 it might be cool to make a variant of this like `PlotFunction`, where you pass in a formula and the lines are generated programmatically...

swharden:
I really want to find something more elegant than passing in a `Vector2[,]` and two `double[]` arrays. This feels convoluted, and it's weirdness is apparent in the quickstart code...

I'm surprised how few structs [system.numerics](https://docs.microsoft.com/en-us/dotnet/api/system.numerics) actually offers. I don't want to make my own struct here, but I'm really limited. `Vector4` is close but no cigar 😝 

I could make an array of tuples or something that pairs a `System.Drawing.PointF` with a `System.Numerics.Vector2` but that seems like a weird mash-up of libraries.

Going the other direction, I'm looking at what matplotlib does. https://matplotlib.org/3.1.0/api/_as_gen/matplotlib.pyplot.quiver.html This is interesting. Their quiver method takes in the equivalent of 4 `double[]`... I might like this 🤔 It inherently supports non-rectangular grids too, though I can't think of a use case there.

```cs
plt.PlotVectorField(posXs, posYs, vecX, vecY);
```

perhaps an optional argument could be used to scale how long the arrows are

```cs
plt.PlotVectorField(posXs, posYs, vecX, vecY, vecScale);
```

I'm between this or writing my own struct to hold these values. Perhaps `PointVector`?

Benny121221:
> I really want to find something more elegant than passing in a Vector2[,] and two double[] arrays. This feels convoluted, and it's weirdness is apparent in the quickstart code...

I think the weirdness comes from how there are two competing ideas of vectors. The CS perspective where they are points defined by the position of their tail, and the Physics perspective where they are quantities defined by their direction and magnitude.

The CS perspective is most convenient for the evolution of a phase space under a differential equation, which is why the pendulum example is so simple, there's no trig required to convert between the two interpretations.

The Physics perspective is more convenient for situations where you care about the direction, not the position of the tail, which is why the slope field example feels klunky.

Personally, my recommendation would be to keep with the CS perspective, and offer a function to convert between, something like `Tools.GetTailPosition()`

Benny121221:
Ah, I misunderstood you, you meant argument list weirdness. My only idea that you didn't already bring up was to do the same as heatmaps, where they take `Vector2[,]`, and then optional parameters to move or scale the axis: `double[] axisOffsets, double[] axisMultipliers`. I don't think that really solves the problem though :/

Benny121221:
> Going the other direction, I'm looking at what matplotlib does. https://matplotlib.org/3.1.0/api/_as_gen/matplotlib.pyplot.quiver.html This is interesting. Their quiver method takes in the equivalent of 4 double[]... I might like this 🤔

That's what the other vector field PR did, it's what I'd lean towards personally. Plus I liked his demo, I think electromagnetism is a lot simpler when you see it as a phase space.

hhubschle:
The reason why my demo looks 'nice' is because I removed the CustomEndCap and drew the arrow head myself. I encountered nothing but pain when I tried to make scaling work nicely with a CustomEndCap.
There is quite a bit more work involved in doing all the trig calculations. However, this shouldn't be much of an issue. Quiver plots with more than 10,000 vectors are usually way too crowded to be useful.

swharden:
> The reason why my demo looks 'nice' is because I removed the CustomEndCap and drew the arrow head myself. I encountered nothing but pain when I tried to make scaling work nicely with a CustomEndCap.
> There is quite a bit more work involved in doing all the trig calculations. However, this shouldn't be much of an issue. Quiver plots with more than 10,000 vectors are usually way too crowded to be useful.

@hhubschle is referring to #440

I didn't notice that before! This is good to take note of. I agree the `System.Drawing` caps are pretty lousy. Thanks for pointing it out!

image

Benny121221:
I ran dotnet-format so idk what that's about

swharden:
> I ran dotnet-format so idk what that's about

This just started happening! #452 was affected too. I think a recent update to dotnet changed the behavior of `dotnet format` to now sort `using` statements (#425).

I just started working on ScottPlot again tonight, so I'm free to merge this if/when you think it's ready. Want me to leave it as it is for you to keep poking at it or do you think it's ready for a merge now?

Benny121221:
I think I'll add hhubschle's demo, to me it looked like two opposite charges but looking at the source I didn't recognize the math.

Other than that I can't think of much which needs adding, idk how you feel about changing the parameter list, I remember you bringing that up earlier?

PR #440, opened 2020-05-30
Add quiver plot, inspired by https://plotly.com/python/quiver-plots/

hhubschle:
**Purpose:**
Utility to simplify drawing of a vector field.

image


swharden:
Thanks for the effort here @hhubschle! Work in this direction is being addressed by #439. I really like the look of your demo vector field though, so maybe we can use this as part of the cookbook in #439

PR #442, opened 2020-06-01
colormap-refactor

Benny121221:
**Purpose:**
@swharden's suggestions here: https://github.com/swharden/ScottPlot/issues/420#issuecomment-636904551

**New functionality (code):**

```cs
(byte r, byte g, byte b) = colormapObject.Lookup(0.3); // From intensity
(byte r, byte g, byte b) = colormapObject.Lookup(77); // From byte
```

N/A

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -1,12 +1,11 @@
-New contributors should review CONTRIBUTING.md:
-https://github.com/swharden/ScottPlot/blob/master/CONTRIBUTING.md
+New contributors should review [CONTRIBUTING.md](https://github.com/swharden/ScottPlot/blob/master/CONTRIBUTING.md)
 
 **Purpose:**
 Describe what is accomplished with this code modification. 
 If this pull request relates to an issue, link to it here.
 
 **New functionality (code):**
-Provide a code example demonstrating new functionality achieved with this pull request (if applicable)
+Provide a code example demonstrating new functionality achieved with this pull request (if applicable):
 
 ```cs
 var newThing = new SpecialObject();
@@ -14,3 +13,11 @@ var newThing = new SpecialObject();
 
 **New functionality (image):**
 If this pull request changes how a plot is rendered, provide an image or screenshot demonstrating the new functionality:
+
+**Autoformat your code:**
+The build will fail if your code is not auto-formatted. Auto-format your code in Visual Studio, or from the console using these commands:
+```
+cd ScottPlot/src/
+dotnet tool install --global dotnet-format
+dotnet format
+```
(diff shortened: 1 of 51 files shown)


swharden:
I spent all weekend working on colormaps in a development branch of my spectrogram library. You may find this page interesting: https://github.com/swharden/Spectrogram/tree/1.2#colormaps

Once I get that finalized, I'll loop back around to ScottPlot and work on this colormap module. I think some of the things I'm learning in Spectrogram's colormaps will translate well here. I thought I'd drop a couple notes for reference:

### class structure

I like having a `Colormap` class which takes an `IColormap` as input, but also has static `Colormap` properties like `Viridis`, `Jet`, etc. This lets you instantiate a colormap with 

```cs
// a standard colormap
Colormap cmap1 = Colormap.Viridis;

// a custom colormap
IColormap customCmap = new CustomCmap(); // custom implementation
Colormap cmap2 = Colormap(myColormap);
```

The biggest advantage here is intellisense gives you a list of colormaps as you start typing `Colormap.`, while avoiding `enum` types entirely.

See https://github.com/swharden/Spectrogram/blob/1.2/src/Spectrogram/Colormap.cs

### storing values

Storing RGB as an `int` seems simplest, where the lowest 3 bytes of the `int` correspond to the RGB values. See: https://github.com/swharden/Spectrogram/blob/1.2/src/Spectrogram/Colormaps/Viridis.cs

### matplotlib colormaps

For reference, mant of matplotlib's colormaps are here, including magma, inferno, plasma, viridis, cividis, twilight, and turbo

https://github.com/matplotlib/matplotlib/blob/4f42c6825fe311f18812f99701701cb484da7670/lib/matplotlib/_cm_listed.py

Note that in-between colors can be calculated if desired

Benny121221:
I like your structure more, enums seem clunky for this one. As for storing them I don't have any preference between byte arrays or ints, bytes are more clear in source code, but System.Drawing uses ARGB ints, so you can save converting. If you want to have users make custom colourmaps then it may be more intuitive to use byte arrays, but it isn't difficult to convert between.

Benny121221:
Also, calculating in-between colours seems unnecessary, as colourmaps are defined for 256 steps, which means adding more between those steps will be rounded away (on an sRGB monitor) unless the colourmap has very uneven steps, in which case it will be indistinguishable to the human eye.

swharden:
> As for storing them

Actually, storage implementation (storing byte arrays, integers, or a formula) is handled in the colormap class itself (e.g., [Viridis.cs](https://github.com/swharden/Spectrogram/blob/1.2/src/Spectrogram/Colormaps/Viridis.cs)) so developers can choose any storage mechanism they like.

The reasons I found `int[]` useful is because I can copy/paste it between languages (like Python) without worrying about syntax. It takes less memory than floating point options, and is also more concise on the page.

> Also, calculating in-between colours seems unnecessary

Good point! A corner case may be some TIFs store pixel intensity as 32-bit values or even floating-point numbers, but I'll ignore that and agree we can keep things simple by assuming we intend to produce 8-bit RGB images.

Issue #451, opened 2020-06-08
AxesChanged - MatchAxis feature has shots and lags


robokamran:
I synchronize 2 or more FormsPlots using AxesChanged and MatchAxis. When I zoom or pan, there is a noticeable shot/lag in synchronization. It is also detectable in the demos. This has happened since the version 4.0.20 (version 4.0.19 is OK). Consider the following code which synchronizes 4 "FormsPlot"s, in previous versions it was perfect for synchronizing 4 FormsPlots each include a 10000 points Signal. But now it has shot/lag. (currently I have to use V4.0.19)

```cs
private void FormsPlot_AxesChanged(object sender, EventArgs e)
{
    foreach (var plot in PlotList)
    {
        plot.plt.MatchAxis((sender as FormsPlot).plt, vertical: false);
        plot.Render();
    }
}
```

StendProg:
Hi @robokamran,
I fixed lagging in #452 
 if you don't want wait for new release, you can grab changes manualy (its only few lines of code).
If you need additional performance, you can render linked plot in low quality
`Render(true, lowQuality:true, processEvents:true)`

swharden:
Hey @robokamran, thanks for reporting this! @StendProg, thank you for your rapid identification of the issue, and your helpful PR (#452). I'll take a closer look and release an updated NuGet package tonight 👍

robokamran:
I have just noticed a key point and somehow my mistake!
In V4.0.19, the axis change update happened on the "right click up" event after the zoom/pan on the original plot is completed. In newer version the synchronization happens in real-time. I write log programs for hardware interfaces and the former type of functionality was so good and applicational that all our guys (the users) didn't care about the real-time synchronization! Their focus was on the original plot, since we have up to 16 plots at the same time with more than 20K points and we need time synchronization for all of them when we zoom in time. Could this sync type be optional? Because the "not real-time" type is very clear, robust, applicational and render OK.

robokamran:
It seems that it does the trick! :sunglasses:
Thanks a lot @swharden :wink: 🙏🏻

PR #452, opened 2020-06-08
Fix lag on linked axes.

StendProg:
**Purpose:**
Fix linked axes lagging. #451 
The reasone is full `Render()` loop in second `formsPlot` on each mouse move.
This PR add `skipIfCurrentRendered = true` to demo,
Add extra param to `Render(bool processEvents)`. This param allow to make `Application.DoEvents()` in `Render()` for not interactive formsPlot (no mouse pan on zoom).

`Resize()` in `MatchAxis()` was commented. I have no idea why it was implemented in this method, But it create new bitmaps for all layers on each `MatchAxis()` call.

//TODO
Demo contain not optimal render call in `formsPlot2.AxisChanged()`.
This is for demonstration benefits of this PR. It must be changed to optimal before merge.

**New functionality (code):**
Extra parameter in `Render()` to process events in not interacted `formsPlot`.
Combined with `skipIfCurrentlyRendered=true` this allow multiple Render calls without lagging for not interactive `formsPlots`. This param may be usefull not only for linked axes plots.
```cs
formsPlot.Render(true, true, processEvents: true)
```

Linked Axes demo demonstrates this PR.
Interactions on upper `formsPlot1` contain this PR changes.
Interactions on lower  `formsPlot2` use old implementation.

### DIFF
--- a/dev/changelog.md
+++ b/dev/changelog.md
@@ -1,5 +1,8 @@
 # ScottPlot Changelog
 
+## ScottPlot 4.0.37
+* Added `processEvents` argument to `formsPlot2.Render()` to provide a performance enhancement when linking axes of two `FormsPlot` controls together (by calling `Plot.MatchAxis()` from the control's `AxesChanged` event, as seen in the _Linked Axes_ demo application) (#451, #452) _Thanks @StendProg and @robokamran_
+
 ## ScottPlot 4.0.36
 * `PlotScatterHighlight()` was created as a type of scatter plot designed specifically for applications where "show value on hover" functionality is desired. Examples are both in the cookbook and WinForms and WPF demo applications. (#415, #414) _Thanks @Benny121221 and @StendProg_
 * `PlotRadar()` is a new plot type for creating Radar plots (also called spider plots or star plots). See cookbook and demo application for examples. (#428, #430) _Thanks @Benny121221_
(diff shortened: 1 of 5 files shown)


StendProg:
It looks like `Azure` dislike imports ordering in whole project. I think it's bad idea fix it in this PR.

swharden:
> It looks like Azure dislike imports ordering in whole project.

This is surprising! It looks like the command `dotnet format` now sorts `using` statements, but it previously did not do this 🤔

It looks like I fixed it: I ran `dotnet format` on the master branch, then re-ran the Azure Pipelines job on this PR. All green! 👍 

Note to self: setting up the auto-formatter is discussed in #425:
```
dotnet format
```

swharden:
### Application.DoEvents

I like your `processEvents` argument. I'm not happy with this system as a long-term solution, but this argument is an excellent immediate improvement! Thank you for this PR - it may have taken me a long time to figure this one out.


### Resize() in MatchAxis()

The purpose is to match the _layout_ (padding and tick density), not just axis limits. A changed layout means new bitmaps, so this is necessary when the source layout changes. This is what it looks when you zoom out really far on the top graph while leaving `Resize()` commented-out:

image

A better solution is probably for me to create a new event "LayoutChanged" (not just "AxisChanged") and only call `Resize()` when the layout changes. I'll do something like this.

### Big Picture

ScottPlot has evolved so much over the last year that its architecture drifted from what the original user controls had in mind when they were designed. This is a topic for another discussion, but my point is there is much room for improvement by refactoring how ScottPlot handles bitmaps (e.g., they should not be stored or resized, IDisposable objects should be better respected, and using separate Bitmaps for the frame vs. data is a bad idea). I plan to improve these things incrementally, but hopefully there will be more time when active issues/PRs slow down again.

Issue #448, opened 2020-06-04
growing data demo breaks after several seconds


swharden:
![image](https://user-images.githubusercontent.com/4165489/83765751-915f8080-a649-11ea-8988-395d5756d94f.png)

StendProg:
Hi @swharden 
This error happens when the `Signal` goes to `RenderHighDensity` mode. It haven't any respects to `maxRenderIndex` and render full signal. But limits calculates correct...
Fixed in #454

swharden:
#454 is now merged

Issue #419, opened 2020-05-23
optional parallel processing for Signal and SignalXY


swharden:
Now that `Settings.GetPixel()` and `Settings.GetLocation()` are thread safe (#245, #72) maybe we can turn on `settings.misc.useParallel` by default.

I created a GUI test app (/dev/Benchmark) which tests performance and assesses stability of the user controls. They seem stable for me, and parallel processing definitely improves performance for large datasets. With 10 million points my FPS doubled (`25.5 FPS` to `57.5 FPS`)

![image](https://user-images.githubusercontent.com/4165489/82717437-13bf6c00-9c6a-11ea-8142-d17678e85400.png)

There's also a pure-bitmap test which doesn't use the user control.

![image](https://user-images.githubusercontent.com/4165489/82717452-2d60b380-9c6a-11ea-88c5-3050af37d6c0.png)

swharden:
@StendProg I enabled parallel processing by default, removed `useParallel` as an option, and even removed the old methods from plottables which don't use parallel processing. 

I think this is ok since the pixel lookup methods are now thread-safe (and your helpful tests confirm this will remain true, #245), but since this was a big change to `Signal` and `SignalConst` I wanted to make sure you were aware of it and find this acceptable.

StendProg:
@swharden 
A bit late sorry,
As i understand main reasone switching to parallel implementation only, is the code duplicates.
I think it's better stay it configurable. Parallel implementation works slower on small data sets, load all processor cores, contains a bunch of potential errors etc..
I will try to remove code duplicates in all parallel/sequental methods, to make it easy to support like in https://github.com/swharden/ScottPlot/pull/454/commits/3d8ae63e39967ed8d26f53049b6a98280ef1fef5 . But in low priority task.

swharden:
Hi @StendProg, thank you for correcting this issue and refactoring the Signal module in #454. I am very happy about your modifications! This is a great example of using LINQ to improve readability and utilize parallel processing.

https://github.com/swharden/ScottPlot/blob/08d42d0447a2cf0097f86c19969f973b0a71fa44/src/ScottPlot/plottables/PlottableSignal.cs#L203-L216

`ScatterXY` (#374) does not use parallel processing, but after 08d42d0447a2cf0097f86c19969f973b0a71fa44 it now accepts `bool useParallel` as a constructor argument. I believe the two slow calls which may benefit from parallel processing are these: 

[...]

PR #454, opened 2020-06-09
Signal RenderHighDensity fix and refactor

StendProg:
**Purpose:**
Fix not used `maxRenderIndex` in `RenderHighDensity()` mode. #448
Additionaly refactored `RenderHighDensity()`.
For now it easy to switch between parallel and sequental implementation without code duplicates.
`useParallel` return back as `Signal` public field. No idea how better implement it in API, but let it stay configurable (imho). Parallel calculations not always better... #419

**New functionality (code):**
User can switch to sequental calculations by
```cs
var signal = plt.PlotSignal(....);
signal.useParallel = false;
```

Growing data demo correct render:
image

### DIFF
--- a/dev/changelog.md
+++ b/dev/changelog.md
@@ -1,5 +1,8 @@
 # ScottPlot Changelog
 
+## ScottPlot 4.0.36 ⚠️ _in active development_
+* Signal plots now have an optional `useParallel` argument (and property) to allow the user to decide whether parallel or sequential calculations will be performed. (#454, #419, #245, #72) _Thanks @StendProg_
+
 ## ScottPlot 4.0.35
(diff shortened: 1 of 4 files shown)


StendProg:
I have no idea which sort order Azure prefers. VS made the right sort order, as I see.

swharden:
> Fix not used maxRenderIndex in RenderHighDensity() mode

Thanks for identifying the problem and correcting it @StendProg!

> Additional refactored RenderHighDensity(). `useParallel` is now public, so it is easy to switch between parallel and sequential implementation without code duplicates. 

This is a wonderful contribution. The file is much easier to read now, and your use of LINQ is excellent! Thanks!

> No idea how better implement it in API, but let it stay configurable (imho). Parallel calculations not always better...

You raise a good point in #419 that I did not consider: the user may have a good reason to _not_ use parallel mode, so we should maintain both parallel and non-parallel code, and make it easy for the user to switch between them.

In this PR you made `useParallel` a public property, and this is a good idea. 

I will reverse a change made by 6e6c86612ba15b02377cc2295588d8ef5a135bff and add `bool useParallel = true` as an optional argument in `Plot.PlotSignal()`

Issue #457, opened 2020-06-11
Checklist for more heatmap features


Benny121221:
**What can we do to make ScottPlot better?**
- [x] Allow user-set intensity scale
- [x] Transparent heatmaps w/ background image
    - [ ] Scaling options? Are these necessary?
    - [ ] ~~Nullables in intensities map to transparent?~~ I don't like this idea
    - [x] Map intensities below threshold to transparent
    - [x] It will be valuable to support transparent image above heatmap as well as transparent heatmap above image
        - [x] For example geographic heatmaps
- [ ] Better axes
- [x] More colourmaps
    - [x] cmocean
    - [ ] ~~plotly~~ doesn't look promising
    - [x] custom
- [x] `ScottPlot.Tools.XYToIntensities()`
    - [x] Gaussian and density-based
- [ ] Strange rounding? #482

Benny121221:
Transparent heatmaps example:

https://github.com/swharden/ScottPlot/pull/424#issuecomment-642446925

swharden:
What do you think about calculating the heatmap from input data? 

On one hand this feels like scope creep, but on the other hand a core goal of ScottPlot is to make visualizing `double[] xs` and `double[] ys` easy so maybe this isn't too far out. 🤔 

data input | heatmap output
---|---
![image](https://user-images.githubusercontent.com/4165489/84423621-bb3a1980-abec-11ea-9fdd-c8f5ab859df6.png) | ![image](https://user-images.githubusercontent.com/4165489/84423634-c12ffa80-abec-11ea-954d-ae092443d1d7.png)

Benny121221:
It doesn't seem difficult but I don't see the use case for it. I assume there is a use case, but if its only applicable to a niche audience, that audience would probably be better served had they implemented it themselves

swharden:
I'll preface this by saying you're probably light, and according to the YAGNI principle we should wait until the need actually arises before implementing this...

...but a use case I could envision is when you have a massively large number of X/Y data points. A traditional scatter plot would look like an opaque blob. Being able to create a heatmap from that X/Y data is a way to present it in a way you can visualize hotspots in the data.

Benny121221:
Oh, I see, I misunderstood input data, I thought that meant mouseclicks. I don't mind the idea, I think it would be best implemented as a `ScottPlot.Tools.XYToIntensities()` sorta thing.

[...]

PR #458, opened 2020-06-11
Heatmap improvements

Benny121221:

**Purpose:**
#457

XYToIntensities creates a rectangular intensities array from xy coordinates:

```cs
public static double[,] XYToIntensities(IntensityMode mode, int[] xs, int[] ys, int width, int height, int sampleWidth)
```
It supports both gaussian and density intensity modes.

The following parameters have been added to `Plot.PlotHeatmap`:

```cs
 double? scaleMin, double? scaleMax, double? transparencyThreshold, Bitmap backgroundImage, bool displayImageAbove
```

`scaleMin` and `scaleMax` allow modifying the scale.

`transparencyThreshold` allows mapping intensities below it to transparent (0x00000000)

`backgroundImage` and `displayImageAbove` allow adding a background image and displaying it above or below the heatmap.

## Modified Heatmap scale:
image

In this instance it looks less vibrant as the scale has effectively capped the range

## Heatmap from XY Data:
Gaussian:
image
Density:
image

Note these are not generated from the same data, these are randomized.

## Background Image:
image

**Autoformat your code:**
```
cd ScottPlot/src/
dotnet format
```


Benny121221:
I didn't add any colourmaps because I looked at how many you had in your spectogram repo and had a stroke. I assume they are generated programmatically, but still. If we add that many I think we should ditch passing them in by enum and just pass an instance of the class (like VectorField does it). Otherwise you'll have a page of switch/case.

As for axes, that sounds like a hefty refactor, and I think heatmaps need axes less than most plot types anyways. I'm not sure if the refactor is likely to be worth the effort.

swharden:
> I didn't add any colourmaps because I looked at how many you had in your spectogram repo and had a stroke. I assume they are generated programmatically, but still. If we add that many I think we should ditch passing them in by enum and just pass an instance of the class (like VectorField does it). Otherwise you'll have a page of switch/case.

Ha, yeah, I'm still working on this. Because of licensing issues with the originals I think I need to re-derive my own colormaps, which probably involves creation of a GUI tool to help this, but I think I can do it better if real curves (or bezier splines?) are used to generate the colormap. Hopefully by doing this I can create some truly open colormaps that can be easily used for other projects too, but I recognize this is a good bit of work and a potential distraction I shouldn't embark on immediately.

And yes, I generated those colormap classes programatically
https://github.com/swharden/Spectrogram/blob/8fa35b398fe3f9cd57bcc8560848009ad2c62848/dev/colormap/cmap2cs.py#L1-L104

> As for axes, that sounds like a hefty refactor, and I think heatmaps need axes less than most plot types anyways. I'm not sure if the refactor is likely to be worth the effort.

Up to you! It's definitely a feature we could implement in the future if we decide to leave it like it is for now.

swharden:
> Because of licensing issues ...

Turbo, viridis, etc are fine. It's the ones derived from "colorbrewer" palettes I feel are obnoxious, and I don't want the end users of ScottPlot to have to carry around colorbrewer license files like matplotlib does 😝 

https://github.com/matplotlib/matplotlib/tree/master/LICENSE

Benny121221:
> Up to you! It's definitely a feature we could implement in the future if we decide to leave it like it is for now.

I'll leave it for now. I just added a parameter to disable the labels because that seems like a good idea.

swharden:
Possible license solution:

Plotly has fantastic colormaps. It seems they are MIT licensed.

https://youtu.be/o9KxYxROSgM

Also, apparently there's a plotcon? 🤣

swharden:
I think cmocean is MIT too, even from matplotlib

https://matplotlib.org/cmocean/

Benny121221:
cmocean is MIT, and so is plotly, but some of plotly's colourmaps are under different licenses.

Her talk was pretty interesting, I think her dissolved O2 one highlights the need for users to make custom colourmaps. I think that specific one is part of cmocean but I wouldn't have thought that there is so much domain-specific nuance. I always wondered why matplotlib has 20 million different colourmaps.

Benny121221:
Also I feel like we'll have to add some documentation for when to use each one

Benny121221:
I think the cookbook could use a little something like this:
image

Just so it's easy to find which colourmap you want without trying each one. I'll look at plotly colourmaps as well but I gotta take care of something right now.

In case you need to change anything (i.e. I made a mistake) the script I used is below:

```py
import sys, argparse

parser = argparse.ArgumentParser()
parser.add_argument('--name')
args = parser.parse_args()


#print(args.name)

cmap_raw = sys.stdin

cmap = []

for line in cmap_raw:
    cmap.append(list(map(lambda l: float(l), line.split(' '))))

output="""//This is a cmocean colormap
//All credit to Kristen Thyng
//This colormap is under the MIT License
//All cmocean colormaps are available at https://github.com/matplotlib/cmocean/tree/master/cmocean/rgb

namespace ScottPlot.Config.ColorMaps
{
"""

output += f"public class {args.name} : ColormapFromByteArray"
output += """
    {
        protected override byte[,] cmap { get { return cmaplocal; } }

        private static readonly byte[,] cmaplocal = {
"""

for i in cmap:
    output+= "\t\t\t{" + f" {str(round(i[0] * 255))}, {str(round(i[1] * 255))}, {str(round(i[2] * 255))} " +"},\n"

output = output[:-2] + "\n"  

output += """
        };
    }
}
"""

print(output)
```

And I used it like this:

```sh
python .\script.py --name Turbid > C:\Users\benny\source\repos\ScottPlot\src\ScottPlot\Config\ColorMaps\Turbid.cs
```

I'm sure I didn't do it the ***p y t h o n i c*** way, seeing as I rarely use python, so don't judge me.

Benny121221:
Two of the colourmaps (Curl and Delta) are snowflakes, they have 512 colours. Which means you should use this script instead:

```py
import sys, argparse

parser = argparse.ArgumentParser()
parser.add_argument('--name')
args = parser.parse_args()


#print(args.name)

cmap_raw = sys.stdin
cmap_list = []

cmap = []
line_count = 0

for line in cmap_raw:
	cmap_list.append(line)
	line_count+= 1


snowflake = line_count == 512 #Curl and Delta are snowflakes

i = 0
for line in cmap_list:
	if(not (snowflake and i % 2)):
		cmap.append(list(map(lambda l: float(l), line.split(' '))))
	i += 1

output="""//This is a cmocean colormap
//All credit to Kristen Thyng
//This colormap is under the MIT License
//All cmocean colormaps are available at https://github.com/matplotlib/cmocean/tree/master/cmocean/rgb

namespace ScottPlot.Config.ColorMaps
{
"""

output += f"public class {args.name} : ColormapFromByteArray"
output += """
    {
        protected override byte[,] cmap { get { return cmaplocal; } }

        private static readonly byte[,] cmaplocal = {
"""

for i in cmap:
    output+= "\t\t\t{" + f" {str(round(i[0] * 255))}, {str(round(i[1] * 255))}, {str(round(i[2] * 255))} " +"},\n"

output = output[:-2] + "\n"  

output += """
        };
    }
}
"""

print(output)

```

I also added a demo which shows all the colourmaps. The only thing is the colourmap demos display with height of 400, when realistically they only need a height of 100.

In the cookbook it looks like this:

image

In my opinion it looks a little comical. I modified the cookbook generator so it creates something like this: 

image

The WPF and Winforms demos are unchanged, they look like this:

image

I think the WPF and Winforms demos are fine but feel free to change them if you like. These demos are mostly for the cookbook anyways.

Benny121221:
I'm looking at [plotly's colourmaps](https://github.com/plotly/plotly.py/tree/master/packages/python/plotly/_plotly_utils/colors), they have these:
- Carto, for mapping, under CC3.0
- cmocean (already added) under MIT
- colorbrewer2, under Apache 2.0
- cyclical, under MIT
- diverging, MIT
- plotlyjs, MIT
- qualitative, MIT
- sequential, MIT

Note that there are a lot of duplicates, for example, qualitative includes colorbrewer and carto colourmaps. And their MIT licensed stuff also includes things like Viridis, which are not under MIT license (although because their license is so lax they can be relicensed under MIT)

Even better, they are in a weird format, they are given in terms of colourstops:
image
 
I couldn't find the code for interpolating between them but it'll be there somewhere. If it's a linear interpolation that wouldn't be so bad but given my experience with their javascript library it probably looks something like this:

image


And I am aware that the sigmoid function is not defined for vector inputs, I was worried that the equation might look to comprehensible.

swharden:
> I'm looking at plotly's colourmaps ...

@Benny121221 There's a lot of great content in this PR here, thanks for all your awesome work on this! I'm burred in axis/tick land this weekend, so I won't dive into the colormap code myself quote yet. 

I just saw that YouTube and thought of you and this PR. Unfortunately it looks like those colormaps aren't all as freely licensed or as simple as I initially thought! That's a bummer. Maybe a good strategy for now is to build in all the simple MIT-licensable colormaps we can find (from matplotlib, plotly, etc.) while also making it easy for users to input their own custom colormaps. 

A little custom tool (perhaps in the /dev folder?) to flip through all the available colormaps would be nice for testing. To populate a listbox of all colormaps and easily retrieve them later I implemented a [GetColormaps() method using reflection](https://github.com/swharden/Spectrogram/blob/d8cddf07eafee160aa720736aff500a03be56357/src/Spectrogram/Colormap.cs#L149-L160) which eliminates the need for if and switch statements as colormaps are added and removed.

Long term it may be nice to create a colormap creation tool in the /dev folder that uses splines (similar to the tool used to create Turbo). Such a tool could utilize plotting to let you visualize the R, G, B, and mean curves as you edit the colormap, as well as display the colormap itself (as a linear strip, and also as a sample heat map). We could use it to create some MIT-licensed colormaps designed to mimic (but not copy) the more restrictively-licensed common alternatives.

swharden:
Not sure how helpful it is, but messing around I created an impromptu colormap generator. It currently doesn't have a means to output values, but it's an easy starting point in case we decide to create original colormaps _inspired_ by ones with complex licenses. I'll upload the code now and tag this PR in the commit message.

It may also benefit from improved interpolation #459 or maybe something smarter than 5 trackbars 😅 

image

swharden:
Okay I'm getting pretty close bringing a bunch of colormap code into this PR. I ended-up deleting most of the colormaps in the spectrogram project, only keeping the ones with public domain (CC0) licenses, and re-created colormaps to mimic my favorite ones from matplotlib that had complex licenses. The new interpolation module #459 helped here.

### Greens

Matplotlib | ScottPlot
---|---
mplgreens|greens

```cs
// original colormap created by Scott W Harden released under a MIT license
private readonly int[] greensRGB =
{
    00015360, 00015361, 00015618, 00015875, 00016132, 16728069, 16728327, 16728584, 
    16728841, 16728842, 16663563, 16663821, 16664078, 16664335, 16664592, 16599313, 
    16599570, 16599572, 16599829, 16600086, 16600343, 16600600, 16600857, 16535579, 
    16535836, 16535837, 16536094, 16536351, 16536608, 16536865, 16537122, 16537124, 
    16537381, 16537638, 16537895, 16538152, 16538409, 16603946, 16604203, 16604460, 
    16604717, 16604974, 16604975, 16670768, 16671026, 16671283, 16737076, 16737077, 
    16737334, 00025911, 00026168, 00026425, 00026425, 00026682, 00092475, 00092732, 
    00158269, 00224062, 00224319, 00289856, 00355649, 00355906, 00421699, 00487236, 
    00487492, 00553285, 00619078, 00684615, 00750408, 00816201, 00881737, 00947530, 
    01013323, 01079116, 01144653, 01210445, 01276238, 01341775, 01407568, 01473360, 
    01539153, 01604690, 01736018, 01801811, 01867604, 01933140, 02064469, 02130262, 
    02196054, 02327383, 02392920, 02458712, 02590041, 02655833, 02787162, 02852699, 
    02984027, 03049820, 03181148, 03312477, 03378269, 03509598, 03640927, 03706463, 
    03837792, 03969120, 04034912, 04166241, 04297569, 04428898, 04494690, 04626019, 
    04757347, 04888676, 05020004, 05151333, 05282661, 05348453, 05479782, 05611110, 
    05742439, 05873767, 06005095, 06136424, 06267752, 06399081, 06530409, 06596201, 
    06727786, 06859114, 06990443, 07121771, 07253099, 07384428, 07515756, 07581549, 
    07712877, 07844205, 07975534, 08106862, 08172655, 08304239, 08435567, 08566896, 
    08632688, 08764017, 08895345, 08961138, 09092466, 09158259, 09289587, 09420916, 
    09486708, 09618293, 09684085, 09749878, 09881206, 09946999, 10012791, 10144120, 
    10209912, 10275705, 10341497, 10472826, 10538619, 10604411, 10670204, 10735996, 
    10801789, 10867582, 10933374, 10999167, 11064960, 11130753, 11196545, 11262082, 
    11327875, 11393668, 11459460, 11525253, 11591046, 11656839, 11722632, 11788169, 
    11853962, 11919755, 11985547, 11985804, 12051597, 12117134, 12182928, 12248721, 
    12314514, 12380051, 12445844, 12446101, 12511638, 12577431, 12643225, 12708762, 
    12774555, 12840348, 12905886, 12971679, 13037473, 13037474, 13103267, 13168805, 
    13234598, 13300136, 13365929, 13431467, 13497260, 13562798, 13628592, 13694129, 
    13759667, 13890997, 13956534, 14022328, 14087866, 14153404, 14219197, 14284735, 
    14350273, 14416067, 14481605, 14547142, 14678472, 14744010, 14809548, 14875086, 
    14940880, 15006418, 15071956, 15203030, 15268824, 15334362, 15399900, 15465438, 
    15596768, 15662306, 15727844, 15793382, 15858920, 15924714, 16055788, 16121326, 
    16186864, 16252402, 16318196, 16449270, 16514808, 16580346, 16645884, 
}
```

### Blues

Matplotlib | ScottPlot
---|---
mplblues|blues

```cs
// original colormap created by Scott W Harden released under a MIT license
private readonly int[] bluesRGB =
{
    00011077, 00011078, 00011336, 00011594, 00011851, 00012109, 00012111, 00012369, 
    00012626, 16724564, 16724822, 16725079, 16725081, 16725339, 16725596, 16725854, 
    16726112, 16726114, 16726371, 16661093, 16661351, 16661608, 16661866, 16662124, 
    16662125, 16662383, 16662640, 16662898, 16663156, 16663413, 16663415, 16663673, 
    16663930, 16664188, 16664445, 16664703, 16664960, 16665218, 16665476, 16665477, 
    16731271, 16731528, 16731786, 16732043, 16732301, 16732558, 16732815, 00021393, 
    00021650, 00021908, 00022165, 00022423, 00022424, 00022681, 00022939, 00088732, 
    00088989, 00089247, 00155040, 00155297, 00155554, 00221348, 00221605, 00287398, 
    00287655, 00353449, 00353706, 00419499, 00419756, 00485549, 00485806, 00551599, 
    00617392, 00617650, 00683443, 00749236, 00749493, 00815286, 00881079, 00946872, 
    01012665, 01012922, 01078715, 01144508, 01210301, 01276093, 01341886, 01407679, 
    01473472, 01539265, 01605058, 01736387, 01802179, 01867972, 01933509, 02064838, 
    02130630, 02196423, 02327752, 02393545, 02459337, 02590666, 02721995, 02787531, 
    02918860, 02984652, 03115981, 03247309, 03312846, 03444175, 03575503, 03706832, 
    03837904, 03903697, 04035025, 04166354, 04297426, 04428754, 04560083, 04691411, 
    04822484, 04953812, 05085141, 05216213, 05347541, 05478870, 05610198, 05741270, 
    05872599, 06003927, 06134999, 06266328, 06397656, 06528728, 06660057, 06791385, 
    06922457, 07053785, 07185114, 07316442, 07447514, 07578843, 07710171, 07841243, 
    07972571, 08103900, 08234972, 08366300, 08497628, 08563421, 08694493, 08825821, 
    08957149, 09088478, 09154014, 09285342, 09351134, 09482462, 09613535, 09679327, 
    09810655, 09876447, 10007520, 10073312, 10204640, 10270432, 10336225, 10467297, 
    10533089, 10598881, 10730210, 10796002, 10861794, 10927331, 11058659, 11124451, 
    11190243, 11256036, 11321828, 11387364, 11518692, 11584485, 11650277, 11716069, 
    11781861, 11847398, 11913190, 11978982, 12044775, 12110567, 12176359, 12242151, 
    12307688, 12373480, 12439272, 12505065, 12570857, 12636649, 12702186, 12767978, 
    12833770, 12899563, 12965355, 13031147, 13162476, 13228012, 13293804, 13359596, 
    13425389, 13491181, 13556717, 13622510, 13688302, 13754095, 13819887, 13885423, 
    13951216, 14017008, 14082800, 14148593, 14214129, 14279921, 14345714, 14411506, 
    14477298, 14608371, 14674163, 14739956, 14805748, 14871540, 14937077, 15002869, 
    15068661, 15134454, 15199990, 15265783, 15397111, 15462903, 15528696, 15594232, 
    15660025, 15725817, 15791609, 15857146, 15922938, 15988731, 16120059, 16185595, 
    16251388, 16317180, 16382973, 16448509, 16514301, 16580094, 16645886, 
}
```

src/ScottPlot/Tools.cs
```diff
@@ -322,6 +322,15 @@ public static string GetOsName(bool details = true)
 
             return name;
         }
+      
+        public static int SimpleHash(double[] input)
+        {
+            byte[] bytes = input.SelectMany(n => { return BitConverter.GetBytes(n); }).ToArray();
+            int hash = 0;
+            foreach (byte b in bytes)
+                hash = (hash * 31) ^ b;
```
Benny121221, line 331: Double.GetHashCode() might be more convenient.

Of course C# has plenty of other hashing algorithms if you feel like it:

image

swharden:
I finally released Spectrogram 1.2 earlier this week which has a nice colormap module. I'm going to mimic that module in ScottPlot. Here's the gist:

* [IColormap.cs](https://github.com/swharden/Spectrogram/blob/master/src/Spectrogram/IColormap.cs) is extremely simple
* Colormaps get dropped in the [Colormaps](https://github.com/swharden/Spectrogram/tree/master/src/Spectrogram/Colormaps) namespace
* The [Colormap](https://github.com/swharden/Spectrogram/blob/master/src/Spectrogram/Colormap.cs) class does all the clever stuff, including listing all the available colormaps using reflection. Common colormaps are coded-in as static properties for quick access, and this file shouldn't require modification in the future. 

I'll start working on this PR this weekend to adapt it to this new colormap structure, pulling-in a lot of the features you wrote for advanced heatmap coloring. It'll probably be easiest if only I make commits to this PR over the next day or two to avoid merge conflicts...

src/ScottPlot/Tools.cs
swharden: Good points! However, I added this to help that an array created by a C# program was identical to one created by a Python program, so this seemed like a simple solution that didn't depend on an an external library or language-specific hashing mechanism.

As an aside, I'm aware that having a "tools", "utils", or "misc" class is a code smell 😅 

Note the MD5 method is used to create Bitmap hashes:
https://github.com/swharden/ScottPlot/blob/85cc7b022e2d26db3ee08933d9614013766f4485/src/ScottPlot/Tools.cs#L52-L61

src/ScottPlot.Demo/ScottPlot.Demo.csproj
```diff
@@ -11,4 +11,10 @@
     <ProjectReference Include="..\ScottPlot\ScottPlot.csproj" />
   </ItemGroup>
 
+  <ItemGroup>
+    <None Update="Images\niceBackground.bmp">
+      <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
+    </None>
+  </ItemGroup>
+
```
swharden, line 19: Rather than distribute a binary file along with this project, perhaps we could generate a nice background Bitmap procedurally. This could also be used by tests in the future, and wouldn't have to get copied from project to project 🤔 

This comes to mind when I think of test textures 🙃
image

src/ScottPlot/Config/ColorMaps/ColormapFromByteArray.cs
```diff
@@ -21,9 +22,9 @@ public abstract class ColormapFromByteArray : Colormap
             return output;
         }
 
-        public override int[] IntensitiesToARGB(double[] intensities)
+        public override int[] IntensitiesToARGB(double[] intensities, double? transparencyThreshold = null)
         {
-            return intensities.AsParallel().AsOrdered().Select(i => RGBToARGB(new byte[] { cmap[(int)(i * 255), 0], cmap[(int)(i * 255), 1], cmap[(int)(i * 255), 2] })).ToArray();
+            return intensities.AsParallel().AsOrdered().Select(i => i < (transparencyThreshold ?? double.NegativeInfinity) ? unchecked((int)0x00000000) : RGBToARGB(new byte[] { cmap[(int)(i * 255), 0], cmap[(int)(i * 255), 1], cmap[(int)(i * 255), 2] })).ToArray();
```
swharden, line 27: I'm a bit confused with this one 🤔 why is the 0 unchecked?

```cs
return intensities
    .AsParallel()
    .AsOrdered()
    .Select(i => i < (transparencyThreshold ?? double.NegativeInfinity) ? 
        unchecked((int)0x00000000) : 
        RGBToARGB(new byte[] { /* cmap RGB values */ }))
    .ToArray();
```

Benny121221: I don't know. It should just be this:
```cs
Select(i => i < (transparencyThreshold ?? double.NegativeInfinity) ? 
        unchecked((int)0x00000000) : 
        RGBToARGB(new byte[] { /* cmap RGB values */ }))
```

I suspect that I copied this without thinking from somewhere which used a constant like this: `0xFFxxxxxx`, and the compiler treats that constant as a long because the largest value an int can hold is `0x7FFFFFFF` because of the sign bit. The unchecked was necessary because it would overflow (intentionally, as ARGBs are unsigned despite being represented as signed ints in .NET). I solved that problem by just writing (0xFF << 24), as bitshifts are not caught by compile-time checking.

So, short story, let the strange code die.

swharden: Sounds good! It made its way into a few files but I'll update this during the colormap refactor 👍 

Thanks for all your work on this! These heatmaps are looking awesome

src/ScottPlot.Demo/ScottPlot.Demo.csproj
Benny121221: Ah, the hammer editor, the bane of my sanity.

I don't know that distributing images with the demo app is too bad, but I could see that it might be useful outside of the demo app, perhaps as part of `ScottPlot.Datagen`. And distributing images (actually anything that isn't .NET Code, even native DLLs) as part of an assembly is often a pain.

swharden:
I took the colormap demo out of the cookbook and gave it its own form
image

I probably won't make the WPF version until heatmaps development is finished. I think this PR is about ready to merge...

Benny121221:
There were a few issues with the colourmap viewer, the big one was it would crash if minimized. There was also an issue which could cause the zoom and axis of the heatmap to be broken if you switched between the image and the gaussian sample data. I fixed them both.

Benny121221:
I'm not a fan of getting rid of the official turbo colourmap, I know you're not a fan of having multiple license documents, but I don't think the NOTICE file was too bad. Regardless, it may be worth ironing out a few inconsistencies:

Original Turbo:
image

Your copy:
image

The differences aren't profound, but they are noticeable. Your greens are darker and take up more of the colourmap. Your oranges also start later and have a peachey tinge.

Benny121221:
> The differences aren't profound, but they are noticeable. Your greens are darker and take up more of the colourmap. Your oranges also start later and have a peachey tinge.

I mean this not as a criticism, but I think trying to mimic a colourmap is going to be very difficult to get perfect, and it would likely be easier to just leave the notice in. The differences probably wouldn't cause any trouble if someone was using only ScottPlot, however, if someone used matplotlib and ScottPlot, or even if they were just comparing data with someone who used another plotting library it could get confusing.

swharden:
Here's my line of thinking on Turbo, and I welcome objective criticism!

It's not difficult for us just to use the real Turbo and add a NOTICE file. However, GitHub says this package is used by 97 others, so I want to make sure nothing we do results in those 97 others having to create "this product contains a modified version of the Turbo colormap" notice file in each of their repos. According to my understanding, that notice file then has to get carried along the build chain, through packaging, and all the way to the end-user, so some the laptop of some person using a stock market analysis app that that has a ScottPlot in it somewhere has to have this notice file on their hard drive or in the application documentation. Including this notice is a small inconvenience, but it ***scales***, so I wish to avoid it.

My goal in mimicking Turbo wasn't to duplicate it, but rather to provide something that was similar to but not derived from it. Copying the magic formula from the [polynomial approximation](https://gist.github.com/mikhailov-work/0d177465a8151eb6ede1768d51d476c7) would have been the simplest way to duplicate it. If you can think of something we can do to make it better (or think we should just remove it), I'm interested in what you think! I'm also colorblind, so maybe I'm not the best person to be trying to mimic a colormap by eye 🙃

swharden:
> There were a few issues with the colourmap viewer, the big one was it would crash if minimized. There was also an issue which could cause the zoom and axis of the heatmap to be broken if you switched between the image and the gaussian sample data. I fixed them both.

Thanks for fixing these! I keep forgetting zero-pixel-size Bitmap instantiation throws an exception 🙈 Great catch!

Benny121221:
>  I want to make sure nothing we do results in those 97 others having to create "this product contains a modified version of the Turbo colormap" notice file in each of their repos

The relevant portion of the Apache 2.0 License is below. Note that "Work or Derivative Works" include both source and object form.

> 4. Redistribution. You may reproduce and distribute copies of the Work or Derivative Works thereof in any medium, with or without modifications, and in Source or Object form, provided that You meet the following conditions:
>
> 
> a) You must give any other recipients of the Work or Derivative Works a copy of this License; and
> 
> b) You must cause any modified files to carry prominent notices stating that You changed the files; and
> 
> c) You must retain, in the Source form of any Derivative Works that You distribute, all copyright, patent, trademark, and attribution notices from the Source form of the Work, excluding those notices that do not pertain to any part of the Derivative Works; and
> 
> d) If the Work includes a "NOTICE" text file as part of its distribution, then any Derivative Works that You distribute must include a readable copy of the attribution notices contained within such NOTICE file, excluding those notices that do not pertain to any part of the Derivative Works, in at least one of the following places: within a NOTICE text file distributed as part of the Derivative Works; within the Source form or documentation, if provided along with the Derivative Works; or, within a display generated by the Derivative Works, if and wherever such third-party notices normally appear. The contents of the NOTICE file are for informational purposes only and do not modify the License. You may add Your own attribution notices within Derivative Works that You distribute, alongside or as an addendum to the NOTICE text from the Work, provided that such additional attribution notices cannot be construed as modifying the License.
> 
> You may add Your own copyright statement to Your modifications and may provide additional or different license terms and conditions for use, reproduction, or distribution of Your modifications, or for any such Derivative Works as a whole, provided Your use, reproduction, and distribution of the Work otherwise complies with the conditions stated in this License.

The sticklers are 4a) and 4d) as they apply to source and object files.

4a means the github repo and NuGet both need to have copies of the license

4d means we do not need to include the notice file as I had thought, as the original does not include one.

In my opinion, distributing the license through NuGet isn't too bad. What would be a real pain is if Steve makes a project which uses ScottPlot and they were required to distribute the license as well. I think a derivative work of ScottPlot would unfortunately count as a derivative work of Turbo. The definition of derivative work is below:

> "Derivative Works" shall mean any work, whether in Source or Object form, that is based on (or derived from) the Work and for which the editorial revisions, annotations, elaborations, or other modifications represent, as a whole, an original work of authorship. For the purposes of this License, Derivative Works shall not include works that remain separable from, or merely link (or bind by name) to the interfaces of, the Work and Derivative Works thereof.

I do have a sneaky way around this but I have a feeling you won't like it (as I don't like it either). You could create another NuGet package which adds Turbo (and any other colourmap with potentially problematic licensing) and clearly outlines the obligations of anyone using the package.

Or you could see if you could get permission to use it under a proprietary license. This seems unrealistic however, as the copyright is to Google LLC, not Anton Mihailov. I'm sure it is worth a try but I wouldn't hold your breath.

Benny121221:
> My goal in mimicking Turbo wasn't to duplicate it, but rather to provide something that was similar to but not derived from it. Copying the magic formula from the polynomial approximation would have been the simplest way to duplicate it. If you can think of something we can do to make it better (or think we should just remove it), I'm interested in what you think! I'm also colorblind, so maybe I'm not the best person to be trying to mimic a colormap by eye 🙃

As for mimicking it, I don't have any ideas other than brute force. You could modify the colourmap generator to have 80 stops and it would probably be identical. But that seems unreasonable. In my opinion it is very close as it is and it would be accurate enough with a few extra stops.

swharden:
> > My goal in mimicking Turbo wasn't to duplicate it, but rather to provide something that was similar to but not derived from it. Copying the magic formula from the polynomial approximation would have been the simplest way to duplicate it. If you can think of something we can do to make it better (or think we should just remove it), I'm interested in what you think! I'm also colorblind, so maybe I'm not the best person to be trying to mimic a colormap by eye 🙃
> 
> As for mimicking it, I don't have any ideas other than brute force. You could modify the colourmap generator to have 80 stops and it would probably be identical. But that seems unreasonable. In my opinion it is very close as it is and it would be accurate enough with a few extra stops.

A few extra spline interpolation stops probably would do the trick. This would require some pretty significant modifications to the spline interpolation colormap generator application (in the dev folder), so rather than halt heatmap progress to work on a colormap generator application I'll go ahead and merge this in and make a note to come back later and create better tools for custom colormap generation. Such a program probably deserves its own project/repository independent of ScotPlot...

Thanks again for all your help on these heatmap and colormap features! This is turning out to be really impressive

swharden:
@Benny121221 I almost missed all the excellent notes you made in https://github.com/swharden/ScottPlot/pull/458#issuecomment-647165266

> If Steve makes a project which uses ScottPlot (a derivative work) it would unfortunately count as a derivative work of Turbo, requiring Steve to distribute Turbo's Apache 2.0 license as well

I'm reassured that your reading of the license resulted in a similar conclusion with mine! Also, what a pain that license is turning out to be. 

After this saga, I _really_ think there's a potential market for a tool (perhaps a SPA?) where users can (1) copy/paste common MIT-licensed colormaps into their language of choice, and (2) use sliders to create their own custom colormaps. Turbo is a nice colormap, but truly free no-obligation resources are better 🤔 

We could invent the next Turbo! TurboCharge colormap? Actually it's free so TurboNoCharge 😜

Benny121221:
> After this saga, I really think there's a potential market for a tool (perhaps a SPA?) where users can (1) copy/paste common MIT-licensed colormaps into their language of choice, and (2) use sliders to create their own custom colormaps. Turbo is a nice colormap, but truly free no-obligation resources are better 🤔

I'm intrigued, and I have a lot of experience with SPAs, so let me know if you want any help.

> We could invent the next Turbo! TurboCharge colormap? Actually it's free so TurboNoCharge 😜

Why not SuperCharge, better low-end torque?

swharden:
> > After this saga, I really think there's a potential market for a tool (perhaps a SPA?) where users can (1) copy/paste common MIT-licensed colormaps into their language of choice, and (2) use sliders to create their own custom colormaps. Turbo is a nice colormap, but truly free no-obligation resources are better 🤔
> 
> I'm intrigued, and I have a lot of experience with SPAs, so let me know if you want any help.

I'm intrigued at the prospect, but considering how much I've got on my plate these days it's something I know will take a few months before I can get to it. I'll keep noodling the idea and if/when it starts moving I'll let you know! It seems like a super specific tool, probably only useful to a small sliver of library developers, so it's hard to assess its impact or true utility. It seems like 99.9% of people can just grab an existing colormap and move on.

> > We could invent the next Turbo! TurboCharge colormap? Actually it's free so TurboNoCharge 😜
> 
> Why not SuperCharge, better low-end torque?

Ha, I like it! "SurperCharge, the public domain alternative to Google's Turbo colormap".

Issue #461, opened 2020-06-15
Bar Plots and Negative Values


Benny121221:
**Describe the bug**
Bar Plots do not play nice with negative values:
- Unless panned/zoomed, the plot will only show y values >0 (Quadrant 1, assuming non-negative x-values)
![image](https://user-images.githubusercontent.com/8635304/84607723-2cd3cb00-ae6c-11ea-877e-589dd11c1871.png)
- Error bars are drawn from the topmost point, which for negative values is 0
![image](https://user-images.githubusercontent.com/8635304/84607734-35c49c80-ae6c-11ea-93ef-75c581dd55cf.png)

Negative values on bar plots are atypical, but I was looking at making waterfall charts:
![image](https://user-images.githubusercontent.com/8635304/84607809-9227bc00-ae6c-11ea-8ab9-ba03c6df4f8f.png)

You might recognize waterfall charts as a generalization of a Gantt chart:
![image](https://user-images.githubusercontent.com/8635304/84607827-ae2b5d80-ae6c-11ea-9504-eff3e8d012b1.png)

I will probably work on this sometime soon-ish. I noticed it because I was thinking about implementing waterfall charts as bar charts with an extra `std::shared_ptr<double> offsets` parameter, but the handling of negative values put a damper on things.

**To Reproduce**
Create a bar plot with negative y values.
```cs
//xs, ys, yError
var barPlot = plt.PlotBar(new double[]{1,2,3}, new double[]{1,0,-1}, new double[]{1,1,1});
```

**Screenshots**
I embedded them in the list :/

Benny121221:
> ... with an extra `std::shared_ptr<double>` offsets parameter ...

Yeesh, I've been doing too much C++, I meant `double[] offsets`

Benny121221:
Also note, negative x values cause no issues

swharden:
Fixed in #462, thanks @Benny121221!

PR #462, opened 2020-06-15
fixed issues w/ barplots and negative values

Benny121221:

**Purpose:**
#461

**New functionality (code):**
N/A

Now bar plots with negative values look as they should:
image

**Autoformat your code:**
```
cd ScottPlot/src/
dotnet format
```

### DIFF
--- a/src/ScottPlot/Plot.cs
+++ b/src/ScottPlot/Plot.cs
@@ -12,6 +12,7 @@ using System.Drawing;
 using System.Drawing.Drawing2D;
 using System.Drawing.Imaging;
 using System.Linq;
+using ScottPlot.Config;
 using ScottPlot.Statistics;
 
 namespace ScottPlot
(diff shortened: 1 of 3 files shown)


src/ScottPlot/Plot.cs
```diff
@@ -1089,7 +1089,7 @@ public void Clear(Predicate<Plottable> plottablesToClear)
 
             settings.plottables.Add(barPlot);
 
-            if (autoAxis)
+            if (autoAxis && !ys.Where(y => y < 0).Any())
```
swharden, line 1092: Is this intentional? I'll probably remove it... my thinking is that if the user set the `autoAxis` argument to `true`, `AxisAuto()` should get called regardless of Y values

swharden: Scratch that - I see the issue now! When negative Y values are used, y2 (not y1) should be set to zero.

src/ScottPlot/Plot.cs
```diff
@@ -1156,7 +1156,7 @@ public void Clear(Predicate<Plottable> plottablesToClear)
                 double[] barYs = ys[i];
                 double[] barYerr = yErr?[i];
                 double[] barXs = DataGen.Consecutive(barYs.Length);
-                containsNegativeY = containsNegativeY ? containsNegativeY : barYs.Where(y => y < 0).Any();
+                containsNegativeY |= barYs.Where(y => y < 0).Any();
```
Benny121221, line 1159: I didn't know C# supported bitwise on booleans? Intriguing, I suspect somewhere buried deep in the MSIL there's a macro like this:

`#define true 1`
`#define false 0`

i.e. C# bools are not truly primitive types, but macros, as is the case in C, and is sort of true in C++.

swharden: I've been doing a lot of bit operations lately (working on audio) and just learned about this one myself!

Don't forget `bool?` which can hold 3 possible states 😝 🙈

Issue #469, opened 2020-06-17
Allow Axis Labels in different radices


Benny121221:
**What can we do to make ScottPlot better?**
In some domains number systems other than decimal are appropriate. For example hexadecimal in Computer Science and Electrical Engineering (and to a lesser extent octal).

You would think this would be provided by the overload which allows the ticks to take format strings, however, the hexadecimal `x` format specifier is only valid for integral types and the ticks are doubles.

I suggest something like this:
```cs
plt.TicksRadixX(int radix = 16, string prefix = "0x");
plt.TicksRadixY(int radix = 16, string prefix = "0x");
```

Another notation for non-decimal radices is to put the base in subscript after the number: ![image](https://render.githubusercontent.com/render/math?math=\mathbf{FFFFFFFF_{16}}&mode=inline)

The prefix notation is going to be easier to implement, even though it doesn't generalize to other radices (except octal with prefix "0" and binary "0b"). In reality bases other than decimal and hex are rare. I'm unsure if the prefix should be on by default or not.

Check the ScottPlot Roadmap to see if your feature is already planned for an upcoming release:
https://github.com/swharden/ScottPlot/blob/master/dev/roadmap.md

swharden:
Thanks @Benny121221! I'll merge #475 in today

PR #475, opened 2020-06-23
Radices

Benny121221:

**Purpose:**
#469 

**New functionality (code):**

```cs
//Base 10 on x axis, base 16 on y axis
int[] radices = { 10, 16 };
string[] prefices = { "", "0x" };

plt.Ticks(radices: radices, prefices: prefices);
```

It also adds the `ScottPlot.Tools.ToDifferentBase` method:

```cs
 public static string ToDifferentBase(double number, int radix = 16, int decimalPlaces = 3, int padInteger = 0, bool dropTrailingZeroes = true)
```

There are some tests for this method in `ScottPlotTests.Tools.ChangingBase`

image

**Autoformat your code:**
```
cd ScottPlot/src/
dotnet format
```

### DIFF
--- a/src/ScottPlot.Demo/Customize/Axis.cs
+++ b/src/ScottPlot.Demo/Customize/Axis.cs
@@ -161,17 +161,40 @@ namespace ScottPlot.Demo.Customize
             public void Render(Plot plt)
             {
                 Random rand = new Random(0);
-                int pointCount = 15;
-                double[] x = DataGen.Consecutive(pointCount);
-                double[] y = DataGen.Consecutive(pointCount);
 
-                plt.PlotScatter(x, y);
+                // create some sample data
+                double[] xs = { 0 };
+                double[] valuesA = { 0x40000000 };
+                double[] valuesB = { 0x40100000 };
+                double[] valuesC = { 0xA0000000 };
+
+
+                // to simulate stacking B on A, shift B up by A
+                double[] valuesB2 = new double[valuesB.Length];
+                double[] valuesC2 = new double[valuesC.Length];
+                for (int i = 0; i < valuesB.Length; i++)
+                {
+                    valuesB2[i] = valuesA[i] + valuesB[i];
+                    valuesC2[i] = valuesC[i] + valuesB2[i];
+                }
+
+                // plot the bar charts in reverse order (highest first)
+                plt.PlotBar(xs, valuesC2, label: "Process C");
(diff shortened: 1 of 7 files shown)


swharden:
Wow @Benny121221, what an interesting idea! And clean implementation (considering the tick system is a little clunky). Thanks for this!

I'd like to add a title, vertical axis label, and horizontal axis label to the demo to make the use case more obvious. What do you think a good labels would be to communicate when this would be used in practice? Memory addresses come to mind, but perhaps those would go on the horizontal axis, so I'm interested in your thoughts about what would make a good demo use case.

Benny121221:
I think a good demo would be stacked bar-plots to show memory usage (e.g. 0x40000000 to 0x400F0000 are allocated to process A, 0x40100000 to 0x50000000 are for process B, etc.

I admit my demo was a little lazy and unimaginative, seeing as I just drew a straight line.

swharden:
> I think a good demo would be stacked bar-plots to show memory usage (e.g. 0x40000000 to 0x400F0000 are allocated to process A, 0x40100000 to 0x50000000 are for process B, etc.
> 
> I admit my demo was a little lazy and unimaginative, seeing as I just drew a straight line.

Sounds good to me! I'll let you update the demo to reflect the multi-process memory idea, then let me know when you think this PR is done and ready to merge

Benny121221:
image

The demo isn't quite as neat as I thought it would be, I think partly because the tick system favours nice numbers in base 10, not in base 16. I don't think this is really worth fixing, especially as there aren't really any nice numbers in base 16 other than full bytes. I expect anyone who was so nitpicky would label the axis themselves anyways.

I think this is all that's needed for this PR, so feel free to merge

swharden:
I'm about to merge this in, 2 thoughts though.
* Should we throw if the base is >16? `char[] symbols = "0123456789ABCDEF".ToCharArray();` makes me think it would crash anyway, so maybe we should make it crash up-front with a descriptive exception message
* Negative values always show as 0. Is this intentional?

Benny121221:
> - Should we throw if the base is >16? 

Yeah we probably should.

>  - Negative values always show as 0. Is this intentional?

No that is not intentional, good catch. I should have included tests with negative values. This does raise a problem as some bases (binary) have multiple ways of writing negatives (one's complement, two's complement, sign and magnitude, etc). I think showing that would be confusing and harmful, so I think it makes the most sense to prepend a negative sign.

Issue #463, opened 2020-06-15
Waterfall Charts


Benny121221:
**What can we do to make ScottPlot better?**
Waterfall charts:
![image](https://user-images.githubusercontent.com/8635304/84610454-e1bfb500-ae77-11ea-8f45-00f8ec36835e.png)

Does your proposed feature allow users to do new things in source code? If so, show us how you'd like to be able to use the new feature:

```cs
var waterfallPlot = plt.PlotWaterfall(double[] xs, double[] ys, double[] yOffsets)
```

I am undecided if it should be done like above, or as an added parameter for `PlotBar`:
```cs
var waterfallPlot = plt.PlotBar(double[] xs, double[] ys, double[] yOffsets)
```

Check the ScottPlot Roadmap to see if your feature is already planned for an upcoming release:
https://github.com/swharden/ScottPlot/blob/master/dev/roadmap.md

swharden:
Looks great, thanks @Benny121221!

![image](https://user-images.githubusercontent.com/4165489/86537227-6754e480-bebb-11ea-9bd9-13cd3d9efc64.png)

PR #476, opened 2020-06-23
bar plot support for Y offset

Benny121221:

**Purpose:**
#463

**New functionality (code):**

```cs
Random rand = new Random(0);
int pointCount = 10;
double[] xs = DataGen.Consecutive(pointCount);
double[] ys = DataGen.RandomNormal(rand, pointCount, 5, 10);
double[] yOffsets = Enumerable.Range(0, pointCount).Select(count => ys.Take(count).Sum()).ToArray();

plt.PlotBar(xs, ys, yOffsets: yOffsets);
```

image

# Opinion Wanted:
I wonder if `PlottableBar` should take a delegate to decide the colour of the bar, something like this:

```cs
Func<double y1, double y2, System.Drawing.Color> delegate = y2 - y1 > 0 ? System.Drawing.Color.Green : System.Drawing.Color.Red;
```

I photoshopped it to emulate the behaviour:
image

I like this personally, however this seems niche and could cause trouble for the legend, as one plottable would be multiple colours.

**Autoformat your code:**
```
cd ScottPlot/src/
dotnet format
```

### DIFF
--- a/src/ScottPlot.Demo/PlotTypes/Bar.cs
+++ b/src/ScottPlot.Demo/PlotTypes/Bar.cs
@@ -1,5 +1,6 @@
 ﻿using System;
 using System.Collections.Generic;
+using System.Linq;
 using System.Text;
 
 namespace ScottPlot.Demo.PlotTypes
@@ -138,6 +139,24 @@ namespace ScottPlot.Demo.PlotTypes
                 // customize the plot to make it look nicer
                 plt.Grid(enableVertical: false, lineStyle: LineStyle.Dot);
                 plt.Legend();
+            }
+        }
+
+        public class Waterfall : PlotDemo, IPlotDemo
+        {
+            public string name { get; } = "Waterfall Plot";
+            public string description { get; } = "You can use the yOffsets parameter to create a waterfall plot";
+
+            public void Render(Plot plt)
+            {
+                // generate random data to plot
+                Random rand = new Random(0);
+                int pointCount = 10;
+                double[] xs = DataGen.Consecutive(pointCount);
+                double[] ys = DataGen.RandomNormal(rand, pointCount, 5, 10);
+                double[] yOffsets = Enumerable.Range(0, pointCount).Select(count => ys.Take(count).Sum()).ToArray();
+
+                plt.PlotBar(xs, ys, yOffsets: yOffsets);
             }
         }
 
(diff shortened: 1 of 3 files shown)


swharden:
Hey @Benny121221, this is an excellent improvement! By adding support for Y offsets you enabled a whole new type of plot type (a waterfall plot).

This one-liner is super useful for creating waterfall plots from a series of Y values.
```cs
double[] yOffsets = Enumerable.Range(0, pointCount).Select(count => ys.Take(count).Sum()).ToArray();
```

What do you think about adding a little function which runs that line under the hood to create the waterfall plot so the user doesn't have to? Something like this (where most of the arguments taken are simply passed to PlotBar):
```cs
plt.PlotWaterfall(xs, ys);
```

Such a method could accept arguments (preferably not delegates though) to customize coloring, and not require modification of the PlotBar method. Perhaps something like:
```cs
plt.PlotWaterfall(xs, ys, Color.Green, Color.Red);
```

Benny121221:
> What do you think about adding a little function which runs that line under the hood to create the waterfall plot so the user doesn't have to? Something like this (where most of the arguments taken are simply passed to PlotBar):
```cs
plt.PlotWaterfall(xs, ys);
```

I like it, I am busy today but I'll certainly work on this when i get home or in the coming days.

swharden:
> I like it, I am busy today but I'll certainly work on this when i get home or in the coming days.

Sounds like a plan!

Benny121221:
I added your suggestions, it looks like this:

image

The demo code is the following:

// generate random data to plot
```cs
Random rand = new Random(0);
int pointCount = 12;
double[] xs = DataGen.Consecutive(pointCount);
double[] ys = DataGen.RandomNormal(rand, pointCount, 5, 10);

System.Drawing.Color profitColor = System.Drawing.Color.Green;
System.Drawing.Color lossColor = System.Drawing.Color.Red;
string[] monthNames = { "Jan", "Feb", "Mar", "Apr", "May", "Jun", "Jul", "Aug", "Sep", "Oct", "Nov", "Dec" };

plt.PlotWaterfall(xs, ys, fillColor: profitColor, negativeColor: lossColor);
plt.Title("Y1 Profits");
plt.XLabel("Month");
plt.XTicks(monthNames);
plt.YLabel("Million USD");
```

The `negativeColor` parameter has been added to `PlotBar` as well, and a user can still manually set `yOffsets` with `PlotBar` should they wish.

Issue #496, opened 2020-07-30
Avalonia Support?


Benny121221:
**Issue to be solved:**
Supporting Avalonia would allow interactive plots on Linux and Mac. If you're not familiar, Avalonia feels very similar to WPF, but with a few features absent. Avalonia is in pre-release, so that could well change.
* If this issue relates to an existing pull request, link to it here.

I have a branch on my fork with a working Avalonia UserControl. Because of the aforementioned absent features, as well as the interest in cross-platform compatibility, not all the same features that WPF and WinForms enjoy are present.

The branch is available here: https://github.com/Benny121221/ScottPlot/tree/avalonia

I can understand that you may be hesitant to support another platform, but I personally believe there is merit in being available on Linux and Mac. Plus, Avalonia feels *very* similar to WPF, most code can be copy-pasted and then find-and-replaced.

Let me know if you want me to open a PR. If you don't want to enhance your maintenance burden, then I'm happy to release an unofficial Avalonia UserControl, as I have a personal need for Linux support.
* If this issue relates to a patch, paste the contents of the patch here.

New contributors should review CONTRIBUTING.md:
https://github.com/swharden/ScottPlot/blob/master/CONTRIBUTING.md

Benny121221:
It turns out that Avalonia is more similar to WPF than I had thought, the codename for WPF was Avalon. Wild.

swharden:
Hey @Benny121221! 

I want to study this topic in a little more detail before I give a thorough response, but perhaps I'll outline my thinking at this point and get your reflection of my thought process. I'm lightly aware of Avalonia (https://avaloniaui.net/), and understand it is already being used to write cross-platform UIs for .NET applications. However, I'm also aware that later this year Microsoft is officially releasing MAUI (https://devblogs.microsoft.com/dotnet/introducing-net-multi-platform-app-ui/) which will be used to write cross-platform UIs for .NET applications. It looks to be a lot like Xamarin.Forms, which in my head I treat as "WPF for mobile".

Knowing MAUI is around the corner, I was planning on pursuing support for it when it's released. It looks like MAUI preview 1 is scheduled for late 2020, with a full release targeted at 2021. https://github.com/dotnet/maui/wiki/Roadmap 

What do you think about the future of Avalonia balanced with the emergence of MAUI?

[...]

PR #503, opened 2020-08-09
Avalonia

Benny121221:

**Purpose:**
Adds Avalonia UserControl.
#496 

**New functionality (code):**

It seems inappropriate to drop several files in here, but I have a repo which I used for testing: https://github.com/Benny121221/ScottPlotTesting/tree/master/src/AvaloniaTesting

You'll notice that Avalonia feels very similar to WPF. Note that you'll have to point to a built ScottPlot dll, otherwise it won't find the assembly.

image

**Autoformat your code:**
```
cd ScottPlot/src/
dotnet format
```


Benny121221:
Looks like I need to rebase :/

Benny121221:
Ah, you made some changes in 2a8c3cfd242b05f7de20f2f19aa378a70642d69e that complicated things, that's all

Benny121221:
I don't think that's my fault because I ran `dotnet-format` locally and it made no changes

swharden:
It's looking good! 

I haven't looked into it too deeply yet, but why doesn't the control resize when the window does? 🤔 

image

Benny121221:
Avalonia does not have a resize event. I noticed this at the time and realized the workaround could be annoying so I decided to leave it for later. I then forgot about it -_-

If you're interested, this is the relevant portion: https://github.com/swharden/ScottPlot/blob/cd6c59a6f0d4b502ba9a7c028730d386c348231e/src/ScottPlot.Avalonia/AvaPlot.axaml.cs#L82:L94

PR #504, opened 2020-08-10
PlottableSignal, PlottableSignalConst refactoring

StendProg:
**Purpose:**
`PlottableSignal` and `PlottableSignalConst` real difference is min/max search algoritms.
So I apply strategy pattern and move this calculations to `IMinMaxSearchStrategy`.
For now only one `PlottableSignalBase<T>` class, that accept `IMinMaxSearchStrategy`.
By passing `SegmentedTreeMinMaxSearchStrategy` we can get `SignalConst` behavior.
By passing `LinearMinMaxSearchStrategy` we get `Signal` behavior.

There are 2 classes to match existing API:
`PlottableSignalConst<T> : PlottableSignalBase<T>` - with `SegmentedTreeMinMaxSearchStrategy<T>`.
`PlottableSignal : PlottableSignalBase<double>` - with `LinearDoubleOnlyMinMaxSearchStrategy : IMinMaxSearchStrategy<double>`
This classes realy simple and it's behavior can be easy get in `Plot.Plot...` api methods`:
```c#
plt.PlotSignal()
{
return new PlottableBase<double>(... , minmaxSearchStrategy: new LinearDoubleOnlyMinMaxSearchStrategy());
}`
```
but it would be API breaking changes.
I think lots of people holding `Plot.PlotSignal()` results in `PlottableSignal` class (Demos make it 3-4 times).

So for now there is no API breaking changes, but there is way to simplify API.
There is one `PlottableSignalBase<T>` can be renamed to `PlottableSignal<>`.
And API which create it with differentFlags:
```c#
PlottableSignal<T> plot.PlottableSignal(..., slowUpdates = true);
or
PlottableSignal<T> plot.PlottableSignal(..., constantValues = true);
or something else
```
`Signal` can be generic, there is ready to use `LinearFastDoubleMinMaxSearchStrategy`, fast for `double` and 30% slower for other types.

Performance of new `PlottableSignal` (its now comes from generic base class, but use `double` math) needs thoroughly tested. I have not seen more than 5% deterioration in simple visual tests on demo. Old `PlottableSignal` moves to `PlottableSignalOld` and have `Plot.PlottableSignalOld` api method for test porpouse. Maybe you can help with testing performance.

This PR in `work in progress` status.
All this strategies should be covered with unit tests. And `PlottableSignalXY` will need to get `IMinMaxSearchStrategy` support (main reasone of this PR). Publish this early to get your feedback. This PR began to make a lot of changes, maybe I am going in wrong direction.

**New functionality (code):**

There is no API breaking changes for now.
```cs
Plot.PlottableSignalOld(ys);
```

The results are too encouraging, but in fact the test results are highly dependent on the order of the signals. Third is always the fastest (I cheated). I'm not sure if it is necessary to open a issue on the `dev/benchmark`, but there are questions to him.
image


swharden:
Hi @StendProg, this is an impressive PR! I can tell this has been a lot of work, and there are great potential improvements here. This code has been a joy to read! I have a few small comments/thoughts (e.g., what do you think about returning tuples instead of `out` arguments?), but before I discuss those details I'd like to have a larger scale discussion and seek your input.

### Strategy pattern

The task of searching arrays using these different algorithms is an excellent use of the strategy pattern! I really like how this refactor isolates tree logic in `Segmentedtree.cs`, isolates min/max searching logic in their own classes, and my favorite part is how graphics code is now isolated in `PlottableSignalBase.cs`. These changes greatly improve "separation of concerns" to improve maintainability and will improve testing too. I love it!

### Allow breaking changes in ScottPlot 4.1?

> there is way to simplify API ... but it would be a breaking change
> ```cs
> plt.PlotSignal(...)
> {
>     return new PlottableBase<double>(... , minmaxSearchStrategy: new LinearDoubleOnlyMinMaxSearchStrategy());
> }
> ```

ScottPlot 4.0 has become much more stable recently, with no large changes in several months. This may be a good time to start version 4.1 and allow any breaking changes we want. The NuGet packages can be published as "pre-release" while we experiment with these API changes to give us time to change them as needed after we start working with them. Because of this timing and the opportunity to start work on 4.1, I'm not at all opposed to the breaking changes in this PR.

### ScottPlot 4.0 -> 4.1 -> 4.2

@StendProg I have listed what I think would be a good set of small but important breaking changes I wish to make when we switch to 4.1, and I am very interested in your feedback.

My primary goals for breaking changes in 4.1 are: 
* `PlotSignal` and `PlotSignalXY` will return a generic object (this PR)
* make all renderable objects (axis labels, ticks, etc) their own classes (like plottables are now)
* make the figure and data area one bitmap (with cropping instead of two bitmaps)
* pass a simple `FigureDimensions` object into each plottable instead of a complex `Settings` object
* more carefully create and destroy all `IDisposable` System.Drawing objects (like brushes) with each render instead of holding them inside plottables for long periods of time
* Keep the public API the same (methods with many optional arguments) but pass a configuration object into the constructor of plottables so they don't require constructors with a large number of arguments (which makes them difficult to subclass).

Most of these refactors will be "invisible" (perhaps using a single bitmap will yield a performance enhancement), but my overall thought is that these things will simplify the rendering system in 4.1 to make it easier to refactor for renderer injection in 4.2 which will add support alternative rendering systems like SkiaSharp, ImageSharp, SMFL, etc. that implement `IRenderer`

If you think these goals and this plan sounds good, I can start work in the next couple weeks. I want to ensure all of your changes are committed before I begin, and as I work it probably makes sense not to start new PRs until my refactoring finishes. If we determine a good time to get started I can promise to complete this refactoring in good time (slow enough to ensure it is done carefully, but I will make it the primary focus of my daily effort until it is done to ensure it stabilizes soon).

What do you think?

StendProg:
Hi @swharden 

Unfortunately, I had planned to finish this PR not earlier than in a week, maybe it will drag on even more. I don’t know how it fits with your plans, but don’t worry, it is always possible to rebase on your changes.

> Allow breaking changes in ScottPlot 4.1?

Well, so far I managed not to break the api and make it simple to make the api breaking change when such a decision is made. So I will try to continue to do so, and shift all the responsibility onto you ;-)

> What do you think?

Sounds good, some comments on it:
>PlotSignal and PlotSignalXY will return a generic object (this PR)

I haven't thought about `PlottableSignalXY` yet. Make it generic for and 'xs' too sounds good, but may require additional work.

> make the figure and data area one bitmap (with cropping instead of two bitmaps)

As far as I remember there were display problems with figure layer with AA disabled. If you know how to render in signle bitmap with mixed AA settings its good idea. Just notice about potential problem, because disabling anti-aliasing during interaction is a very good solution that I would not want to lose

> more carefully create and destroy all IDisposable System.Drawing objects (like brushes) with each render instead of holding them inside plottables for long periods of time.

I did not fully understand. It is profitable to reuse the object while preserving it than to recreate it every time. GC call `Dispose()` anyway before destruction, when all reference to him will be lost. I think that the real resource that these objects occupy is also memory, it's just that it is allocated in unmanaged code with `WinApi` call, therefore, it is wrapped in `Dispose` to free it with correct `WinApi` call on `Dispose`.

>Keep the public API the same (methods with many optional arguments) but pass a configuration object into the constructor of plottables so they don't require constructors with a large number of arguments (which makes them difficult to subclass).

This is the problem that has been haunted for a long time, and there are many thoughts on this score.
As I see it, a fairly simple solution (.net way which is possible thanks to initializers) is to make public properties for all arguments, and check them in `setter`. In constructor you just init them with defaults. For example:
```c#
private double _lineWidth;
public double LineWidth
{
get => _lineWidth;
set
{ 
  if (value <0}
   {
     throw new Exception();
      //or just set to default
     _lineWidth = 1;
   }
  else
    _lineWidth = value;
}
...

public PlottableSignal()
{
  LineWidth = 1; // default value if user not provide it
}

//Client code will be:
var signal = new PlottableSignal()
{
   LineWidth = 7,
   PenColor = ColorRed,
} // not provided properties will fill with defaults in Constructor
```
The only problem here is the impossibility of complex checks at once for a group of arguments in a setter. And there are no guarantees in passing required parameters. Since this code will be hidden behind the user api. You can remove most of these complex checks and make them part of your developer contract.

Another solution could be `Fluent Builder` pattern, whom I hardly know.
Client code can look like this:
```c#
var signal = PlottableSignal.Builder()
                                           .Data(ys)
                                           .WithLineWidth(5)
                                           .WithColor(Color.Red)
                                           .RenderRange(1000, 9000)
                                           .Parallel()
                                           .Build();
```
It looks nice, but the implementation of such a builder will be much more complicated and bulky than a constructor with a bunch of arguments available at the moment.

swharden:
> Unfortunately, I had planned to finish this PR not earlier than in a week, maybe it will drag on even more. I don’t know how it fits with your plans, but don’t worry, it is always possible to rebase on your changes.

No worries! I am not in a rush, and I can wait for your work on this PR to complete before I move to the next big step. I won't make any large changes until this PR is complete and merged. Take as long as you need to move forward with confidence, and when you think this is in a good place let me know and I will merge it in 👍 

> Well, so far I managed not to break the api and make it simple to make the api breaking change when such a decision is made. So I will try to continue to do so, and shift all the responsibility onto you ;-)

This is a good plan! If you continue to make non-breaking changes I will merge them into 4.0.x soon and apply breaking changes later.

> disabling anti-aliasing during interaction is a very good solution that I would not want to lose

I agree! I will make sure this functionality is not lost.

> [regarding disposing System.drawing objects] I did not fully understand. It is profitable to reuse the object while preserving it than to recreate it every time. GC call Dispose() anyway before destruction, when all reference to him will be lost. I think that the real resource that these objects occupy is also memory, it's just that it is allocated in unmanaged code with WinApi call, therefore, it is wrapped in Dispose to free it with correct WinApi call on Dispose.

Compared to the other topics (like improving styling API), this is low priority. However, this is my thinking:

Preserving GDI objects may be a good idea if we know there will be a limited number of them and are certain we will re-use them. However, I think there is a maximum number of open objects GDI will allow (default is 10,000 and can be changed with a registry modification). This may be a concern if we collect too many GDI objects in an application (especially if it already uses GDI for other things). This problem may get even worse in multi-threaded code, such as console applications that generate many plots in parallel. Also, it may not make sense for plottables to store GDI objects at all if they will be rendered using a different system (like SkiaSharp). For these reasons I think storing GDI objects for long periods of time is discouraged. I can run benchmark tests to measure how creating/disposing these objects with every rendered frame changes performance. If it is slow, maybe we should keep it like it is, but my understanding is that this is a limitation of GDI and keeping objects open for a long time is not recommended.

> [regarding styling changes] This is the problem that has been haunted for a long time, and there are many thoughts on this...

Thank you for sharing your thoughts! I think adding error checking to auto-properties, using builder pattern, or making properties `get; private set;` and creating smart setter methods to update them are all interesting possibilities. On one hand all these options add some complexity to the plottables, but on the other hand they simplify the constructor arguments which is very desirable. I will continue to think about options as well!

Thank you for sharing your thoughts on these issues! I find your ideas very interesting and helpful to consider.

swharden:
Following-up, limitations of GDI (maximum number of open handles, concurrency issues, etc) are better described in this article: https://photosauce.net/blog/post/5-reasons-you-should-stop-using-systemdrawing-from-aspnet (although it is a bit old)

For desktop applications that just involve a single user control these are probably not a problem, but if ScottPlot is used to create a large number of charts as part of an automated report pipeline in a scientific application, or dynamically create plots in from a web server, GDI-related leaks may be more of a concern.

I started performance testing System.Drawing and I'm finding that creating/destroying GDI objects is extremely fast. I'll post those results in #505, but if GDI objects like brushes and fonts can be created and destroyed in each plottable's render method with negligible effect on performance, I'd rather be safe and do it this way. It also makes sense to work toward not storing GDI objects at the plottable class level, as a SkiaSharp rendering system would have no need for a System.Drawing.Font 😎

StendProg:
SignalXY changes:

For now there is PlottableSignalXY with no api breaking changes, but based on PlottableSignalXY<Tx,Ty> supports different generic types for `xs` and `ys` arrays, and `IMinMaxSearchStrategy`.
Added PlottableSIgnalXYConst<Tx, Ty> in `Plot` api. It have generic interface and use `SegmentedTreeMinMaxSearchStrategy`.
Added some demos for `PlottableSignalXYConst`.

Breaking changes direction, as i see:
Remove `PlottableSignalXY` and `PlottableSignalXYConst`. Rename `PlottableSignalXYGeneric` to `PlottableSignalXY`.
Make `Plot` api method(methods?) which create `PlottableSignalXY` with different `IMinMaxSearchStrategies` and return `PlottableSIgnalXY`.

The same for `PlottableSignal`...

swharden:
Hi @StendProg, this is wonderful! Thank you for working so hard to make these improvements to SignalXY.

> Suggested breaking changes:
> `PlottableSignalXY` → `PlottableSignalXYConst`
> `PlottableSignalXYGeneric` → `PlottableSignalXY`
> The same for `PlottableSignal`
> Update `Plot` API somehow to create objects (by passing in the appropriate strategy) and return generic objects

This sounds good! I will make these breaking changes after starting version 4.1.

Are you finished with this PR / are you ready for me to merge it now?

StendProg:
Hi @swharden, It's quite finished, additionaly I planed write Unit tests for `MinMaxSearchStrategies`
аnd check all MinMaxRangeQuerry calls.  I have a suspicion that there is confusion with upper boundary inclusive/exclusive sometimes.
I'll try to do it today, but I won't promise

swharden:
@StendProg that sounds good to me! I'll wait to merge this PR until you are finished with those tests and are satisfied that the strategies are working well.

In the mean time I will begin refactoring other plottables to work toward goals in #505 but I will not modify the signal classes until this PR is merged.

Thanks again for all your work in this area!

StendProg:
I think it's finished.
There are a lot of changes, I tested it a little, but I am not completely sure that there are no errors. I think to leave it as it is, and solve problems as they become available.

swharden:
This looks great, and the unit tests are reassuring. Thank you @StendProg!

swharden:
@StendProg I get a compile warning that `TreesReady` covers the one in the base class, and to use `new` if this is intended.

image

Since `PlottableSignalBase` already has `public bool TreesReady => true`, can this line be deleted?

StendProg:
@swharden, I think `PlottableSignalBase` don't need `TreesReady` property at all, it can be safety deleted.
`PlottableSignalConst.TreesReady`  also looks unfinished, but I can't tell you how easy it is to fix it.

Issue #505, opened 2020-08-14
ScottPlot 5


swharden:
ScottPlot 5 is being actively developed (#521) to achieve the following goals:

* Figure items (ticks, axis labels, grid, etc.) will implement `IRenderable` and track state (styling) and handle rendering in the same class. This makes experimental renderables (like new tick systems) easy to implement and test.
* Support for an arbitrary number of coordinate planes (allowing plots with two different Y scales)
* Improved rendering system
  * The Plot module will not store Bitmap objects
  * Existing Bitmaps can be passed-in to render onto, improving performance of user controls
  * Architecture is designed for eventual support of System.Drawing alternatives like SkiaSharp and ImageSharp
* Mouse interactivity of all user controls will be abstracted into a `PlotControl` module, ensuring all user controls to have the same functionality and options without requiring maintenance of duplicate code for each control.
* All plottables will have improved error-checking when data is modified

### Strategy  / Timeline
* ScottPlot 5 will be developed in a separate folder from ScottPlot 4
* A minimum viable mouse-interactive plotting system will be created using the new architecture
* Features will be migrated from 4 one-by-one, refactoring to achieve the design goals above
* Unit tests will be added for each feature as it is migrated from 4 to 5
* After things settle down, work through every item in #412 to ensure nothing was forgotten

swharden:
progress toward ScottPlot5 was copied to `dev/ScottPlot5` so it can be experimented with in parallel with ScottPlot

PR #506, opened 2020-08-16
4.1

swharden:
Experimenting with new render system. Not sure if I'll merge this yet or start over 🤔 #505

### DIFF
--- a/src/ScottPlot.Avalonia/ScottPlot.Avalonia.NUGET.csproj
+++ b/src/ScottPlot.Avalonia/ScottPlot.Avalonia.NUGET.csproj
@@ -17,9 +17,9 @@ Release demo: http://swharden.com/scottplot/demo</PackageReleaseNotes>
     <GeneratePackageOnBuild>true</GeneratePackageOnBuild>
     <IncludeSymbols>true</IncludeSymbols>
     <SymbolPackageFormat>snupkg</SymbolPackageFormat>
-    <Version>4.0.39</Version>
-    <AssemblyVersion>4.0.39.0</AssemblyVersion>
-    <FileVersion>4.0.39.0</FileVersion>
+    <Version>4.1.0</Version>
+    <AssemblyVersion>4.1.0.0</AssemblyVersion>
+    <FileVersion>4.1.0.0</FileVersion>
     <SignAssembly>false</SignAssembly>
     <AssemblyOriginatorKeyFile>Key.snk</AssemblyOriginatorKeyFile>
     <DelaySign>false</DelaySign>
--- a/src/ScottPlot.WPF/ScottPlot.WPF.NUGET.csproj
+++ b/src/ScottPlot.WPF/ScottPlot.WPF.NUGET.csproj
@@ -19,9 +19,9 @@ Release demo: http://swharden.com/scottplot/demo</PackageReleaseNotes>
     <GeneratePackageOnBuild>true</GeneratePackageOnBuild>
     <IncludeSymbols>true</IncludeSymbols>
     <SymbolPackageFormat>snupkg</SymbolPackageFormat>
-    <Version>4.0.39</Version>
-    <AssemblyVersion>4.0.39.0</AssemblyVersion>
-    <FileVersion>4.0.39.0</FileVersion>
+    <Version>4.1.0</Version>
+    <AssemblyVersion>4.1.0.0</AssemblyVersion>
+    <FileVersion>4.1.0.0</FileVersion>
     <SignAssembly>true</SignAssembly>
     <AssemblyOriginatorKeyFile>Key.snk</AssemblyOriginatorKeyFile>
     <DelaySign>false</DelaySign>
--- a/src/ScottPlot.WinForms/ScottPlot.WinForms.NUGET.csproj
+++ b/src/ScottPlot.WinForms/ScottPlot.WinForms.NUGET.csproj
@@ -18,9 +18,9 @@ Release demo: http://swharden.com/scottplot/demo</PackageReleaseNotes>
     <IncludeSymbols>true</IncludeSymbols>
     <SymbolPackageFormat>snupkg</SymbolPackageFormat>
     <GeneratePackageOnBuild>true</GeneratePackageOnBuild>
-    <Version>4.0.39</Version>
-    <AssemblyVersion>4.0.39.0</AssemblyVersion>
-    <FileVersion>4.0.39.0</FileVersion>
+    <Version>4.1.0</Version>
+    <AssemblyVersion>4.1.0.0</AssemblyVersion>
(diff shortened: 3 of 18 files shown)


StendProg:
Hi Scott (@swharden) ,

I check this PR and I want to share my thoughts if you will.

1. You need to utilize benefits of this `IRenderable` interface.
`Plot.Render()` don't need to know anything about  plot  structure.
Instead of : 
https://github.com/swharden/ScottPlot/blob/f5dbf82c5a3d53bae8c4f6b3d9e7e4c67ca3d4fc/src/ScottPlot/Plot.cs#L262-L267
you can process `List<IRenderable>` similar to `IPlottable`, and form this list somethere on plot construction.
```c#
foreach (IRenderable renderable in PlotElements)
    renderable.Render(settings)
```

2. It's better  tho hide this call on abstraction:
```c#
using (Graphics gfx = Graphics.FromImage(settings.Bitmap))
// can be replaced with:
using (Graphics gfx = settings.GetGraphics();
...
public class Settings
{
  public Graphics GetGraphics()
  {
    return Graphics.FromImage(settings.Bitmap);
  }
}
```
3. If we go further, we can abstract from the specific Graphics. Return some `IRenderer` interface and call it.
```c#
using (IRenderer renderer = settings.GetRenderer())
{
  renderer.DrawLine(...);
  rendeerr.RotateTransform(90);
  ....
}
```
GDIRenderer just delegate all required calls to Grapics
```c#
public class GDIRenderer : IRenderer
{
  private Grapchics GDI;
  DrawLine(Point a, Point b)
  {
    GDI.DrawLine(Points a, Point b)
  }

  RotateTransform(...)
  {
    GDI.RotateTransform(...);
  }
}
``` 
Yes, this goes beyond 4.1 goals, but it can simplify further work. And it's easy to do now for `IRenderable` only, since they are rewriting.

4. If you move part of `Renderable.Render()` code  to helper method:
```c#

        public void Render(Graphics gfx, Font font, Brush brush, StringFormat sf)
       {
                float centerX = CenterToDataArea ? settings.DataWidth / 2 + settings.DataL : settings.Width / 2;
                float centerY = CenterToDataArea ? settings.DataHeight / 2 + settings.DataT : settings.Height / 2;

                gfx.TextRenderingHint = System.Drawing.Text.TextRenderingHint.AntiAlias;

                switch (Edge)
                {
                    case Edge.Left:
                        gfx.TranslateTransform(Padding, centerY);
                        gfx.RotateTransform(-90);
             ....
       }

        public void Render(Settings settings)
        {
            using (Graphics gfx = Graphics.FromImage(settings.Bitmap))
            using (Font font = new Font(Config.Fonts.GetDefaultFontName(), FontSize, FontStyle))
            using (Brush brush = new SolidBrush(Color.Black))
            using (StringFormat sf = new StringFormat())
            {
                Render(gfx, font, brush, sf)
            }
        }
```

It would be easy to make later GDI caching version for `Renderables` which not destroy GDI objects on each `Render` call:
```c#
public class AxisLabelCached: AxisLabel
{
  public Brush brush = new SolidBrush(Color.Black);
  public Font font = //fill in constructor, recreates on font properties changes

  public overide void Render(Settings settings)
 {
   using (Graphics gfx = Graphics.FromImage(settings.Bitmap))
   {
    Render(gfx, font, brush, sf);
   } 
}
```
You made a compelling case for GDI `Dispose`, but still this is not always critical, and it would be convenient to leave the option to configure (IMHO).

5. Confused by a large number if/case, this can be solved by polymorphism, but I'm not sure that it will be simpler.


In general, only 1 point is principled, it is wrong not to use the capabilities of the new `IRenderable` interface.
Good work!

swharden:
Hi @StendProg, thank you very much for your input! This branch/PR is experimental and I don't think I'll merge it, but instead use it as a reference to carefully refactor the existing Render() method (rather than replace it). I learned a lot from this refactor/PR, and I really appreciate your feedback as well - your input is really helpful!

I now realize that a "4.1" PR will be too large to review easily, so I will probably create feature branches/PRs and merge them in one at a time. Refactoring the Render method will probably be its own branch/PR, and it may be possible to do without breaking changes. I have edited the sequence of refactor plans on #505 based on what I learned with this experimental branch.

### 1) Use a List to store `IRenderable` objects

This will allow code like:

```cs
foreach (IRenderable renderable in PlotElements)
    renderable.Render(settings)
```

I agree! This also makes it easier for users to add their own custom renderable objects into the render queue. 

I wrote code like this...

```cs
 settings.FigureBackground.Render(settings); 
 settings.TitleLabel.Render(settings); 
 settings.AxisLabelX.Render(settings); 
 settings.AxisLabelY.Render(settings); 
 settings.TicksX.Render(settings); 
 settings.TicksY.Render(settings); 
```

...because I am thinking about making these changes directly in the existing Render method using **strangler pattern** to replace these items one at a time. After all legacy render methods have been replaced with new `IRenderable` objects, I can then put them all in a List and render them as you suggest. My plan is to carefully refactor the existing Render() method (committing after every large step, and ensuring all tests pass) instead of creating a new Render method to replace it.

### 2-4) Eventual Use of `IRenderer`

By your comments I can tell you know _exactly_ what I am trying to work toward! 

Your examples with `IRenderer` demonstrate how such an interface will make it easy to create renderers for GDI, SkiaSharp, ImageSharp, etc. In the case of System.Drawing's renderer, I also recognize that this may be a good place to cache GDI objects. I agree with each of your suggestions: (2) use this abstraction to avoid putting many `using` statements inside plottables, (3) abstract graphics method calls by putting them in the renderer, (4) move graphics code out of plottables and into the abstracted renderer.

This architecture will be good, but I am still trying to figure out the best path to take to get there. I don't want tests to break along the way, and I think several small PRs is better than completely replacing the rendering system in one large PR. However, because of this some of the intermediate PRs may have code that looks somewhat awkward. For example, in some of these intermediate steps plottables will have many System.Drawing calls and lots of `using` statements. I'm trying to focus my effort narrowly onto one task at a time, so I don't get burned-out trying to refactor too many large things at once. Hopefully this all will be complete within the next few weeks!

PS: These articles are a work in progress, but I explored this topic a few months ago and made some notes about how to create an abstract Renderer and use it to render animations identically with System.Drawing and SkiaSharp. It's a strategy like this I hope to apply to ScottPlot:
* [Constellation Animation](https://swharden.com/CsharpDataVis/constellation/readme.md.html) 
* [Isolating the Renderer](https://swharden.com/CsharpDataVis/constellation/isolate-the-renderer.md.html)
* [Creating and Injecting Abstract Renderers](https://swharden.com/CsharpDataVis/constellation/inject-the-renderer.md.html) ← ultimate goal

---

I am continuing to experiment and explore the best path to proceed, and I'm excited to lean more about your thoughts along the way. Thank you again for all of your help and input! -Scott

Issue #518, opened 2020-08-23
Refactor Coordinate System


swharden:
This is a chicken and egg problem! In order to properly refactor the layout system (#513), the coordinate system must be refactored too. This is because coordinates depend on the layout system (e.g., `GetPixel()` needs to know the layout). 

I think it makes sense to put Layout state and axis state into a single module (with coordinate-to-pixel methods), and to pass this object to plottables (instead of the whole settings module). This refactor will be extensive, and should probably happen in its own branch.

Benefits from this refactor are: simplified layout system, faster unit/pixel transformations, better support for second Y axis, potential support for non-linear axes.

swharden:
After coding this up a bunch of different ways, I think the best way to do this is to have an object which just contains figure dimensions (pixel units) and axis dimensions (non-pixel units) all in one.

An ultimate goal will be for this object to have a `List<Plane>` where different plottables can support different planes (perhaps with a `PlaneIndex` property) allowing some plottables to be displayed using different Y scales. Note that a `Plane` has two `Axis` objects, and it's easy to create a `LinearAxis` and `LogAxis` with this system.

Magenta items are defined by the user, green items are calculated when needed. The new Layout procedure (#513) will simply look at the size of all the axis labels (#513) and tick labels (#514) to determine how to best define these values.

![image](https://user-images.githubusercontent.com/4165489/90990508-2b133b80-e570-11ea-9374-a49c467a6e24.png)

swharden:
Excellent progress has been made on this front (#519). Work will continue in #521

Issue #516, opened 2020-08-22
Refactor entire rendering system to use IRenderable


swharden:
**Goal:** Refactor _renderable_ plot items (anything that's not plotted on the coordinate space) into single modules which implement `IRenderable` (#511) that store configuration and rendering code in the same class. 
- [x] Figure background (#511)
- [x]  Data background (#511)
- [x]  Grid lines (#511)
- [x]  Benchmark (#511)
- [x]  Legend (#515, #517)
- [ ] Refactor the coordinate system (#518)
- [ ]  Refactor the layout system (#513)
- [ ]  Ticks and tick labels (#514)
- [ ]  Axis labels and title (#513)

These non-breaking changes should occur prior to the breaking changes introduced by ScottPlot 4.1 (#505, #506).

**Motivation:** Examples of tasks which are currently hard but will become easy after this refactor include alternative tick calculations (#394, #456), second Y axis (#180, #213, #478, #510), and placement customization of above-plot annotations (#321) and legends (#388).

swharden:
Excellent progress has been made on this front (#519). Work will continue in #521

Issue #513, opened 2020-08-22
Refactor Layout System


swharden:
### The Problem
It's time to refactor the axis label system (#505, #511) but the layout module [Layout.cs](https://github.com/swharden/ScottPlot/blob/master/src/ScottPlot/Config/Layout.cs) is highly coupled to it.

Axis label placement depends on the figure layout which depends on axis label size. Because of this circular problem, the axis label system cannot be refactored without also refactoring the Layout system. 

### Current System
The current layout system was created about a year ago (#156) and has many strengths:
* `Plot.Layout()` has many useful options so the user can adjust size and padding of layout elements
* `Plot.TightenLayout()` auto-adjusts the layout based on axis and tick label sizes

However, it has some weaknesses:
* Layout configuration code is separated from rendering code
* Layout auto-sizing is getting called to much. String measurement means this method is slow.
* Layout module stores dimensions rather than calculating them from plot objects (axis and tick labels)

### New System
The Layout system needs to be refactored with the goal of:
* Only recalculate layout when one of these occurs: 
  * A plot is rendered for the first time
  * `Plot.TightenLayout()` is called
  * A user control triggers a MouseUp event
* Rather than store label dimensions as properties, retrieve this information from the new AxisLabel classes.
* Same for tick label dimensions
* This refactor should make it easier to support a second Y axis

swharden:
Excellent progress has been made on this front. Work will continue in #521

PR #519, opened 2020-08-23
new coordinate and mouse interaction system

swharden:
As described in #518, the coordinate system will be refactored to combine it into a common module with the layout system. 

This will be done to improve the layout system (#513) and make it easier for the plot to auto-size itself based on axis label and tick dimensions, leading to the ability for these things to eventually implement `IRenderable` (#516)

The new coordinate system object (not the Settings module) will then be passed into the Render method of plottables (instead of the Settings module) which is a pretty big change so the version has been rolled from 4.0 to 4.1


StendProg:
Hi Scott, (@swharden )
It looks so promising that I can't get past it.
1. It will be possible to create `LogAxis` later. So maybe extract `IAxis` interface from `LinearAxis` and make dependency inversion now. 
```c#
public class Plane
{
  public IAxis X{get;set;}
  public IAxis Y{get; set;}
}
```

2. If you take the creation of the axes outside `Plane` it will make linking the `X` axes easier.
```c#
public FigureInfo()
{
     var XAxis = new LinearAxis();
     Planes.Add(new Plane(XAxis, new LinearAxis())); // primary Y
     Planes.Add(new Plane(XAxis, new LinearAxis())); // secondary Y
}
```
Although hiding the `axes` specific from the `FigureInfo` also good disign.

3. If you think about something really bold, for example, the future implementation of polar axes. You can't just delegate `GetPixel` from `FigureInfo` to `Plan.Axis.GetPixel()`. You need layer methods `Plane.GetPixelX()`, `Plane.GetPixelY()` because GetPixelX() will depend on both `X` and `Y` axes for polar axes.

4. `LinearAxes.GetUnit(float pixel, bool inverted)` suggests itself.  It may never be used, but it must be.

5. This `inverted` flag can be removed by creating `LinearAxisInverted` class
```c#
public class LinearAxisInverted: LinearAxis
{
  public override float GetPixel()
  {
      ...
  }
}
```
But you can leave it as it is, everything is also very clear.

swharden:
Hi @StendProg I am very happy you like this design!

1) **Interfaces:** Using interfaces here is definitely a good idea. I had `IAxis` and `IPlane` interfaces in earlier implementations but haven't abstracted them here yet. I will certainly do this.

2) **Two planes with a shared axis:** I really like your idea of passing in axes when creating a plane! I will implement this next. Not only does this make it easier for two planes to share a Y axis, but it provides a path for _two different plots_ to share an axis too. Another goal of this refactor is to make it easier to link axes from multiple plots together (by sharing a reference to the common axes).

3) **Polar Coordinates:** I did not consider this! However when I try to think of class structures that support this, it is not obvious whether or not this is achievable. It seems `PolarPlane` would be an `IPlane` that does not hold any `IAxis` objects (it tracks its own coordinates in the class), but panning and zooming get confusing. I may choose not to support polar coordinates now, and come back to this idea in the future.

4) **`GetUnit()` to compliment `GetPixel()`:** Indeed, this is a good idea! Similar to point 1, I'll try to implement abstractions and functionality only when I'm ready to use them. This will be useful for things like "get position under cursor".

5) **`LinearAxisInverted` instead of passing a `bool`:** Thank you for this great idea! I will do this

Thank you for reviewing this design! I am very happy about the direction this is going, and I look forward to implementing some of your excellent suggestions 👍

swharden:
This work will be continued in #521

PR #528, opened 2020-09-04
Plot.PlotBitmap()

AlexFsmn:
**Purpose:**
Added functionality to plot images

**New functionality (code):**
```cs
var image = new Bitmap("img.bmp");
plt.PlotImage(image, 20, 0.3);
```

**Discussion:**
(1) Im reusing TextAlignment for the image alignment. Might be good to use another name?
(2) PlottableImage row 60-105 was duplicated from PlottableText. Maybe this can be restructured in a better way?
(3) A better image should be used in the cookbook probably :)


swharden:
Hey @AlexFsmn thanks for adding this! On first pass it's looking great!

One thing I'm hesitant about is requiring a bitmap file to be carried around. You can get a 2D array image data from the `DataGen.SampleImageData()`, so maybe you can use this to create a bitmap without storing one on the disk. Something like:

```cs
double[,] bmpValues = ScottPlot.DataGen.SampleImageData();
Bitmap bmpIndexed = BitmapFrom2DArray(bmpValues);
Bitmap bmpColor = ScottPlot.Colormap.Viridis.Apply(Bmp bmpIndexed);
```

What do you think of this strategy for avoiding storing a binary file to disk (and also having a more interesting image to look at)?

AlexFsmn:
Sounds good!

...And about my discussion topics:
(1) I could also rename "TextAlignment" to something like "ContentAlignment" if you don't have a better name for this.
(2) Since this is my first commit in this project I didn't want to make too many changes at once. I think you would agree that this code could be used in a more generic way. Do you have any suggestion where I should add this?

swharden:
> (1) I could also rename "TextAlignment" to something like "ContentAlignment" if you don't have a better name for this.

For backwards comparability let's leave this one how it is. This will get a little cleaner in ScottPlot 4.1 (#521/#505), but for now let's try not to break anything that currently works.

> This code could be used in a more generic way

I agree! In ScottPlot 4.1 I've created for enums this:

```cs
public enum HorizontalAlignment  { Left, Center, Right }
public enum VerticalAlignment { Top, Center, Bottom }
```

If you want to create something similar and use it for this plottable, a simple place to put it would be right after `TextAlignment` in Style.cs:

https://github.com/swharden/ScottPlot/blob/109946132e8751a548e77d71e74590fe84189d4b/src/ScottPlot/Style.cs#L25-L37

Congratulations on the first commit! This is looking really nice 😄

swharden:
Also, curious, is there a strong reason to accept `Image` instead of `Bitmap`? Most of ScottPlot uses `Bitmap` objects so it may be more consistent to have `PlotBitmap()` instead of `PlotImage()` 🤔 

Interested to hear your thoughts on this!

AlexFsmn:
I used Image since the class "Graphics" "DrawImage" method accepts Image.
Bitmap also inherits from Image so I guess they are pretty similar. However, if you don't make use of any bitmap properties you might as well go with Image.
I don't know the libraries well enough to know what the big differences is though. But I guess you had a reason for using Bitmap in the first place?
If you want me to change this and stay consistent just tell me and I'll change this in the next commit.

swharden:
> I don't know the libraries well enough to know what the big differences is though. But I guess you had a reason for using Bitmap in the first place?

I had to look this one up... I've always used `Bitmap` because wanted to work with files like bmp, jpeg, gif, png, etc. From what I'm reading now it seems that an alternative class is `Metafile` which supports creating WMF files, although both these classes inherit from `Image`. 

Since I'm not really interested in supporting WMF format in ScottPlot 4.0, and I use `Bitmap` a lot elsewhere, let's stick with `Bitmap` in this case. This makes me think the best changes would be:
* Rename `PlotImage()` to `PlotBitmap()` and have it accept a `Bitmap` to be consistent with the rest of ScottPlot
* `PlottableImage` can keep its name, and the property can stay `Image` like it is.

> If you want me to change this and stay consistent just tell me and I'll change this in the next commit.

That would be great! After you're finished let me know and I'll merge it in! 👍

AlexFsmn:
Ok I made some changes according to your comments.
Could you fix the image loading part from memory?
It doesn't look like there is a finished method for doing this in the standard library?

swharden:
Very cool! Thanks @AlexFsmn! I look forward to publishing this on NuGet later today 👍 

image

Issue #255, opened 2020-02-14
Shade area in-between curves


ckovamees:
Hi Scott,
Amazing activity in this project. If you ever lack things to do here is a feature request.... :-)

For my hobby project I would find it useful to render an area between curves in a light shade of green or red/or any other color for that matter.
![image](https://user-images.githubusercontent.com/9557072/74552681-c1acd300-4f23-11ea-91d3-69d7176f24cd.png)

_Edited by Scott to add todo list:_
- * ~~add arguments to PlotSignal() to support shading under the curve~~
- * ~~add arguments to PlotSignalConst() to support shading under the curve~~
- * ~~add arguments to PlotScatter() to support shading under the curve~~
- [ ] create PlotPolygon() to allow rendering of arbitrary shapes
- [ ] create PlotRectangle() to allow easy outlining of data areas

swharden:
This is an excellent idea @ckovamees. I'll look into it shortly!

- [ ] add arguments to PlotSignal() to support shading under the curve
- [ ] add arguments to PlotSignalConst() to support shading under the curve
- [ ] add arguments to PlotScatter() to support shading under the curve
- [ ] create PlotPolygon() to allow rendering of arbitrary shapes
- [ ] create PlotRectangle() to allow easy outlining of data areas

swharden:
It took a while, but there's some movement on this. I just finished creating a `Polygon` object, and this code can be extended to make it easier to plot shaded areas between curves. I'll keep working on this.

![image](https://user-images.githubusercontent.com/4165489/78502807-f0c80f80-7730-11ea-81db-a6de7fa6ea58.png)

```cs
var plt = new ScottPlot.Plot(600, 400);

plt.PlotPolygon(
    xs: new double[] { 2, 8, 6, 4 },
    ys: new double[] { 3, 4, 0.5, 1 },
    label: "polygon A", lineWidth: 2, fillAlpha: .8,
    lineColor: System.Drawing.Color.Black);

plt.PlotPolygon(
    xs: new double[] { 3, 2.5, 5 },
    ys: new double[] { 4.5, 1.5, 2.5 },
    label: "polygon B", lineWidth: 2, fillAlpha: .8,
    lineColor: System.Drawing.Color.Black);

plt.Title($"Polygon Demonstration");
plt.Legend();
```

swharden:
[...]

PR #529, opened 2020-09-05
Added support to draw signal plots using gradient fill.

AlexFsmn:
**Purpose:**
Added support to draw signal plots using gradient as well as solid fill.

**New functionality (code):**
```cs
double[] data = new double[] { 51, -5, 6, 12, 3 };
plt.PlotSignal(data, minRenderIndex: 1, maxRenderIndex: 4, fill: true, fillColor1: Color.Blue, fillColor2: Color.Transparent);
```

GradientPlottableSignal

### DIFF
--- a/dev/changelog.md
+++ b/dev/changelog.md
@@ -5,6 +5,7 @@
 * Holding shift while left-click-dragging the edge of a span moves it instead of resizing it (#509) _Thanks @Torgano_
 * CSV export is now culture invariant for improved support on systems where commas are decimal separators (#512) _Thanks Daniel_
 * Added fill support to scatter plots (#529) _Thanks @AlexFsmn_
+* Fix bug that occurred when calling GetLegendBitmap() before the plot was rendered (#527) _Thanks @el-aasi_
 
 ## ScottPlot 4.0.39
 * Legend now reflects LineStyle of Signal and SignalXY plots (#488) _Thanks @Benny121221_
--- a/dev/changelog.md
+++ b/dev/changelog.md
@@ -4,6 +4,7 @@
 * Added user control for Avalonia (#496, #503) _Thanks @Benny121221_
 * Holding shift while left-click-dragging the edge of a span moves it instead of resizing it (#509) _Thanks @Torgano_
 * CSV export is now culture invariant for improved support on systems where commas are decimal separators (#512) _Thanks Daniel_
+* Added fill support to scatter plots (#529) _Thanks @AlexFsmn_
 
 ## ScottPlot 4.0.39
 * Legend now reflects LineStyle of Signal and SignalXY plots (#488) _Thanks @Benny121221_
--- a/src/ScottPlot.Demo/PlotTypes/Polygon.cs
+++ b/src/ScottPlot.Demo/PlotTypes/Polygon.cs
@@ -1,6 +1,8 @@
-﻿using System;
+﻿using ScottPlot.Demo.Customize;
+using System;
 using System.Collections.Generic;
 using System.Drawing;
+using System.Linq;
 using System.Text;
 
 namespace ScottPlot.Demo.PlotTypes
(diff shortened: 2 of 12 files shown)


swharden:
Hi @AlexFsmn, thanks for this PR! The aesthetic is fantastic!

### Determining the Edge

It might be nice to have the shaded area extend to the bottom of the view area, rather than the minimum value of the signal itself. The current method produces an unexpected interactive experience:
fill


The easiest way to do this is is probably to replace those lines:
https://github.com/swharden/ScottPlot/blob/88319299a1ef9ce9f4289fafdf1e59b8bfce1a40/src/ScottPlot/plottables/PlottableSignalBase.cs#L285-L286

With something like:

```cs
float minVal = 0;
float maxVal = settings.dataSize.Height;
```

I made the change and committed to this PR. What do you think?

This always looks nice as a single image, but interactively the effect is slightly unexpected. I think this is a step in the right direction though...

fill2

### Supporting FillAbove and FillBelow

Similarly, I'm curious if the `bool fill` argument should change to some type of `enum` like `NoFill, FillBelow, GradientBelow, FillAbove, GradientAbove`. Supporting "fill above" probably just involves tweaking the two lines of code cited above too 🤔 

### Supporting FillAboveAndBelow

How much work do you think it would take to achieve an effect like this? This was previously requested (#255) and I made it work with Scatter plots, but I can see this being requested around the corner if we add shaded support to signal plots.

image

Note that in this example the "split" between the colors is 0, but it would be fantastic if the user could define that too.

### Let's keep the constructor simple

We're experimenting with additional styling options options for edge cases. I recognize the existing style is to have methods with long lists of optional arguments and plottables with huge constructors, however I'm trying to move away from this in favor of plottables with exposed public properties the user can tweak to achieve advanced styling functionality. 

As you found, those huge constructors are a pain to update as you add little styling tweaks, so I'm hoping this will be an easier format to maintain moving forward. I modified this PR to use this newer style, and updated the cookbook demos accordingly. This PR is much simpler now, with only 2 modified files 👍 

I look forward to hearing your thoughts on the above! Thanks again for all your work here, this is looking _really_ nice!

AlexFsmn:
Yes, I think the change on how the edge is handled is better now.
Another way of handling this is would be to set the global minimum instead of the local minimum. However, this might affect performance(?).
About fill above & fill above & below: I made a few changes that should take care of this.
You can have a look at this and tell me what you think:
GradientPlottableSignalAboveAndBelow

I also added a fill type as you suggested. Seems like the better choice if multiple types should be supported :)
..Also tell me if the location of the enum should be in another place.

AlexFsmn:
Ah.. Seems like I didn't zoom in enough when testing.
I think this is a good start but has a few problems. Gradient colors doesn't seem to work as good as I though. There is also a problem with the color rendering around the baseline. Will look into this later.

swharden:
> Another way of handling this is would be to set the global minimum instead of the local minimum. However, this might affect performance(?)

I think if the min/max calculation is performed on `linePoints` (rather than `ys`) it will be very fast. I like this idea!

> You can have a look at this and tell me what you think

> Also tell me if the location of the enum should be in another place

I think this is looking wonderful! Thanks again for working toward this excellent improvement.

> I think this is a good start but has a few problems. Gradient colors doesn't seem to work as good as I though. There is also a problem with the color rendering around the baseline. Will look into this later.

Graphics programming can be finicky sometimes! I'll let you work toward putting the finishing touches on this, and I'll take a closer look and merge it in when you think it's ready 👍

AlexFsmn:
> I think if the min/max calculation is performed on linePoints (rather than ys) it will be very fast. I like this idea!

Ok! Might change to this later on.
I changed the new rendering "above and below" to work as the other types for now.
A also fixed the rendering issue around the baseline. 😀
Have a look and let me know what you think. The gradient example creates a kind of "focus" effect if you choose transparent as your second color. Looks kinda cool even though this wasn't the plan.

AlexFsmn:
Add some styling and it looks pretty good!
DarkGradientPlottableSignalAboveAndBelow

Issue #533, opened 2020-09-06
[New plot] Doughnut plots


AlexFsmn:
**What can we do to make ScottPlot better?**
Pie plots are available, but what about doughnut plots?

Is there any plan on adding these?

swharden:
I like this idea! It would be nice if this effect could be achieved like
```cs
double[] values = { 778, 283, 184, 76, 43 };
var pie = plt.PlotPie(values);
pie.explode = true;
pie.donut = true;
``` 

![image](https://user-images.githubusercontent.com/4165489/92335486-4b112780-f065-11ea-8e98-c65f3d861305.png)

The various labeling options are already in the pie plottable class, so this would probably be pretty easy.
https://swharden.com/scottplot/cookbooks/4.0.38/#plottypes-pie-show-everything

@Benny121221 recently created the pie chart (#423) so they might want to add this functionality, but if not @AlexFsmn I'd be happy to review a PR. Extra points if the donut can be achieved using clipping instead of drawing an opaque circle on top of the plot (to accomodate transparent plot backgrounds).

swharden:
This is in a pretty good place now! This feature (and a new cookbook) will be be packaged with the next release on NuGet 👍

PR #534, opened 2020-09-06
Donut pie chart

Benny121221:

**Purpose:**
#533

**New functionality (code):**

```cs
plt.PlotPie(values, donut: true);
```

image
Note this screenshot uses both an exploded and a donutted pie chart.

**Autoformat your code:**
```
cd ScottPlot/src/
dotnet format
```

### DIFF
--- a/dev/changelog.md
+++ b/dev/changelog.md
@@ -5,6 +5,7 @@
 * Holding shift while left-click-dragging the edge of a span moves it instead of resizing it (#509) _Thanks @Torgano_
 * CSV export is now culture invariant for improved support on systems where commas are decimal separators (#512) _Thanks Daniel_
 * Fix bug that occurred when calling GetLegendBitmap() before the plot was rendered (#527) _Thanks @el-aasi_
+* Pie charts now have an optional hollow center to produce donut plots (#534) _Thanks @Benny121221 and @AlexFsmn_
 
 ## ScottPlot 4.0.39
 * Legend now reflects LineStyle of Signal and SignalXY plots (#488) _Thanks @Benny121221_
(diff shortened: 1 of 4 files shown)


Benny121221:
I did it with clipping, not an opaque blob. This means that you can have things poking out from under it.

AlexFsmn:
Looks good!

2 questions:
1. Is it possible to choose the thickness?
2. Would it be easy to add a similar plot where you only have one value in %?
Similar to this:
Chart

Benny121221:
Custom Donut Size (size of 0.9 for demonstration purposes):

image

Benny121221:
With percentage in the middle:

image

AlexFsmn:
Great job!
Maybe the single value plot should be a separate class though since you only need to specify a value between 0-100?
@swharden, what do you think?

swharden:
This came together nicely! I exposed some properties to make it easier to customize (without modifying the original Plot module that was there before). I think I'm going to try to slowly move in this direction, probably marking the old constructors obsolete in ScottPlot 4.1

It was a bit of a surprise that clipping masks aren't anti-aliased... I know I suggested using them, but it's hard not to feel a little let down that an anti-aliased donut has an aliased hole! It's fine how it is, but I thought I'd mention it.

Thanks for all your work on this @Benny121221!

Regarding @AlexFsmn's original suggestion:
image

I don't think it makes sense for an image like this to have a "coordinate system", so it probably shouldn't be a "plot" in the first place. An image like this could be created with just a few lines of C# using System.Drawing to draw some circles (totally eliminating the massive complexities that come with plotting a mouse-interactive coordinate system)... so extensive support for these types of panels probably won't be a priority of ScottPlot development moving forward. That said, we got a pretty cool effect here and I'm happy where this ended up!

Issue #537, opened 2020-09-09
Support timecode-style DateTime tick labels


vrdriver:
Hi Scott, 

I've got a bit of a curly question.

I was wondering if there is a way to reformat the data on the X Axis?

I'm able to open audio files and display their waveforms, however I'd like to be able to show the X Axis in a time code, instead of samples. 

I've looked at the format-strings for plt.Ticks(numericFormatStringX: "t"), however it's just reformatting values in the millions and trying to get a time, which isn't working.

With 44100 samples a second, you can imagine the numbers don't work well at all...

Something like this maybe, I don't know...

`x value = formatstring( x value / samplerate, "d:hh:mm:ss:mmm")`

Is there a way to convert these values to time properly? Or can you point me in the right direction?

I'm at this so far:
```
formsPlot1.Configure(lockVerticalAxis: true);                    
formsPlot1.plt.PlotSignal(dataX1);
formsPlot1.plt.AxisAuto();
formsPlot1.plt.AxisBounds(0, samples, -1, 1);
formsPlot1.plt.Ticks(numericFormatStringX: "c"); // Obviously wrong
formsPlot1.Render();
```

Cheers,
Steve

StendProg:
Hi @vrdriver,

You can provide sampleRate to `PlotSingal()` and set `XAxis` to date time by:
```c#
double SampleRateDays = 44100.0 * 3600 * 24;
plt.PlotSingal(dataX1, sampleRate: SampleRateDays);
plt.Ticks(dateTimeX: true);
```
Please note that the sampling rate is provided in samples per day, not samples per second.
The disadvantage of this solution is that the minimum tick size is 1 second.

Maybe for your task it is enough just to specify the sampling rate, and then 1 count will mean 1 second without converting the `XAxis` to `DateTime` format?
```c#
plt.PlotSignal(dataX1, sampleRate: 44100);
```

vrdriver:
I've just tried that, and while it didn't work, I'll keep plugging away. You may have pointed me in the right direction though.

vrdriver:
I've discovered that while the above suggestions do change the axis, it also seems to change the entire data set when it's plotted, rather than just the time axis. But alas It won't format for me at "hh:mm:ss:mmm" on the X Axis.

StendProg:
PR #539 brings millisecond ticks on `DateTimeAxis`.

vrdriver:
[...]

PR #539, opened 2020-09-10
Add Milliseconds unit to DateTime Axis

StendProg:
**Purpose:**
Current `DateTime Axis` minimal unit is second.
This PR add `decisecond`, `centisecond` and `millisecond` units to DateTimeAxis. 
There is possible to  zoom into second interval and get millisecond ticks on DateTimeAxis. #537 
There is no culture invariant time specifier to display `milliseconds`, so i hardcoded this formats:
 `hh:mm:ss.f` - decisecond
 `hh:mm:ss.ff` - centisecond
 `hh:mm:ss.fff` - millisecond

Additional refactor DateTime Ticks module.

**New functionality (code):**
There is no new `API` in this PR.

image


StendProg:
After a little thought I came up with the idea that it would be nice to split the `millisecond` into `hundredsOfMilliseconds`, `TensOfMilliseconds` and `Millisecond`. This allow remove ugly zeroes at the end. 1.500, 1.600, 1.700 for now, will be 1.5, 1.6, 1.7. And will return small arrays in intermediate calculations. Up to 1000 ticks intermediate arrays for now in worst case.

I'm ready to do this, but I can't decide how best. I'm waiting for your input.

swharden:
Hi @StendProg, your idea to add subdivisions of seconds to the DateTimeUnit enum is perfect. I was not familiar with these words (I had to look them up to be sure they were real), but it appears these units of times have names we can use:

* second (1 sec)
* [decisecond](https://en.wiktionary.org/wiki/decisecond) (0.1 sec)
* [centisecond](https://en.wiktionary.org/wiki/centisecond) (0.01 sec)
* millisecond (0.001 sec)

Thank you for working on this! I see how this PR relates to #537, and I also very much appreciate your recent efforts responding to issues while I was busy over the last few days.

StendProg:
The naming system you suggested made the confusion in my head surprisingly simple.
We can, of course, go down to nanoseconds, I often deal with this, but personally I am satisfied with the usual numerical values.

Benny121221:
I would have use for microseconds (µs) at 10^-6 seconds.

And since this library is targeted towards science I expect that nano (10^-9), pico, femto (10^-12), and atto (10^-15) prefices are likely to be useful. There are prefices for even smaller numbers but I've never seen them used. It's also worth noting that in some contexts it's more clear to write `x * 10^6 µs` than it is to write `x fs`. So this should probably be configurable.

I don't think you can even go to attosecond precision thanks to the width of a double, which means anyone going this precise should probably use a conventional axis and rescale it (i.e. multiply by 10^-x) instead of using a DateTime axis. I don't know how OLE Date runs its affairs and I would suggest not changing to another system (e.g. Unix time) for backwards compatibility. So my recommendation is to support the smallest prefix that can be reliably expressed as an OLE Date with an FP64 double. If it's a very small prefix it might need configuration so people don't get confused by attoseconds.

swharden:
> anyone going this precise should probably use a conventional axis ... instead of using a DateTime axis

☝️  this! I agree that some scientists may want to plot data with femtosecond precision. They will probably use `double[]` to store time points, not `DateTime[]` 🙃 

This raises an interesting question though, how precise is a `DateTime`? I looked it up. The [DateTime Struct](https://docs.microsoft.com/en-us/dotnet/api/system.datetime?view=netcore-3.1) stores `ticks` which are accurate to one tenth of a millisecond. Further, "The value of this property represents the number of 100-nanosecond intervals that have elapsed since 12:00:00 midnight, January 1, 0001 in the Gregorian calendar"

swharden:
~~@StendProg if you want to add one more level of precision, you could probably call it `ticks`. ~~
EDIT: I misunderstood the length of a `tick` by several orders of magnitude - nevermind!

* second (1 s)
* [decisecond](https://en.wiktionary.org/wiki/decisecond) (0.1 s)
* [centisecond](https://en.wiktionary.org/wiki/centisecond) (0.01 s)
* millisecond (1 ms)
* microsecond (0.001 ms)
* nanosecond (0.001 µs)
* tick (0.1 ns)

Benny121221:
Yes, a tick is precise to 0.1µs, but an OLE date is stored with the integer part being the days since December 30, 1899, with the fractional part representing the time of day (i.e. 6 AM on January 1 1900 is 2.25)

FP64 gives (at least) 15 significant digits. For dates this year 5 of them are going to be taken up by the day (120 * 365 = 43800). So we can express fractions of a day to 10 significant digits, which puts the smallest amount guaranteed to be preserved 8.64 * 10^-6 seconds (86400 * 10^-10). So this means OLE dates can guarantee at best 5 significant digits for seconds (including whole seconds), which puts microseconds out of reach.

I had assumed it would be a lot better with a Unix timestamp, but if you express that time in seconds you use 10 significant digits on the integer part, leaving 5 significant digits for fractions of a second, so this gains one significant digit in precision.

So that means that no commonly used time convention (in FP64) can offer the precision for microseconds (other than Windows File Time, which is similar to `DateTime.Ticks` except it starts in 1601 AD instead of 1 AD). So I think unless you want to create a helper function (and rewrite the axes accordingly) to use `System.Decimal`, fixed point, or extended precision floating point (FP80 or FP128), you're not going to be able to get microsecond precision.

If I did any of the math wrong, I apologize, but I think it should be correct.

vrdriver:
I'll just add my little bit, since my suggestion surprisingly kicked off this thread. I'm presently working in samples of audio, which when displayed should labelled as hours, minutes, seconds and milliseconds - with the leading zeros included.

Regardless of the file format (MP3 or wave), the samples rates of audio files can vary anywhere from the common ones as 48000, 44100 and all the way up to 192kHz per second. 

I would think that the display conversion for only the X-axis text label should be enough, while still keeping the plot unchanged while still being able to show and zoom in to the individual samples.

I'm amazed that you are all jumping in on this. Go team!

StendProg:
>FP64 gives (at least) 15 significant digits. For dates this year 5 of them are going to be taken up by the day (120 * 365 = 43800). So we can express fractions of a day to 10 significant digits, which puts the smallest amount guaranteed to be preserved 8.64 * 10^-6 seconds (86400 * 10^-10). So this means OLE dates can guarantee at best 5 significant digits for seconds (including whole seconds), which puts microseconds out of reach.

I am also thinked that precise of OA date low, but found on StackOverflow:

>Near the date and time we're considering, the precision of an OA date is:
Math.Pow(2.0, -37.0) days, or circa 0.6286 µs
We conclude that in this region a DateTime is more precise than an OA date by (just over) a factor six.

Benny121221:
> I am also thinked that precise of OA date low, but found on StackOverflow:
> 
> > Near the date and time we're considering, the precision of an OA date is:
> > Math.Pow(2.0, -37.0) days, or circa 0.6286 µs
> > We conclude that in this region a DateTime is more precise than an OA date by (just over) a factor six.

I got a little bit bamboozled, but I think the difference comes from me rounding the number of significant digits midway through. If you do as he did you find that you need 16 bits for the day and that leaves 37 bits for the fractions of a day. I assumed it would always be 15 significant digits, but with FP64 it can be 15-17, so my rounding came back to bite me.

StendProg:
As far as I understand OA date, at one moment the accuracy will be exactly the same as you indicated initially (sooner or later this moment will come) :-)

swharden:
Hi @StendProg, this is an excellent refactor of the DateTimeTicks module! I am happy to see the factory pattern here, and these small classes really help readability and maintainability. This improvement is useful for #537 too.

It's been a few days since your last commit - is this ready for me to review and merge into master, or do you intend to continue working on it?

StendProg:
Hi Scott (@swharden),
>  is this ready for me to review and merge into master, or do you intend to continue working on it?

It's ready.
 I decided not to introduce microsecond labels. This will increase the `maxLabelSize`,  because of which every `TickUnit` will suffer. But really, microsecond ticks are unlikely to be needed by anyone.

There are 2 unresolved issues that I want to pass on to you.
1. For now `DateTimeUnit` enum name may be misleading. A more accurate name should be something like`DateTimeUnitKind`. And for good it would be moved to `ScottPlot.Config.DateTimeTickUnits` namespace/folder. But this will be a breaking API change. Anyone who has used manual ticks will have to manually apply these corrections to their projects.
2. `MothTicks` return 3 lines labels, all others 2 lines. It was in the original code. Maybe just a bug, or was it intended?

swharden:
> 2. `MothTicks` return 3 lines labels, all others 2 lines. It was in the original code. Maybe just a bug, or was it intended?

I think the line that caused that behavior is:

https://github.com/swharden/ScottPlot/blob/c5e634448ce7759ac4fd620df065f3a1549089fe/src/ScottPlot/Config/DateTimeTicks.cs#L93-L94

The 3 line behavior was not intended, and is probably the result of locales where the string contains more than 1 space.  I recognize that hard-coding a string search/replace with a line break is bad practice for internationalization, and am happy the new code no longer does things like this! Thanks again for this excellent PR

https://docs.microsoft.com/en-us/dotnet/standard/base-types/standard-date-and-time-format-strings#YearMonth

StendProg:
> the new code no longer does things like this

In fact, I tried to keep the original behavior, because I did not fully understand it, and it is difficult to test for many cultures at once. I think it's a good idea to write unit tests that count the number of lines in different cultures, with the current design it should be easier. I'll do it some other day.

PR #545, opened 2020-09-15
Improve mouse wheel zoom speed

StendProg:
**Purpose:**
Greatly improve render speed for `ScrollWheel` interactions #543.
1. Fix multiple event handler subscibe bug (`Control.Reset`).
2. All Scroll Wheel events puts in Render requests queue. At the end of the current rendering, all new requests are selected, the axes are scaled for each request, and then a new render is started. All renders are done in low quality. If no new request is received within 500 ms, then high quality rendering starts.

The code is durty, and shoud be refactored. But the functionality can already be tested.

**New functionality (code):**
There is no new `API` for now. But control settings like `HQDelayTime`, `LQDuringScrollWheel` should appear.

**New functionality (image):**
FastZoom

PR #550, opened 2020-09-16
Improve mouse wheel zoom speed (WPF control)

StendProg:
**Purpose:**
Continuing a #545 .
1. Implement the same Render queue for WPF Control.
2. Small refactoring.
3. WPF Control `Render()` now have `bool processEvents` argument like WinForms Control. Improve WPF control pan/zoom speed.

**New functionality (code):**
```c#
WPFControl.Configure(lowQualityOnScrollWheel: true, lowQualityScrollWheelDelay: 500);
Render(processEvents: true);
```

### DIFF
--- a/src/ScottPlot.WPF/WpfPlot.xaml.cs
+++ b/src/ScottPlot.WPF/WpfPlot.xaml.cs
@@ -467,14 +467,6 @@ namespace ScottPlot
             return null;
         }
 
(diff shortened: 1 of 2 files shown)


swharden:
Hi @StendProg, thanks for this PR! 

I see the `ScrollWheelProcessor()` method is better than what I originally implemented for WPF, and I am happy the WinForms and WPF controls now use a similar process.

I also think changing the List to a Queue was an excellent idea! I often forget about the Queue, but this is a perfect use for it.

Issue #546, opened 2020-09-15
SignalPlot fails when arrays contain double.NaN


bukkideme:
**Describe the bug**
When I create a second SignalPlot using a List, it is only shown after zoom in. I guess I only miss something basic...?

**To Reproduce**
Create a SignalPlot with rand numbers, then a List based on this array, and send it to a second SignalPlot.

```cs
// include code to help us reproduce this bug
using MathNet.Numerics.Statistics;
using ScottPlot;
using System;
using System.Collections.Generic;
using System.ComponentModel;
using System.Data;
using System.Drawing;
using System.Linq;
using System.Text;
using System.Threading.Tasks;
using System.Windows.Forms;

namespace ScottPlot_PlotSignalNotShown
{
    public partial class Form1 : Form
    {
        public Form1()
        {
            InitializeComponent();
        }

        private void Form1_Load(object sender, EventArgs e)
        {
            Random rand = new Random(0);
            int pointCount = (int)1e5;
            double[] array1 = DataGen.RandomWalk(rand, pointCount);
            
            formsPlot1.plt.PlotSignal(array1, 1, 0, 0, color: Color.Blue, 2, 6, "Signal 1");
            
            List<double> smooth = new List<double>();
            MovingStatistics moving_aver = new MovingStatistics(24);
            foreach (double value in array1)
            {
                moving_aver.Push(value);
                smooth.Add(Math.Round(moving_aver.Variance));
            }

            formsPlot1.plt.PlotSignal(smooth.ToArray(), 1, 0, 0, color: Color.Green, 2, 6, "Signal 1 variance");

            formsPlot1.plt.Legend(true, fontSize: 18, location: legendLocation.upperRight);
            formsPlot1.Render();
        }
    }
}
```

**Screenshots**
![image](https://user-images.githubusercontent.com/51174971/93259334-004b8a00-f7a0-11ea-91a7-b7e5ba807164.png)

After zoom in:

![image](https://user-images.githubusercontent.com/51174971/93259362-093c5b80-f7a0-11ea-91a4-4307c43e83fb.png)

StendProg:
Hi @bukkideme,
Probably `smooth` array contains `NAN` or `Infinity` values. Check for this especially on borders.
All others looks fine.

bukkideme:
Ups! Thanks, i totally forgot that the variance calculation needs at least 2 values, it returns NaN for the first iteration at the single element :)
[...]

PR #553, opened 2020-09-19
Add diagnostic mode

StendProg:
**Purpose:**
Sometimes users provide incorrect data and blame library for incorrect results #546.
This PR allow user to enable `Diagnostic mode` at init of their project.
As a result, multiple data checks are enabled before each render, and detailed exceptions are thrown if the data is not in the correct format.
With diagnostic mode turned off, there are absolutely no negative performance impacts compared to the existing code.

**Added a stub for `IValidatableData`. If some plottable implements it, instead of attribute checks simple call `IValidatableData.ValidateData()`**

PR contains 3 `Diagnostic mode` demo.
It needs a good names and descriptions. My bad English won't let me do it myself.
Also feel free to change any namings in this PR.

**Expansion of this PR**
To enable existing types of check on new plottable fields, just mark it with existing attribute:
```c#
public class SomePlottable
{
  [NotNAN]
  double SomeDataField;
}
```
this field will start to automaticaly be checked for NAN value in `Diagnostic mode`

To make new Type of check:
1. Create new `Attribute` discribing this check. A short name is very preferred because attributes create a lot of visual noise in `Plottables`.
2. Create new `FieldsChecker` derived from `FieldsCheckerBase`.
3. Add new check to `DiagnosticDataChecker.Checks`.
4. Mark required `Plottable` fields with new `Attribute`

**New functionality (code):**

```cs
plt.DiagnosticMode = true;
plt.PlotSignal(new double[]{12, 45, double.NAN, 54});

// Exception wil be thrown on first Render.
```


swharden:
Hi @StendProg, I really like the direction of this PR!

Having decorators like these that we can place on plottable data variables, combined with methods which can check them for validity, is an excellent combination. This will indeed make it easier for users to troubleshoot plots that fail due to bad data.

> PR contains 2 Diagnostic mode demo.
> It needs a good names and descriptions. My bad English won't let me do it myself.

Actually, it may be best to delete these demos since they throw exceptions which crash the cookbook and demo application.

image

 I'm not sure of the best way to communicate that this feature exists though. I wonder if it would be a good idea to put `diagnosticMode = false` right in the constructor?

https://github.com/swharden/ScottPlot/blob/4d06b49aa8ce85139fbc22a3a4990e311700ff6e/src/ScottPlot/Plot.cs#L25

This would make it easier to see for new users creating console applications, but unfortunately it would not be seen by people using the user control. 

I will continue to think about this problem!

StendProg:
Hi Scott (@swharden ),
I'm really glad you liked it.
> Actually, it may be best to delete these demos since they throw exceptions which crash the cookbook and demo application.

The sole purpose of these demos is to let the user know about this mode, I think you will find a beautiful solution without them.
We can make custom error notification, say:
1. Throw
2. Raise event
3. Write to log file.

Demo can use second variant, and replace demo description with error detail on event.

But I think most users will be satisfied with getting an exception to diagnose and fix the problem...

src/ScottPlot/Diagnostic/CheckFiledsOrderedAccendingDecorator.cs
```diff
@@ -0,0 +1,40 @@
+﻿using ScottPlot.Diagnostic.Attributes;
+using System;
+using System.Linq;
+using System.Reflection;
+
+namespace ScottPlot.Diagnostic
+{
+    public class CheckFiledsOrderedAccendingDecorator : PlottableDecoratorBase
```
Benny121221, line 8: This is a typo, it should be `CheckFieldsOrderedAscendingDecorator`, although the "Ordered" is probably redundant, I would suggest `CheckFieldsAscendingDecorator`.

src/ScottPlot/Plot.cs
```diff
@@ -225,7 +227,10 @@ public void SaveFig(string filePath, bool renderFirst = true)
 
         public void Add(Plottable plottable)
         {
-            settings.plottables.Add(plottable);
+            if (DiagnosticMode)
+                settings.plottables.Add(new DecoratorFactory().WrapWithAll(plottable));
```
Benny121221, line 231: This *is* a breaking change even though `PlottableDecoratorBase` extends `Plottable`

For example, if the user has code like this:
```cs
if(plottables[0].GetType() == typeof(PlottableOHLC)){
    //Foo
}
```

This is why C# has the `is` operator, which works correctly for childclasses:

```cs
if(plottables[0] is PlottableOHLC){
    //Foo
}
```

Now, ScottPlot doesn't make many guarantees about backwards compatibility when it comes to the actual plottables, because users are expected to use `Plot.PlotOHLC()` instead of using the constructors, and those methods are guaranteed to be backwards compatible. Furthermore, I don't think `settings.plottables` makes any backwards compatibility guarantees because users aren't expected to use the `Settings` object at all.

I understand this concern is pretty minor, and it's unclear whether this is supposed to be guaranteed backwards compatible code. Especially given that these issues will only exist in diagnostic mode which is for development only. And if these changes are too breaking, it could be merged into ScottPlot 5.0 instead of 4.0.40.

src/ScottPlot/Diagnostic/CheckFiledsOrderedAccendingDecorator.cs
StendProg: Your absolutely right. At first i plan to make universal `CheckFieldsOrdered`, but then I realized that in fact only Accending needed, and decided not to complicate things, somehow it happened.

src/ScottPlot/Plot.cs
swharden: Hey @Benny121221, this is a good point to be aware of, and your message is well-received. Breaking changes are something to keep an eye-out for, and _run time_ breaking changes could be especially difficult to troubleshoot for users not expecting it.

A discussion could be had about whether this is or is not the best architecture to use for data verification (e.g., adding a simple `bool verifyBeforeRender` and `VerifyNow()` to the `Plottable` class or maybe to a new `IVerifiable` interface may be preferred to this inheritance system), but maintaining strict backward-compatibility is not one of my chief concerns.

> ScottPlot doesn't make many guarantees about backwards compatibility

> it's unclear whether this is supposed to be guaranteed backwards compatible code

Backwards compatibility is something we prefer, but it's not a hardened rule. ScottPlot has a bit of a history of sneaking small breaking changes into the code base when a breaking change confers a real benefit to reducing bugs, improving performance, adding new useful features, or improving maintainability of the code base. I think the number of users is still small enough that I don't mind making small breaking changes which may only affect fringe use cases. 

The expectation of possible slight breaking changes is one reason [every version of ScottPlot has its own cookbook](https://swharden.com/scottplot/cookbook).

All that said, I think I'm going to release ScottPlot 4.0.40, merge this into ScottPlot 4.1.0, and make a couple larger breaking changes at the same time.

StendProg: > This is a breaking change even though PlottableDecoratorBase extends Plottable

Not really, it wraps around a real object. So you can't find it with `is` too. `Decorator` contain source plottable in protected `sourcePlottable` field. 
I proceeded from the assumption that the `settings.plottables` list is not available to the user and no one manipulates it manually in their projects. But you gave me the idea that he was additionally manipulated by Clear() methods and they check for type like you provided. I think I'll need to tweak them too.
>Now, ScottPlot doesn't make many guarantees ...

The key here is that when the diagnostic mode is turned off, ScottPlot works exactly the same as before. `settings.plottables` contain real plottable classes like `PlottableOHLC`. And none of these decorators are used at all. Only one check of the DiagnosticMode flag when adding Plottable. 

Regardless of the diagnostic mode, the user can add Plottable through the `Plot.Plot....` as before. api, or create manually using constructors and Add it with 'Plot.Add(new PlottableOHLC())'.

The diagnostic mode itself is designed for single use only.
When the user sees an incomprehensible library behavior, he turns `DiagnosticMode`, checks his data with it, fix errors in the data if any, and turns it off. If someone decides to enable it by default, then I am strongly opposed, this is a bunch of costly, unnecessary checks also through reflection, I do not approve of this.

> And if these changes are too breaking, it could be merged into ScottPlot 5.0 instead of 4.0.40.

Yes, this is not an urgent fix, and can be postponed for a long time. All the more, on good terms, it should be expanded to dozens of checks and go through all the existing Plotables.

I'm sorry if I didn't understand you (my poor English), but you didn't convince me about breaking changes.

Benny121221: `PlottableDecoratorBase` inherits from `Plottable`, which means the `is` operator will return `true` but `typeof`/`GetType()` will not. An example is available [here](https://tio.run/##dY49CsJAEIVr9xRTbgpzAEMKEbRREBQsxGKJmzCQ7ISdiRJCzr6uSSzt3s/34BW8LsjbEDpGV8GlZ7FNporaMMPZU@VNA4NasRjBAl6ETzgZdJrFx8H9AcZXnHyR1bZtawslEeTg7Bsmr5MsVjtyTLVNbx7FHtFZHbH0YOXatxGBPAeJikq99x1K8n8EyDAzERnVqJavUzaMi5uvbH5pCB8)

In my opinion, using the `is` operator is probably best practice anyways, especially when it comes to third-party libraries that may not guarantee types but only type compatibility. Plus the `is` operator is more convenient. I have seen style guides forbid `typeof`/`GetType()` except in situations where you cannot accept subclasses (which should never exist as it violates polymorphism). However, Microsoft doesn't give any warning against its use (that I can find).

As for situations where this matters, I can't think of very many. The best example I have is this code which would break if it used `typeof`/`GetType()` in diagnostic mode: https://github.com/Benny121221/WPlot/blob/master/MainWindow.xaml.cs#L388L399

Given that it only happens in diagnostic mode with questionable code, I don't think it matters.

Benny121221:
To make sure the user knows that diagnostic mode is not intended to be used for non-debugging purposes for performance reasons, it may make sense to log a warning when it is enabled?

swharden:
I like the idea...

Maybe something like Debug.WriteLine("WARNING: diagnostic mode is enabled, reducing performance")

Perhaps with some logic so the message is only shown once?

StendProg:
@Benny121221 pointed to the thing I missed. There is a `GetPlottables()` API method that gives the user full access to the list. In particular, he can delete, add, overwrite Plottable as he wants, and check for the existence of specific types in the list. This breaks the whole concept of decorators. Keeping another list and trying to synchronize them will make it all very difficult. The beauty of decorators is that you don't need to control all the Render calls, which can be scattered throughout the code in a chaotic manner. In reality, we have a clear rendering traversal of the list, and it might be worth really adding checks there without all those wraps.

What can be left of all this is clever universal checks through reflection on attributes. This will greatly reduce code cohesion and eliminate duplication. The only drawback is the visual pollution of the code with attributes.

src/ScottPlot/Plot.cs
StendProg: @Benny121221 , I understood what you meant, and you are absolutely right (I missed existing `GetPlottables()`), 
If the user can do something using the existing api, it is wrong to impose additional implicit restrictions.

swharden:
> What can be left of all this is clever universal checks through reflection on attributes. This will greatly reduce code cohesion and eliminate duplication. The only drawback is the visual pollution of the code with attributes.

I agree this is clever code, but I'm not totally convinced yet that decorators are the best way to solve this problem. A `CheckValidation()` method still is needed for things like comparing the length of `Xs` and `Ys` are equal on scatter plots, or that the number of errorbars is equal to the number of points. Using decorators means some rules would be defined as decorators on properties, and others in `CheckValidation()`.

It might be nicer to put all validation information for a plottable inside a `CheckValidation()` method (which would be required by `IPlottable`). We could eliminate duplication by putting assert methods in a static `Validate` class. Interestingly the validation method would resemble a series of unit tests. Perhaps this is a good thing? If we wanted to resemble nUnit we could create an `Assert` class and have similarly named methods which also take a message to provide if an exception occurs.

```cs
void CheckValidation(){
    Assert.NotNull(Xs, "Xs cannot be null");
    Assert.NotNull(Ys, "Ys cannot be null");
    Assert.EqualLength(Xs, Ys, "Xs and Ys have different lengths");
    Assert.IsReal(Xs, "Xs cannot contain infinity or NaN");
    Assert.IsReal(Ys, "Ys cannot contain infinity or NaN");
    Assert.Ascending(Xs, "Xs cannot contain any descending values");
}
```

I will continue to think about this! These are all interesting ideas.

StendProg:
> CheckValidation() method still is needed for things like comparing the length of Xs and Ys are equal on scatter plots

We can mark `xs` and `ys` with [EqualLength] attirbute. This will provide enouth information for check.

> Using decorators means some rules would be defined as decorators on properties, and others in CheckValidation().

Decorators, in my terms, are the objects that wrapped the plottable, I have already abandoned them, at least as long as `GetPlottable ()` API exists.

But mark proprtyes with attributes is another concept. The only thing that keeps this PR from closing.
The big advantage of this approach is that it is very easy to add checks to the newly created Plottable object. 
No matter how we simplify the creation of checks using helper methods when implementing the `IValidate` interface, it will still take more lines of code than simply marking up the required properties.
Considering the fact that the number of plotables, even at the moment, is already quite large, any savings have the right to be discussed.
The `IValidate` interface approach is also very clean, but you need to implement it right now in 20+ classes.

A significant drawback of attribute-based checks is that you will have to request properties through reflection, which is tens of times slower than the usual approach. But since these checks are only for special mode, where performance is not significant, this does not matter much.

StendProg:
I think it's finished.
I some `Plottables` like `PlottableVectorField`, `PlottableHeatmap` stay without attributes because their behavior not obvious for me.
During tests on demos i found that `Finance->SMA` demo contains NAN on `SMA Scatters`. I didn’t touch it since `SMA()` return `NAN` for elements outside the window...

swharden:
Thank you for your hard work on this @StendProg! As I work toward refactoring the constructors for all plottables (#573) I'll review their public properties and continue to build upon this work.

PR #559, opened 2020-09-22
ScottPlot 4.1.0-alpha (and new src/ folder structure)

swharden:
* version `4.0.40` → `4.1.0-alpha`
  * `-alpha` indicates prerelease
  * NuGet hides prereleases by default so we are more free to experiment with breaking API changes
* `src/` now contains `ScottPlot4` and `ScottPlot5` folders
  * development occur in parallel without requiring longstanding branches
* Tests have been moved from `tests/` to `src/ScottPlot4/ScottPlot.Tests/`
* Demo projects have been moved inside the `src/ScottPlot4/ScottPlot.Demo/` folder


StendProg:
Hi Scott (@swharden ),
As for me, the project is already very large, in order to manually rewrite it, something is sure to be lost and forgotten.
Although I remember this already said for version 3...

I have a suggestion on how to gradually implement `IRenderer` on the current version. It is possible to create a `ScottPlot4Renderer: IRenderer` that will build on existing `settings.gfxData` `settings.bmpData` and eventually cause rendering on them. 
Thus, both calls will be correct. Old directly in `gfxData` or new through an interlayer in the form of `ScottPlot4Render`. 
We can gradually replace direct calls to calls through the interface step by step while maintaining functionality. When all the calls have been replaced, all that remains is to replace `ScottPlot4Render` with another one that is no longer associated with `setting.gfxData`.

swharden:
Hi @StendProg, I agree with your assessment. Fully re-writing large code bases is typically a bad idea and usually not justified.  I think I did a poor job expressing the intent of this new folder structure. However, I also now think this PR should not be merged.

My original intent is to provide a place for extremely large changes to be experimented with, and only after a new system is determined to be ideal those changes can be implemented in ScottPlot 4. Many of the things the experimental ScottPlot 5 does well (improved coordinate system, multi-Y axes, and linked plots) can probably be applied to ScottPlot 4 as non-breaking changes, but it's very hard to test these in ScottPlot 4 because it could involve editing so many files (plottables, tests, demos, controls, etc).

My thought was to use the ScottPlot 5 folder to test/demonstrate new very large features (like the new coordinate system that allows multi-Y axes), but actually implement them in the ScottPlot 4 code base once they have stabilized. Features that cannot be implemented in ScottPlot 4 will stay in the ScottPlot 5 folder until the time comes to upgrade 4->5, and at that time implementation would occur in the ScottPlot 4 folder (it would get renamed to ScottPlot 5, and the experimental ScottPlot 5 folder would be deleted). This way there is no re-writing, and every plottable (for example) would have a continuous git history.

A better and less disruptive solution is probably to preserve `/src` and do this work in `/dev/ScottPlot5` 🤔

Issue #581, opened 2020-10-05
More flexible tick spacing


Benny121221:
**What can we do to make ScottPlot better?**
Currently automatic tick spacing is based on "nice numbers". These numbers are defined as a literal here: https://github.com/swharden/ScottPlot/blob/f9b4b2ad5520610c1d8bce4bba8e33bf3e318876/src/ScottPlot/Config/TickCollection.cs#L203

These give nice numbers in decimal, but not any other base. When I worked on #475 I neglected to work on this and then I forgot about it for several months :/ This is theoretically a very easy change, perhaps add another parameter for radix? Then nice numbers in hex could be `[2, 2, 2, 2]`, giving nice numbers of `[16, 8, 4, 2, 1]`. And mentions to `Log10` would be replaced with a call to the log function with the appropriate base.

Admittedly this is a pretty niche feature and while a function to create "nice" numbers for any radix is possible I'm unsure that it would really be used. As it is the only commonly used radices are decimal, hex, and binary. You can't really create a nice looking number in binary anyways so there's probably not much downside to just hardcoding in nice numbers for hex like was done for decimal. Might get nicer results that way too if it's hand done for each radix. If there's a compelling case that other radices might be used then I may change my mind. Keep in mind that [radices above 16 are not currently supported](https://github.com/swharden/ScottPlot/blob/master/src/ScottPlot/Tools.cs#L420L424).

Benny121221:
I implemented this on my branch https://github.com/Benny121221/ScottPlot/tree/nice-hexnumbers, however, I added some new tests and they didn't pass. I'll try and fix that or rewrite the offending code.

swharden:
Hey @Benny121221, this is an interesting idea! I agree that on one hand this is a pretty obscure use case, but on the other hand it seems that supporting it looks to be pretty simple.

As you refine this you could add another optional argument at the end of the `Plot.Ticks()` list

https://github.com/swharden/ScottPlot/blob/f9b4b2ad5520610c1d8bce4bba8e33bf3e318876/src/ScottPlot/Plot/Plot.Ticks.cs#L40-L42

to store a feature toggle in the end of the ticks config module

swharden:
Oops, I hit enter with the comment button selected and I guess I hit send too early. I pretty much captured all the technical details with the message above, but I wanted to follow-up with a "thanks for your effort on this one"! 

[...]

Issue #578, opened 2020-10-05
Road to ScottPlot 4.1


swharden:
### Update on November 17, 2020

**⚠️ I am reducing effort on issues and PRs while I work on this daily.**

* See [changelog.md](https://github.com/swharden/ScottPlot/blob/master/dev/changelog.md) for a summary of major changes.
* See [roadmap.md](https://github.com/swharden/ScottPlot/blob/master/dev/roadmap.md) for a big-picture discussion of plans and goals
* I merged #605 which accomplishes most of the goals described on this page
* The master branch is now ScottPlot `4.1-beta`. Old source code has its own [4.0-stable](https://github.com/swharden/ScottPlot/tree/4.0-stable) branch.

### Plottables to Refactor
- [x] Candlestick / OHLC - [PlottableOHLC.cs](https://github.com/swharden/ScottPlot/blob/master/src/ScottPlot/plottables/PlottableOHLC.cs)
- [x] Annotation - [PlottableText.cs](https://github.com/swharden/ScottPlot/blob/master/src/ScottPlot/plottables/PlottableText.cs)
- [x] Bar - [PlottableBar.cs](https://github.com/swharden/ScottPlot/blob/master/src/ScottPlot/plottables/PlottableBar.cs)
- [x] ErrorBar - [PlottableErrorBars.cs](https://github.com/swharden/ScottPlot/blob/master/src/ScottPlot/plottables/PlottableErrorBars.cs)
- [x] Function - [PlottableFunction.cs](https://github.com/swharden/ScottPlot/blob/master/src/ScottPlot/plottables/PlottableFunction.cs)
- [x] HLine - [PlottableAxisLine.cs](https://github.com/swharden/ScottPlot/blob/master/src/ScottPlot/plottables/PlottableAxisLine.cs)
- [x] HSpan - [PlottableAxisSpan.cs](https://github.com/swharden/ScottPlot/blob/master/src/ScottPlot/plottables/PlottableAxisSpan.cs)
- [x] VLine - [PlottableAxisLine.cs](https://github.com/swharden/ScottPlot/blob/master/src/ScottPlot/plottables/PlottableAxisLine.cs)
- [x] VSpan - [PlottableAxisSpan.cs](https://github.com/swharden/ScottPlot/blob/master/src/ScottPlot/plottables/PlottableAxisSpan.cs)
[...]

PR #582, opened 2020-10-06
Nice hexnumbers

Benny121221:
**New Contributors:**

**Purpose:**
#581 

**New Functionality:**
This is a comparison from the cookbook. Note that the only radices which currently have "nice" numbers are decimal and hexadecimal.

Before:
image

After:
image

**Other Concerns**
I noticed when writing this PR that in some situations the helper function `ScottPlot.Tools.ToDifferentBase` gave wrong answers, most noticeably in the case of numbers of the form r^k where r is the radix and k is some integer. It's hard to notice these cases when the numbers are not nice because the odds of that number also being a nice number in decimal are low; therefore it will not show up on the tick labels.

### DIFF
--- a/dev/changelog.md
+++ b/dev/changelog.md
@@ -6,6 +6,7 @@
 * Improved appearance of semi-transparent legend items (#567)
 * Improved tick labels for ticks smaller than 1E-5 (#568) _Thanks @ozgur640_
 * Improved support for Avalonia 0.10 (#571) _Thanks @Benny121221 and @apkrymov_
+* Improved positions for base16 ticks (#582, #581) _Thanks @Benny121221_
 
 ## ScottPlot 4.0.42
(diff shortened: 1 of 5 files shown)


Benny121221:
image

Well, I didn't change the file that is now failing, and this branch is behind master so I think it's the upstream branch.

swharden:
> It looks like Azure is running tests on a merge commit with the master branch. This is a weighty argument not to leave the master in a broken state...

That was definitely my fault! The last commit of the day broke the build (#578) and I didn't notice. I re-ran the failed pipelines and they pass now 👍

swharden:
Hey @Benny121221, thanks again for this! Sorry it took me so long to get to. #605 has absorbed most of me free time over the last couple weeks! Hopefully it'll get merged soon then I can clear out the rest of these issues.

Issue #586, opened 2020-10-07
Plottable H and V Span draglimits while drag fixed size


citizen3942:
When a plottable **H-** or **VSpan** is set up and it is configured to be a **DragFixedSize** span, and **dragLimitLower** and/or **dragLimitUpper** are set, the span can exit those limits while dragging. 
Currently, if I drag on the 2nd edge the 1st edge can get lower than the **dragLimitLower** value and the same opposite with the 1st edge dragging.
I believe this is not the kind of reaction one would expect.

I believe this part:
https://github.com/swharden/ScottPlot/blob/0fb684146dceb2c4963c9b2289ba9e8494d160df/src/ScottPlot/plottables/PlottableHSpan.cs#L95-L120

should check for boolean:
https://github.com/swharden/ScottPlot/blob/0fb684146dceb2c4963c9b2289ba9e8494d160df/src/ScottPlot/plottables/PlottableHSpan.cs#L71

And adapt to **dragLimitLower** and **dragLimitUpper** values.

citizen3942:
Even if I disable dragFixedSize it still is fixed size span? Why is that? Technically it should not....

StendProg:
Hi @citizen3942,
#589 fix this issue.

citizen3942:
> Hi @citizen3942,
> #589 fix this issue.

Yes indeed commits 6f0091db8ccf4633a5ed1d9a403df05c9a2280e7 and 2860ea53c2d75b5927ff2bd3a1762c25e459f148 worked perfectly.
I see that there is some refactoring for PlottableAxisSpans.
Merging with refactored function, I get:
```csharp
public void DragTo(double coordinateX, double coordinateY, bool isShiftDown = false, bool isAltDown = false, bool isCtrlDown = false)
        {
            if (!DragEnabled)
                return;

            double sizeBeforeDrag = position2 - position1;

            if (DragFixedSize || isShiftDown)
            {
                double AbsSize = Math.Abs(sizeBeforeDrag);
                bool DragMaxEdge = (position2 >= position1 && edgeUnderMouse == Edge.Edge2)
                                || (position1 >= position2 && edgeUnderMouse == Edge.Edge1);
                if (DragMaxEdge)
                {
                    if (coordinateX < dragLimitX1 + AbsSize) coordinateX = dragLimitX1 + AbsSize;
                    if (coordinateX > dragLimitX2) coordinateX = dragLimitX2;
[...]

PR #589, opened 2020-10-08
Drag fixed size fix

StendProg:
**Purpose:**
Drag `PlottableSpan` with fixed size now check for both Edges inside drag limits. Previosly it check only for dragged edge. #586


swharden:
fixed by 7b33ee5 in #586

Issue #592, opened 2020-10-09
Draggable DragEnter and DragExit Event 


charlescao460:
Hi,

First of all, thank you for this wonderful library!!

I am now working on a program which will play video, and in the meantime, show corresponding time-series data. 

And, I'd like to use `PlottableHSpan` as a slider to control the video playing, therefore user can use to navigate on displayed data. (See picture attached at the end) 

Unfortunately, I didn't find such drag events in the API. I'm using WPF, so I implemented it by myself in the way below, which is basically filtering `MouseButtonUp` event by myself. 
```C#
WpfPlot Plot = new WpfPlot();
Plot.MouseLeftButtonUp += OnXSpanMouseUp;
```
```C#
        private Point GetPixelPosition(MouseButtonEventArgs e)
        {
            float dpiScale = Plot.plt.GetSettings(false).gfxFigure.DpiX / 96;
            Point pos = e.GetPosition(this);
            pos.X *= dpiScale;
            pos.Y *= dpiScale;
            return pos;
        }
```
```C#
     private void OnXSpanMouseUp(object sender, MouseButtonEventArgs e)
     {
         var mousePixel = GetPixelPosition(e);
         var draggable = Plot.plt.GetDraggableUnderMouse(mousePixel.X, mousePixel.Y);
         if (draggable == null)
         {
             return;
         }
         OnXSpanDragged(); // Actual handler to control video
     }
```

It works fine, but I am wondering if it is possible to add **DragEnter** and **DragExit** event into the API? 

I know how to implement it in WPF, and I can create a PR for it. But I don't know how to implement for other framework. 

It would be great if someone can help. 

![image](https://user-images.githubusercontent.com/33741854/95623437-3dbfd200-0a43-11eb-8000-f5eac1c4ffc5.png)

StendProg:
Hi @charlescao460,
`WinForms control`(`FormsPlot`) already contain `MouseDragPlottable`, `MouseDragged` and `MouseDropPlottable` events. But `WPF control`  does not have this implementation.
>  know how to implement it in WPF, and I can create a PR for it

You are welcome, any helps matter.
The only thing I want to note is that it is important that different controls have a similar API, so try to repeat the events already available in `WinForms`.

charlescao460:
Hi @StendProg 

Thank you for your information!

[...]

PR #593, opened 2020-10-11
Add PlottableDragEventArgs

charlescao460:
**New Contributors:**

**Purpose:**
As mentioned in #592 , added events and handler when draggables are dragged.

**New Functionality:**
Added PlottableDragEventArgs in `ScottPlot` namespace
Such event was raised in `WpfPlot` and `FormsPlot`, but not implemented in Avalonia.

Tested in WPF: See demo codes here: https://github.com/charlescao460/ScottPlot/tree/DragEvent-Demo/src/ScottPlot.Demo.WPF.DraggableEvent
```cs
_hSpan = Plot.plt.PlotHSpan(0, 0.1, Color.Crimson, draggable: true, dragFixedSize: true);
Plot.MouseDragPlottable += OnMouseDrag;

...

private void OnMouseDrag(object sender, PlottableDragEventArgs agrs)
{
   if (agrs.Draggable == _hSpan)
   {
      TextBox.Text += "\nDrag H-Span!";
   }
}
```

image

**Note:**

Not sure if I need to change WinForm's event handler to `PlottableDragEventHandler`, because it will break backward compatibility. 

And in WPF, the mouse up event is named `OnMouseUpPlottable` rather than  `OnMouseDropPlottable`, which makes more sense to me.

This feature is not implemented in Avalonia. And I didn't find automated UI test so I didn't add test.


charlescao460:
src/ScottPlot.Demo.WPF/AssemblyInfo.cs
And
src/ScottPlot.WPF/AssemblyInfo.cs 
Are formatted by autoformat.bat

StendProg:
Hi @charlescao460, it looks fine.

There is some questions:
1. Does `plot` field in `PlottableDragEventArgs` realy needed? You can easy find it in event `source`, `source->plt`.
2. Does `EventType` field in `PlottableDragEventArgs` really needed? You already have different types of events which indicate event type.

There is probably nothing wrong with duplication, it can be convenient in some scenarios.

charlescao460:
Hi @StendProg 

Good point! I've changed my codes according to your suggestion. They look nicer now.

swharden:
Thanks @charlescao460 and @StendProg! Sorry I've taken a while to get to this (I've been pretty busy with #578).

I agree with @StendProg it's a good idea to isolate GUI code (events and mouse/keyboard interactions) to the GUI packages, and I can see that is what was done here with support for both WinForms and WPF. Thanks again!

PR #605, opened 2020-10-23
ScottPlot 4.1.0

swharden:
This PR extends work that started in #578 where the final steps are to replace the `Plottable` abstract class with an `IPlottable` interface, improve data validation for all plottables, improve the rendering system (render onto a single Bitmap and never store Bitmaps), and also add the framework to eventually support for multiple axes.

Since this involves small breaking changes, I'll roll version 4.0 -> 4.1 and mark NuGet packages as pre-release (beta) until development stabilizes.


StendProg (commented):
IsVisible -> Visible.
`Is` - commonly used for check methods.

src/ScottPlot/Validate.cs
```diff
@@ -29,6 +29,22 @@ public static void AssertAllReal(string label, double[] values)
                     throw new ArgumentException($"{label} index {i} is invalid ({values[i]})");
         }
 
+        /// <summary>
+        /// Throw an exception if the array is null or contains NaN or infinity
+        /// </summary>
+        public static void AssertAllReal<T>(string label, T[] values)
+        {
+            if (values is null)
```
StendProg, line 37: label also can be null or empty,
Maybe add some check for this?

src/ScottPlot/Validate.cs
```diff
@@ -29,6 +29,22 @@ public static void AssertAllReal(string label, double[] values)
...
+                throw new ArgumentException($"{label} must not be null");
+
+            for (int i = 0; i < values.Length; i++)
+            {
+                double val = Convert.ToDouble(values[i]);
```
StendProg, line 42: This translation is very controversial. There are only 2 types of floating point numbers (double, float).
It is more correct to organize switching by type.

src/ScottPlot/plottables/IPlottable.cs
```diff
@@ -1,19 +1,20 @@
...
+    [Obsolete("Use IPlottable", true)]
+    public abstract class Plottable { }
+
     public interface IPlottable
     {
         void Render(PlotDimensions dims, Bitmap bmp, bool lowQuality = false);
-
-        bool IsValidData(bool deepValidation = false);
-        string ValidationErrorMessage { get; }
+        string ErrorMessage(bool deepValidation = false);
```
StendProg, line 14: It is very unclear that this is a method with a bunch of checks. At first glance, it looks a lot like a simple query for a string. Maybe you should come up with a more descriptive name?

src/ScottPlot/plottables/IPlottable.cs
```diff
@@ -1,19 +1,20 @@
...
+        bool IsVisible { get; set; }
```
StendProg, line 15: `Visible` is more clear for me.

src/ScottPlot/plottables/IPlottable.cs
```diff
@@ -1,19 +1,20 @@
...
+        AxisLimits2D GetLimits();
+        int GetPointCount();
```
StendProg, line 17: int PointCount{get;}

src/ScottPlot/plottables/IPlottable.cs
```diff
@@ -1,19 +1,20 @@
...
+        LegendItem[] GetLegendItems();
```
StendProg, line 18: LegendItem[] LegendItems{get;} ?

src/ScottPlot/Plot/Plot.cs
```diff
@@ -134,16 +135,48 @@ private void RenderBitmap()
...
+
+                        if (ValidationErrorAction == ErrorAction.DebugLog)
+                            Debug.WriteLine(error);
+
+                        if (ValidationErrorAction == ErrorAction.ShowErrorOnPlot)
+                            errorMessages.AppendLine(error);
+
+                        if (ValidationErrorAction == ErrorAction.ThrowException)
+                            throw new InvalidOperationException(error);
+                    }
```
StendProg, line 165: This code can be moved inside some subsystem.

StendProg:
Hi @swharden, good job.
I maked review, but it does not contain any significant comments.

The only thing is, I would move the `deepValidation` into a separate method, and a separate interface. And would call it from Plot level as needed. This would allow simple Plottalbe to not implement this interface if they don't need it.

src/ScottPlot/Plot/Plot.Settings.cs
```diff
@@ -3,24 +3,24 @@
...
+using System.Security.Cryptography.X509Certificates;
 using System.Text;
 
 namespace ScottPlot
 {
     partial class Plot
     {
-        private bool diagnosticMode = false;
-        public bool DiagnosticMode
+        public enum ErrorAction { Render, SkipRender, DebugLog, ShowErrorOnPlot, ThrowException }
```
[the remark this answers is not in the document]
swharden: Thanks for the suggestion! There are several top-level enums scattered throughout the project. I'll create `Enums.cs` and put all the enums there. https://stackoverflow.com/questions/2282976/should-enums-in-c-sharp-have-their-own-file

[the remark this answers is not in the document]
StendProg: very very controversial, it's better to leave it as it is than to mix everything in one file.
Files are cheap these days, why not put it in a separate file with a good name?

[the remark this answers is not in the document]
swharden: > very very controversial

Ha, I agree! I didn't find a clear best practice when I googled how to do this. Common options are:
* one file per enum
* one file with all enums
* put emums in the same file as classes that use them

Files are cheap, but I don't want to clutter the top level of the project with many small files. Maybe a better solution is to make every enum its own file, but put them all inside an `Enums` folder?

src/ScottPlot/Plot/Plot.cs
```diff
@@ -203,12 +236,12 @@ public void SaveFig(string filePath, bool renderFirst = true)
...
         }
 
-        public void Add(Plottable plottable)
+        public void Add(IPlottable plottable)
         {
             settings.plottables.Add(plottable);
         }
 
-        public List<Plottable> GetPlottables()
+        public List<IPlottable> GetPlottables()
```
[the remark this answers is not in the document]
swharden: I like this idea! This seems like a good time to switch to a property with a simpler name too.

```cs
[Obsolete("Access the 'Plot.Plottables' array instead", true)]
public List<IPlottable> GetPlottables() => settings.plottables;

public IPlottable[] Plottables { get => settings.plottables.ToArray(); }
```

src/ScottPlot/Validate.cs
swharden: Good catch! I can add something simple like:

```cs
label = string.IsNullOrWhiteSpace(label) ? "[unknown variable]" : label;
```

src/ScottPlot/Validate.cs
swharden: Is something like this what you are suggesting?
```cs
public static void AssertAllReal<T>(string label, T[] values)
{
    if (typeof(T) == typeof(double))
        AssertAllReal(label, (double[])(object)values);
    else if (typeof(T) == typeof(float))
        AssertAllReal(label, (float[])(object)values);
    else
        throw new ArgumentException("values must be float[] or double[]");
}
```

src/ScottPlot/Plot/Plot.Settings.cs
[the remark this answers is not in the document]
StendProg: The only reason for the right choice is convenience in further work. I think you have the most important right to vote.

swharden:
**This is almost ready to merge into master.** The project will remain as "4.1-beta" (pre-release) for a while as I continue to make improvements. However, merging now means new PRs will be easier to manage. The last stable version of 4.0 can be accessed on https://github.com/swharden/ScottPlot/releases/tag/4.0.42

**Different systems (like ticks, axis labels, grids, plottables, etc) are now _much_ easier to work on** because they are `IRenderable` objects with fields to control their styling (rather than the old system where a settings module stored configuration objects that were read by a separate render method in another class).

**Multiple axes are supported,** but I'm still deciding the best way to expose this with the public API. Plottabes have a `xAxisIndex` and `yAxisIndex` (which defaults to `0`), and users can add an arbitrary number of axes. Currently only 4 axes are supported, but the framework is there to support more (stacking them horizontally or vertically). Better support for multiple axes will require a lot more work, but I can pause that effort for now and come back to it.

multi-axis

swharden:
Multi-axis plots are now fully supported, including support for an arbitrary number/combination of axes and automatic layout adjustment based on axis label and tick size. 

I'm still thinking about how to best present this using a simplified API, but for now this is functional.

https://github.com/swharden/ScottPlot/blob/14390b9beaad37c2c0fd331f350ac87d919cbc6e/tests/AxisRenderTests/MultiAxis.cs#L10-L60

image

swharden:
Large-scale refactoring is now mostly complete. Cookbook and demos work acceptably. Several improvements still need to be made, but they can achieved by modifying just a few modules, so I think now is a good time to merge this PR. 

See /dev/changelog.md for a summary of major changes.

Remaining high priority tasks include:
* Refactor layout API for improved simplicity
* Refine multi-axis API for improved simplicity
* Improve data validation for plottables
* Create new plottables specifically for growing data (mark public data fields readonly and provide setters to replace data arrays)
* Refactor the cookbook generator to create a simpler multi-page HTML (not markdown) static website
* Refactor user controls for improved multi-threaded rendering by making `Render()` non-blocking and thread-safe

Issue #628, opened 2020-11-25
Additional customizations for radar charts


SommerEngineering:
Dear team,

Thank you very much for this great library 🙂

I would like to see the possible font options (e.g. size, bold, font) for the radar charts as well. As far as I can see (cf. [line 121 in RadarPlot.cs](https://github.com/swharden/ScottPlot/blob/63cc24c05912db00b1ea42b856869fe76ef923ae/src/ScottPlot/Plottable/RadarPlot.cs#L121)) the font cannot be configured right now. I would like to format the font for the axis and the category names differently. For me, this would be a valuable addition.

Also, the different axis of the radar seems to be automatic. I would need the opposite: Manually fixed axis. Let's say something like this:

```
var allValues = new [,] { { 5, 357, 0.244 } };
var categoryNames = new[] { "A", "B", "C" };
var fillColors = new[] {Color.FromArgb(128, Color.CornflowerBlue)};

plt.PlotRadar(values: allValues, categoryNames: categoryNames, categoryFontSize: 26, fillColors: fillColors);

plt.RadarAxis(axisIndex: 1, min: 1, max: 7, fontSize: 21, bold: true);
plt.RadarAxis(axisIndex: 2, min: 10, max: 750, fontSize: 21, bold: true);
plt.RadarAxis(axisIndex: 3, min: -1, max: 1, fontSize: 21, bold: true);
```

swharden:
Hey @SommerEngineering, thanks for the suggestions!

Radar plots were first discussed in #428 and #430, and I'll tag @Benny121221 in case they want to comment.

#### Font Customization
I'll add some public fields to [RadarPlot.cs](https://github.com/swharden/ScottPlot/blob/master/src/ScottPlot/Plottable/RadarPlot.cs) so users can tweak the font. It'll be in ScottPlot 4.1 which might not get released for a little while though, so if you want an immediate solution you can download the ScottPlot 4.0 source code (it has its own branch) and make this modification yourself.

#### Radar Axis
I don't think I fully understand your goal here... Is your goal to be able to individually customize font options for each "arm" of the radar plot? Does your question somehow relate to the coordinate system in 2D space?

Benny121221:
## Fonts

[...]

PR #634, opened 2020-11-28
Radar multi axis

Benny121221:
**New Contributors:**

**Purpose:**
#628 

**New Functionality:**
Adds two options, a parameter `bool independentAxes` which determines whether the axes should be separate. And a public property `bool showAxisLabels` which determines whether axis labels are shown.

The following demo shows both:
```cs
double[,] values = { { 5, 3, 10, 15, 3, 2, 256 }, { 5, 2, 10, 10, 1, 4, 252 }, };
string[] categories = { "Wins", "Poles", "Podiums", "Points Finishes", "DNFs", "Fastest Laps", "Points" };
string[] groups = { "Sebastian Vettel", "Fernando Alonso" };

var radarPlot = plt.PlotRadar(values, categories, groups, independentAxes: true);
radarPlot.showAxisLabels = false;
plt.Legend();

// customize the plot
plt.Grid(false);
plt.Frame(false);
plt.Ticks(false, false);
plt.Title("2010 Formula One World Championship");

/* Data represents the 2010 Formula One World Championship
 * https://en.wikipedia.org/wiki/2010_Formula_One_World_Championship
 * Note: Alonso did not finish (DNF) in the Malaysian GP, but was included 
 * here because he completed >90% of the race distance.
 */
```

Labeled:
image

Unlabeled:
image

As you might expect, the graph is rather cluttered with all the axis labels turned on.

**Other Notes**
The axis limits for radar plots were adjusted in 2d4f388e3937042805b82995a30713ff12ab801c to make it look nicer.

I accidentally created this when changing stuff, and I actually thought it looked cool:
image
Makes me wonder if there should be an option to remove the circular rings? Or replace them with polygonal ones, like in this one, for example: 
image

### DIFF
--- a/dev/changelog.md
+++ b/dev/changelog.md
@@ -1,6 +1,7 @@
 # ScottPlot Changelog
 
 ## ScottPlot 4.1.0 ⚠️ in development
+* Additional customizations for radar charts (#634, #628) _Thanks @Benny121221_
 
 Work toward ScottPlot 4.1 began in October, 2020 and merged into the master branch one month later ([#605](https://github.com/swharden/ScottPlot/pull/605)). Improvements are focused at enhanced performance, improved thread safety, support for multiple axes, and options for data validation. See [roadmap.md](roadmap.md) for details.
 
--- a/src/ScottPlot/Plot/Plot.PlotMisc.cs
+++ b/src/ScottPlot/Plot/Plot.PlotMisc.cs
@@ -65,13 +65,14 @@ namespace ScottPlot
             string[] groupNames = null,
             Color[] fillColors = null,
             double fillAlpha = .4,
-            Color? webColor = null
+            Color? webColor = null,
+            bool independentAxes = false
             )
         {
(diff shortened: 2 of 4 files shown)


Benny121221:
I don't know why the code format failed, I ran `dotnet-format` locally.

swharden:
> I don't know why the code format failed

Demo/PlotTypes/Radar.cs
```cs
- double[,] values = { { 5, 3, 10, 15, 3, 2,  256}, { 5, 2, 10, 10, 1, 4, 252 }, };
+ double[,] values = { { 5, 3, 10, 15, 3, 2, 256 }, { 5, 2, 10, 10, 1, 4, 252 }, };
```

RadarPlot.cs: the whole file is tabs instead of spaces

> I ran dotnet-format locally.

Try running it from the `/src/` folder? Is there another `.editorconfig` hiding in the subfolders?

Aside from the formatting formality, this is looking excellent! Thank you @Benny121221!

Benny121221:
> Try running it from the /src/ folder?

I did.

>  Is there another .editorconfig hiding in the subfolders?

There is not.

¯\_(ツ)_/¯ very strange...

Issue #649, opened 2020-12-09
Resource leak in Demos when opening/closing Live Data forms


philfontaine:
**I have a PR for this issue #654**

The Live Data demos all use 2 timers, one to update the data and one to re-render. Both of these timers are not stopped when closing the forms, so they keep running in the background. 

Problems that this is causing:
- CPU is still high, even after having closed the Live Data examples
- The more you open/close the dialog, the more laggy it will get
- The more you open/close the dialog, the more memory it leaks

This affects WinForms, WPF and Avalonia.

PR #650, opened 2020-12-09
Fix some resource leaks (CPU and mem) when using Live Data forms

philfontaine:
**New Contributors:**

**Purpose:**
#649

For WPF and Avalonia:

I used the `Closed` event to stop the timers. This fixes the issue.

For WinForms:

Timers are linked to the window, which is exactly what we want. The only problem is that we did not call `Dispose` on the form after it was closed. By calling dispose on the form, it will take care of its timers.

Note: I did not know if you wanted me to target master or this branch. If it was master, you could just cherry-pick the commit.

### DIFF
--- a/src/demo/ScottPlot.Demo.Avalonia/AvaloniaDemos/LiveDataFixed.axaml.cs
+++ b/src/demo/ScottPlot.Demo.Avalonia/AvaloniaDemos/LiveDataFixed.axaml.cs
@@ -17,6 +17,9 @@ namespace ScottPlot.Demo.Avalonia.AvaloniaDemos
         DataGen.Electrocardiogram ecg = new DataGen.Electrocardiogram();
         Stopwatch sw = Stopwatch.StartNew();
 
+        private Timer _updateDataTimer;
+        private DispatcherTimer _renderTimer;
+
         public LiveDataFixed()
         {
             this.InitializeComponent();
@@ -33,23 +36,25 @@ namespace ScottPlot.Demo.Avalonia.AvaloniaDemos
             avaPlot1.plt.SetAxisLimits(yMin: -1, yMax: 2.5);
 
             // create a traditional timer to update the data
-            _timer = new Timer(_ => UpdateData(), null, 0, 5);
+            _updateDataTimer = new Timer(_ => UpdateData(), null, 0, 5);
 
             // create a separate timer to update the GUI
-            DispatcherTimer renderTimer = new DispatcherTimer();
-            renderTimer.Interval = TimeSpan.FromMilliseconds(10);
-            renderTimer.Tick += Render;
-            renderTimer.Start();
+            _renderTimer = new DispatcherTimer();
+            _renderTimer.Interval = TimeSpan.FromMilliseconds(10);
+            _renderTimer.Tick += Render;
+            _renderTimer.Start();
 
+            Closed += (sender, args) =>
+            {
+                _updateDataTimer?.Dispose();
+                _renderTimer?.Stop();
+            };
         }
 
         private void InitializeComponent()
         {
             AvaloniaXamlLoader.Load(this);
         }
-
-
-        private System.Threading.Timer _timer;
 
         void UpdateData()
         {
--- a/src/demo/ScottPlot.Demo.Avalonia/AvaloniaDemos/LiveDataGrowing.axaml.cs
+++ b/src/demo/ScottPlot.Demo.Avalonia/AvaloniaDemos/LiveDataGrowing.axaml.cs
@@ -20,6 +20,9 @@ namespace ScottPlot.Demo.Avalonia.AvaloniaDemos
         TextBox LatestValueTextbox;
         CheckBox AutoAxisCheckbox;
 
+        private DispatcherTimer _updateDataTimer;
+        private DispatcherTimer _renderTimer;
+
         public LiveDataGrowing()
         {
             this.InitializeComponent();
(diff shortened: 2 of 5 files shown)


philfontaine:
BTW I ran the code formatter. There were some files that were modified, but they had nothing to do with this PR, so I did not commit them.

Benny121221:
Yeah, currently checks do not run on commits to branches other than master, but they do run on PRs to every branch. So the code formatting thing wouldn't be caught until your PR, even though it's likely existed for a while.

Note the second last line in this log (from the Job Build and Test on MacOS pipeline, the Build ScottPlot.Demo task)
```
2020-12-09T20:01:11.0927960Z ##[section]Starting: Build ScottPlot.Demo
2020-12-09T20:01:11.0941880Z ==============================================================================
2020-12-09T20:01:11.0942230Z Task         : .NET Core
2020-12-09T20:01:11.0942590Z Description  : Build, test, package, or publish a dotnet application, or run a custom dotnet command
2020-12-09T20:01:11.0942980Z Version      : 2.179.2
2020-12-09T20:01:11.0943230Z Author       : Microsoft Corporation
2020-12-09T20:01:11.0943610Z Help         : https://docs.microsoft.com/azure/devops/pipelines/tasks/build/dotnet-core-cli
2020-12-09T20:01:11.0944030Z ==============================================================================
2020-12-09T20:01:11.6342140Z Info: .NET Core SDK/runtime 2.2 and 3.0 are now End of Life(EOL) and have been removed from all hosted agents. If you're using these SDK/runtimes on hosted agents, kindly upgrade to newer versions which are not EOL, or else use UseDotNet task to install the required version.
2020-12-09T20:01:11.6463460Z Info: Azure Pipelines hosted agents have been updated and now contain .Net 5.x SDK/Runtime along with the older .Net Core version which are currently lts. Unless you have locked down a SDK version for your project(s), 5.x SDK might be picked up which might have breaking behavior as compared to previous versions. You can learn more about the breaking changes here: https://docs.microsoft.com/en-us/dotnet/core/tools/ and https://docs.microsoft.com/en-us/dotnet/core/compatibility/ . To learn about more such changes and troubleshoot, refer here: https://docs.microsoft.com/en-us/azure/devops/pipelines/tasks/build/dotnet-core-cli?view=azure-devops#troubleshooting
2020-12-09T20:01:11.6526070Z ##[error]Project file(s) matching the specified pattern were not found.
2020-12-09T20:01:11.6547420Z ##[section]Finishing: Build ScottPlot.Demo
```

It looks like the build failed because `ScottPlot.Demo` could not be found. This often happens if you move/rename project files without updating the solution file (for example you do it from the command line or a file manager not from Visual Studio). Or the changes to the solution file were not committed. This could also be because the build pipelines themselves have paths to the project files and those paths are now incorrect.

I'm tempted to say it may be a good idea to look into running checks on commits to other branches. However, this is unconventional and it may get expensive in Azure credits. Generally one only *really* cares if master (or release branches if you're using something like gitflow) is in a good state, not dev branches.

I feel like the problem is the inconsistency, not the less strict checking; checks are required on PRs to dev branches but not on normal commits, however it would be less confusing if it were either required on all commits or no commits.

swharden:
Hey @philfontaine, thanks for catching this and suggesting a fix!

The branch this PR was opened on was a short-lived branch that has since been merged and deleted, so I'm going to try to re-target this PR and merge it into master. If there's a merge conflict, I'll try to resolve it on my end and get it merged in tonight.

@Benny121221 thanks for your feedback about Azure Pipelines too. I knew the branch would break the Azure Pipelines test system so I didn't open a PR for it until I was done with it. It's since finished and merged, and I'll publish a new pre-release (hopefully with a nice new cookbook) later tonight

PS: GitHub discussions came out this week and I started one (#653) so feel free to comment there if you have thoughts about the new API or cookbook

swharden:
Hey @philfontaine, sorry for the inconvenience here. It looks like I can't re-target this PR since the original branch has been deleted. If you create a new PR targeting master, I'd be happy to merge this in!

Also, thinking out loud, I wonder if:
```cs
var liveDataIncoming = new WinFormsDemos.LiveDataIncoming();
liveDataIncoming.ShowDialog();
liveDataIncoming.Dispose();
```

could be simplified to:
```cs
using (var liveDataIncoming = WinFormsDemos.LiveDataIncoming())
    liveDataIncoming.ShowDialog();
```

(with or without curly braces, I'm indifferent to one-line `using` statements)

swharden:
For the record, an alternative solution could be to have the form that creates the timers also be responsible for disposing them when the window closes. The advantage here is that the caller wouldn't have to remember which forms require disposal and which ones don't. I think something like this would work, but haven't tested it 🤔 

```cs
protected override void OnFormClosing(FormClosingEventArgs e)
{
    _updateDataTimer.Dispose();
    base.OnFormClosing(e);
}
```

philfontaine:
Sure, I'm gonna make another PR. 

There is also the `Closed` event on WinForms, so it could work the same way as WPF and Avalonia. However, the reason I did not do it is because the timers are linked to the form (i.e. `this.timerUpdateData = new System.Windows.Forms.Timer(this.components);`). 

So, if someone calls dispose on the form + if we ourselves dispose the timers in the `Closed` event, then they are gonna be disposed twice. Of course, they should tolerate that just fine, so it should not be an issue. 

So, for consistency, I will dispose the timers the same way as we do in WPF and Avalonia.

Let me know if that sounds good @swharden

swharden:
> for consistency, I will dispose the timers the same way as we do in WPF and Avalonia.

> I'm just gonna call Stop on the timer and that should be good enough (No need to dispose it).

Both sound like good plans! Thanks for making a new PR, and I'll merge it when it comes in 👍

Issue #639, opened 2020-12-01
New Cookbook / FAQ / Quickstart website generator


swharden:
I am working to improve the cookbook system
* The ultimate goal is a simpler cookbook
* The new cookbook will use generated HTML (not markdown)
* The new cookbook will have multiple pages 
* Every `IRecipe` has a `String Category` to indicate what page it belongs to
* A converter isolates recipe code into flat text files to simplify source code distribution with the demo program

swharden:
The new cookbook is live:
https://swharden.com/scottplot/cookbook

Because every plot type has its own page I'm less pressured to keep the total number of recipes low and I am more open to showcasing more advanced styling examples, edge cases, unique use cases, etc.

The FAQ section will be added after the new user control system is developed #637

Issue #571, opened 2020-09-29
Support for Avalonia 0.10


apkrymov:
System.MissingMethodException: "Method not found: 'Void Avalonia.Controls.Image.set_Source(Avalonia.Media.Imaging.IBitmap)'." occured when I try to run application on Avalonia 10.0 instead 9.11.

Are there any plans to make the library compatible with the latest version of Avalonia?

The problem in AvaPlotBackend.cs.
```c#
public override void SetImagePlot(bool lowQuality)
{
    view.Find<Ava.Controls.Image>("imagePlot").Source = BmpImageFromBmp(plt.GetBitmap(true, lowQuality));
}
```

swharden:
@Benny121221 (#496) I haven't looked into this yet... do you have any suggestions?

Benny121221:
Firstly, sorry for the slow response, midterms are getting wild.

@apkrymov The issue does not appear when the `ScottPlot.Avalonia` library is compiled with Avalonia 0.10.0-preview5, the most recent preview version available as long as you remove/comment out this [line](https://github.com/swharden/ScottPlot/blob/master/src/ScottPlot.Avalonia/HelpWindow.axaml.cs#L16). Since Avalonia 0.10 is in preview I don't think it makes sense to publish a new release to nuget for support unless @swharden wants to go down the road of shipping preview versions to nuget (which I suspect he doesn't but he may surprise me).

Hopefully by the time Avalonia reaches 1.0 there'll be better compatibility between versions but this is what 0.x versions are for, they're expected to change often. In case you're curious the issue was caused by the relevant function going from taking an `IBitmap` to an `IImage`.

So in summary, I've set up github to give me a notification on new Avalonia releases and when 0.10.0 comes out of preview I'll make a pull request to add support. It seems like it's only changing the dependencies and removing one line. If Scott is ok with shipping preview releases to NuGet it may happen sooner but that could mean a lot more work for him. If you need Avalonia 0.10.0 features right away then compiling the library your self probably makes the most sense, and I suppose at least we know that it does work with only the minor changes above. Sorry for any inconvenience.

swharden:
Hi @Benny121221, no worries about the delay! Thank you for your thorough answer.

### Supperted Version

Re-stating @Benny121221's answer, ScottPlot currently targets/supports Avalonia 0.9.11 as defined by these lines. These lines can be changed to control which version of Avalonia is used when the `ScottPlot.Avalonia` NuGet package is installed.

[...]

PR #656, opened 2020-12-10
Avalonia 0.10.0

Benny121221:
**New Contributors:**

**Purpose:**
Adds support for Avalonia 0.10.0rc1. #571

**New Functionality:**
Works correctly on the latest Avalonia Release Candidate.

I created this pull request in anticipation of an imminent non-prerelease version of Avalonia. In theory, all one would have to do is change the package references to the new version. I said earlier that I would do this once it had actually been released, however, I currently have finals and I would like to have the actual work done in advance (turned out there wasn't much work to do anyways) so I can focus on those. So even if the new version drops right in the middle of finals no one would have to wait for support on the latest version. I'm not expecting any breaking changes between now and the release of 0.10.0, so it *should* just be as simple as changing the package versions.

Also, the Avalonia cookbooks do not work, I believe this is from #639, not anything specific to the new Avalonia version. I have elected not to touch the cookbooks at all for this reason. I did touch one demo because there was a minor change to the `PlotViewer` class in e182ae3.

### DIFF
--- a/src/controls/ScottPlot.Avalonia/AvaPlotViewer.axaml.cs
+++ b/src/controls/ScottPlot.Avalonia/AvaPlotViewer.axaml.cs
@@ -29,5 +29,10 @@ namespace ScottPlot.Avalonia
         {
             AvaloniaXamlLoader.Load(this);
         }
+
+        public void SetWindowOwner(WindowBase owner)
+        {
+            this.Owner = owner;
+        }
     }
 }
--- a/src/controls/ScottPlot.Avalonia/ScottPlot.Avalonia.NUGET.csproj
+++ b/src/controls/ScottPlot.Avalonia/ScottPlot.Avalonia.NUGET.csproj
@@ -33,8 +33,8 @@ Release demo: http://swharden.com/scottplot/demo</PackageReleaseNotes>
     <None Include="icon.png" Pack="true" PackagePath="\" />
     <None Include="icon.ico" Pack="true" PackagePath="\" />
     <None Include="README.txt" Pack="true" PackagePath="\" />
-    <PackageReference Include="Avalonia" Version="0.10.0-rc1" />
-    <PackageReference Include="Avalonia.Desktop" Version="0.10.0-rc1" />
+    <PackageReference Include="Avalonia" Version="0.10.0" />
+    <PackageReference Include="Avalonia.Desktop" Version="0.10.0" />
   </ItemGroup>
   
 </Project>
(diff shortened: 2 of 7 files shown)


Benny121221:
When the new release drops I can change the package references in `ScottPlot.Avalonia.csproj` and `Avalonia Demo.csproj` and assuming there are no problems I will mark it as not a draft PR.

swharden:
Hey @Benny121221, I recognize a lot of the code base is pretty active right now, but thanks for working around it and issuing this PR!

> Also, the Avalonia cookbooks do not work, I believe this is from #639

They seem to be working on the latest version of the master branch... but this is probably a separate topic we can explore deeper as we get closer to a release.

image

> I currently have finals and I would like to have the actual work done in advance (turned out there wasn't much work to do anyways) so I can focus on those

I really appreciate all the work you've put in to make this library better! Take all the time you need to ensure you'll do great on those finals, and fingers crossed I'll continue to have made good progress between now and then 🤞

Benny121221:
> They seem to be working on the latest version of the master branch... but this is probably a separate topic we can explore deeper as we get closer to a release.

Yeah, you are correct. I just saw `NullReferenceException` on the template tree and assumed that you changed its structure without reading deeper. Turns out it's an issue where `TreeView.SelectedItem` can be null even when there is a selection. I fixed this by using `TreeView.SelectedItems` (a list).

@FoggyFinder knows more about Avalonia than me, this isn't listed anywhere in the [breaking changes ](https://github.com/AvaloniaUI/Avalonia/wiki/Breaking-Changes#09---010).

There are similar issues listed in this gitter: https://gitter.im/AvaloniaUI/Avalonia?at=5e479a1fd56ddb68a4a79fcd

My observed behaviour was clicking the first time would set `SelectedItem` to null, and the second click would set it to the expected value. If I'm doing something wrong I would be happy to open an issue in the Avalonia repo.

Benny121221:
@swharden Avaloniao 0.10.0 was released earlier this morning. I have rebased this branch to match up with master.

swharden:
> @swharden Avaloniao 0.10.0 was released earlier this morning. I have rebased this branch to match up with master.

Awesome, is this ready to merge now?

I plan to make another beta release on NuGet this weekend so this can get out there in the wild pretty soon!

swharden:
> is this ready to merge now?

I'm pretty sure it is and you're done with this PR, so I'll save you the reply an go ahead and merge, then I can start working on #700 👍

PR #662, opened 2020-12-18
Fix AxisSpan disapear on extreme zoom in

StendProg:
**Purpose:** (bug fix)

Then zoom to `AxisSpan`, it shoud fill whole plot with its color. But on extreme zoom `AxisSpan` disapear.
This is a known `GDI+` bug with clipping objects that go far beyond the clipping region.
To fix bug, this PR manually clip `AxisSpan` by 1 pixel more then `Plottable area` before passing it to` GDI+`.

Additional added 2 integration tests that reproduce this bug.
This tests fails without this PR fix.


swharden:
Hi @StendProg, thank you for noticing this problem and fixing it!

PR #667, opened 2020-12-20
Implemented separable gaussian blur for improved performance.

Benny121221:
In ScottPlot.Tools.XYToIntensities

**New Contributors:**

**Purpose:**
Improved performance for `ScottPlot.Tools.XYToIntensities` when using gaussian intensity mode.

The gaussian blur is separable, instead of using an n x n matrix as the kernel one can use a 1 x n matrix followed by an n x 1 matrix. This gives O(n) complexity instead of O(n^2).

Note that the previous code did not explicitly use a blur filter, but instead applied the gaussian after adding each coordinate. This is still O(n^2).

This code also precomputes the kernel, which ought to bring a performance boost of its own.

**New Functionality:**
N/A

### DIFF
--- a/src/ScottPlot/Tools.cs
+++ b/src/ScottPlot/Tools.cs
@@ -289,18 +289,57 @@ namespace ScottPlot
             double[,] output = new double[height, width];
             if (mode == IntensityMode.gaussian)
             {
+                double[,] intermediate = new double[height, width];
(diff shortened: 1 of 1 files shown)


Benny121221:
This does not address the concerns raised in #666, although this PR could be expanded to do so.

PR #668, opened 2020-12-20
Fixes transparency threshold to compare with normalized values

Benny121221:
**New Contributors:**

**Purpose:**
Currently when using the `TransparencyThreshold` parameter on heatmaps it compares the intensity on a range from 0 to 255 to the given threshold. This threshold, however, is given on a range from 0 to 1. This PR corrects that issue.

**New Functionality:**
N/A

### DIFF
--- a/src/ScottPlot/Drawing/Colormap.cs
+++ b/src/ScottPlot/Drawing/Colormap.cs
@@ -131,9 +131,31 @@ namespace ScottPlot.Drawing
             {
                 byte pixelIntensity = (byte)Math.Max(Math.Min(intensities[i] * 255, 255), 0);
                 var (r, g, b) = colorMap.GetRGB(pixelIntensity);
-                byte alpha = intensities[i] < minimumIntensity ? (byte)0 : (byte)255;
+                byte alpha = pixelIntensity < minimumIntensity ? (byte)0 : (byte)255;
                 byte[] argb = { b, g, r, alpha };
                 rgbas[i] = BitConverter.ToInt32(argb, 0);
+            }
+            return rgbas;
+        }
+
+        public static int[] GetRGBAs(double?[] intensities, Colormap colorMap, double minimumIntensity = 0)
+        {
+            int[] rgbas = new int[intensities.Length];
+            for (int i = 0; i < intensities.Length; i++)
+            {
+                if (intensities[i].HasValue)
+                {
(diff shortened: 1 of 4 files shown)


Benny121221:
4fd5ea6 breaks backwards compatibility, heatmap data was changed from `double[,]` to `double?[,]`

Benny121221:
I avoided breaking backwards compatibility with the `Colormap.GetRGBAs` module, as it has an overload for `double[]` and for `double?[]`

swharden:
The use of `double?[,]` is interesting - I like it!

Let me know when you're done working on this PR and I'll merge it in 👍

Benny121221:
The issue in `ScottPlot.Demo.WinForms\WinFormsDemos\ColormapViewer.cs` is interesting, all the methods like `Tools.XYToIntensities` and `DataGen.SampleImageData` return `double[,]`

I think the best solution here is to let heatmaps take both, so I added an overload to `Plot.Obsolete.PlotHeatmap` and `Heatmap.UpdateData` to accommodate non-nullable doubles. This brings back backwards compatibility, and it means that `Tools.XYToIntensities` and similar can be used without needing to convert in user code (or without changing those functions to return nullable doubles when they never return null, which would have a larger compatibility footprint as well as introducing conversions if they were to be used outside of heatmaps).

Benny121221:
@swharden I think it's ready to merge. I added a NaN check which throws an `ArgumentException` if there's a NaN value. I elected not to check for +/- infinity as they may be useful as sentinels, but I can add that check easily if you think it makes more sense.

swharden:
> can be used without needing to convert in user code

I've been changing a lot between 4.0 and 4.1, and I'm typically ok with this as long as an [Obsolete] decorator to an old method or class indicates what to use instead. I'll probably end-up renaming `UpdateData()` to just `Update()` to make it consistent with scatter and signal plots...

On one hand I don't want to cause pain for people who have written existing code and want to upgrade, but on the other hand I want the API to be as intuitive and discoverable as possible for new users (because there will probably be many more of those than existing users today), so I'm incentivized to make changes now if I think it'll make the software easier to learn in the future.

I'm reminded of a quote I heard something like, "We could force everything to always be back-compatible so old users will never have to change their code and they will get all the new features for free, but then everyone gets worse API" 🤔

PR #671, opened 2020-12-21
Fix display bug in WinForms MouseTracker demo

MosheBarkan:


swharden:
Good catch @MosheBarkan! I double-checked the WPF and Avalonia demos and they are implemented correctly.

Issue #678, opened 2020-12-24
Colormap scalebar Bitmap generation should be abstracted


swharden:
It may be useful to allow colormaps to generate their own scalebars as standalone bitmaps. Colormap-scalebar-generating code should be removed out of the heatmap module and placed elsewhere...

This refactoring can be performed in parallel with #677 and #675

Benny121221:
It may be useful to model this off of the `Renderable.Legend`, or if you want it to be an `IPlottable` then `Plottable.ScaleBar` might make sense. Conceptually they are similar, although getting the colours could be interesting. `ScaleBar` gets passed in `width` and `height` explicitly, while `Legend` has its items updated on every [render](https://github.com/swharden/ScottPlot/blob/3bf13aa5a64b2d7cbdeb1e27c73bb245b9cc07d3/src/ScottPlot/Plot/Plot.Render.cs#L88).

I think getting the colourmap information in the same way that `Legend` does is better as it's much more convenient from a user perspective, although it does introduce extra coupling.

swharden:
@Benny121221 thanks for the idea! I think a standalone plot type for this would be excellent. I'll call it a `Colorbar`, borrowing the term from matplotlib. This plot type (with its own methods for getting the bitmap with/without labels) will indeed help solve this issue, while also removing a bit of complexity from the heatmap module. I'll explore this idea shortly...

I just finished refactoring the heatmap module a bit and adding recipes to the 4.1 cookbook. (#679) I'm curious if you think the "Scale Limits" example is working as expected. The cookbook is generated whenever the tests are run, but I'm attaching the heatmap page in case you don't want to fire that up: [Cookbook: Plottable Heatmap.zip](https://github.com/swharden/ScottPlot/files/5740598/Cookbook_.Plottable_.Heatmap.zip)

It might benefit from refactoring so _styling_ options are public fields but options which affect the data are passed in as arguments to the `Update()` function 🤔

Benny121221:
>  I'm curious if you think the "Scale Limits" example is working as expected.

Yep, it looks right. It does look a little strange because you're clipping half the range, but that's what I would expect it to look like.

> It might benefit from refactoring so styling options are public fields but options which affect the data are passed in as arguments to the Update() function 🤔

[...]

Issue #677, opened 2020-12-24
Heatmap scale is broken


swharden:
This is surely a mistake I made somewhere between 4.0.40 and 4.0.44, but it looks like the heatmap scalebar no longer displays

https://swharden.com/scottplot/cookbooks/4.0.40/#experimental-heatmap-heatmap-quickstart
https://swharden.com/scottplot/cookbooks/4.0.47/#experimental-heatmap-heatmap-quickstart

This would have been noticed faster if heatmap examples were added to the ScottPlot 4.1 cookbook (#675)

swharden:
I fixed this in ScottPlot 4.1 with #679 (it was an error caused by clipping)

This is not easy to fix in 4.0 though. In 4.0 only the data area bitmap is passed into the Render() function:

https://github.com/swharden/ScottPlot/blob/343f74864cebb3cde3cef579bab7f8fc48bded5f/src/ScottPlot/Renderer.cs#L40

This is already working pretty well in 4.1 and this will continue to mature rapidly there, so I probably won't go back and fix this in 4.0. I'll post a new 4.1 pre-release on NuGet (with accompanying cookbook) shortly showing how to work with heatmaps and colorbars.

Issue #675, opened 2020-12-24
Add heatmap recipes to cookbook


swharden:
ScottPlot 4.1 cookbook lacks heatmap recipes, although the ScottPlot 4.0 cookbook has many of them. Add new recipes and publish a new pre-release with updated cookbook.

swharden:
done in #679 and comments in #678

PR #679, opened 2020-12-24
Refactor heatmaps

swharden:
refactor heatmaps to address #677, #675, and #678

### DIFF
--- a/src/tests/Cookbook/Validate.cs
+++ b/src/tests/Cookbook/Validate.cs
@@ -1,6 +1,7 @@
 ﻿using NUnit.Framework;
 using ScottPlot.Cookbook.Site;
 using System;
+using System.Collections.Generic;
 using System.Linq;
 using System.Text;
 using System.Threading.Tasks;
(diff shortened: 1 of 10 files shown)


src/tests/Cookbook/Validate.cs
```diff
@@ -74,6 +75,21 @@ public void Test_CookbookRecipes_IDsAreValid()
             }
         }
 
+        [Test]
+        public void Test_CookbookRecipes_TitlesInsideACategoryAreUnique()
+        {
+            foreach (var cat in ScottPlot.Cookbook.Locate.GetCategorizedRecipes())
+            {
+                List<string> seen = new List<string>();
```
Benny121221, line 83: This is admittedly nitpicky considering this likely has a meaningless performance impact as the categories likely contain less than a dozen demos. However, as the order is unimportant to this problem, and the only operations applied to the collection are adding elements and checking for membership, a `HashSet` would be a better choice.

`HashSet` uses black magic (hash functions) to give constant time membership checking. A `List` gives linear time. If you're unfamiliar a set basically can be thought of as a dictionary with only keys, no values.

Benny121221: For what it's worth, if you need ordered elements you can use a `SortedSet` which stores elements in _sorted_ order (but you can provide whichever `IComparer` you like). This is (presumably) implemented as a Binary Search Tree, as it gives O(log n) search. `SortedSet` still does not allow duplicates, you'd have to create your own search tree to do that.

swharden: Hi @Benny121221, thanks for the suggestion! I made this change in e2275a8

I don't usually try to optimize cookbook code (or most code outside hot loops), but this conversation is an excellent reminder that a strong understanding of the `System.Collections.Generic` types can be very useful! Thanks for bringing it up

https://docs.microsoft.com/en-us/dotnet/api/system.collections.generic?view=net-5.0

PR #680, opened 2020-12-25
Colour Bar Interpolation Should be Nearest Neighbour

Benny121221:
**New Contributors:**

**Purpose:**
Some tweaks after #679 

**New Functionality:**
Previously: 
image

Now:
image

Part of the issue was that the black rectangle around the colourbar was twice the width of the colourbar. I elected to half the width of the rectangle, but you could also double the width of the colourbar. As for the interpolation, that fixes the colour seeming to fade to white at the right edge of the colourbar

This also changes the description of a demo: 

> This option results in a colormap with 4x the number of squares.

is replaced with:
> This option results in a heatmap with a standard deviation of 4.

I believe this mistake may have come from the confusingly named `sampleWidth` parameter. I think it could be split up into two parameters (with appropriate defaults) instead of meaning either the width of the square used for calculating density, or standard deviation, depending on context. Or it could be split into two different functions.

There was a similar issue with the density mode, the demo originally said it used bilinear interpolation to construct the heatmap, when it actually counts the number of points within a square of fixed size.

### DIFF
--- a/src/ScottPlot/Plottable/Heatmap.cs
+++ b/src/ScottPlot/Plottable/Heatmap.cs
@@ -218,11 +218,12 @@ namespace ScottPlot.Plottable
             using (var font = GDI.Font(null, 12))
             using (var sf2 = new StringFormat() { LineAlignment = StringAlignment.Far })
             {
+                gfx.InterpolationMode = System.Drawing.Drawing2D.InterpolationMode.NearestNeighbor;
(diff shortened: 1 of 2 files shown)


swharden:
@Benny121221 Thanks for updating the descriptions! I implemented these changes directly in #681. Since over there I implemented a colorbar as its own plot type (and pulled that code out of the heatmap module) I'll close this PR, but I very much welcome your input on the direction the new Colorbar plot type is going and how it interacts with heatmaps! It feels a little clunky right now, and I'm curious what you think might make it feel more natural

Issue #637, opened 2020-11-30
Refactor all controls to use a common back-end


Benny121221:
**What can we do to make ScottPlot better?**
This was originally the intention, however, when the Avalonia module was created for 4.0.39 it was decided that it would be annoying to bump the package version of the rest of ScottPlot. This change would make a move to supporting another platform easier (or indeed updating the WPF and WinForms controls to use `ControlBackend` which would help in maintainability). I think I said I would backport `ControlBackend` to WPF and WinForms when I first created the Avalonia control, 4 months ago.........

The [ScottPlot Roadmap](https://github.com/swharden/ScottPlot/blob/master/dev/roadmap.md) may provide additional context regarding current goals and future directions.

swharden:
Hey @Benny121221! 

**A central control backend is still the plan/goal,** but I don't think right now is the best time to be refactoring the controls. The core library is being deeply refactored daily (#578), and I'm devoting all my free time right now to getting the core library stabilized and documented (#572). After that settles down I'll focus my free time to refining the controls, including refactoring toward a central backend that lives in the `ScottPlot` namespace (probably using your [ScottPlot.Avalonia/ControlBackend.cs](https://github.com/swharden/ScottPlot/blob/master/src/controls/ScottPlot.Avalonia/ControlBackend.cs) as a starting point). 

**Advantages of a central control backend are:** it allows controls to be as thin as possible, greatly reduces code duplication across the controls, makes it easy to implement controls for new frameworks (WinUI and MAUI are on the horizon), and makes it easier to test GUI functionality with traditional unit tests.

**My immediate next steps** are refining and documenting multi-axis support, and refactoring the website and documentation away from huge single-page cookbooks toward multi-page, version-specific, auto-generated documentation pages (#572). I'd also like to get most of the website content (minus images) back into source control ([/dev/website](https://github.com/swharden/ScottPlot/tree/master/dev/website)). I think after that would be a good time to start devoting primary focus toward refactoring the controls. I'll update the roadmap soon to reflect this (I recognize it's fallen behind).

**I have two questions that may be worth exploring before refactoring the controls.** If we decide to make a large-scale change to how user controls display plots, it may be good to make the change at the same time we refactor to use a common backend. Those questions are:

[...]

PR #681, opened 2020-12-25
Colorbar Plottable

swharden:
Previously colorbar rendering was achieved inside the heatmap module. This PR gives colorbar its own plot type, and `AddColorbar()` can be called on any plot. This addresses #678

The colorbar could benefit from continued refinement/refactoring (especially the tick system), but this is a strong start.

```cs
var hm = plt.AddHeatmap(intensities);

var cb = plt.AddColorbar(hm.Colormap);
cb.SetTicks(hm.ColorbarMin, hm.ColorbarMax);
```

image

This PR also makes some changes suggested by #680

### DIFF
--- a/src/ScottPlot/Plot/Plot.Add.cs
+++ b/src/ScottPlot/Plot/Plot.Add.cs
@@ -168,6 +168,17 @@ namespace ScottPlot
         }
 
         /// <summary>
+        /// Add a colorbar to display a colormap beside the data area
+        /// </summary>
+        public Colorbar AddColorbar(Drawing.Colormap colormap = null, int space = 100)
+        {
+            var plottable = new Colorbar(colormap);
+            Add(plottable);
+            YAxis2.SetSizeLimit(min: space);
+            return plottable;
+        }
+
+        /// <summary>
         /// Create a polygon to fill the area between Y values and a baseline.
         /// </summary>
         public Polygon AddFill(double[] xs, double[] ys, double baseline = 0, Color? color = null)
@@ -273,13 +284,15 @@ namespace ScottPlot
         /// <summary>
         /// Add a heatmap to the plot
         /// </summary>
-        public Heatmap AddHeatmap(double[,] initialData, int scalebarSpace = 100, bool lockScales = true)
+        public Heatmap AddHeatmap(double[,] intensities, bool lockScales = true, Drawing.Colormap colormap = null)
         {
+            if (lockScales)
+                AxisScaleLock(true);
+
             var plottable = new Heatmap();
-            plottable.Update(initialData);
+            plottable.Update(intensities, colormap);
             Add(plottable);
-            YAxis2.SetSizeLimit(min: scalebarSpace);
-            AxisScaleLock(lockScales);
+
             return plottable;
         }
 
--- a/src/ScottPlot/Plot/Plot.Add.cs
+++ b/src/ScottPlot/Plot/Plot.Add.cs
@@ -172,10 +172,23 @@ namespace ScottPlot
         /// </summary>
         public Colorbar AddColorbar(Drawing.Colormap colormap = null, int space = 100)
         {
-            var plottable = new Colorbar(colormap);
-            Add(plottable);
+            var cb = new Colorbar(colormap);
+            Add(cb);
             YAxis2.SetSizeLimit(min: space);
-            return plottable;
+            return cb;
+        }
+
+        /// <summary>
+        /// Add a colorbar initialized with settings from a heatmap
+        /// </summary>
+        public Colorbar AddColorbar(Heatmap heatmap, int space = 100)
+        {
+            var cb = new Colorbar(heatmap.Colormap);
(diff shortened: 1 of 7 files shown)


swharden:
After separating the heatmap from the colorbar I was surprised how clunky it feels to ask the heatmap what the min/max limits are then load them into the colormap. It feels like the API could be improved still 🤔

Benny121221:
>  how clunky it feels to ask the heatmap what the min/max limits are then load them into the colormap.

One option would be to grab the information from the `Plot` object, but that introduces a lot of coupling. I bring it up because that's similar to what the `Legend` renderable does, but the legend is pretty intrinsically tied to a `Plot` object whereas a heatmap is not.

You could give a heatmap a `ColorBar` plottable as one of its fields. Then the heatmap could populate all the relevant state. The user would be free to create their own `ColorBar` on their own, but they would also have a convenient method to create a `ColorBar` from a heatmap. This method probably could be written pretty simply as well. Whether the heatmap holds the colorbar itself and is responsible for rendering it or a factory method returns the colorbar for the user to render is a design choice.

Note that if you go the factory method route there are a couple of snakes waiting should the heatmap be modified, I don't know if you want the user to take care of that or if you want to create something (such as a class inheriting from `ColorBar`) which can "sync up" with the parent heatmap. If you go that route you're going to have to make a choice about how to treat orphan `ColorBar`s should their parent heatmap be disposed.

If you go with the observer pattern then that choice is delightfully known as "dead [objects] tell no tales", i.e. the colorbar will never change after its parent heatmap is disposed. However if you use another route, such as checking the parent heatmap has not changed before every render of the colorbar than that choice is simply to prevent the parent heatmap from being disposed by the garbage collector because there is a reference to the parent heatmap held by the colorbar. There is another option with this approach, if you create a weak reference (a reference that does not prohibit garbage collection) then you get a dangling reference, which would create a `NullReferenceException` when dereferenced. One could, in principle, detect this and instead choose any other behaviour, which is why I bring this up, this method is the most flexible. C# offers the `System.WeakReference` class so there's no need to mess about with pointers (or `IntPtr`). This class conveniently offers the `IsAlive` method to determine whether it's target has been garbage collected, so fancy deconstructors are not necessary.

Benny121221:
I implemented this with the observer pattern in #682.

Basically, the end result is you can just do this:

```cs
Colorbar cb = plt.AddColorbarFromHeatmap(hm);
```

If you want to disable syncing (i.e. create a colorbar from the heatmap and don't modify it on subsequent changes) you can do this:

```cs
Colorbar cb = plt.AddColorbarFromHeatmap(hm, synced: false);
```

swharden:
Hey @Benny121221, thanks for your input! There is a lot of good discussion here.

### Syncing

You gave a lot of thoughts about how to keep a colorbar and heamap sync'd. At the end of the day, I don't think I'm in favor of this though... It adds a lot of coupling and complexity to this library, all so the end-user doesn't have to write an extra line or two of code in the niche case that they want to change colormaps or tick labels after a heatmap is displayed.

You pointed out a lot of "snakes" above and gave some good ideas about how to overcome them, but altogether I think I'm leaning away from this type of syncing in general in an effort to keep the code minimal and well isolated.

However, observer pattern would be a good way to accomplish syncing, and the implementation in this PR is a good demonstration of this pattern! Interestingly, a strong argument could be made that the control should the other way around: A colorbar should be a type of `Axis`, and a heatmap should update its image any time the colorbar's colormap settings changes... 

Although I don't think we should go down that road at this time, if we did want to pursue syncing I think we should go all-in and have an `IHasColormap` that plottables could implement so all coupling points to the interface (not to a concrete plot type).

### Simpler API to Create a Heatmap and Colorbar

> You could give a heatmap a ColorBar plottable as one of its fields or allow it to create one via a public factory method. Then the heatmap could populate all the relevant state.

Maybe we could collapse this into a single method? This could create both items at the same time and initialize the colorbar the appropriate state, and return both in case the user wants to fine-tune something 🤔 

```cs
var (hm, cb) = plt.AddHeatmapAndColorbar(intensities);
```

Benny121221:
Of the options you listed my preference would be allow a user to create a colorbar with the ticks up front:

```cs
plt.AddColorbar​(​hm​.​Colormap​, ​hm​.​MinLabel​, hm.MaxLabel);
```

Or to create a colorbar from a heatmap:

```cs
plt.AddColorbar(hm);
```

On the topic of syncing, I mostly agree. While it is convenient, it's not a likely issue to come up, and if it does it's not difficult for a user to maintain the two objects with the same state. If they were so inclined they could even use the observer pattern with something like a decorator class.

swharden:
> Or to create a colorbar from a heatmap:
> `plt.AddColorbar(hm);`

@Benny121221 I created this method like you suggested and it really simplified the cookbook examples! I added a new cookbook section for Colorbar too.

There is certainly room for improvement/refinement for both `Heatmap` and `Colorbar`, but both seem usable as they are and the cookbook examples are looking much cleaner. Unless you have any other suggestions, I'll merge this and make a NuGet package later today then go back to working on #637

PR #682, opened 2020-12-26
AddColorbarFromHeatmap and syncing Colorbars with Heatmaps

Benny121221:
**New Contributors:**

**Purpose:**
Adds a method `plt.AddColorFromHeatmap` which allows optional syncing Colorbars with a "parent" Heatmap (I don't know if parent is the right term for the observer pattern). Any call to the `Heatmap.Update` method will trigger this syncing.

**New Functionality:**

```cs
Colorbar cb = plt.AddColorbarFromHeatmap(hm);
```

The signature of this method is `public Colorbar AddColorbarFromHeatmap(Heatmap hm, int space = 100, bool synced = true)`, so setting `synced` to `false` lets the user disable syncing (the colorbar will be synced with the heatmap on its creation but it will not be synced to reflect further updates to the heatmap).

### DIFF
--- a/src/ScottPlot/Plot/Plot.Add.cs
+++ b/src/ScottPlot/Plot/Plot.Add.cs
@@ -179,6 +179,18 @@ namespace ScottPlot
         }
 
         /// <summary>
+        /// Add a colorbar to display a colormap beside the data area
+        /// </summary>
+        /// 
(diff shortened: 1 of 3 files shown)


swharden:
We decided not to sync settings between plottables and colorbars at this time (see discussion in #681), although this PR is a useful demonstration of how to use observer pattern to achieve this effect 👍

PR #690, opened 2021-01-09
Used QuickSelect to add a NthOrderStatistic method

Benny121221:
**New Contributors:**

**Purpose:**
An ith order statistic is an element of an array that is the ith smallest value. It is a generalization of the min (the 0th order statistic*), the max (the last order statistic), the median (the n/2 th order statistic), as well as all quantiles.

I also added a test for the selection algorithm. It validates the algorithm against naïve algorithm (sorting and then indexing) for a large array of doubles.

*it is the 1st order statistic if you index from one. The algorithm I implemented indexes from zero.

**New Functionality:**

```cs
double nthOrderStatistic = ScottPlot.Statistics.Common.NthOrderStatistic(values, n);
```

This is implemented using QuickSelect, the younger brother of QuickSort. It gives O(n) expected time, but O(n^2) worst case. The worst case time complexity (like in QuickSort) is exceedingly unlikely when using a randomized algorithm. 

The naïve algorithm is O(n log n), where one sorts the array first and then indexes into it. C# uses IntroSort on `Array.Sort()`, which is essentially QuickSort but which switches to another sorting algorithm to maintain O(n log n) at all times (QuickSort has a O(n^2) worst case). [Array.Sort Remarks](https://docs.microsoft.com/en-us/dotnet/api/system.array.sort?redirectedfrom=MSDN&view=net-5.0#System_Array_Sort_System_Array_ ), [Wikipedia on Introsort](https://en.wikipedia.org/wiki/Introsort)

The official IntroSelect uses a hybrid of QuickSelect and Median of Medians (an algorithm which approximates a median to choose a better pivot). Median of Medians always yields O(n) time complexity, but has higher overhead. There are other "IntroSelects" which can promise O(n) expected and O(n log n) worst case (by using sorting). If it becomes appropriate I can take a look at implementing an IntroSelect algorithm. There is also Floyd-Rivest which is a more complicated algorithm which may be more appropriate than either type of IntroSelect

### DIFF
--- a/src/ScottPlot/Statistics/Common.cs
+++ b/src/ScottPlot/Statistics/Common.cs
@@ -61,6 +61,20 @@ namespace ScottPlot.Statistics
             return Quantile(values, i, 100);
         }
 
+        public static double Median(double[] values)
+        {
+            // Note, this is one of many definitions of a median. It is the most common.
+
+            if (values.Length % 2 == 1)
+            {
+                return NthOrderStatistic(values, values.Length / 2 + 1);
+            }
+            else
+            {
+                return (NthOrderStatistic(values, values.Length / 2) + NthOrderStatistic(values, values.Length / 2 + 1)) / 2;
+            }
+        }
+
         private static double QuickSelect(double[] values, int begin, int end, int i)
         {
             // QuickSelect (aka Hoare's Algorithm) is a selection algorithm (i.e. given an integer i it returns the ith smallest element in a sequence) with O(n) expected time.
--- a/src/ScottPlot/Statistics/Common.cs
+++ b/src/ScottPlot/Statistics/Common.cs
@@ -51,6 +51,34 @@ namespace ScottPlot.Statistics
                 return values[begin];
             }
 
+            if(i == 0)
+			{
+                double min = values[begin];
+                for(int j = begin; j <= end; j++)
+				{
+                    if(values[j] < min)
+					{
+                        min = values[j];
+					}
+				}
+
+                return min;
+			}
+
+            if(i == end - begin)
+			{
+                double max = values[begin];
+                for (int j = begin; j <= end; j++)
+                {
+                    if (values[j] > max)
+                    {
+                        max = values[j];
+                    }
+                }
+
+                return max;
+            }
+
             int pivot_index = Partition(values, begin, end);
             int k = pivot_index - begin;
 
--- a/src/ScottPlot/Statistics/Common.cs
+++ b/src/ScottPlot/Statistics/Common.cs
@@ -36,7 +36,7 @@ namespace ScottPlot.Statistics
             double[] copied_values = new double[values.Length];
             Array.Copy(values, copied_values, values.Length); // QuickSelect mutates the array
 
-            return QuickSelect(copied_values, 0, values.Length - 1, n);
+            return QuickSelect(copied_values, 0, values.Length - 1, n - 1);
         }
 
         private static double QuickSelect(double[] values, int begin, int end, int i)
--- a/src/ScottPlot/Statistics/Common.cs
+++ b/src/ScottPlot/Statistics/Common.cs
@@ -33,7 +33,10 @@ namespace ScottPlot.Statistics
 
         public static double NthOrderStatistic(double[] values, int n)
         {
-            return QuickSelect(values, 0, values.Length - 1, n);
+            double[] copied_values = new double[values.Length];
+            Array.Copy(values, copied_values, values.Length); // QuickSelect mutates the array
+
+            return QuickSelect(copied_values, 0, values.Length - 1, n);
         }
 
         private static double QuickSelect(double[] values, int begin, int end, int i)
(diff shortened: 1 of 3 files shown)


Benny121221:
I created a demo which looks like this:

image

swharden:
Hey @Benny121221, thank you for this PR which offers improved performance for finding the Nth-smallest element of an array!

### Increased testing speed

I modified the tests in a1a6f2d to only test 100 random `n`s rather than all `n`s, increasing the speed of this test from half a second to about ten milliseconds. However, while writing tests for the extremes I got confused about what the range of valid `n`s should be.

### What is the valid range of `n` values?

* Presently `n` must be from `0` through `values.Length - 1`
* Should `n` be from `1`  through `values.Length`?

Looking over some of the references you linked to, `n` often starts at 1, so I'm interested in your input here. Perhaps an XML doc on top of the `NthOrderStatistic()` method could clarify this, and an exception could be thrown if `n` is outside the expected range.

### Percentile()

Since this offers an improved method for finding the percentile, could you also add a `Percentile` method which uses your improved algorithm? We could mimic [numpy's signature](https://numpy.org/doc/stable/reference/generated/numpy.percentile.html) where the first argument is an array and the second argument is a floating point number between 0 and 100 inclusive.

Thanks for your input on these topics!

Benny121221:
> ### What is the valid range of `n` values?
> * Presently `n` must be from `0` through `values.Length - 1`
> * Should `n` be from `1`  through `values.Length`?

I think order statistics are often 1 indexed in math/stats, as the interpretation as the kth smallest value makes more sense because ordinals start at 1, there is no "0th". In CS it's sometimes 1 to comply with the math/stats crowd, but as is often the case 0 indexing is more convenient to use in an algorithm. I'll change it so that it starts at 1.

> ### Percentile()
> Since this offers an improved method for finding the percentile, could you also add a `Percentile` method which uses your improved algorithm?

Good idea, I'll work on that.

> We could mimic numpy's signature where the first argument is an array and the second argument is a floating point number between 0 and 100 inclusive.

I'm a little spooked by fractional percentiles, as there are already multiple definitions of percentiles. Perhaps a `Quantile` function may make more sense, which can take a parameter (typically named `q`) and give the nth q-quantile.

I see you added some tests for the min and max value, worth noting that QuickSelect (or indeed any other selection algorithm) will perform worse than a simple for loop for finding the mins and maxes. It is still O(n) expected time, but the constant there is higher, whereas for a for loop the constant there is 1.

I may consider adding a special case for these in `QuickSelect` itself, since it's recursive it may be looking for the min/max of a partition even if not the min/max of the array, so this minor performance benefit will help for more than just the min/max values of n.

Benny121221:
I am hesitant to offer median as it has a few meanings:

image

However I recognize that most user, would use sorting were they to compute the median themselves. Or use a library like `Math.Net Numerics`, which uses the [average of the upper and lower medians](https://github.com/mathnet/mathnet-numerics/blob/master/src/Numerics/Statistics/ArrayStatistics.cs#L416-L418)

I think the best option is to go with the average, as it's the most common definition, and I believe it's what the rest of ScottPlot uses.

swharden:
Hey @Benny121221, thanks for your work on this! I've been a bit occupied with other projects lately but I'm starting to work on this one again. I was close to merging this PR but I had 1 thing I wanted to check with you first about...

I moved the random integer generator out of the quickselect method and noticed it was generating numbers with an _inclusive_ lower bound but an _exclusive_ upper bound. I suspect this was not intended, and may affect the quicksort algorithm. 

On Wikipedia I see quicksort uses a random number:

```cs
left + floor(rand() % (right − left + 1))
```

And I noticed the `+1` was omitted from your code but I think your `end` and `begin` are both inclusive.

https://github.com/ScottPlot/ScottPlot/blob/2aa51d351ade0b98ed351ff7c3d7048210c1f7a3/src/ScottPlot/Statistics/Common.cs#L140

Want to double-check and let me know what you think? I updated the XML documentation for all the methods based on my understanding, but let me know if I got anything wrong there.

Thanks again for this great work!

Benny121221:
> I moved the random integer generator out of the quickselect method and noticed it was generating numbers with an inclusive lower bound but an exclusive upper bound. I suspect this was not intended, and may affect the quicksort algorithm.

You are right, this is incorrect, thank you.

>  I updated the XML documentation for all the methods based on my understanding, but let me know if I got anything wrong there.

No big issues, I think you said somewhere that QuickSelect sorted the array when it does not (although it does mutate the array in a way that kinda looks like sorting, if you look at this visualization: https://en.wikipedia.org/wiki/Quickselect#/media/File:Selecting_quickselect_frames.gif)

The only other changes I made were to the percentile overload which takes a `double`, it was previously using sorting instead of QuickSelect. And I added some tests for that same method, they just check that it gets the quartiles right. Tt's not a high coverage test because it's assumed that the ways that method could go wrong are covered by the general NthOrderStatistic tests.

swharden:
Looks great, I'll merge it in now. Thanks for all the work on this @Benny121221!

PR #692, opened 2021-01-11
Add nullable support where possible

### LINKED ISSUE #691: C# 8 nullable support
**What can we do to make ScottPlot better?**

Consider adding nullable annotations as they will increase productivity of using library.
This will clarify what you may expect from function.

For example:
https://github.com/ScottPlot/ScottPlot/blob/6e9a9ac35254210636be16abecff4ad147f8370e/src/ScottPlot/DataGen.cs#L100
Only those who read source code would know that they may pass null as first parameter

Or:
https://github.com/ScottPlot/ScottPlot/blob/6e9a9ac35254210636be16abecff4ad147f8370e/src/ScottPlot/Ticks/TickCollection.cs#L334
This may not be obvious that if you pass array with less than 2 elements it will return null

**Is this breaking change?**

No. Nullable annotation are not breaking existing code. Types remain same, but with some attributes, not changing behavior. It will work with .Net Framework, I've tested it with .Net Framework 4.0 and 4.6 in different projects

jl0pd:
**New Contributors:**

**Purpose:**
Add nullable reference types support
Resolves #691 

**New Functionality:**
None

### DIFF
--- a/src/ScottPlot/DataGen.cs
+++ b/src/ScottPlot/DataGen.cs
@@ -9,6 +9,8 @@ using System.Runtime.InteropServices;
 using System.Text;
 using System.Threading.Tasks;
 
+// TODO: Need to inspect Random instantiation because some of them is seed (new Random(0)), some not (new Random())
+
 namespace ScottPlot
 {
     public static class DataGen
@@ -87,7 +89,7 @@ namespace ScottPlot
             return ys;
         }
 
-        public static double[] Random(Random rand, int pointCount, double multiplier = 1, double offset = 0)
+        public static double[] Random(Random? rand, int pointCount, double multiplier = 1, double offset = 0)
         {
             if (rand is null)
                 rand = new Random();
@@ -97,7 +99,7 @@ namespace ScottPlot
             return ys;
         }
 
-        public static int[] RandomInts(Random rand, int pointCount, double multiplier = 1, double offset = 0)
+        public static int[] RandomInts(Random? rand, int pointCount, double multiplier = 1, double offset = 0)
         {
             if (rand is null)
                 rand = new Random();
@@ -133,7 +135,7 @@ namespace ScottPlot
             return values;
         }
 
-        public static double[] NoisyLinear(Random rand, int pointCount = 100, double slope = 1, double offset = 0, double noise = 0.1)
+        public static double[] NoisyLinear(Random? rand, int pointCount = 100, double slope = 1, double offset = 0, double noise = 0.1)
         {
             if (rand is null)
                 rand = new Random(0);
@@ -145,7 +147,7 @@ namespace ScottPlot
             return data;
         }
 
-        public static double[] NoisySin(Random rand, int pointCount, double oscillations = 1, double noiseLevel = .5, double mult = 1)
+        public static double[] NoisySin(Random? rand, int pointCount, double oscillations = 1, double noiseLevel = .5, double mult = 1)
         {
             if (rand is null)
                 rand = new Random(0);
@@ -155,7 +157,7 @@ namespace ScottPlot
             return values;
         }
 
-        public static Color RandomColor(Random rand, int min = 0, int max = 255)
+        public static Color RandomColor(Random? rand, int min = 0, int max = 255)
         {
             if (rand is null)
                 rand = new Random();
@@ -211,7 +213,7 @@ namespace ScottPlot
         /// <summary>
         /// Return OHLC array with random prices X positions as sequential numbers (0, 1, 2, etc.)
         /// </summary>
-        private static OHLC[] RandomStockPrices(Random rand, int pointCount, double mult = 10, double startingPrice = 123.45)
+        private static OHLC[] RandomStockPrices(Random? rand, int pointCount, double mult = 10, double startingPrice = 123.45)
         {
             if (rand is null)
                 rand = new Random(0);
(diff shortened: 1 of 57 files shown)


src/ScottPlot/Plottable/DataStructures/Segmentedtree.cs
```diff
@@ -53,14 +52,20 @@ public SegmentedTree(T[] data)
 
         public async Task SetSourceAsync(T[] data)
         {
-            if (data == null)
-                throw new Exception("Data cannot be null");
-            sourceArray = data;
-            await Task.Run(() => UpdateTreesInBackground());
+            sourceArray = data ?? throw new ArgumentNullException("Data cannot be null");
+            await Task.Run(() => UpdateTrees());
```
jl0pd, line 56: This should be fixed in different PR
With old solution we're not awaiting real operation, only creation of task, that will be scheduled
As the result we might begin to work with data, which is not yet updated properly

src/ScottPlot/Plottable/FinancePlot.cs
```diff
@@ -28,12 +28,12 @@ public class FinancePlot : IPlottable
         /// <summary>
         /// If True then OHLC timestamps are ignored and candles are placed at x= 0, 1, 2, 3, etc.
         /// </summary>
-        public bool Sqeuential;
+        public bool Sequential;
```
jl0pd, line 31: Renaming public field/property - breaking change

src/ScottPlot/Plottable/RadarPlot.cs
```diff
@@ -10,21 +11,22 @@ public class RadarPlot : IPlottable
...
-        public string[] GroupLabels;
+        private readonly double[]? NormMaxes;
+        public string[]? CategoryLabels;
+        public string[]? GroupLabels;
         public Color[] FillColors;
         public Color[] LineColors;
         public Color WebColor = Color.Gray;
-        public readonly bool IndependentAxes;
+        [MemberNotNull(nameof(NormMaxes))]
+        public bool IndependentAxes { get; }
```
jl0pd, line 21: Converted from read-only field to get-only property. Binary breaking change. Recompilation should work

src/ScottPlot/Statistics/Interpolation/SplineInterpolator.cs
```diff
@@ -13,13 +13,13 @@ public abstract class SplineInterpolator
...
-        protected MatrixSolver gauss;
+        protected Matrix? m;
+        protected MatrixSolver? gauss;
 
         protected readonly int n;
-        protected double[] a, b, c, d, h;
+        protected double[] a = null!, b = null!, c = null!, d = null!, h = null!; // they're assigned in derived classes
 
-        public SplineInterpolator(double[] xs, double[] ys, int resolution = 10)
+        protected SplineInterpolator(double[] xs, double[] ys, int resolution = 10)
```
jl0pd, line 22: Might be breaking change. It's was public constructor in abstract class. Only derived classes could call it.
Binary breaking change anyway

jl0pd:
This is draft of what need to be done in order to support nullable reference types.
This commit actually should be split into several smaller commit:
1. Annotate folder after folder, starting with signatures
1. Annotate top-level files
1. Annotate demo-applications (right now only core library is annotated)

Amount of warnings should be reduced to 0. Right now ~200 of them.

Some classes should be reworked, so they wouldn't get into invalid state, for example [Image](https://github.com/ScottPlot/ScottPlot/blob/bb58c3606c249451b6b0b95648c12753d07234fc/src/ScottPlot/Plottable/Image.cs). It should have Bitmap set, but it's not enforced by constructor

Also there's couple of fixes that should be in different PRs

src/ScottPlot/DataGen.cs
```diff
@@ -9,6 +9,8 @@
 using System.Text;
 using System.Threading.Tasks;
 
+// TODO: Need to inspect Random instantiation because some of them is seed (new Random(0)), some not (new Random())
```
jl0pd, line 12: This question is open for discussion. I'm not sure whether `new Random(0)` should be replaced. Maybe this is by design?

```csharp
public static double[] NoisyLinear(Random? rand, int pointCount = 100, double slope = 1, double offset = 0, double noise = 0.1)
```
For example if call it twice I'll get same results. This may be problem

src/ScottPlot/Plottable/FinancePlot.cs
swharden: Thank you for identifying this issue! I'll track its discussion in #705

swharden: Thanks @jl0pd! This is a breaking change, but will be released with 4.1-beta which has other breaking changes so this is an excellent time to fix it. I corrected this in c3e1062fe1f8a07a3e3519e9cdf62f40ac155cd4

src/ScottPlot/DataGen.cs
swharden: This is the intended behavior, but I recognize this is unclear. Soon I will mark the `DataGen` class obsolete and replace it with with a `Generate` class which will have better method names and documentation to make this easier to understand.

swharden:
This PR is a great example of how to nullables (and some switch expressions) in the core ScottPlot code base. Thanks @jl0pd! Unfortunately there remains ambiguity as to whether or not .NET Standard libraries (which use C# 7.3 by default) should be compiled using features of higher language versions, as discussed in #691. I'll close this PR for now, and we may open it up again depending on how that conversation progresses.

This PR is very insightful, and the steps described by https://github.com/ScottPlot/ScottPlot/pull/692#issuecomment-758347281 are excellent advice. Thanks @jl0pd for all the work you put into this! If/when the language version issue is resolved (or when we deprecate .NET Framework support) and the language version can be increased with confidence, this will be a great starting point to add nullable support throughout the library 👍

Issue #651, opened 2020-12-10
Triaged Tasks and Features (Dec 2020)


swharden:
**This page is a collection of low-priority tasks and features** I hope to accomplish eventually, but are not the primary focus of my development efforts at this time. This list originated from #412

### General
- [ ] Refactor markers ([#386](https://github.com/swharden/ScottPlot/pull/386)) to be classes instead of defined in an enum. Markers will implement `IMarker`, live in `ScottPlot.Markers`, markers have the ability to draw themselves. This makes it easy to switch between markers and even create custom markers that work for all plot types without modifying any ScottPlot code.
- [ ] Turn `Tools` static class into a `Tools` namespace with smaller static classes
- [x] Make it easier to reset the plot by creating `Plot.Reset()` (#407)
- [ ] Option for non-overlapping manual tick labels (#648, #644)
- [ ] Support for image as axis label (with demo) (#446, #696)

### Plottable-Specific
- [x] heatmaps (#420, #441, #458, #424, #482, #457) need refinement and better cookbook examples
- [x] vector fields (#438, #439) need refinement and better cookbook examples
- [ ] Add options to customize location of L-shaped scale bar (#363)
- [ ] Add minRenderIndex and maxRenderIndex to scatter plots (#566)
- [ ] Extend the legend to support horizontal orientation or columns (#580)
- [ ] OHLC bars should show values in a tooltip when the mouse is near them (#696). When this is added it seems useful to make a static class with static methods for drawing tooltips, and it can simply be called from any plottable which needs a tooltip drawn.

### Control-Specific
- [x] Add a right-click menu to WpfPlot (like FormsPlot's)
- [ ] Improve WPF Performance with BitmapSource (#522)
- [ ] Add print and print preview to user control right click menus (#524)
- [x] Use delegate and/or timer to automatically render after plottables are added (#525, #637)
- [x] Use a `Configuration` class (not a `Configure()` method) to customize control settings (#547, #637)

### Documentation
- [ ] Create a Finance Demo (not a cookbook example, but a full demo program) to consolidate common finance-related tasks in one program. This demo should describe how to add live incoming data (#395, #403, #637)
- [ ] figure out how to implement better shared axes in user controls (#399, #429)
- [ ] add "view source code" button to all demos (#434)
- [ ] create MVVM control demo (#633)

swharden:
I made a new list of triaged tasks/goals (#716) and am closing this one

Issue #446, opened 2020-06-04
Allowing images as plot/axis labels?


Benny121221:
**What can we do to make ScottPlot better?**
Allowing images as plot titles or axis labels allows a user to do their own formatting should they wish, including TeX.

```cs
plt.YLabelImage("label.png")
```

For example, a user could get a result similar to this:
![image](https://user-images.githubusercontent.com/8635304/83700686-b52cb300-a5c4-11ea-8848-2d3e4d1c3c9b.png)

Pardon my poor handwriting, it's not so great when it isn't converted to mathprint :/ If you couldn't read it, it says I forgot to add a vector hat.

Also, don't bother trying to find out what that graph is supposed to be, it's meaningless.

Benny121221:
I blame my 30 dollar drawing tablet for my poor handwriting.

And that's Canadian dollars, so that's probably 5 dollars in freedom land

swharden:
I'll close this issue while I tackle some of the more pressing topics, but this feature has been added to #412 so it won't be forgotten, and this issue can be re-opened as work toward this resumes 👍

Issue #412, opened 2020-05-20
Upcoming Changes


swharden:
# UPDATE (August, 2020)
Many of these issues are being tackled as I begin work on ScottPlot 4.1 (#505). \
I'll close this issue for now, and re-open / reassess it after 4.1 is released.

See the [ScottPlot Roadmap](https://github.com/swharden/ScottPlot/blob/master/dev/roadmap.md) for a larger-scale description of the current state and future goals of ScottPlot development.

## Task List (to be revisited after 4.1 is released)
- [ ] Make `Plottable` an interface (don't rename it though) and turn `visible` to `IsVisible()`
- [ ] Make `SignalDensity` and pull density-specific code out of `Signal` so `Signal` can stay simple and easy to extend or hack on. Will this affect `SignalConst`?
- [ ] See if we can use parallel by default, then delete the non-parallel methods from plottables (Scatter, Signal, and SignalConst) #419
- [ ] Create static methods to aid GetLimits code. It shouldn't have to be re-implemented discretely in every plottable.
- [ ] Create a central `Transform` module and have all plottables use its `GetPixel(x, y)` to convert data positions to screen pixels
- [ ] Add log support to `Transform`, magically adding log support to all plottables #207 #41 #29
- [ ] Add "show value under mouse" support #115 
- [ ] Add a text rendering method to `Drawing` and funnel all `Graphics.DrawString()` calls through it. Add alignment support, not just for text alignment but also for anchor alignment (aligned how, with reference to what?). Add arguments to Title, XLabel, and YLabel to control alignment of those labels #411
- [ ] Make grid a special type of plottable (`Renderable`?)
- [ ] Make labels (title, xlabel, ylabel) a special type of plottable (`Renderable`?)
- [ ] Make axes a special type of plottable (`Renderable`?)
- [ ] Support multiple Y axes #180, #213, #478
- [ ] Make legend a special type of plottable (`Renderable`?) so it's easier to customize #388 #318 #253 #45
- [ ] Don't use separate bitmaps for the figure, data, and legend. Instead use 1 bitmap and clipping to only draw on the desired component. This increases simplicity and performance.
- [ ] Make render system that accepts an existing Bitmap and renders onto it. Eventual goal is to not hold bitmaps in memory in ScottPlot. Ultimate goal is to remove system.drawing dependency.
- [ ] The render method should be really simple: Just iterate through a `List<Renderable>` and call the `Render()` method for each object (passing in the bitmap to render on).
[...]

PR #696, opened 2021-01-15
Tooltip and Annotation Renderables

Benny121221:
**New Contributors:**

**Purpose:**
Allows the user to add tooltips to the plotting surface. I elected to make this an `IRenderable` instead of an `IPlottable` like the conceptually similar `Annotation` class. This however could be amended easily if you think that would be better. This choice of an `IRenderable` gave a simpler class, although it did necessitate adding special methods to add or remove tooltips instead of being able to use the ones for `IPlottables`.

It does mean tooltips always draw overtop plottables, regardless of which order they are added in.

**New Functionality:**

```cs
plt.AddTooltip("This is the max of the\nsine function.", Math.PI / 2, 1);
```

### Preferred Drawing Orientation:
image

### Reverse Drawing Orientation (i.e. not enough space to the right):
image

### DIFF
--- a/src/ScottPlot/Plot/Plot.Add.cs
+++ b/src/ScottPlot/Plot/Plot.Add.cs
@@ -24,21 +24,12 @@ namespace ScottPlot
 {
     public partial class Plot
     {
-
         /// <summary>
         /// Display text in the data area at a pixel location (not a X/Y coordinates)
         /// </summary>
-        public Annotation AddAnnotation(string label, double x, double y, float fontSize = 12, Color? fontColor = null, Color? fillColor = null)
+        public Annotation AddAnnotation(string label, double x, double y)
         {
-            var plottable = new Annotation()
-            {
-                Label = label,
-                X = x,
-                Y = y,
-                FontSize = fontSize
-            };
-            plottable.Font.Color = fontColor ?? plottable.Font.Color;
-            plottable.BackgroundColor = fillColor ?? plottable.BackgroundColor;
+            var plottable = new Annotation() { Label = label, X = x, Y = y };
             Add(plottable);
             return plottable;
         }
--- a/src/ScottPlot/Plot/Plot.Add.cs
+++ b/src/ScottPlot/Plot/Plot.Add.cs
@@ -29,43 +29,19 @@ namespace ScottPlot
         /// <summary>
         /// Display text in the data area at a pixel location (not a X/Y coordinates)
         /// </summary>
-        public Annotation AddAnnotation(string label, double x, double y, float size = 12, Color? color = null, Color? backColor = null)
+        public Annotation AddAnnotation(string label, double x, double y, float fontSize = 12, Color? fontColor = null, Color? fillColor = null)
         {
             var plottable = new Annotation()
             {
                 Label = label,
                 X = x,
                 Y = y,
-                FontSize = size
+                FontSize = fontSize
             };
-            plottable.Font.Color = color ?? plottable.Font.Color;
-            plottable.BackgroundColor = backColor ?? plottable.BackgroundColor;
+            plottable.Font.Color = fontColor ?? plottable.Font.Color;
+            plottable.BackgroundColor = fillColor ?? plottable.BackgroundColor;
             Add(plottable);
             return plottable;
-        }
-
-        public Tooltip AddTooltip(string contents, double x, double y, float size = 12, Color? color = null, Color? backColor = null)
-        {
-            var tooltip = new Tooltip()
-            {
-                Contents = contents,
-                Coordinates = (x, y),
-                FontSize = size,
-            };
-
-
-            if (color.HasValue)
-            {
-                tooltip.FontColor = color.Value;
-            }
-
-            if (backColor.HasValue)
-            {
-                tooltip.BackgroundColor = backColor.Value;
-            }
-
-            settings.ToolTips.Add(tooltip);
-            return tooltip;
         }
 
         /// <summary>
@@ -665,6 +641,22 @@ namespace ScottPlot
             };
             Add(plottable);
             return plottable;
+        }
+
+        public Tooltip AddTooltip(string contents, double x, double y, float fontSize = 12, Color? fontColor = null, Color? fillColor = null)
+        {
+            var tooltip = new Tooltip()
+            {
+                Contents = contents,
+                Coordinates = (x, y),
+                FontSize = fontSize,
+            };
+
+            tooltip.FontColor = fontColor ?? tooltip.FontColor;
+            tooltip.BackgroundColor = fillColor ?? tooltip.BackgroundColor;
+
+            settings.ToolTips.Add(tooltip);
+            return tooltip;
         }
 
         /// <summary>
--- a/src/ScottPlot/Plot/Plot.Add.cs
+++ b/src/ScottPlot/Plot/Plot.Add.cs
@@ -13,7 +13,6 @@
  */
 
 using ScottPlot.Plottable;
-using ScottPlot.Renderable;
 using ScottPlot.Statistics;
 using System;
 using System.Collections.Generic;
@@ -643,21 +642,14 @@ namespace ScottPlot
             return plottable;
         }
 
-        public Tooltip AddTooltip(string label, double x, double y, float fontSize = 12, Color? fontColor = null, Color? fillColor = null)
+        /// <summary>
+        /// Display a text bubble that points to an X/Y location on the plot
+        /// </summary>
+        public Tooltip AddTooltip(string label, double x, double y)
         {
-            var tooltip = new Tooltip()
-            {
-                Label = label,
-                X = x,
-                Y = y
-            };
-
-            tooltip.Font.Size = fontSize;
-            tooltip.Font.Color = fontColor ?? tooltip.Font.Color;
-            tooltip.FillColor = fillColor ?? tooltip.FillColor;
-
-            settings.ToolTips.Add(tooltip);
-            return tooltip;
+            var plottable = new Tooltip() { Label = label, X = x, Y = y };
+            Add(plottable);
+            return plottable;
         }
 
         /// <summary>
--- a/src/ScottPlot/Plot/Plot.Add.cs
+++ b/src/ScottPlot/Plot/Plot.Add.cs
@@ -13,6 +13,7 @@
  */
 
 using ScottPlot.Plottable;
+using ScottPlot.Renderable;
 using ScottPlot.Statistics;
 using System;
 using System.Collections.Generic;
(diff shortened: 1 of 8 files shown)


Benny121221:
Long term I think it would be interesting for plottables to be able to give their own tooltips (probably following the mouse around). However I wanted to see if you would consider this in scope for ScottPlot. I also didn't want to add mouse following in case it needs to be changed for #683.

In my mind, plots would be able to do something a little like this on mouseover:
image

Now, there are two tooltips because plotlyjs is like that, but it is what it is.

swharden:
Hey @Benny121221, thanks for this! I don't think I would have thought to come up with an arrow tip shaped like that, but it looks really clean and I like it a lot!

The architecture raises a few interesting possible directions:

- Currently you have a special list of tooltips that gets rendered last, but we could further embrace the idea and have a list of "annotations" that are always displayed above the plot. This could include things like `Annotation`, `Arrow`, `Tooltip`, etc.

- Another option is to keep everything a `Plottable` in one plottable list and make the user arrange the order of things.

- Another option is to to add a flag to the `IPlottable` to indicate that it should be rendered on top. I keep putting this in than taking it out. A style I like here is to have an enum that says all the possible layers: figure background, data background, plottables, data frame, axes, annotations

- I originally intended `Renderable` for things that did not use coordinates, and `Plottable` for things that used the axis system. I guess technically this is one reason the tooltip is broken as-is, because it doesn't support multiple axes. This might justify making it a `Plottable` after all 🤔 

I'll keep exploring these ideas and work on this PR some more tomorrow! 

Thanks again
Scott

swharden:
Thinking out loud... OHLCs might benefit from something like "highlight candle nearest cursor" (easy because only the X axis would have to be considered), and the highlight could be a colored tooltip with the corner symbol that looks just like the picture you showed. Probably low priority, but it might be pretty straightforward to implement.

Benny121221:
> Currently you have a special list of tooltips that gets rendered last, but we could further embrace the idea and have a list of "annotations" that are always displayed above the plot. This could include things like Annotation, Arrow, Tooltip, etc.

I like this idea, it makes it feel less like a one-off exception and more like a fleshed-out part of the API.

> I originally intended Renderable for things that did not use coordinates, and Plottable for things that used the axis system. I guess technically this is one reason the tooltip is broken as-is, because it doesn't support multiple axes. This might justify making it a Plottable after all 🤔

That's a good point, although can "normal" plottables do this either? I thought secondary axes are just visual, a plottable drawing at point (x, y) will always draw at (x, y) on the primary axes?

> Thinking out loud... OHLCs might benefit from something like "highlight candle nearest cursor" (easy because only the X axis would have to be considered), and the highlight could be a colored tooltip with the corner symbol that looks just like the picture you showed. 

Yeah, I think OHLCs and Candlesticks are prime candidates for this because it's difficult to read precise values off of them (or maybe I'm just not paying enough attention in my ECON classes). Plus things like trading volume aren't shown by a candlestick or an OHLC plot, but is certainly useful information. Although I know that candlesticks are often used more for technical analysts to see patterns than to read off the 3rd decimal place.

And I think the axes themselves might benefit from tooltips? For example, you could have a short form date like 1/20/21 which expands to January 20, 2021 in the tooltip. Or scientific notation expanding to decimal form, log axes to conventional axes, or like I saw in a lecture today which converted what are decimals on the axis to exact value forms.

swharden:
@Benny121221 Thanks for your input!

### Multiple Axes

> I thought secondary axes are just visual, a plottable drawing at point (x, y) will always draw at (x, y) on the primary axes?

Actually, `IPlottable` has two numbers that define which axes the data lives on.

https://github.com/ScottPlot/ScottPlot/blob/540219671b1e78a55595e3d847ebcec9bb5114da/src/ScottPlot/Plottable/IPlottable.cs#L8-L9

Those numbers get read right before calling `Render()` and are used to build the `dims` object that gets passed in. `dims` is automatically adjusted so `GetPixel()` methods already know which axis they're working at. Therefore, anything that implements `IPlottable` automatically has support for multiple axes.

### Financial Charting

> my ECON classes

Haha I'm going to try not to get bated/triggered into writing a huge text block here. Suffice to say I'm passionate about economics, finance, personal finance, investing, and I'm morbidly fascinated by the intersection of meme culture and day trading and swing trading (and don't get met started on crypto!). I will say that learning how to manage investments (deciding savings rate, asset allocation, risk tolerance, tax strategies, etc.) is super intimidating at first, but amazingly empowering once you start to figure it out. At the same time virtually the entire "financial services" industry is out to get you, so it's quite a minefield for people getting started. After losing a fortune investing in a popular company of the day, Isaac Newton wrote "I can calculate the motions of the heavenly bodies, but not the madness of the people". I really enjoyed [A Random Walk Down Wall Street](https://www.amazon.com/dp/B0118LNMA4/) and its punny title. I'll just leave it there I guess.

I'm torn on whether or not to improve financial charting. It's easy to add little things here and there, but the more we say ScottPlot supports financial charting the more we will attract a whole crowd who may ask for really important financial features that aren't really at the current core of what ScottPlot is made for. Check out some YouTube videos showing people using TC2000, probably the most popular trading software for retail investors. Most of those features could probably be implemented by ScottPlot, but may be best created as its own control specifically for financial charting. I've gotten a few pings about this, and there may be some commercial opportunity here. Since such software would obviously be used to make money (or at least attempt to), perhaps the financial control could be licensed as GLPL and companies would have to pay for a custom license. I'm not really interesting pursuing this at this time, but the idea may be worth exploring as the primary library stabilizes (if that ever happens).

### Tooltips on axes

This is an interesting idea! I'd probably prioritize it pretty low compared to the other things on my plate, but it certainly seems possible.

### This PR

I'll keep working on this over the next couple days and hopefully come up with a solution that feels natural.

Thanks again for all your input along the way!

Benny121221:
> ...but may be best created as its own control specifically for financial charting...

I'm intrigued, I do think a dedicated financial charting control would be interesting, but I'm also intrigued just by the idea of a control for a specific field.

I vaguely thought about creating something which wraps ScottPlot to add a couple math-specific features. I never did because I couldn't think of enough reasons to do so (and most of them got merged into ScottPlot proper anyways). The features were:
- [x] Vector Fields
- [x] Function Plots
- [ ] TeX rendering on axes (briefly mentioned in #439)
- [ ] Slope fields (probably better done by hooking up a CAS system to vector fields anyways)

#446 brought up allowing images as axis labels which would allow (La)TeX rendering to be handled by the user. I wonder if a simple package like ScottPlot.TeX could be created to do something like this:

```cs
plt.YAxis.Image = ScottPlot.TeXRender("\int \sec\theta \: d\theta")
```

Or, if you wanted to get flashy with extension methods:

```cs
plt.YAxis.Image.FromLaTeX("\int \sec\theta \: d\theta");
```

image

This package wouldn't necessitate adding another dependency to ScottPlot, although I confess that just adding one feature does sound pretty overkill for just one package.

In #412 adding images as axis labels is checked, but I don't know any way to do this? Is this an oversight? If not I think a demo would be fitting.

Benny121221:
> I'm torn on whether or not to improve financial charting. It's easy to add little things here and there, but the more we say ScottPlot supports financial charting the more we will attract a whole crowd who may ask for really important financial features that aren't really at the current core of what ScottPlot is made for.

Yeah I get that, I generally try to contribute to the library in areas where I feel I can help, but unfortunately I'm not much into trading (and certainly not technical analysis). My interest in economics is mostly in macroeconomics and public policy (I used to work in a provincial legislature).

I think a common rule in open source when it comes to feature requests that are relevant but not the main focus of a project is that they get given a pretty low priority unless the feature request is accompanied by a PR. After all, code review and testing generally doesn't take as long as developing it from scratch. I don't know if that's the line you want to take, it's just the most common view that I've seen before.

swharden:
There's some excellent discussion here! I'll try to wrap-up this PR but we can continue to discuss here (or in Discussions?)

### This PR
I made the tooltip a plottable after all, and its height in the plottable stack is determined by its order. I'm trying to not keep long-running PRs, so I'll go ahead and merge this in but you're welcome to offer suggestions/improvements.

### Ideas for more tooltip features
* If you want to add tooltip support to OHLC charts you're welcome to (although it's low priority, so only if you're interested). Since multiple plottables will use tooltips, it may be time to create a static class for drawing tooltips that the plottables can call.
* I noticed rendering artifacts at the end of thick lines for closed paths. I hacked a workaround here by duplicating two points instead of just one, but there is likely a more elegant solution and a standalone class could explore this better
* You may consider using [Graphics.TranslateTransform](https://docs.microsoft.com/en-us/dotnet/api/system.drawing.graphics.translatetransform) to center the Graphics at the X/Y point at the tip of the arrow before drawing the tooltip because then the math gets a lot easier. Those edges and corners are confusing!
* I called [Graphics.DrawClosedCurve](https://docs.microsoft.com/en-us/dotnet/api/system.drawing.graphics.drawclosedcurve) on the `PointF[]` just to see what would happen, and it was an interesting effect. With a little tweaking, it might be a way to make a softer tooltip, or one that resembles a thought bubble.

### Financial control
> I'm intrigued, I do think a dedicated financial charting control would be interesting, but I'm also intrigued just by the idea of a control for a specific field.

I think the reason it merits its own control is because of how the X axis is handled in financial charts. It's always DateTime, but they're unevenly spaced. That's a lot for the end-user to keep track of, but a control which wraps a Plot and has methods to set data and styling would make this way easier. 

A control could wrap most or all of the ScottPlot complexity, and all the functions the user interacts with could be named correctly according to their financial terms. This becomes especially clear when you think about methods like `AddSMA(10)` or `AddBollinger(10)`.

A financial control could also automatically handle all the mouse tracking so the user never has to think about that. In summary, the reason I think a control is good here is so it could just expose the features relevant to "chartists", and not require the adopter to learn the ins and outs of ScottPlot and the little hacks required to make financial charts look good.

### Image Axes

> adding images as axis labels is checked, but I don't know any way to do this? Is this an oversight? If not I think a demo would be fitting

Yeah, this is an oversight on my part. It may have been supported at some point, but when I re-wrote the Axis module I didn't add support for images. Reconsidering this now, I wonder if `AddImage()` could be most easily modified to do this using some type of configuration to (1) not clip to the data area and (2) center the image where the axis label would be. I agree though that a demo is fitting! I added this to #651 but I'll probably work on some higher priority tasks first.

### Finance

> unfortunately I'm not much into trading (and certainly not technical analysis). My interest in economics is mostly in macroeconomics and public policy (I used to work in a provincial legislature).

Not sure if you're familiar with Reddit's wallstreetbets (WSB), but if you ever want to see some craziness look them up. There are some YouTube videos where people highlight some of the craziest stuff that goes on there. I'm a pretty conservative hands-off bogleheads.org style investor, but I bet the majority of people who would use a financial charting tool we make would be more of the WSB crowd (and perhaps overlap with people deep into trading crypto), and I'm hesitant to attract that kind of drama here. I'll continue to think about it

### Prioritizing Feature Requests with PRs
> I think a common rule in open source when it comes to feature requests that are relevant but not the main focus of a project is that they get given a pretty low priority unless the feature request is accompanied by a PR. After all, code review and testing generally doesn't take as long as developing it from scratch. I don't know if that's the line you want to take, it's just the most common view that I've seen before.

I agree this is a common strategy for mature libraries. My goal for the last few months (and one I hope to wrap up here in the next couple weeks) is to get this new ScottPlot 4.1 API stable, then have _really nice_ documentation, and make a few YouTube videos walking through how to do common things with ScottPlot. I'm going full speed ahead prioritizing whatever I think will make ScottPlot easier to document and figure out. Hopefully after that it'll settle down a little, and we can switch to the slower PR/review style of feature inclusion rather than the 4.0->4.1 hurricane that's been going on lately. That's my plan at least, we'll see how it goes!

Benny121221:
> ... relevant to "chartists" ...

Wow, that took me back to AP history, those were the days.

> Not sure if you're familiar with Reddit's wallstreetbets

Yeah, I have had the (mis)fortune of encountering it before. I think that if you're worried about attracting the WSB types that being a library is fortunate for ScottPlot. Most people aren't developers, so they'll use someone's app which may use ScottPlot and that app will attract most of the drama. However, I don't know how much second-hand drama there'd be from the app developer.

PR #700, opened 2021-01-20
Avalonia refactor onto new Backend

Benny121221:
**New Contributors:**

**Purpose:**
#637 

**New Functionality:**
N/A

**Known Issues:**
- The Linked Axis demo, Custom Right-Click Menu and Custom Layout demos do not work.
- Copy Image to Clipboard does not exist (Avalonia requires native code for binaries in the clipboard, at least as far as I know)


Benny121221:
I have yet to test it on Linux so I can't comment on that experience but so far it's pretty hard to tell any difference. I think the scrolling is less sensitive now, but that's quite possibly just my mind playing tricks on me.

Benny121221:
I added one of Eric's suggestions in #637 and it feels better on Linux. I never had much of an issue except in VMs but better is better.

Benny121221:
I don't know how you'd feel about this, but I experimented with adding filtering on input, you can see that commit here (it's not on this branch): https://github.com/Benny121221/ScottPlot/commit/235b75bc1cb1047dda4b7ed7603eafc467400651

Let me know if you want this commit brought into this PR. It mildly increases performance and the change is imperceptible (although maybe if you have a 480P screen you might notice).

swharden:
Hey @Benny121221, I'm hesitant to resolve this merge conflict because I'm slightly insure about `plotViewer.Owner = this` vs. `plotViewer.SetWindowOwner(this)`.

Although I see how mouse filtering could slightly improve performance, I think unless there's a compelling improvement it's worth keeping as simple as possible. Maybe one day if there's a strong desire for it we could enable it with a Configuration flag, but for now I'm happy to leave this one out

swharden:
Also, way to go on the quick turnaround here! Overall the code seemed to shrink considerably, and there's a lot of commented-out code in there too.

image

Do you have any big picture impressions about the design of this backend system that relies on passing events around? I'll probably eventually write controls for MAUI, WinUI, and who knows what Microsoft will come up with later, so I'm interested if anything jumped out as you as particularly good or challenging about the current design.

Benny121221:
> Hey @Benny121221, I'm hesitant to resolve this merge conflict because I'm slightly unsure about plotViewer.Owner = this vs. plotViewer.SetWindowOwner(this). Do you want to make the final call here?

Yeah, sure, the `Owner` property was made `{get; protected set;}` in Avalonia 0.10.0. So there does need to be the special method to set that, as otherwise the class does not have access.

> there's a lot of commented-out code in there too.

Yeah, I think a lot of it is related to features that I'd like to add but am unsure how to do with the current backend (i.e custom context menus) or unsure how to do with current Avalonia (i.e. copying images to clipboard).

> Do you have any big picture impressions about the design of this backend system that relies on passing events around? I'll probably eventually write controls for MAUI, WinUI, and who knows what Microsoft will come up with later, so I'm interested if anything jumped out as you as particularly good or challenging about the current design.

I think that the backend was pretty easy to work with, currently, my only confusion is how custom context menus work. I think that it should be relatively easy to add another control when the time comes.

swharden:
Sounds good!

I know some people try to avoid committing commented-out code but I don't mind it because this control is certainly a work in progress and the technology is evolving quickly. I'm also glad to hear the backend was pretty straightforward! If you come up with any ideas for how to simplify the back-end in the future, it should be pretty easy to refactor as needed now.

Thanks again for your great work on this @Benny121221!

Issue #695, opened 2021-01-13
What is the best way to add plottables to listboxes?


Sedighzade:
Each plot should have a Name property and the ToString method overrided to return that Name. This way, a plot can be added to list controls directly.
This is specially needed When we want to toggle visibility for multiple plots. Each plot can be added to a CheckedListbox/Listview directly

swharden:
Thanks for the suggestion @Sedighzade! I'll look into this after merging #683.

I'll explore this idea by adding a demo that shows how to use a `CheckedListbox` and `ListView` to control visibility of multiple `FormsPlot` controls.

Sedighzade:
I just added a Name property to the SignalPlot class. 
This is my code I'm using now:


                for (int i = 0; i < ys.GetLength(0); i++)
                {
                    double[] y = ys[i];
                    SignalPlot sp = plt.AddSignal(y);
                    sp.Name = yLable[i];

                    if (i > 1)
                    {
                        var yAxis3 = plt.AddAxis(ScottPlot.Renderable.Edge.Left, i + 2);
                        yAxis3.Label(yLable[i]);
                    }
                    else
                        plt.YAxis.Label(yLable[i]);

                    int index = checkedListBox1.Items.Add(sp);
                    checkedListBox1.SetItemChecked(index, true);
                }
                Plot1.Reset(plt);
                Plot1.Render();

Benny121221:
I like this idea. May I suggest that if no name is specified the name is still guaranteed to be distinct?

i.e. don't simply set it to null or an empty string, but a unique identifier. This could be using [`System.Guid`](https://docs.microsoft.com/en-us/dotnet/api/system.guid?view=net-5.0#methods).

GUIDs have a benefit over just counting up in case someone creates two plottables at the same time in separate threads, with a simple counter those two plottables could have the same ID. The other benefits of GUIDs are mostly related to security, and I don't think they're likely to apply to ScottPlot.

Benny121221:
Hmmm, this gets confusing with the presence of the `Label` property in many of the plottables.

Perhaps we shouldn't muddy the waters, and instead give each plottable an automatically assigned ID.

Benny121221:
I added this ID property in https://github.com/Benny121221/ScottPlot/tree/plottable-id but in the middle of writing it, I noticed that I can't seem to see any benefit to maintaining a list of IDs over maintaining a list of references to plottables.

[...]

PR #704, opened 2021-01-22
Adds Label to IPlottable interface

Benny121221:
**New Contributors:**

**Purpose:**
#695, adds a `Label` property to `IPlottable` and to all classes which implement `IPlottable`

**New Functionality:**
N/A

### DIFF
--- a/src/ScottPlot/Plottable/Annotation.cs
+++ b/src/ScottPlot/Plottable/Annotation.cs
@@ -22,7 +22,7 @@ namespace ScottPlot.Plottable
         /// <summary>
         /// Text displayed in the annotation
         /// </summary>
-        public string Label;
+        public string Label { get; set; }
 
         public readonly Drawing.Font Font = new Drawing.Font();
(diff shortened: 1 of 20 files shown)


swharden:
Thanks for looking into this @Benny121221! However, I'm going to close this PR until/unless #695 moves forward.

I don't think we should encourage application developers to store their custom business logic classes inside of ScottPlot objects. I recognize this is a common practice in the Visual Basic and Windows Forms worlds, but this doesn't strike me as a pattern we want to encourage. I welcome discussion though, and my mind is open to being changed. We will see how #695 evolves.

PS: A `string Label` field makes sense for most plottables (which is why so many have that field already), but it doesn't make sense for all plottables (e.g., grouped bar plots)

swharden:
> I don't think we should encourage application developers to store their custom business logic classes inside of ScottPlot objects

... I recognize that's not what this PR did, but this PR emerged from roughly the same conversation (#695) and I conflated the topics in my last comment. Sorry if I made that confusing

Issue #699, opened 2021-01-17
Tick mark label sizes are incorrectly calculated resulting in a poor layout


MisterRedactus:
I am having an issue with a plot cast as SignalPlotXY. Everything works well, but the Y-Axis label is on top of the Y-Axis tick labels, which are about 6-7 places in length. Is this a gap in my understanding or a bug, and, if the former, how do I address this?

Benny121221:
Hello, since it's a visual issue a screenshot would help a lot in understanding or diagnosing your issue. If possible, the relevant code would also be a plus.

If you're plotting sensitive data then a screenshot of just the affected axis would be fine, or an example with the same issue but meaningless data.

MisterRedactus:
Certainly. My original code was part of a much larger app and would have been difficult to easily summarize. However, I have been able to duplicate the issue in a much smaller app. My xaml:

`
<Window x:Class="SagacityPipeline.Designer.Windows.SequentialChartWindow"
        xmlns="http://schemas.microsoft.com/winfx/2006/xaml/presentation"
        xmlns:x="http://schemas.microsoft.com/winfx/2006/xaml"
        xmlns:d="http://schemas.microsoft.com/expression/blend/2008"
        xmlns:mc="http://schemas.openxmlformats.org/markup-compatibility/2006"
        xmlns:local="clr-namespace:SagacityPipeline.Designer.Windows"
        xmlns:ScottPlot="clr-namespace:ScottPlot;assembly=ScottPlot.WPF"
        mc:Ignorable="d"
        Title="SequentialChartWindow" Height="450" Width="800">
    <Grid>
        <Grid.RowDefinitions>
            <RowDefinition Height="1*" />
            <RowDefinition Height="10*" />
        </Grid.RowDefinitions>
        <Button Grid.Row="0" Click="ShowLabel" >Button 1</Button>
        <ScottPlot:WpfPlot Name="wpfPlot1" Grid.Row="1"/>
    </Grid>
</Window>
`

Code behind:


            // generate random, unevenly-spaced data
            Random rand = new Random(1);
            int pointCount = 500_000;
            double[] ys = new double[pointCount];
            double[] xs = new double[pointCount];

            ys[0] = 50000;
            for (int i = 1; i < ys.Length; i++)
            {
                ys[i] = ys[i - 1] + rand.NextDouble() - .5;
                xs[i] = xs[i - 1] + rand.NextDouble();
            }

[...]

Issue #701, opened 2021-01-20
Coordinated Heatmap


StendProg:
**What can we do to make ScottPlot better?**

For now `PlottableHeatmap` binds to coordinate system in simple way `x = heatmap_x_index`, `y = heatmap_y_index`.
Zooming and panning works fine.

It would be useful to set the boundaries for the map along the x and y axes.
Let's say `Xmin`, `Xmax`, `Ymin`, `Ymax`.
Coordinate step will be calculated using heatmap dimensions and provided boundaries.

For example i want to plot heatmap from x=-100 to x = 200, from y = 100 to y = 101;
Heatmap size is 300x1000.
```c#
double[,] intensities = new double[300, 1000];
plt.PlotHeatmap(intensities, xmin:-100, xmax: 200, ymin: 100, ymax:101);
```
xStep will be 300 / 300 = 1,
yStep will be 1 / 1000 = 0.001.

At first glance, this does not look like a difficult task for the Heatmap developer, but it is quite difficult for me since I need to understand the existing code...

Benny121221:
This was actually in #457 over six months ago. I ended up mapping the bitmap to fixed coordinates because I didn't know how to treat axes that don't preserve the aspect ratio.

Here are the options that I can think of:
- Simply stretch or crop the heatmap, which is not ideal.
- Coerce axes to ones that preserve aspect ratio:
    - e.g. if the user passed in a 200x100 bitmap then attempting to rescale the axes to x ∈ [0, 400] and y ∈ [0, 250] actually yields x ∈ [0, 400] and y ∈ [0, 200]
    - I think this flies in the face of the whole point of being able to change the axes. This is only actually useful if they only care about the numbers on one axis.
- "Poke through" into the axis and rewrite the numbers to be disconnected from the actual rendering
    - One might be able to turn the axis into a facade and plottables can arbitrarily remap coordinates (messy)
    - Probably better achieved by turning off the "real" axes and letting them draw their own axes with their own numbers
 
I think "poking through" is the only option which helps the user and it may be more viable now that multiple axes are supported. I have not looked very hard at the new `Renderable.Axis` class but my understanding is that the axes with `AxisIndex` of 0 will be matched to the bitmap's dimensions, and other indices will not be (unless the plottable's `XAxisIndex` or `YAxisIndex` specifies otherwise). `AxisIndex` of 1 is also taken by the top and right axes.

## My Attempt:

This was the best I could do in a demo: 
[...]

PR #706, opened 2021-01-23
Invisible axes should have no size

Benny121221:
**New Contributors:**

**Purpose:**
Makes two adjustments to the axis system:
- Invisible axes take up no space, allowing someone to overwrite the existing axes (related to #701)
- Large numbers on the Y axis will get more space to prevent overlapping with axis labels (previously this only happened on the X axis) #699

**New Functionality:**
#### Collapsing Axes:
**Before:**
image

**After:**
image

**Code:**

```cs
public class HeatmapCustomAxes : IRecipe
{
    public string Category => "Plottable: Heatmap";
    public string ID => "heatmap_customaxes";
    public string Title => "Heatmap with Custom Axes";
    public string Description =>
        "The axes of a heatmap can be easily customized.";

    {
        plt.AddHeatmap(imageData);

        plt.XAxis.IsVisible = false;
        plt.YAxis.IsVisible = false;

        var newYAxis = plt.AddAxis(Renderable.Edge.Left, 2);
        var newXAxis = plt.AddAxis(Renderable.Edge.Bottom, 2);

        newXAxis.Dims.SetAxis(-70, 70);
        newYAxis.Dims.SetAxis(0, 500);
    }
}
```

### Accommodating Large Numbers

**Before:**
image

**After:**
image

**Code:**

```cs
Random rand = new Random(1);
int pointCount = 500_000;
double[] ys = new double[pointCount];
double[] xs = new double[pointCount];

ys[0] = 50000;
for (int i = 1; i < ys.Length; i++)
{
  ys[i] = ys[i - 1] + rand.NextDouble() - .5;
  xs[i] = xs[i - 1] + rand.NextDouble();
}

plt.YAxis.Label("Test Label");
plt.AddSignalXY(xs, ys);
```

I normally would split these PRs into two, but I think because they're so conceptually similar and they're so small that it makes more sense to review them as one. However I can make two separate PRs if you would prefer.

### DIFF
--- a/src/ScottPlot/Renderable/Axis.cs
+++ b/src/ScottPlot/Renderable/Axis.cs
@@ -374,26 +374,21 @@ namespace ScottPlot.Renderable
 
                 if (AxisTicks.TickLabelVisible)
                 {
+                    // determine how many pixels the largest tick label occupies
                     float maxHeight = AxisTicks.TickCollection.maxLabelHeight;
                     float maxWidth = AxisTicks.TickCollection.maxLabelWidth * 1.2f;
-                    float sizeNeeded = IsHorizontal ? maxHeight : maxWidth;
-                    float diff = Math.Max(maxWidth, maxHeight) - Math.Min(maxWidth, maxHeight);
 
-                    switch (Edge)
-                    {
-                        case Edge.Top:
-                        case Edge.Bottom:
-                            sizeNeeded = Math.Min(maxWidth, maxHeight) + diff * (float)Math.Sin(AxisTicks.TickLabelRotation * Math.PI / 180);
-                            break;
+                    // calculate the width and height of the rotated label
+                    float largerEdgeLength = Math.Max(maxWidth, maxHeight);
+                    float shorterEdgeLength = Math.Min(maxWidth, maxHeight);
+                    float differenceInEdgeLengths = largerEdgeLength - shorterEdgeLength;
+                    float sinRotation = IsHorizontal ?
+                        (float)Math.Sin(AxisTicks.TickLabelRotation * Math.PI / 180) :
+                        (float)Math.Cos(AxisTicks.TickLabelRotation * Math.PI / 180);
+                    float rotatedLabelSize = shorterEdgeLength + differenceInEdgeLengths * sinRotation;
 
-                        case Edge.Left:
-                        case Edge.Right:
-                            sizeNeeded = Math.Min(maxWidth, maxHeight) + diff * (float)Math.Cos(AxisTicks.TickLabelRotation * Math.PI / 180);
-                            break;
-
-
-                    }
-                    PixelSize += sizeNeeded;
+                    // add the rotated label size to the size of this axis
+                    PixelSize += rotatedLabelSize;
                 }
 
                 if (AxisTicks.MajorTickVisible)
--- a/src/ScottPlot/Renderable/Axis.cs
+++ b/src/ScottPlot/Renderable/Axis.cs
@@ -374,21 +374,26 @@ namespace ScottPlot.Renderable
 
                 if (AxisTicks.TickLabelVisible)
                 {
-                    // determine how many pixels the largest tick label occupies
                     float maxHeight = AxisTicks.TickCollection.maxLabelHeight;
                     float maxWidth = AxisTicks.TickCollection.maxLabelWidth * 1.2f;
+                    float sizeNeeded = IsHorizontal ? maxHeight : maxWidth;
+                    float diff = Math.Max(maxWidth, maxHeight) - Math.Min(maxWidth, maxHeight);
 
-                    // calculate the width and height of the rotated label
-                    float largerEdgeLength = Math.Max(maxWidth, maxHeight);
-                    float shorterEdgeLength = Math.Min(maxWidth, maxHeight);
-                    float differenceInEdgeLengths = largerEdgeLength - shorterEdgeLength;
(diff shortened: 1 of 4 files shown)


Benny121221:
I may look at adding images as axis labels in #651, and there are other axis concerns in https://github.com/ScottPlot/ScottPlot/issues/701#issuecomment-764199068:

> As an aside, in addition to the stacking on invisible axes, it turns out that when you scroll the fake axis is also zoomed in/out, and this is not reset by a middle click. My potential solutions are either add an auto-axis behaviour on the Axis renderable, which sounds like mixing concerns, or to create a special heatmap axis class (probably a subclass of Axis), whose instances need to be treated specially by the backend, which is better but still inelegant.
> 
> Also the demo maps the whole window to the given coordinates, not the area taken up by the heatmap. This is solvable, but not the priority of the demo.

Looking forward to your feedback.

swharden:
### Rotated Ticks

Hey @Benny121221, thanks for your work on this! I feel like you did all the hard work by figuring out where this error was and how to fix it, and I extended it slightly farther by adding support and a cookbook example for rotated ticks on the vertical left axis. I replaced your switch statement with this. What do you think?
https://github.com/ScottPlot/ScottPlot/blob/3c19169af0de86af23306aeab1ecfe00bf5a1c63/src/ScottPlot/Renderable/Axis.cs#L377-L390

### Multi-Axis Auto-Axis

> I normally would split these PRs into two, but I think because they're so conceptually similar and they're so small that it makes more sense to review them as one. However I can make two separate PRs if you would prefer.

I think a separate PR is the way to go. I'd love to merge this PR in, close #699, and release a NuGet package in the next day or two. Work related to multi-axis and auto-axis behavior has a tendency to explode into complexity, so if that happens it would be nice to have that work contained in its own PR in case it gets complicated.

Benny121221:
> I think a separate PR is the way to go

Cool, I used cherry-pick to create a new PR with the same commits except for the axis collapsing stuff. You can see that PR in #708

swharden:
Hey @Benny121221, thanks for cherry-picking #708! I'm happy that's merged in now.

> https://github.com/ScottPlot/ScottPlot/pull/706#issuecomment-765851826: it turns out that when you scroll the fake axis is also zoomed in/out, and this is not reset by a middle click. My potential solutions are either add an auto-axis behaviour on the Axis renderable, which sounds like mixing concerns, or to create a special heatmap axis class (probably a subclass of Axis), whose instances need to be treated specially by the backend, which is better but still inelegant.

I think I got confused by this text. It sounded like you were describing a bug in where plots with multiple axes aren't properly panned/zoomed or reset by middle-clicking. This is why I thought this PR might explode in complexity (because multi-axis auto-axis is a bit thorny).

However, when I flip through the multi-axis cookbook demos things seem all right to me. Is the text above referring to undesired behavior just in the heatmap module, or something that causes undesired behavior for all plot types? If the latter, I might need clarification on what the issue is and how I can recreate it...

Benny121221:
The issue is any axis that isn't the primary one will not be reset by a middle-click. However zooming and panning will be reflected properly on all axes.

swharden:
On my system middle-click auto-axis works for multi-axis demos. I confirmed this on the WinForms, WPF, and Avalonia controls.

I'm still a bit confused on what the issue is, and how it differs from what I observe here 🤔 

middle-click

Benny121221:
@swharden I see where I went wrong, my second axis wasn't associated with any plottables, so it was never reset.

It looks like the best way to give a heatmap arbitrary axes is to allow a plottable to hold 2 sets of axes, with independent tick labels. I don't know how helpful this is for other plottables, I suppose it could be used to have axes in two different units, i.e. one axis has distance in metres and the other in feet.

Not sure if this is an option that should be put on all `IPlottable`s or just the heatmap.

Now, there *is* another option, make heatmaps unpannable and unzoomable. I do wonder if panning/zooming really makes sense for heatmaps or pie or radar charts. They're not plottables that need to be zoomed or panned in and of themselves, and they are unlikely to share a plotting window with other plottables (not least because the user doesn't actually know the coordinates where they are drawn).

I don't know what other libraries do with these sorts of plottables, and I think the flexibility of being able to zoom does mean that if something is covered by the legend the user can just zoom out instead of needing to fix it in their code.

swharden:
Awesome - it sounds like our misunderstanding is resolved, and thankfully the multi-axis / auto-axis system isn't broken! I'll go ahead and merge this PR, and I'm relieved it didn't "explode in complexity" after all.

Similar to my sentiment in https://github.com/ScottPlot/ScottPlot/discussions/666#discussioncomment-306408, I'll probably withhold engaging in heatmap discussion today, focus on getting a NuGet package published today, then hopefully we can aggregate all the various heatmap discussions into one somehow and focus on that exclusively for a bit.

@Benny121221 I'm not sure how interested you are in exploring GitHub's features, but now that ScottPlot is an organization we have access to projects/tasks to organize progress on the heatmap front https://github.com/orgs/ScottPlot/projects ... I'm not sure if that would be helpful for organizing those efforts, or hurtful because it would further dilute the conversation with yet another location for it to occur in, but I thought I'd mention it in case you had an opinion about it or were interested in checking it out.

Benny121221:
@swharden I created a public project https://github.com/orgs/ScottPlot/projects/1 for heatmap features.

I've used Azure's version which has issue tracking and kanban boards, it looks like githubs is more barebones (which is not necessarily a bad thing, Azure's version is very bloated). I think it's a decent enough way of planning things.

PR #707, opened 2021-01-23
Add CoordinatedHeatmap

StendProg:
**Purpose:**
Allows to use an existing coordinate system to get additional coordinate information about `HeatMap`. #701

Additionaly Fix `ColorBar`, with this PR it will not affect AxisLimits calculation.

**New Functionality:**
User can provide coordinated boundaries for provided rectangular distribution.
Missing bounds are calculated based on the size of the `intensities` array.

```cs
double[,] intensities = ....; 

var hmc = plt.AddHeatmapCoordinated(intensities, -100, 500, 200, 201);

var hmc1 = plt.AddHeatmapCoordinated(intensities, xMin: -100, yMax: 50);
```

image

### DIFF
--- a/src/ScottPlot/Plot/Plot.Add.cs
+++ b/src/ScottPlot/Plot/Plot.Add.cs
@@ -16,7 +16,6 @@ using ScottPlot.Plottable;
 using ScottPlot.Statistics;
 using System;
 using System.Collections.Generic;
-using System.Diagnostics;
 using System.Drawing;
 using System.Linq;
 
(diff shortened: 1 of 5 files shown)


swharden:
Thanks @StendProg! This behavior is a lot more consistent with heatmaps I've interacted with in the past. This is a very useful addition.

https://github.com/ScottPlot/ScottPlot/blob/141c6c317fa403c3d1b697344e6d7bd97c0769d8/src/ScottPlot/Plot/Plot.Add.cs#L276-L330

I am considering removing the nullable min/max arguments, and forcing the user to define all 4 values when adding a coordinated heatmap. This could remove a lot of that null-checking code, but the main reason I think this may be a good idea is because it is not very obvious what passing null actually does without reading the source code.

Perhaps an intermediate option could be to force the user to define the upper limits, while the lower limits default to zero.

```cs
AddHeatMapCoordinated(double[,] intensities, double xMax, double yMax, double xMin = 0, double yMin = 0) 
```

I probably won't make the change until later when I finish working on other parts of the API, but if you feel strongly about what you think the best type of arguments I'd appreciate your input! Thanks again for this PR 👍

PR #713, opened 2021-01-25
Heatmap edges fix

StendProg:
**Purpose:**
Current `Heatmap` implementation render wrong on the edges, This is especially noticeable at small `Heatmaps`.
This PR setup additional params for Graphics.DrawImage() to make correct edges render.
Also more accurate pixel calculation were done, as result `Heatmap` perfect align with axis grid.

Negative side: I remove minScale calculation, as result `Heatmap` cells may be not square if `EqualAxis` will be disabled.
All Demo already have equal axis scales, and additional math not needed.

**New Functionality:**
Old version (incorrect)  3x2 HeatMap from cookbook:
image

The same demo with this PR:
image

### DIFF
--- a/src/ScottPlot/Plottable/Heatmap.cs
+++ b/src/ScottPlot/Plottable/Heatmap.cs
@@ -194,20 +194,20 @@ namespace ScottPlot.Plottable
 
                 if (BackgroundImage != null && !DisplayImageAbove)
                     gfx.DrawImage(
-                            BackgroundImage,
-                            destRect,
-                            0,
-                            0,
-                            BackgroundImage.Width,
-                            BackgroundImage.Height,
-                            GraphicsUnit.Pixel,
-                            attr);
+                            image: BackgroundImage,
+                            destRect: destRect,
+                            srcX: 0,
+                            srcY: 0,
+                            srcWidth: BackgroundImage.Width,
+                            srcHeight: BackgroundImage.Height,
+                            srcUnit: GraphicsUnit.Pixel,
+                            imageAttr: attr);
 
                 gfx.DrawImage(
-                        BmpHeatmap,
-                        destRect,
-                        0,
-                        0,
+                        image: BmpHeatmap,
+                        destRect: destRect,
+                        srcX: 0,
+                        srcY: 0,
                         BmpHeatmap.Width,
                         BmpHeatmap.Height,
                         GraphicsUnit.Pixel,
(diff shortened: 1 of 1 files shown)


Benny121221:
Thank you for getting this working, although I'm curious about https://github.com/ScottPlot/ScottPlot/blob/b560eee6abb85efe626a808ef11e9037f41cb329/src/ScottPlot/Plottable/Heatmap.cs#L193

It looks like this is for tiling images? But we're not trying to tile images at all.

Benny121221:
If we only need that one line then I think one should only change what you need to. This would avoid changing the rendering of heatmaps to be stretched without being specified by the user.

For example, `RenderHeatmap` can become this:

```cs
private void RenderHeatmap(PlotDimensions dims, Bitmap bmp, bool lowQuality)
{
	using (Graphics gfx = GDI.Graphics(bmp, dims, lowQuality))
	{
		gfx.InterpolationMode = System.Drawing.Drawing2D.InterpolationMode.NearestNeighbor;
		gfx.PixelOffsetMode = System.Drawing.Drawing2D.PixelOffsetMode.Half;

		double minScale = Math.Min(dims.PxPerUnitX, dims.PxPerUnitY);

		if (BackgroundImage != null && !DisplayImageAbove)
			gfx.DrawImage(
				BackgroundImage,
				dims.GetPixelX(0),
				dims.GetPixelY(0) - (float)(Height * minScale),
				(float)(Width * minScale), (float)(Height * minScale));

		gfx.DrawImage(
			BmpHeatmap,
			dims.GetPixelX(0),
			dims.GetPixelY(0) - (float)(Height * minScale),
			(float)(Width * minScale),
			(float)(Height * minScale));

		if (BackgroundImage != null && DisplayImageAbove)
			gfx.DrawImage(BackgroundImage,
				dims.GetPixelX(0),
				dims.GetPixelY(0) - (float)(Height * minScale),
				(float)(Width * minScale),
				(float)(Height * minScale));
	}
}
```

This change was made in https://github.com/Benny121221/ScottPlot/tree/heatmap-edges, so far it seems ok.

StendProg:
Hi @Benny121221,
You are right, PixelOffset.Half make most of the things. But if You plan plan to use different interpolation mode you need to set wrap mode too. I plan use bicubic interpolation in coordinated heat map. Maybe it not needed for base heatmap.

More accurate image coordinates also make some effect. With old ones it visible that heatmap draw 1-2 pixel less then expected on right and bottom edges.

Issue #709, opened 2021-01-24
Highlight point near cursor with pixel space awareness


oszymczak:
Hi,

I have an issue that visually it appears that the mouse is closer to a point but is highlighting another point because of the way the calculation is made and doesn't take into account the axis scale. Looking into the method HighlightPointNearest the formula is only looking at the actual coordinates of the points (x1 - x) * (x1 - x) + (y1 - y) * (y1 - y) and not looking how the points are drawn on the screen.

Note how visually the mouse appears closure to (2, 200) point
![image](https://user-images.githubusercontent.com/16024904/105620411-f245fe00-5e50-11eb-84af-5f99ac1a13ce.png)

As I move the mouse vertically the highlighted point eventually moves from (1, 100) to (2, 200)
![image](https://user-images.githubusercontent.com/16024904/105620429-2a4d4100-5e51-11eb-92fe-642752b09150.png)

Is there way to factor in the pixel location or the scale of the X and Y axis when different so that the point is highlighting correctly?

### Sample Project
* [TestHighlight.zip](https://github.com/ScottPlot/ScottPlot/files/5861534/TestHighlight.zip)

### MainWindow.xaml.cs
```cs
/// <summary>
/// Interaction logic for MainWindow.xaml
/// </summary>
public partial class MainWindow : Window
{
    ScottPlot.PlottableScatterHighlight sphTest;
    PlottableVLine vLine;
    PlottableHLine hLine;

    public MainWindow()
    {
        InitializeComponent();

        double[] xs = new double[] { 1, 2 };
        double[] ys = new double[] { 100, 200 };

        sphTest = pltTest.plt.PlotScatterHighlight(xs, ys, lineWidth: 0, color: System.Drawing.Color.Blue, markerSize: 10);

        vLine = pltTest.plt.PlotVLine(1, color: System.Drawing.Color.Red, lineStyle: LineStyle.Dash);
        hLine = pltTest.plt.PlotHLine(100, color: System.Drawing.Color.Red, lineStyle: LineStyle.Dash);

        pltTest.Render();
    }

    private void pltTest_MouseMove(object sender, MouseEventArgs e)
    {

        (double mouseX, double mouseY) = pltTest.GetMouseCoordinates();
        sphTest.HighlightClear();

        var (x, y, index) = sphTest.HighlightPointNearest(mouseX, mouseY);


        int pixelX = (int)e.MouseDevice.GetPosition(pltTest).X;
[...]

PR #719, opened 2021-01-27
Moved to HashSet in ScatterHighlight

bclehmann:
**New Contributors:**

**Purpose:**
Uses `HashSet<int>` in ScatterHighlight instead of `bool[]` which saves some memory. https://github.com/ScottPlot/ScottPlot/issues/709#issuecomment-767032539

**New Functionality:**
N/A


swharden:
Hey @bclehmann,

This PR suffers from some unfortunate timing - I just deprecated this plot type 😅

Point distance logic now lives in the scatter plot (as defined by `IHasPoints`) and fancy things like managing lists/maps/arrays of multiple highlighted points is no longer handled by the plottable (that's the a user to implement according to their needs).

I updated the WinForms, WPF, and Avalonia "highlight point under cursor" demos to use the latest strategy. I think the new system makes more sense, but I certainly welcome feedback!

PR #727, opened 2021-01-31
Add API documentation to Cookbook

### LINKED ISSUE #726: Add API Documentation to Cookbook
It would be nice if XML documentation for the public API were generated automatically and displayed on the ScottPlot website as part of the version-specific cookbooks.

@bclehmann previously suggested using [DocFX](https://dotnet.github.io/docfx/) (a static site generator) to generate these pages (#617). This is pretty close to ideal, but I'd like to more tightly control which classes/methods are documented, and have more control over how to break some modules apart into multiple pages (e.g., the "AddSomething" methods in the Plot module should have their own page). I don't have it all figured out yet, but this is what I'm thinking I'll do:

## Exploratory PR #727

- [x] Generate XML documentation on every build

- [ ] Tests should confirm all public methods are documented

- [x] Tests will generate API markdown tracked by git

- [x] Tests will generate API cookbook HTML pages

swharden:
This PR seeks to automatically generate API documentation and add it to the Cookbook. It also allows tests to test for documentation. This will be achieved by a combination of reflection and XML parsing.

Closes #726


bclehmann:
I wonder if adding an annotation such as `[PublicAPI]` would be a good way to specify which classes get included in the XML documentation.

You could do something like `[PublicAPI(Category="Plottables")]` to split the classes up into categories if need be?

swharden:
@bclehmann interesting idea! I thought about this for a bit, and will probably lean away from it. I do see their potential value in that they could add a lot of functionality, and most interestingly be a nice way to define which public methods are described on which documentation pages.

However, I'm generally am not a big fan of custom attributes for public code bases because people can't quickly hop in the code and know how to make new properly-annotated methods. This was the primary reason I really didn't like the use of attributes to define "data validation" behavior in plottable code (#533), and more recently am apprehensive about requiring the whole code base to have special attributes to provide back-compatible nullable annotations (#691).

Custom attributes can add a lot of potential functionality, but they increase the learning curve for people new to the code base, and I'm trying to make this library as simple to jump into as possible. There are a few pockets of complexity still (e.g., the tick system) but I hope to slowly massage those out with time.

swharden:
I'm amazed how difficult this turned out to be. I think it's starting to look nice though. It makes it easier to identify parts of the API that could use improvement. This is what the markdown looks like:
https://github.com/ScottPlot/ScottPlot/blob/d35e183048132d35ce5687781f8ae182adb7bac2/dev/API-Plot.md

PR #738, opened 2021-02-02
Improve cookbook documentation

swharden:
#727 was a strong start at automated documentation using XML docs as a source.

This PR relies more heavily on XML documentation and less on reflection.

To achieve this, **tests were added to fail if XML documentation is lacking** in public Plot or Plottable modules

This PR improves support for auto-properties as well (which Reflection sees as methods).


swharden:
This PR brought the cookbook forward a long way by adding API information from XML documentation. I can see that the ability to link to a method's URL will be very useful in writing documentation or responding to questions.

However, the cookbook generation code and also the XML/Reflection code were thrown together as an experiment. Although they work somewhat, they have large shortcomings and will be hard to maintain as they are now. Since the cookbook/API documentation style has now been mostly figured out, I think these modules should be re-written with this goal in mind. Templating should probably be separated from page generation too.

Issue #741, opened 2021-02-03
Alt left click zoom


JS-BGResearch:
Is there a way to disable the Alt left click and drag to zoom into an area of a WPF Signal plot?

bclehmann:
### 4.0.x

Not without modifying the control yourself, the relevant code is in https://github.com/ScottPlot/ScottPlot/tree/4.0-stable/src/ScottPlot.WPF

### 4.1.x (Beta)

```cs
wpfPlot.Configuration.MiddleClickDragZoom = false;
```

**Please note, this option is broken!** If this is set it will draw the rectangle as normal, and once it's released it will not zoom and the rectangle will not disappear.

I'll make a PR to address this. I'm glad I tested this before just telling you it would work.

JS-BGResearch:
Thanks for the quick response. 

I've found a simple work around for 4.0.x, if I add an event handler for the plot control's PreviewMouseDown event and set e.Handled to true if the alt keys are pressed, then the feature is disabled.

        private void Plot_PreviewMouseDown(object sender, System.Windows.Input.MouseButtonEventArgs e)
        {
            if ((Keyboard.IsKeyDown(Key.LeftAlt)) || (Keyboard.IsKeyDown(Key.RightAlt)))
            {
                e.Handled = true;
            }
        }

bclehmann:
Interesting, reminds me of the old days of Javascript development where you'd need to sprinkle some `e.preventDefault()` around.

I know @swharden is considering backporting small features from 4.1.x to 4.0.x so it's possible that you get a more official solution before 4.1.x leaves beta.

swharden:
Hey @JS-BGResearch, thanks for the question and sharing your findings! @bclehmann thanks for your input too, and for #742

Since a workaround is found and 4.1 is pretty close to ready, I'll probably leave 4.0 like it is. Setting `e.Handled` to `true` seemed to do the trick, so I'll convert this issue to a discussion and mark that the answer 👍

PR #742, opened 2021-02-03
Fix Configuration.MiddleClickDragZoom and rendering artifact if Alt is lifted prematurely

bclehmann:
**New Contributors:**

**Purpose:**
Referenced in https://github.com/ScottPlot/ScottPlot/issues/741#issuecomment-772740349

Currently when `Configuration.MiddleClickDragZoom` is disabled the rectangle will still be drawn, and it will not disappear upon release of the mouse. With this fix the rectangle will not be drawn at all.

**New Functionality:**
N/A

### DIFF
--- a/src/ScottPlot/Control/Backend.cs
+++ b/src/ScottPlot/Control/Backend.cs
@@ -219,7 +219,7 @@ namespace ScottPlot.Control
         private bool IsZoomingRectangle;
         public void MouseMove(InputState input)
         {
-            IsZoomingRectangle = IsMiddleDown || (IsLeftDown && input.AltDown);
+            IsZoomingRectangle = (IsMiddleDown || (IsLeftDown && input.AltDown)) && Configuration.MiddleClickDragZoom;
             MouseLocationX = input.X;
             MouseLocationY = input.Y;
             if (PlottableBeingDragged != null)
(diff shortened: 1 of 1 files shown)


bclehmann:
In testing this I noticed another issue, when using Alt + Left Click the box will stop expanding if alt is released, and then the control will be panned as if alt was never pressed.

This may be desired behaviour, except that the box will not disappear.

bclehmann:
I changed it so if the user releases Alt before they release mouse1 they zoom to that box. Unfortunately if the mouse cursor is still at the point they release Alt it will only do this once they move the mouse again (because it's in the MouseMove handler). Makes me wonder if ScottPlot should support KeyUp/KeyDown events? Although I should admit that I don't know how necessary it is, even in this situation it's not that noticeable unless you're looking for it. Still, if there were a KeyUp event the fix would be much simpler.

swharden:
Thanks for this PR @bclehmann, and I'm happy you discovered the secondary bug! 

In the case of prematurely lifting Alt, I called `Settings.ZoomRectangle.Clear();` so the axes aren't changed and it acts like the zoom was never started.

Probably the keypress capture is unnecessary at this time, but something interesting to consider! It would certainly be possible to listen to keypresses, but I suspect this is a pretty niche use case so I'm happy to leave this one like it is 👍

Issue #716, opened 2021-01-26
Triaged Tasks and Features (Feb 2021)


swharden:
**This page is a collection of tasks and features I want to focus on.** I hope to accomplish everything eventually, but I will arrange this list according to what I think the primary focus of my development efforts should be. This list carries over from #651 and #412.

## Immediate Priority

**⚠️ My highest priority at this time is releasing ScottPlot 4.1** and I will try to minimize effort devoted to new features while I work to complete these tasks:

- [x] Fix warnings due to tests and demos referencing the old API (#718)
- [x] Create a repo for the website #714 [ScottPlot/Website](https://github.com/ScottPlot/Website)
- [x] Create NuGet organization https://www.nuget.org/profiles/ScottPlot
- [x] Implement pixel-aware highlightable plot with demo (#709, #722)
- [x] Add auto-generated API documentation to cookbook (#726, #727)
- [x] Require every Plot method and parameter to have XML documentation (#738)
- [x] Require every Plottable to have an XML summary (#738) 
- [x] Add Plot and Plottable API information to cookbook (#738) 
- [x] Refactor the cookbook to add API documentation support. (#727, #738, #756)
- [x] Improve website documentation: Quickstarts
- [x] Improve website documentation: FAQ (https://github.com/ScottPlot/Website/issues/6)
- [x] Implement render queue #736, #761, #813
- [x] Refactor bar plots #842
- [x] Improve DateTime tick density calculations (#725, #925, #828)
- [ ] Create a modern demo/cookbook for ScottPlot 4.1 👈⚠️ ***WORKING ON THIS NOW***
- [ ] Improve website documentation to better document user controls (#887)
- [x] Add "open in new window" to right-click menu (#958)
- [ ] ✨🚀  Remove the beta/pre-release designation and release ScottPlot 4.1 on NuGet (#715)
- [ ] Release a 4.0 package with a pop-up readme informing people about 4.1 (#949)
- [ ] Review GitHub Issue templates to ensure they're up to date (#966)

## High Priority
- [ ] Improve async rendering support and documentation/demonstration of the render queue (#736, #761, #813)
- [ ] Add methods to make it easier to add live data to scatter, signal, bar, and finance plots. This may or may not involve a data object to manage array resizing and data validation. (#815, #841, #827)
- [ ] Improve API consistency for deep styling options (#981)

## Moderate Priority
- [ ] Use nullable annotations #691
[...]

PR #759, opened 2021-02-08
Images as Axis Labels

bclehmann:
**New Contributors:**

**Purpose:**
#446 #716
Allows bitmaps to be used as axis labels.

**New Functionality:**
image

Code:
```cs
plt.XAxis.ImageLabel(new Bitmap("Images/theta.bmp"));
plt.XAxis2.ImageLabel(new Bitmap("Images/theta.bmp"));
plt.YAxis.ImageLabel(new Bitmap("Images/d_theta_dt.bmp"));
plt.YAxis2.ImageLabel(new Bitmap("Images/d_theta_dt.bmp"));
```

I had wanted to use a LaTeX package for the demo but as it turns out there isn't a Nuget package which does LaTeX rendering that can easily interop with `System.Drawing`, so the images are shipped with the demo, which isn't ideal.

One may also want to remove the labels on `XAxis2` and `YAxis2`, as they were only added for debugging purposes.


bclehmann:
I'm also wondering if this feature would make sense for the title as well, such as in this picture from matplotlib:

image

bclehmann:
I believe the reason it fails tests on Linux and Mac is because they have a different bitmap format than Windows (the header is different) which means that this test will only pass on Windows. I had hoped using the resources instead of the filesystem would help but it turns out the same problem persists.

So I guess I can either find a way to specify a cross-platform bitmap to pass into `System.Drawing`, for example by stripping out the whole header and just specifying the pixels in a `byte[]`, similar to how heatmaps are done. Or one can draw the bitmap with `System.Drawing` at runtime. Fortunately I chose a pretty simple demo so the second option is not that much work.

I think either way the code for generating the image shouldn't be shown as part of the demo, as a) it'd be confusing and b) it's not relevant.

StendProg:
Hi @bclehmann,
You shoud try different bitmaps formats.
Say 'png', or bmp with different color maps.
I think problem is on concrete `.bmp` format you use, just try to change something.

bclehmann:
> You shoud try different bitmaps formats.
Say png, or bmp with different color maps.
I think problem is on concrete .bmp format you use, just try to change something.

Yeah, that's a lot easier. For some reason I got it in my head that the `Bitmap` class wouldn't support importing from a different file format.

swharden:
I added a jpeg example, and also demonstrated how transparency in PNGs is respected. This looks great! Thanks @bclehmann 

Regarding the matplotlib-style example, I think the example used in this PR is better because it's simpler. I don't want this recipe be to flashy, because at a glance it looks like this library supports LaTeX but fundamentally it doesn't, so I think a subtle recipe is the way to go here 👍

swharden:
@bclehmann I refactored the AxisLabel model but left all the logic intact. I'm a bit confused about the two `3`s and one `4` in the image offset calculations though... Is the intent to pad images by a fraction of how large they are?

https://github.com/ScottPlot/ScottPlot/blob/d1e2d3cc072414348bc08283d352c2269428b078/src/ScottPlot/Renderable/AxisLabel.cs#L88-L113

I added an `ImagePadding` field. Depending on your answer to the question above, you or I can modify that method to use it instead. I think it makes sense to let the user define the padding in pixels (with a default of 5 or something)

https://github.com/ScottPlot/ScottPlot/blob/d1e2d3cc072414348bc08283d352c2269428b078/src/ScottPlot/Renderable/AxisLabel.cs#L34-L37

Also, do you think images should be rotated automatically for vertical axes like text labels are? I guess a bool could let the user control whether this happens or not. It's a pretty niche feature, but while we're in here maybe it's a good time to add it in.

Interested in your thoughts!

bclehmann:
I didn't bring up that matplotlib example because I wanted to steer the demo that way, I brought it up because I noticed that while this PR allows images as axis labels it doesn't allow them for the plot title.

bclehmann:
> Is the intent to pad images by a fraction of how large they are?

The numbers seem arbitrary but this is for positioning, if you replaced the fraction with 1/2 it would clip the plotting window. Image padding in pixels may make sense for aesthetic reasons (although I might suggest having separate padding for x and y).

> Also, do you think images should be rotated automatically for vertical axes like text labels are? I guess a bool could let the user control whether this happens or not. It's a pretty niche feature, but while we're in here maybe it's a good time to add it in.

I believe it's confusing if images are rotated by default, where it's not confusing for text. Text doesn't prescribe how it should be layouted other than the order of the characters. Images have a top and a bottom and changing those to left and right is disorienting. For example, it would be confusing to render the d_theta/dt sideways. Now it may be a decent idea as an optional setting, although I can't say that I can think of a situation where the user would want their image rotated, and I think if they wanted to draw it sideways they'd rotate it themselves.

swharden:
Sounds good! I implemented options for custom padding (in pixels) and give the user the ability to customize it separately for each side of the bitmap.

https://github.com/ScottPlot/ScottPlot/blob/97677145fb9f03abe591fccfc40f054b6da37631/src/ScottPlot/Renderable/Axis.cs#L142-L148

Good points about rotation - It's so easy to use `System.Drawing` to rotate a bitmap anyway, if a user wants a rotated image axis they have an easy way to do it themselves. 

This PR is looking pretty good! Thanks for all your worn on this @bclehmann! 👍

Issue #767, opened 2021-02-10
GetColormaps() throws on second call


StendProg:
`ScottPlot.Drawing.Colormap.GetColormaps()` throws on second call.
There is related with `Reflection` exception.

```cs
var colormaps = ScottPlot.Drawing.Colormap.GetColormaps();
var colormaps1 = ScottPlot.Drawing.Colormap.GetColormaps();
```

swharden:
Hi @StendProg, I added some tests for this but am unable to reproduce the exception. What version of ScottPlot does this occur on, and are you running it from source or from the NuGet package?

I have learned a lot more about C# since I first wrote this method, so I refactored this method to be simpler and use Reflection. Perhaps something in this change will improve the issue?

```cs
public static Colormap[] GetColormaps() =>
    Assembly
        .GetExecutingAssembly()
        .GetTypes()
        .Where(type => type.Namespace == "ScottPlot.Drawing.Colormaps")
        .Select(colormapType => (IColormap)Activator.CreateInstance(colormapType))
        .Select(colormapInstance => new Colormap(colormapInstance))
        .ToArray();
```

```cs
[Obsolete("https://github.com/ScottPlot/ScottPlot/issues/767")]
public static Colormap[] GetColormapsOld()
{
    IColormap[] ics = AppDomain.CurrentDomain.GetAssemblies()
                        .SelectMany(s => s.GetTypes())
                        .Where(p => p.IsInterface == false)
                        .Where(p => p.ToString().StartsWith("ScottPlot.Drawing.Colormaps."))
                        .Select(x => x.ToString())
                        .Select(path => (IColormap)Activator.CreateInstance(Type.GetType(path)))
                        .ToArray();

    return ics.Select(x => new Colormap(x)).ToArray();
}
```

StendProg:
Hi @swharden,
sorry for missed information,
I get it in Winforms app, .NET Framework 4.7, `4.1.6-beta` nuget package.

StendProg:
I just want to fill combobox with available `Colormaps` names.
Then user change selection i want to create new `Colormap` by provided name.
It's possible to achive with existing `GetColormaps()` method, but i get `LoadType` exception.

Maybe it worth to make hard-coded list of available colormaps, and realise 2 additional api methods:
1. `GetAvailableColormapsNames()`;
2. `CreateColormap(string name)`; 

[...]

PR #773, opened 2021-02-12
Extend Colormap API

StendProg:
**Purpose:**
Extends Colormap API with `ColormapFactory` class. #767
`ColormapFactory` allow get available colormaps names without creating `Colormap` instances.
And then create only `Colormap` with requested name.
For adding new types of `IColormap` it needs manual add single line in `availableColormaps` dictionary initializer.

Failed pipline report:
`Unrecognized command or argument '--dry-run'
`
It looks like not my fault.

**New Functionality:**
```cs
var factory = new ColormapFactory();
var names = factory.GetAvailableNames();
var lastColorMap = factory.Create(names.Last());
var oxyColormap = factory.Create("Oxy");
```

### DIFF
--- a/dev/devops/azure-pipelines.yml
+++ b/dev/devops/azure-pipelines.yml
@@ -53,7 +53,7 @@ steps:
   inputs:
     command: 'custom'
     custom: 'format'
-    arguments: '-w src/ScottPlot.sln --dry-run --check'
+    arguments: '-w src/ScottPlot.sln --check'
 
 ### INSTALL NUGET AND RESTORE PACKAGES
 
--- a/src/ScottPlot/Drawing/Colormap.cs
+++ b/src/ScottPlot/Drawing/Colormap.cs
@@ -56,26 +56,30 @@ namespace ScottPlot.Drawing
         }
 
         public static Colormap[] GetColormaps() =>
-            Assembly
-                .GetExecutingAssembly()
-                .GetTypes()
-                .Where(type => type.Namespace == "ScottPlot.Drawing.Colormaps")
-                .Select(colormapType => (IColormap)Activator.CreateInstance(colormapType))
-                .Select(colormapInstance => new Colormap(colormapInstance))
+                typeof(Colormap)
+                .GetProperties()
+                .Where(x => x.PropertyType == typeof(Colormap))
+                .Select(x => (Colormap)x.GetValue(x))
                 .ToArray();
 
-        [Obsolete("https://github.com/ScottPlot/ScottPlot/issues/767")]
-        public static Colormap[] GetColormapsOld()
-        {
-            IColormap[] ics = AppDomain.CurrentDomain.GetAssemblies()
-                                .SelectMany(s => s.GetTypes())
-                                .Where(p => p.IsInterface == false)
-                                .Where(p => p.ToString().StartsWith("ScottPlot.Drawing.Colormaps."))
-                                .Select(x => x.ToString())
-                                .Select(path => (IColormap)Activator.CreateInstance(Type.GetType(path)))
-                                .ToArray();
+        public static string[] GetColormapNames() =>
+            typeof(Colormap)
+            .GetProperties()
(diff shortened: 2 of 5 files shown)


swharden:
Hi @StendProg, 

I added these new methods below. It now seems the Colormap class has similar functionality to the ColormapFactor class, the only difference being reflection vs. hard-coded dictionary. Do you think one method has a strong advantage over the other? Perhaps both can remain in the code base, but one method should be favored over the other for code examples. I'll go ahead and merge this, but I'm interested to learn your thoughts.

https://github.com/ScottPlot/ScottPlot/blob/4c52ce4b71274bd0a0a513e64df3acd32d69ec64/src/ScottPlot/Drawing/Colormap.cs#L58-L63

https://github.com/ScottPlot/ScottPlot/blob/4c52ce4b71274bd0a0a513e64df3acd32d69ec64/src/ScottPlot/Drawing/Colormap.cs#L65-L70

https://github.com/ScottPlot/ScottPlot/blob/4c52ce4b71274bd0a0a513e64df3acd32d69ec64/src/ScottPlot/Drawing/Colormap.cs#L72-L83

Thanks again!

StendProg:
Hi @swharden,
> the only difference being reflection vs. hard-coded dictionary. Do you think one method has a strong advantage over the other?

What are the disadvantages of reflection and the `static` keyword for me?
This is outside of the object oriented programming paradigm. With all the ensuing consequences.
I'm afraid I won't undertake to argue the details, but I think it was done long ago by much more capable people.
Nevertheless, both of these thing's have the right to be, and allow you to solve problems that are unsolvable or difficult to solve using OOP. When using reflection, one must clearly understand what this threatens, and when the rules can be `broken`. I don't understand such things well, so I prefer to avoid using them. In addition to everything, reflection is slow, just dozens of times slower than native calls.

As an illustrative example for comparing the 2 described approaches, I came up with the following example:
Let's say tomorrow I want to have 3 colormap categories `Nice`, `Medium`, `Bad`.
All I need to do with `ColormapFactory approach`:
This is to inherit 3 factories from `ColormapFactory` and in each just to correct the existing dictionary. The work is done, you can instantiate a new `NiceColormapFactory` and ask for `nice` colormaps. At the same time, the existing code is not affected in any way, and remains guaranteed to work.

How to deal with the approach using reflection, but most likely you will have to edit the implementation. If you inherit from Colormap (I don't know how static properties will behave.). You will have to do something with this line, because it will not affect the inherited classes, and you will not be able to inherit the behavior.
```c#
typeof(Colormap)
```

All these approaches to programming that are written in smart books, they are not about how to solve a problem. They are about how to solve the problem so that in the future it can be expanded, modified without large costs.

Although, of course, for this task I am quite satisfied with the approach, `it works - do not touch`. It's just that I got incomprehensible errors with reflection on an old project, i.e. it didn't work. Therefore, I solved the problem as I was used to, and tried to explain why I was so used since you asked.

swharden:
Hi @StendProg, thank you for your extensive answer!

The points that stand out most to me are

> dozens of times slower than native calls

> you will not be able to inherit the behavior ... solve the problem so that in the future it can be expanded

> incomprehensible errors with reflection

These are excellent points! In summary, it makes sense to solve problems using OOP where possible and reserve reflection for special cases where it is not possible to use OOP. This is good enough justification for me to stop using reflection, and instead prefer OOP approaches where possible.

I come to C# from a Python (data science / scripting / non-OOP) background. Python has lots of magic. For example, a running program can easily read its own code comments (that functionality is built into the language). I am attracted to reflection because it feels very natural to me, but I now recognize that in a large OOP system it should be avoided to improve extension and maintainability.

Thank you again for your comments! I've been learning a lot about C#/OOP by working on this library, and this discussion was very helpful! I really value the advice you have provided along the way, so thanks!

I'll refactor the colormap to use the factory for lookups, and only put reflection inside tests to ensure every colormap is present in the factory (#797).

PR #780, opened 2021-02-15
Remove a Console.WriteLine() Call in ScatterPlotList.Render()

Scr0nch:
**Purpose:**

Removes an unnecessary, performance-degrading `Console.WriteLine()` call inside the `ScatterPlotList`s `Render` method.


swharden:
Good catch! Thanks @Scr0nch! This will get packaged in the next NuGet release in a few days :thum

Issue #800, opened 2021-02-16
Small candlesticks render without a body


AlgoExecutor:
It seems that some candles are not displayed correctly. This affects only candles without a body. So-called Dojis. You can see the differences on the MS Reference Chart: 

![image](https://user-images.githubusercontent.com/79081847/108119973-daa11480-70a0-11eb-8f79-c50ecf134785.png)

Version: **ScottPlot 4.1.8-beta**

Code:
```cs
 pData.Add(new OHLC((double) candle.open,
                        (double) candle.high,
                        (double) candle.low,
                        (double) candle.close,
                        candle.opentime,
                        1.0 / (24 * 60)
formsPlot.Plot.AddCandlesticks(pData.ToArray());
formsPlot.Plot.XAxis.DateTimeFormat(true);
formsPlot.Render();
```

Data:
OHLC: open=0,1353, high=0,1354, low=0,1353, close=0,1353, timestamp=44243,7229166667
OHLC: open=0,1347, high=0,1353, low=0,1346, close=0,1349, timestamp=44243,7222222222
OHLC: open=0,1344, high=0,1347, low=0,1344, close=0,1347, timestamp=44243,7215277778
OHLC: open=0,1339, high=0,1344, low=0,1339, close=0,1344, timestamp=44243,7208333333
OHLC: open=0,1334, high=0,1342, low=0,1334, close=0,1339, timestamp=44243,7201388889
OHLC: open=0,1334, high=0,1334, low=0,1329, close=0,1334, timestamp=44243,7194444444
OHLC: open=0,1334, high=0,1334, low=0,133, close=0,133, timestamp=44243,71875
OHLC: open=0,133, high=0,1334, low=0,133, close=0,1334, timestamp=44243,7180555556
OHLC: open=0,1333, high=0,1334, low=0,1329, close=0,1334, timestamp=44243,7173611111
OHLC: open=0,1337, high=0,1341, low=0,1329, close=0,1335, timestamp=44243,7166666667
OHLC: open=0,1338, high=0,1341, low=0,1334, close=0,1334, timestamp=44243,7159722222
OHLC: open=0,133, high=0,1339, low=0,133, close=0,1339, timestamp=44243,7152777778
OHLC: open=0,1334, high=0,1334, low=0,133, close=0,133, timestamp=44243,7145833333
OHLC: open=0,1341, high=0,1341, low=0,1334, close=0,1334, timestamp=44243,7138888889
[...]

PR #803, opened 2021-02-17
Improve small candlestick rendering and support custom wick color

bclehmann:
**New Contributors:**

**Purpose:**
Fixes the drawing of "Doji" candlesticks (i.e. open == close) #800 

**New Functionality:**
Adds a check for the Doji case and adds a red line where otherwise the candle body would not be drawn.

```cs
OHLC[] arr = DataGen.RandomStockPrices(new Random(0), 10);
arr[5] = new OHLC(arr[5].open, arr[5].high, arr[5].low, arr[5].open, arr[5].time); // Ensure a doji session exists

plt.AddCandlesticks(arr);
```

Before:
image

After:
image

### DIFF
--- a/src/ScottPlot/Plottable/FinancePlot.cs
+++ b/src/ScottPlot/Plottable/FinancePlot.cs
@@ -38,6 +38,8 @@ namespace ScottPlot.Plottable
         public int YAxisIndex { get; set; } = 0;
         public Color ColorUp = Color.LightGreen;
         public Color ColorDown = Color.LightCoral;
+        public Color WickColor = Color.LightGray;
+        public bool SeparateWickColor = true;
 
         public AxisLimits GetAxisLimits()
         {
@@ -110,10 +112,10 @@ namespace ScottPlot.Plottable
                 boxWidth = (float)(spacingPx / 2 * fractionalTickWidth);
             }
 
+
             using Graphics gfx = GDI.Graphics(bmp, dims, lowQuality);
             using Pen pen = new Pen(Color.Magenta);
             using SolidBrush brush = new SolidBrush(Color.Magenta);
-
             for (int i = 0; i < OHLCs.Length; i++)
             {
                 var ohlc = OHLCs[i];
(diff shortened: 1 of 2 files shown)


bclehmann:
Currently the body of the candle is the colour passed into `ColorDown` (generally red), like the rest of the candlestick. I don't know if this is preferred, after all red implies that the price went down over the session, but the point of this PR is for when the price doesn't change. Perhaps someone with more interest in technical analysis can weigh in here.

AlgoExecutor:
That is good point @bclehmann. Often the candle wick is kept neutral. Black is often used against a white background. Here is an example with different backgrounds:
image

You could also leave it up to the user by making this color configurable and overwriting the existing settings. 
I would not discard the green/red wicks completely. Some people like that. If it would be controllable via options that would be great ;)

bclehmann:
image

I added a `WickColor` field which is by default `System.Drawing.Color.LightGray`. If `SeparateWickColor` is true then this colour is used instead of either `ColorUp` or `ColorDown`. This doesn't impact OHLCs at all.

Currently `SeparateWickColor` is true by default, which I leave to others to decide on which they think is better. I have not added a demo to illustrate the use of `SeparateWickColor` because if it gets changed to being off by default that demo would likely need to be changed to reflect that.

bclehmann:
The grey wick is making me wonder if the candle bodies should have an outline?

I tried it and got this:
image

Perhaps `LightGrey` is too light, but I actually quite like this.

AlgoExecutor:
Hi @bclehmann 
thank you for the quick feedback. The wicks look good 👍 . Normally they are relatively dark. Super cool if you can customize them. The outline I would make optional or very thin? Looks like the MS chart outline was no bigger than 1 px? You can see it here #800 in the reference image. If it's not complicated, you could make it customizable as well ;)

swharden:
Hey @bclehmann and @AlgoExecutor, thanks for your input on this one!

* Default styling is designed to closely mimic https://www.tradingview.com/chart

* Customizing the wick (and border) color was a great idea, but I struggled with how to best expose this. I'm typically against using nullable value types because they can be confusing, but I think they make sense here (compared to a separate bool flag variable, or a magic color like Transparent to indicate no border).

I added some new examples to the cookbook demonstrating what it can do.

finance_color

I'm worried I'm inviting chaos into the ScottPlot community with this one... but I made a demo designed to look like TC2000 😅 We'll see how much craziness it invites...

image

I love how this example shows a clear "consolidation period" and periods of "momentum", all resulting from a random number generator 😈

As a neuroscientist, I find the combination of chart reading and pattern hallucination fascinating.

Again I'm reminded of [A Random Walk Down Wall Street](https://www.amazon.com/dp/B07MWDML9V/)

AlgoExecutor:
Thank you both @swharden and @bclehmann for the good work! I also had to laugh about the TC2000 chart :joy:. Looks like chaos is really coming - love it ;-)

You make a really good point. Reading charts is some kind of art. You just have to believe in it hard enough. Always reminds me of the tea cup pattern: https://www.investopedia.com/terms/c/cupandhandle.asp

Looks like the inventor had a deeper meaning in mind :upside_down_face:

PR #807, opened 2021-02-18
Add support of negetive OLE Automation Date values

StendProg:
**Purpose:**
OLE Automation date support negative values, and it's minimum value is midnight, 1 January 0100.##806
This PR allow to use negative values and expand avalable range.

**New Functionality:**
image

### DIFF
--- a/src/ScottPlot/Ticks/TickCollection.cs
+++ b/src/ScottPlot/Ticks/TickCollection.cs
@@ -1,5 +1,5 @@
-﻿using ScottPlot.Ticks.DateTimeTickUnits;
-using ScottPlot.Drawing;
+﻿using ScottPlot.Drawing;
+using ScottPlot.Ticks.DateTimeTickUnits;
 using System;
 using System.Collections.Generic;
 using System.Globalization;
@@ -131,7 +131,7 @@ namespace ScottPlot.Ticks
 
             if (low < high)
             {
-                low = Math.Max(low, new DateTime( 0100, 1, 1, 0, 0, 0).ToOADate()); // minimum OADate value
+                low = Math.Max(low, new DateTime(0100, 1, 1, 0, 0, 0).ToOADate()); // minimum OADate value
                 high = Math.Min(high, DateTime.MaxValue.ToOADate());
 
                 var dtManualUnits = (verticalAxis) ? manualDateTimeSpacingUnitY : manualDateTimeSpacingUnitX;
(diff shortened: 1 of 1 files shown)


audun:
Note that this doesn't work correctly for the date 1899-12-30. OLE Automation dates have surprising conversions between -1.0 and 1.0:

```
DateTime.FromOADate(-1.0).ToString("s") "1899-12-29T00:00:00"
DateTime.FromOADate(-0.9).ToString("s") "1899-12-30T21:36:00"
DateTime.FromOADate( 0.9).ToString("s") "1899-12-30T21:36:00"
DateTime.FromOADate( 1.0).ToString("s") "1899-12-31T00:00:00"
```

StendProg:
Hi @audun,
Thanks for the fun fact, really as you wrote.
But I think it's worth leaving everything as it is, it is unlikely that many will honor 30 December 1899 with an attentive look.

audun:
True, and it works okayish. 

image

However, it appears that there are more issues with negative OADates. Note that the two first time ticks above are reversed. It becomes very obvious if you change the date generation in the TicksDateTime recipe to show a few data points per date in a monotonously increasing fashion, like: `myDates[i] = new DateTime(1899, 12, 20, 1, 0, 0).AddDays(i / 4.0);`

image

If I change it to use the "Excel" conversion, which is very close to OADate, it looks like what you'd expect, and is much easier to reason about.

image

```
        // From OxyPlot.Axes.DateTimeAxis
        private static readonly DateTime TimeOrigin = new DateTime(1899, 12, 31, 0, 0, 0, DateTimeKind.Utc);
        public static double ToDouble(DateTime value)
        {
            var span = value - TimeOrigin;
            return span.TotalDays + 1;
        }

        {
            // create data sample data
            DateTime[] myDates = new DateTime[100];
            for (int i = 0; i < myDates.Length; i++) 
                myDates[i] = new DateTime(1899, 12, 20, 1, 0, 0).AddDays(i / 4.0);

            // Convert DateTime[] to double[] before plotting
            //double[] xs = myDates.Select(x => x.ToOADate()).ToArray();
            double[] xs = myDates.Select(x => ToDouble(x)).ToArray();
            double[] ys = DataGen.RandomWalk(myDates.Length);
            plt.AddScatter(xs, ys);

            // Then tell the axis to display tick labels using a time format
            plt.XAxis.DateTimeFormat(true);
        }
```

Issue #816, opened 2021-02-22
OHLC should implement IComparable


ticool:
Hello,

Im just saving your OHLC  to Entity framework  with Sqllight Core and when im trying to make queries using linkq i got this :

Property 'HistoricRates.OhlcItem' cannot be used as a key because it has type 'OHLC' which does not implement 'IComparable<T>', 'IComparable' or 'IStructuralComparable'. Use 'HasConversion' in 'OnModelCreating' to wrap 'OHLC' with a type that can be compared.'

```cs
public static List<OHLC> GetOhlcList(string cryptoCoinName, Granularity granularity)
{
     using (SqliteDbContext dbContext = new SqliteDbContext())
     {

      var c =  dbContext.HistoricRates.Where(s => s.MarketItemName == MarketItemName )     // Multiple market products
                                                    .Where(s => s.GranularityItem == granularity)                   // Multiple timeSpan  5 min 1 hour etc
                                                    .Select(s => new
                                                    {
                                                         s = s.OhlcItem
                                                    }).ToList();
     }
}


public class HistoricRates
{
        public string MarketItemName { get; set; }
        public Granularity GranularityItem { get; set; }
        public OHLC OhlcItem { get; set; }
}
```

Would be cool if OHLC  had IComparable<T> interface !

If you have better idea on how to implement that im open to new ideas ! :)

bclehmann:
I agree it would be useful for OHLCs to implement `IComparable`, although I think it could be ambiguous whether it should compare them based on open or closing price.

For the timebeing, you can try following the error message by creating a value converter and registering it with `HasConversion`: https://docs.microsoft.com/en-us/ef/core/modeling/value-conversions?tabs=data-annotations. I haven't messed with that specific aspect of EF Core so I don't have any further advice. I believe this assumes you have control over `SqliteDbContext` and that it's not autogenerated, although even in that case you may be able to extend that class and override `OnModelCreating` regardless. If that class overrides `OnModelCreating` itself (often used for seeding data) then that could get hairy.

Another option instead of setting up value conversions between a wrapper class that implements `IComparable` is to just put that wrapper class in the database and not bother registering an automatic conversion.

ticool:
Ok thx for suggestion !

swharden:
It seems easy enough to implement comparison functionality in OHLCs. Rather than price, I think comparing the `DateTime` makes sense (especially when considering sorting).

Would something like this address the issue? (#818)
[...]

PR #818, opened 2021-02-23
OHLC implement IComparable

swharden:
Discussion in  #816


swharden:
see https://github.com/ScottPlot/ScottPlot/issues/816#issuecomment-783796158

PR #839, opened 2021-02-27
Updated Plot.Add.cs to return a correct array of BarPlots

rhys-wootton:
Adds bar to BarPlot array so it now returns an array of BarPlots instead of a null array.

**Purpose:**
Beforehand, calling AddBarGroups would result in it returning an array with a single null element in it as seen below:

image

All that had to be changed was actually adding the bar to the array within the loop! A nice quick fix and now I can change the styles of the bars in that group, which I wasn't able to do before. I created a quick extension method that copied the original with that addition and now you get the expected result:

image

### DIFF
--- a/src/ScottPlot/Plot/Plot.Add.cs
+++ b/src/ScottPlot/Plot/Plot.Add.cs
@@ -129,6 +129,7 @@ namespace ScottPlot
                     ErrorCapSize = errorCapSize,
                     FillColor = GetNextColor()
                 };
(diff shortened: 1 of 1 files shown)


swharden:
Good catch @rhys-wootton! I'll merge this is now, and I plan to release a NuGet package tomorrow. Thanks for the PR!

rhys-wootton:
No problem! This is a great bit of kit, thanks for actively developing it :)

Issue #844, opened 2021-03-02
PiePlot with individual text colors


Timothy343:
Hi Scott,
currently the PiePlot class allows individual fill colors, but no font colors.

```
    public PiePlot(double[] values, string[] groupNames, Color[] colors)
    {
        SliceFillColors = colors;
        SliceFont.Color = Color.White;

```
For a good contrast, a possibility to choose proper label colors would be helpful. May I ask you to add this to the class?

swharden:
Hi @Timothy343, thanks for reporting this! I just merged this feature in #883 and will release it on NuGet tonight and it will show up in the cookbook. Thanks for the suggestion!

PR #883, opened 2021-03-18
Custom pie chart slice and label colors

bclehmann:
**New Contributors:**

**Purpose:**
#844 

**New Functionality:**
A new `SliceLabelColors` array is exposed.

```cs
double[] values = { 778, 43, 283, 76, 184 };
string[] labels = { "C#", "JAVA", "Python", "F#", "PHP" };
var pie = plt.AddPie(values);
pie.SliceLabels = labels;
pie.ShowLabels = true;
pie.SliceFillColors = new Color[] { // From https://github.com/ozh/github-colors
	Color.FromArgb(0x17, 0x86, 0x00),
	Color.FromArgb(0xB0, 0x72, 0x19),
	Color.FromArgb(0x35, 0x72, 0xA5),
	Color.FromArgb(0xB8, 0x45, 0xFC),
	Color.FromArgb(0x4F, 0x5D, 0x95)
};

pie.SliceLabelColors = new Color[] { Color.White, Color.Silver, Color.Silver, Color.Silver, Color.Silver }; // Highlight the largest section
```

image

### DIFF
--- a/src/ScottPlot/Plottable/PiePlot.cs
+++ b/src/ScottPlot/Plottable/PiePlot.cs
@@ -17,6 +17,7 @@ namespace ScottPlot.Plottable
         public string[] SliceLabels;
 
         public Color[] SliceFillColors;
+        public Color[] SliceLabelColors;
         public Color BackgroundColor;
 
         public bool Explode;
@@ -86,7 +87,7 @@ namespace ScottPlot.Plottable
             using (Pen outlinePen = GDI.Pen(OutlineColor, OutlineSize))
             using (Brush sliceFillBrush = GDI.Brush(Color.Black))
             using (var sliceFont = GDI.Font(SliceFont))
-            using (Brush sliceFontBrush = GDI.Brush(SliceFont.Color))
+            using (SolidBrush sliceFontBrush = (SolidBrush)GDI.Brush(SliceFont.Color))
             using (var centerFont = GDI.Font(CenterFont))
             using (Brush centerFontBrush = GDI.Brush(CenterFont.Color))
             using (StringFormat sfCenter = new StringFormat() { LineAlignment = StringAlignment.Center, Alignment = StringAlignment.Center })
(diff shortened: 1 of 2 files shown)


bclehmann:
I elected to change from `Brush` to `SolidBrush` for `sliceLabelBrush` to allow reusing the same brush instead of creating one for each slice. I also chose to only expose an array of colours instead of fonts, as I don't know that having labels of different fonts or sizes makes sense. Although now that I'm thinking about it, perhaps allowing the user to change the size could let them fit longer text in?

As an aside, I typed "color" instead of "colour" in my commit message without thinking, and I already use color within code for consistency's sake. I guess I'm no longer a Canadian :(

StendProg:
Hi @bclehmann,

I have a couple of notes on the code:
1. It is necessary to check for equality of the size of `SliceLabelColors` and `Values`. Otherwise, you can get an exception in your code.
2. Using null propagation will make this line look a little prettier, although this is a matter of taste.
```c#
//sliceFontBrush.Color = SliceLabelColors is null ? SliceFont.Color : SliceLabelColors[i];
sliceFontBrush.Color = SliceLabelColors?[i] ?? Slice.Font.Color;
```

bclehmann:
Interesting, I knew about `?.` but not `?[]`. I elected not to use it though as there's no point separating the null check and the compatible length check.

StendProg:
It's possible to move evaluating expression outside loop in variable say `useCustomLabelColors`.  Although it is quite possible that the compiler will do this, and the loop is unlikely to be greater than 100.

bclehmann:
I don't know how linking works in C# or how much optimization can be done once you already have it in MSIL, but my first instinct is to say that it cannot optimize before linking because it is a public variable and thus subject to change. Since it's exposed as part of an API it is considered volatile, even if our code never writes to it the user could (although for it to happen between loop iterations the user would need to do so on another thread).

I don't know if they can do any optimization after linking if they determine that the user code can never change that variable after the first loop iteration. Regardless of whether it's optimized (and whether the optimization would be noticeable), it's an easy fix.

swharden:
> As an aside, I typed "color" instead of "colour" in my commit message without thinking, and I already use color within code for consistency's sake. I guess I'm no longer a Canadian :(

Let's go with `color` since it's consistent spelling with `System.Drawing.Color`

You're still Canadian! It seems _grey_ is your _favourite_ font _colour_ for demos and the new _behaviour_ of this PR is _marvellous_ 😜

swharden:
```cs
sliceFontBrush.Color = SliceLabelColors?[i] ?? SliceFont.Color;
```

Thanks for sharing this @StendProg! Even if this implementation is not used here, I learned something new by reading this, and this looks technique looks very useful

Issue #809, opened 2021-02-18
Incorrect IHasPoints.GetPointNearestX results using SignalPlotXY 


at2software:
ScottPlot version v4.1.8-beta

_Note: this might be related to #548 and maybe this function is just not yet supported for `SignalPlotXY` (then this is not a bug of course, but probably this function should then not be available to `SignalPlotXY` instead of returning false values)._

There seems to be an issue with the results returned by `IHasPoints.GetPointNearestX` with a `SignalPlotXY` when the x values aren't consecutive integers starting at zero. 
Here is an example to highlight the nearest point of a ScatterPlot (adapted from [here](https://github.com/ScottPlot/ScottPlot/blob/b09e6dd902378c5818e64ed8efb81ae408b40e37/src/sandbox/WinFormsFrameworkApp/Form1.cs))

```cs
public partial class MainWindow : Window
{
	private readonly IHasPoints plottable;
	private readonly ScatterPlot HighlightedPoint;
	private int LastHighlightedIndex = -1;

	public MainWindow()
	{
		InitializeComponent();

		int sampleCount = 50;
		double[] xs = new double[sampleCount];
		double[] ys = new double[sampleCount];

		Random rand = new Random(0);
		for (int i = 0; i < 50; i++)
		{
			xs[i] = DateTime.Now.ToOADate() + i*10;
			ys[i] = i > 0 ? ys[i - 1] + rand.NextDouble() - .5 : 0;
		}

		//plottable = plot.Plot.AddScatterPoints(xs, ys);    // <-- this works as expected
		plottable = plot.Plot.AddSignalXY(xs, ys);
			
		HighlightedPoint = plot.Plot.AddPoint(0, 0);
		HighlightedPoint.Color = System.Drawing.Color.Red;
		HighlightedPoint.MarkerSize = 10;
		HighlightedPoint.MarkerShape = MarkerShape.openCircle;
		HighlightedPoint.IsVisible = false;

		plot.Plot.XAxis.DateTimeFormat(true);
	}

	private void plot_MouseMove(object sender, MouseEventArgs e)
	{
		(double mouseCoordX, double mouseCoordY) = plot.GetMouseCoordinates();

		// these results are incorrect when plottable is SignalPlotXY:
		(double pointCoordX, double pointCoordY, int pointIndex) = plottable.GetPointNearestX(mouseCoordX);

		// highlight the point of interest
		HighlightedPoint.Xs[0] = pointCoordX;
[...]

PR #886, opened 2021-03-18
Improve GetPointNearest support for generic signal plots

StendProg:
**Purpose:**
`PlottableSignalXY` used the implementation from `PlottableSignal` due to inheritance, which led to inconsistent results when calling the `GetPointNearestX`. #809, #882
This PR brings to `PlottableSignalXY` it's own `GetPointNearestX` implementation with correct behavior.

This pull request consists of 2 commits. The very first one fixes the described problem.
In the second commit, I decided to go further, split the original interface into several, and make it so that `PlottableSignals` explicitly implement only what it is capable of `IHasPointsXCalculatable` with a single `GetPointNearestX()` method
On top of that, I made interfaces generic, since signals inherently support generics and this is more natural for them.

Old nongeneric `IHasPoints` are derived from these interfaces and are essentially unchanged. These changes should not affect the `PlottableScatter` in any way. But `PlottableSignals` lose API compatibility because they now implement a different interface, and users will have to change the interface type, although the method has remained with the same name.

If you think I went too far in the second commit, just revert it (I'm not entirely sure myself).
If you find the changes acceptable, I will be glad to have better names for the interfaces. And I would like to have time until tomorrow to once again take a fresh look at all this (I really do not want to drag out as last time, I am still ashamed).

### DIFF
--- a/src/ScottPlot/Plot/Plot.Obsolete.cs
+++ b/src/ScottPlot/Plot/Plot.Obsolete.cs
@@ -3,14 +3,13 @@
  *   - Long lists of optional arguments (matplotlib style) are permitted.
  *   - Use one line per argument to simplify the tracking of changes.
  */
+using ScottPlot.Plottable;
+using ScottPlot.Statistics;
 using System;
 using System.Collections.Generic;
 using System.Diagnostics;
 using System.Drawing;
 using System.Linq;
-using System.Text;
-using ScottPlot.Plottable;
-using ScottPlot.Statistics;
 
 namespace ScottPlot
 {
@@ -742,7 +741,7 @@ namespace ScottPlot
             T[] ys,
             double sampleRate = 1,
             double xOffset = 0,
-            double yOffset = 0,
+            T yOffset = default(T),
             Color? color = null,
             double lineWidth = 1,
             double markerSize = 5,
--- a/src/ScottPlot/Plottable/IHasPoints.cs
+++ b/src/ScottPlot/Plottable/IHasPoints.cs
@@ -1,9 +1,21 @@
 ﻿namespace ScottPlot.Plottable
 {
-    public interface IHasPoints
+    public interface IHasPointsXCalculatable<TX, TY>
     {
-        (double x, double y, int index) GetPointNearestX(double x);
-        (double x, double y, int index) GetPointNearestY(double y);
-        (double x, double y, int index) GetPointNearest(double x, double y, double xyRatio = 1);
+        (TX x, TY y, int index) GetPointNearestX(TX x);
+    }
+
+    public interface IHasPointsYCalculatable<TX, TY>
+    {
+        (TX x, TY y, int index) GetPointNearestY(TY y);
+    }
+
+    public interface IHasPointsGeneric<TX, TY> : IHasPointsXCalculatable<TX, TY>, IHasPointsYCalculatable<TX, TY>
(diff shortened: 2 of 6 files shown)


bclehmann:
Ah, I've gotten that error before. Looks like you added a public method but not documentation for it (or removed documentation from a public method).

From a quick glance it looks like it might bet this one: https://github.com/StendProg/ScottPlot/blob/201c85c0385515510aed526af71534faa9627a2e/src/ScottPlot/Plottable/SignalPlotBase.cs#L167

However running the test in debug mode will track it down if I'm wrong.

StendProg:
Thanks for the tip, I spent some time trying to figure out this test.Checked all public methods, but somehow forgot about constructors. I miscalculated over time a bit, I can only continue to deal with this tomorrow.

StendProg:
It looks like the problem here is completely different. On my local machine, I am getting an error test because there is no `summary` for `SignalXY.GetPointNewestX()`. This method is not even declared in `SignalXY`, but inherits from `SignalXYGeneric`. Maybe this is the case and this test does not take into account the possibility of `summary` inheritance. IntelliSense does an excellent job with this.

StendProg:
I made private new (double x, TY, int index) GetPointNearestX(double X)
This do the trick for local tests, but online stil need documentation.
So I documented this method.
I do not have enough knowledge to understand all the nuances of hiding the interface with a more general interface. So I've tested a bit and it seems to work.

swharden:
## Required XML Documentation

> I spent some time trying to figure out this test

Sorry this took a while to figure out! A few weeks ago I added tests to ensure _plottables_ are documented (without requiring every module to be documented). I think the tests require public methods and constructors to have XML, but ignores public fields and properties. 

I see that links to [https://www.fuget.org/packages/ScottPlot](https://www.fuget.org/packages/ScottPlot) now appear on official NuGet pages, so I'm hoping that better XML documentation may make the API easier to figure out just using this website. For example, [the ScatterPlot API](https://www.fuget.org/packages/ScottPlot/4.1.9-beta/lib/netstandard2.0/ScottPlot.dll/ScottPlot.Plottable/ScatterPlot). It's nice for flat classes, but not quite as easy to navigate for complex inheritance trees.

Eventually I'd like to work toward enabling [#1591](https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/compiler-messages/cs1591) at the project level...

## Generic Support for IHasPoints

Thanks for this PR! At first I was reluctant to include the second commit because those interfaces seemed pretty complex and the generic lambda expression was somewhat hard to understand at first. I think I made it a little easier to read by changing some of the names, making the expression readonly, adding XML documentation, and wrapping the expression in typed functions calls so it doesn't have to be directly interacted with in the rendering/searching routines.

This extra function call may slightly impair performance, but these methods are only called from `RenderLowDensity()` and `GetPointNearestX()` so I think it is ok. Let me know if you disagree and I can refactor these to interact with the expression directly.

https://github.com/ScottPlot/ScottPlot/blob/a1ff357d036dc95eed583ba2979532f4bebe595d/src/ScottPlot/Plottable/SignalPlotBase.cs#L167-L188

## OffsetX and OffsetY are broken in SignalPlotXY

I don't think this is a new bug, but I created a cookbook example to demonstrate it in this PR. It should probably be addressed in another PR (I opened an issue #890)

Issue #889, opened 2021-03-19
Scatter plot render artifacts


grabul:
If you have different densities on the x-axis of your scatter data and you zoom into a high density area there are render errors drawing additional horizontal lines where we shouldn't have there.
example code:

```cs
var plt = new ScottPlot.Plot(600, 400);
int pointCount = 1000;
double[] ys = new double[pointCount];
double[] xs = new double[pointCount];
double currentX = 0;
 for (int i = 0; i < pointCount; i++)
 {
    if ((i % 100) < 10)
       currentX += 10;
    else
       currentX += 0.001;
    xs[i] = currentX;
    ys[i] = i % 2 == 0 ? 1.0 : 0.0;
 }
plt.AddScatterStep(xs, ys);
```

Also belongs to a normal ScatterPlot (no step style) where you add additional data points to create the steps "by hand" within the loop like:

```cs
var plt = new ScottPlot.Plot(600, 400);
int pointCount = 1000;
var xs = new List<double>();
var ys = new List<double>();
double last_y = -1.0;
double currentX = 0;
for (int i = 0; i < pointCount; i++)
{
  if ((i % 100) < 10)
     currentX += 10;
  else
     currentX += 0.001;

  if (last_y >= 0.0)
  {
     xs.Add(currentX);
     ys.Add(last_y);
  }

  xs.Add(currentX);
  last_y = i % 2 == 0 ? 1.0 : 0.0;
  ys.Add(last_y);
}
plt.AddScatter(xs.ToArray(), ys.ToArray());
```

![scatter](https://user-images.githubusercontent.com/80955161/111768050-b6a33f80-88a7-11eb-8f90-7946649ba3d0.jpg)

swharden:
Hi @grabul, thanks for reporting this!

## The Problem is GDI+

This looks like a rendering bug related to GDI+. It occurs when lines are drawn that extend extremely long distances off the screen.

![artifact](https://user-images.githubusercontent.com/4165489/111889958-7d5bf400-89bb-11eb-97df-6ea4b024a821.gif)

## Immediate Fix / Workaround

An immediate fix would be for the user to dice-up their data a bit more to ensure there aren't data points with extremely long gaps. It's not elegant, but it would work right away.

## Best Solution

[...]

PR #893, opened 2021-03-22
Fix SignalXY right edge clipping bug

StendProg:
**Purpose:**
Fix for `SignalXY` wrong clipping on right edge. https://github.com/ScottPlot/ScottPlot/issues/889#issuecomment-803582419

The problem was not at incorrect clipping, but that `float` `DataWidth`, when cast to `int`, is not correctly rounded. thus, for example, with a width of `627.998`, a simple conversion to int gives 627 instead of 628.
And the points that should have gotten into the last column did not get there. as a consequence, interpolation did not go from the last point of the last column, but from the last point of the previous colomn. Which gave such an effect if points in last and previous columns had many difference.

In general, it is very strange to use `float` variable for the pixels count, a quick glance at the project showed that cast to int is still used in many places, they can also play a cruel joke.

### DIFF
--- a/src/ScottPlot/Plottable/SignalPlotXYGeneric.cs
+++ b/src/ScottPlot/Plottable/SignalPlotXYGeneric.cs
@@ -163,7 +163,7 @@ namespace ScottPlot.Plottable
                 IEnumerable<PointF> VisiblePoints;
                 if (UseParallel)
                 {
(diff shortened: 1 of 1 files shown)


swharden:
> that float DataWidth, when cast to int, is not correctly rounded

Thank you for figuring this one out @StendProg!

PR #901, opened 2021-03-24
Optimizations to Tests and Cookbook Generation

bclehmann:
**New Contributors:**

**Purpose:**
This is a PR to make running tests faster. My test results are below:

Cookbook Tests | Avg Time (s)
-- | --
Before | 7.5
After | 5.2

In theory, these optimizations should affect other tests but in reality I couldn't notice any meaningful difference.

I know optimizing tests is a pretty low priority (especially as the whole suite takes under 20 seconds to run) but it's a pretty straightforward set of optimizations: all of them are simply caching the results of pure functions. This can be especially fruitful with reflection as it can be quite slow. Turning linear searches into dictionary lookups is also quite easy.

I haven't looked at optimizing other tests but most that take up a lot of time are slow because they are benchmarks, i.e. they're supposed to be slow.

**New Functionality:**
N/A

### DIFF
--- a/src/cookbook/ScottPlot.Cookbook/Chef.cs
+++ b/src/cookbook/ScottPlot.Cookbook/Chef.cs
@@ -6,6 +6,7 @@ using System.IO;
 using System.Linq;
 using System.Text;
 using System.Text.RegularExpressions;
+using System.Threading.Tasks;
 
 namespace ScottPlot.Cookbook
 {
@@ -27,13 +28,16 @@ namespace ScottPlot.Cookbook
         /// </summary>
         public void CreateCookbookImages(string outputPath)
         {
+            List<Task> tasks = new List<Task>();
+
             outputPath = Path.GetFullPath(outputPath);
             if (!Directory.Exists(outputPath))
                 Directory.CreateDirectory(outputPath);
 
             var recipes = Locate.GetRecipes();
             Console.WriteLine($"Cooking {recipes.Length} recipes in: {outputPath}");
-            foreach (var recipe in recipes)
+
+            Parallel.ForEach(recipes, recipe =>
             {
                 Debug.WriteLine($"Executing {recipe.ID}");
                 var plt = new Plot(Width, Height);
@@ -52,7 +56,8 @@ namespace ScottPlot.Cookbook
                 string thumbFileName = (recipe.ID + ExtThumb).ToLower();
                 string thumbFilePath = Path.Combine(outputPath, thumbFileName);
                 thumb.Save(thumbFilePath, System.Drawing.Imaging.ImageFormat.Jpeg);
-            }
+            });
+
         }
 
         /// <summary>
@@ -67,7 +72,7 @@ namespace ScottPlot.Cookbook
 
             var sources = GetRecipeSources(sourcePath);
             Console.WriteLine($"Creating source code for {sources.Count} recipes in: {outputPath}");
-            foreach (var recipe in sources)
+            Parallel.ForEach(sources, recipe =>
             {
                 StringBuilder sb = new StringBuilder();
                 sb.AppendLine("// " + recipe.title);
@@ -77,7 +82,7 @@ namespace ScottPlot.Cookbook
                 string fileName = recipe.id + ".cs";
                 string filePath = Path.Combine(outputPath, fileName.ToLower());
                 File.WriteAllText(filePath, sb.ToString());
-            }
+            });
         }
 
         /// <summary>
--- a/src/cookbook/ScottPlot.Cookbook/Chef.cs
+++ b/src/cookbook/ScottPlot.Cookbook/Chef.cs
@@ -28,8 +28,6 @@ namespace ScottPlot.Cookbook
         /// </summary>
         public void CreateCookbookImages(string outputPath)
         {
-            List<Task> tasks = new List<Task>();
-
             outputPath = Path.GetFullPath(outputPath);
             if (!Directory.Exists(outputPath))
                 Directory.CreateDirectory(outputPath);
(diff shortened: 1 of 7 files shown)


bclehmann:
I ran `dotnet-format` locally which caused two files to be autoformatted incorrectly. Not sure where this issue came from :/

bclehmann:
Saved ~1 more second by using `Parallel.ForEach` instead of normal `foreach`. I actually resorted to it after getting no benefit from making a bunch of tasks and using `Task.WhenAll` -_-

bclehmann:
I'm a little curious about this parameter: https://github.com/ScottPlot/ScottPlot/blob/master/src/tests/Benchmark/Scatter.cs#L15

I'm not convinced that rendering speed is variable enough that 1000 iterations makes sense, and I also think that if it were the sliding scale would make the larger datasets (where performance matters more anyways) less accurate.

bclehmann:
Currently, the performance improvements for the _entire_ test suite look like this:


Full Test Suite | Avg Time (s)
-- | --
Before | 18.2
After | 11.6

I did not change what any of the tests were testing except `Test_scatter_10kPoints`, which now always runs 10 iterations for each size of dataset. That saved ~3 seconds.

src/cookbook/ScottPlot.Cookbook/Chef.cs
```diff
@@ -27,13 +28,16 @@ public class Chef
         /// </summary>
         public void CreateCookbookImages(string outputPath)
         {
+            List<Task> tasks = new List<Task>();
```
StendProg, line 31: It seems that this is no longer needed.

src/cookbook/ScottPlot.Cookbook/Locate.cs
```diff
@@ -10,66 +10,96 @@ namespace ScottPlot.Cookbook
...
         {
-            var categorizedRecipeList = new List<KeyValuePair<string, IRecipe[]>>();
+            InitializeDictionaries();
+            InitializeCategoryList();
+        }
 
-            foreach (string category in GetCategories())
+        private static void InitializeDictionaries()
+        {
+            Dictionary<string, List<IRecipe>> byCategory = new Dictionary<string, List<IRecipe>>();
```
StendProg, line 48: This dictionary can be initialized easier using LINQ:
```c#
Dictionary<string, List<IRecipe>> byCategory = GetRecipes.GroupBy( x => x.Category).ToDictionary(  gk => gk.Key, gk => gk.ToList());
```

src/cookbook/ScottPlot.Cookbook/Locate.cs
```diff
@@ -10,66 +10,96 @@ namespace ScottPlot.Cookbook
...
+                if (!byCategory.ContainsKey(curr.Category))
+                {
+                    byCategory.Add(curr.Category, new List<IRecipe>());
+                }
+
+                byCategory[curr.Category].Add(curr);
             }
 
-            string[] topCategories =
+            foreach (string category in GetCategories())
```
StendProg, line 61: If we go further, then in fact all we need to do is sort the existing dictionary.
This can be done by defining an additional selector:
```c#
Func<string, int> CategoryIndex(string category)
{
if ( topCategories.Contains(category))
  return 0;
if (bottomCategories.Contains(category))
  return 2;
return 1; // other Category
}
```

add then just sort using this selector
```c#
_recipesByCategory = byCategory.OrderBy(CategoryIndex);
```
I'm not sure if the above code will work (browser programming) but I think you get the idea.
Then it will be possible to throw out(simplify) all the code below `InitializeCategoryList()`

src/tests/Plot/Guid.cs
```diff
@@ -13,15 +13,15 @@ class Guid
         public void Test_Guids_AreUnique()
         {
             int reps = 100;
-            var guids = new List<string>();
+            var guids = new HashSet<string>();
```
StendProg, line 16: Does it really matter?
`Distinct` also uses hash, and if it slower then not many times.
100 is a small number...

src/cookbook/ScottPlot.Cookbook/Chef.cs
bclehmann: Good catch.

src/cookbook/ScottPlot.Cookbook/Locate.cs
bclehmann: Cool, I hadn't known about this. Reminds me of dictionary comprehension in Python.

src/cookbook/ScottPlot.Cookbook/Locate.cs
bclehmann: That code works, with minor adjustments. Main thing is it needs to take a `KeyValuePair<string, IRecipe>` parameter, not `string`. And then we need to maintain the order within the categories, so that's just a `ThenBy(...)`

bclehmann: This _almost_ obviates the need for the static constructor, but since this initialization depends on `_recipesByCategory` this has to run afterwards. There are other solutions such as initializing it the first time it's needed, but for now I kept the static constructor.

src/tests/Plot/Guid.cs
bclehmann: No, it doesn't really matter. It saves a little bit of time on my machine but it's mostly inconsequential.

swharden:
Thanks for optimizing this @bclehmann and for your input @StendProg!

I'm happy with the reduction in benchmarking replicates too. It may be useful to create a benchmarking module in the future we can tactically enable only when we are interested in actually inspecting the output.

Issue #948, opened 2021-04-06
PlotScatter fails plotting a single point when line width is >0


avshiza:
We are facing the same issue [PlotScatter fails plotting a single point #126
](https://github.com/ScottPlot/ScottPlot/issues/126) in ScottPlot nuget version 4.0.48 (.net framework 4.7.2) when using PlotScatter and line width is set to >1. Is this fix mentioned in the related link avaiable in this version?
Error Details
```
Parameter is not valid.
   at System.Drawing.Graphics.CheckErrorStatus(Int32 status)
   at System.Drawing.Graphics.DrawLines(Pen pen, PointF[] points)
   at ScottPlot.PlottableScatter.Render(PlotDimensions dims, Bitmap bmp, Boolean lowQuality)
   at ScottPlot.Renderer.DataPlottables(Settings settings)
   at ScottPlot.Plot.RenderBitmap()
   at ScottPlot.Plot.SaveFig(String filePath, Boolean renderFirst)
```

This is the code we are using:
```cs
var pltNew = new ScottPlot.Plot(640, 480);
pltNew.PlotScatter(new double[1] { 0 }, new double[1] { 2 }, markerSize: 1, lineWidth: 1, color: ColorTranslator.FromHtml("#FF4848"));
pltNew.SaveFig(string.Format(@"{0}test.png", Program.pathPrefix));
```

swharden:
Thanks for noting this @avshiza! This just got fixed in #949 by https://github.com/ScottPlot/ScottPlot/pull/949/files#diff-1ea41615b8a5a5a7bae1acbce3a617b8dd1d3562ed62b1a95f89bffbc99391a1

I won't release a new 4.0 package on NuGet for several more days though (#716 ), but if you want to make this change yourself you can download the 4.0 branch of ScottPlot here https://github.com/ScottPlot/ScottPlot/tree/4.0-stable

PS: These issues are all connected: #948, #949, #126, #822, #764

PR #949, opened 2021-04-06
4.0.x Fixes crash with ScatterPlots containing a single point

bclehmann:
**New Contributors:**

**Purpose:**
#948 

I don't know how you feel about backporting fixes to 4.0.x right now, I guess it depends on how soon you plan to release 4.1. In my opinion, even if you released 4.1.0 tomorrow trivial fixes like this should continue for at least a little while because 4.0 -> 4.1 does entail a sizeable number of breaking changes, even if many of them can be alleviated by search and replace. This is made easier by the fact that 4.0.48 seems pretty stable, so it shouldn't be too difficult to maintain.

The only difficulty is that 4.0.x and 4.1.x have different directory structures and the solution file has a different name. Which means you have to close Visual Studio and normally run `git clean -df` and sometimes `git reset --hard` when switching between the branches. I think this is supposed to be what git worktrees are for but I've never used them. Regardless, more an annoyance than anything else.

**New Functionality:**
N/A

### DIFF
--- a/dev/devops/azure-pipelines.yml
+++ b/dev/devops/azure-pipelines.yml
@@ -52,7 +52,7 @@ steps:
   inputs:
     command: 'custom'
     custom: 'format'
-    arguments: '-w src/ScottPlotV4.sln --dry-run --check --verbosity diagnostic'
+    arguments: '-w src/ScottPlotV4.sln --check --verbosity diagnostic'
 
 ### INSTALL NUGET AND RESTORE PACKAGES
 
--- a/src/ScottPlot/plottables/PlottableScatter.cs
+++ b/src/ScottPlot/plottables/PlottableScatter.cs
@@ -176,7 +176,7 @@ namespace ScottPlot
                 }
 
                 // draw the lines connecting points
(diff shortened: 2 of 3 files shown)


bclehmann:
I ran `dotnet-format` locally so I'm unsure what the issue is :(

StendProg:
> I ran dotnet-format locally so I'm unsure what the issue is :(

The reason is that the 4.0 branch does not contain this fix https://github.com/ScottPlot/ScottPlot/commit/08812ef370caaae55aa64dae0c07b8d3e3792795.

swharden:
Hi @bclehmann, thanks for this fix!

> I don't know how you feel about backporting fixes to 4.0.x right now

I'm happy to continue to maintain 4.0 to make small changes that fix bugs (though I don't intend to add new features there). 

My plan is to merge this is now but not release a new 4.0 package yet. I'll get 4.1 released, then add a pop-up NuGet readme to the 4.0 branch informing people about 4.1, directing people to the FAQ/cookbook, etc., and then upload that package

Issue #955, opened 2021-04-07
Consider renaming AddHeatMapCoordinated


bclehmann:
**What can we do to make ScottPlot better?**
I think `AddHeatmapCoordinated` is better because it matches the capitalization used everywhere else. `AddCoordinatedHeatmap` is another good option, especially as the type is `CoordinatedHeatmap`, not `HeatmapCoordinated`.

The [ScottPlot Roadmap](https://github.com/swharden/ScottPlot/blob/master/dev/roadmap.md) may provide additional context regarding current goals and future directions.

StendProg:
Hi @bclehmann,
You're absolutely right, it's just a mistake.
But fixing it is no longer so easy, at least you need to leave the old version marked `obsolete` with a migration hint. I think already many projects that reference to the current version.
As for changing the word order, then the logical grouping with `AddHeatmap` is broken. I once considered current option.

If we go further, then in fact `CoordinatedHeatmap` is not much differs from the original `Heatmap`. And it would be quite reasonable to combine both behaviors into one object. I chose a separate type just so as not to break the original behavior. Maybe you, as an author, will be able to summarize everything together.

bclehmann:
> As for changing the word order, then the logical grouping with AddHeatmap is broken.

Good point, I think that's more important than matching the name of the type. The type could be renamed instead? Still, the word order is a minor issue.

> But fixing it is no longer so easy, at least you need to leave the old version marked obsolete with a migration hint. I think already many projects that reference to the current version.

CoordinatedHeatmaps have not made their way into a non-beta build, so I don't think this is strictly necessary. Although it would certainly be considerate to at least give a migration hint.

As for merging the two into one class, I think that's the best solution here. I think a boolean property such as `StretchToFill` would suffice? Perhaps we could even take an enum to specify from amongst many different scaling options? Such as stretching one axis, cropping, zooming, etc.

StendProg:
> The type could be renamed instead? Still, the word order is a minor issue.

We can rename it, just `CoordinatedHeatmap` sounds more natural even to me. For types, logical grouping is not as important as for API methods (imho).

> Although it would certainly be considerate to at least give a migration hint.

It is very annoying when, after updating a third-party library, the working code stops compiling. And an additional hint on what to do very much smoothes this moment. I just reminded that we must not forget about this opportunity.

[...]

PR #963, opened 2021-04-10
AddHeatMapCoordinated → AddHeatmapCoordinated

bclehmann:
**New Contributors:**

**Purpose:**
#955 

This renaming preserves the previous method signature but marks it as obsolete. The warning points to the new method.

I also used ` [EditorBrowsable(EditorBrowsableState.Never)]` so that IDEs which respect this attribute will not show it in code completion. Note it will always be shown on the ScottPlot assembly, but not in projects which use the library. I think this attribute can be controversial, I'm using it here because the previous method signature is nearly identical and does the same thing, which can make the two difficult to differentiate. I wouldn't use this attribute on obsolete methods like `PlotHeatmap` because that method does actually do something different and thus shouldn't be hidden. I think it's okay in this case since anyone who is looking for `AddHeatMapCoordinated` should use `AddHeatmapCoordinated` and they wouldn't notice the difference, while still protecting people from accidentally using the deprecated method through code completion.

There was discussion on #955 about refactoring the heatmap plottables and potentially adding other scaling options such as cropping, zooming, scaling only one axis, etc. That might be interesting for a future PR, depending on your opinion.

**New Functionality:**
N/A

### DIFF
--- a/src/ScottPlot/Plot/Plot.Add.cs
+++ b/src/ScottPlot/Plot/Plot.Add.cs
@@ -368,7 +368,7 @@ namespace ScottPlot
         /// Returns the heatmap that was added to the plot.
         /// Act on its public fields and methods to customize it or update its data.
         /// </returns>
-        public CoordinatedHeatmap AddHeatMapCoordinated(double?[,] intensities, double? xMin = null, double? xMax = null, double? yMin = null, double? yMax = null, Drawing.Colormap colormap = null)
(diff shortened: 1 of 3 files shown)


swharden:
Hi @bclehmann, I agree with this - thanks for the PR!

Also I learned about `EditorBrowsable(EditorBrowsableState.Never)` and this has a lot of utility! There are many objects in the `Plot` module that I think could benefit from this 🤔

PR #966, opened 2021-04-10
Adds request for environment info on applicable issue templates.

bclehmann:
**New Contributors:**

**Purpose:**
I've noticed that a lot of recent issues have stemmed from users attempting to use ScottPlot 4.1 methods in 4.0 or vice versa. 

Additionally, when answering a lot of questions one has to preface them with "If you're using 4.1.x do ... If you're using 4.0.x do ...". This also happens to a lesser degree when answering GUI questions, one has to modify their answer for WinForms or WPF or Avalonia.

I think to help mitigate this we should edit the issue templates so that they ask the user for this information up front. I modified the `bug_report` and the `question-about-scottplot` templates to ask for this information:
- ScottPlot version
- Operating System
- Which platform (WPF, WinForms, Avalonia)

Let me know what you think.
**New Functionality:**
N/A

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.md
+++ b/.github/ISSUE_TEMPLATE/bug_report.md
@@ -6,6 +6,11 @@ labels: 'BUG'
 assignees: ''
 
 ---
+### Your Environment
+- ScottPlot Version:
+- Operating System:
+- Is it a GUI app? What platform?
+    - Platforms include WinForms, WPF, and Avalonia
 
 Describe the bug and how we can reproduce it.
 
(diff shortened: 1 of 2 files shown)


bclehmann:
As an aside, I've been burned by issue templates not turning out how I expect so I would check after merging to make sure they look right.

swharden:
Thanks @bclehmann! This is an improvement and I'll merge it in now (and double-check it worked). I also want to revisit/review issue templates after the 4.1 release to perhaps add a note about this version change and provide updated links to resources related to the switch (#716)

Issue #958, opened 2021-04-09
ScottPlot 4.1 needs "open in new window" right-click menu option


ademkaya:
Hello;

In the previous version I use (it was 4.0.48 I believe) there was "open in a new window" choice when we right click on the graph. It was so useful for me because I was able to enlarge the graph. Now I updated in to 4.1.11 and saw that It was removed.

Could you add it back in future revisions ?
Thank you.

Screenshot from 4.0.48 demo
![2](https://user-images.githubusercontent.com/4140956/114148118-1da19a80-9922-11eb-9d40-6cd9dd586e6f.png)

screenshot from 4.1.11
![3](https://user-images.githubusercontent.com/4140956/114148367-56417400-9922-11eb-8545-f13a3be1ed03.png)

swharden:
Hi @ademkaya, thanks for noting this! This was an oversight in the transition from ScottPlot 4.0 to 4.1. I'll add this to the triage list #716 and fix it before I remove the "beta" tag from this new version

swharden:
Thanks for the quick fix @bclehmann! For some reason I thought this would be a lot more complex. A true/deep fix means solving #819, but this is a strong improvement as it is so thanks!

PR #969, opened 2021-04-11
Adds "Open in New Window" to context menus

bclehmann:
**New Contributors:**

**Purpose:**
#958 #716 

**New Functionality:**
image

### DIFF
--- a/src/controls/ScottPlot.Avalonia/AvaPlot.axaml.cs
+++ b/src/controls/ScottPlot.Avalonia/AvaPlot.axaml.cs
@@ -172,7 +172,7 @@ namespace ScottPlot.Avalonia
             return bitmapImage;
         }
 
-        public void DefaultRightClickEvent(object sender, EventArgs e)
+        private ContextMenu GetDefaultContextMenu()
         {
             MenuItem SaveImageMenuItem = new MenuItem() { Header = "Save Image" };
             SaveImageMenuItem.Click += RightClickMenu_SaveImage_Click;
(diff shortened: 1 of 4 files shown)


bclehmann:
I believe I have discovered why I have had issues recently with `dotnet format`, I had version 4.1.131201 installed and Azure Pipelines uses the most recent (5.0.211103). I assume there's some subtle difference between the two.

swharden:
Thanks @bclehmann, you beat me to this! I tried to reduce some of the complexity in the WpfPlot control by simplifying construction of the menu, but if there's something I'm overlooking using this strategy let me know.

>  issues recently with dotnet format

Apparently .NET 5 formats things a little differently than .NET 4 🙄

You probably figured this out, but the latest way to update the formatter and run it is:
```
dotnet format
```

bclehmann:
> I tried to reduce some of the complexity in the WpfPlot control by simplifying construction of the menu, but if there's something I'm overlooking using this strategy let me know.

I made it so that it would only create the default context menu once, and just keep reusing the same thing after that. I admit the code is not as clear, and the performance difference is likely minuscule. Depending on how you feel you can more or less just copy-paste the code over to Avalonia, I think the only changes you need to make is to replace `IsOpen = true` with `Open(this)`, and then comment out the copy image stuff because Avalonia doesn't offer an easy way to add photos to the clipboard.

swharden:
> the code is not as clear, and the performance difference is likely minuscule

Okay cool - that's what I thought, but I just wanted to double check. My goal is to keep the WinForms and WPF controls as simple and easy to read as possible, since common advice I give to people asking for advanced features is "make your own control based on the existing one - it's only a couple hundred lines". Those programmers may not be that experienced with C#, so I try extra hard to keep those two modules as minimalistic and easy to read as possible

> Depending on how you feel you can more or less just copy-paste the code over to Avalonia

I'm less worried about the Avalonia control code because I assume Avalonia users are stronger programmers 🙃

Thanks again for your quick work on this! The PR came in just in time to include it in tonight's release on NuGet 🚀

Issue #982, opened 2021-04-18
Better support for high-resolution images


PeterDavidson:
As discussed in #981, the current library makes it difficult to produce high-resolution images. If large values are given for width/height then the plot will resize to the new values, with the result that text is very small, and lines are very thin. This can be corrected by manually choosing widths and sizes for all plot features, but this is tedious and does not allow for multiple image sizes to be produced without significant effort.

Ideally there would be a separation of plot dimensions from image dimensions, and adjusting the image dimensions would result in a higher resolution image without resizing. In terms of the API, the main change would be a new version of Plot.SaveFig that would allow the width/height of the output image to be specified, along with a setting that would allow the user to choose whether to resize the plot, or just change the resolution. 

The implementation of this is reasonably simple, as the GDI library used for plotting does allow for general transformations between the world coordinates of the plots and the device coordinates. The only problem is that ScottPlot does not have a single location where the GDI Graphics object can be configured, except for GDI.Graphics, which does not have the information necessary to configure the Graphics transform (it is a static function).

There is another static function in GDI.Graphics which requires passing in the PlotDimensions, but many functions do not use it. I propose to include a TranformScale in PlotDimensions which can be used by GDI.Graphics(Bitmap bmp,PlotDimensions dims...) to set the ScaleTransform of the GDI Graphics object. All calls to GDI.Graphics should then be changed to use the version of the function that requires the PlotDimensions. 

The Settings class would also be changes, to have two new optional properties - OutputWidth and OutputHeight. These would be used by Settings.GetPlotDimensions to configure the TranformScale of the PlotDimensions if needed. 

I have made these changes in my fork and they all seem to work, and they allow high resolution images to be created. I have tested the changes in the CookBook and they do not seem to break anything (at the moment the only impact is in the new Plot.SaveFig). But I have only just started looking at ScottPlot and I don't know if there are any other impacts of my proposed changes. There were still some calls to GDI.Graphics that I couldn't change to use the version that required the PlotDimensions, so these will not have the ScaleTransform applied - these seem mostly to be associated with measuring text. I am not sure if there will be problems with these. Ideally I would have liked to remove the GDI.Graphics function that didn't have the passed in PlotDimensions but I that would have increased the scope of my changes to even more classes/functions and I am not confident enough in my understanding of the library to to that.

[...]

PR #983, opened 2021-04-18
Add support for image scaling

PeterDavidson:
**Purpose:**
As discussed in #982 and #981, these changes allow Plot.SaveFig to adjust to the new output width and height without resizing the plot. This allows high resolution images to be saved without changing the size/width of any plot elements. This required a new property in PlotDimensions (TransformScale) and two new properties in Settings (OutputWidth and OutputHeight). It also required converting most GDI.Graphics calls to be adjusted to use the version that has access to the PlotDimensions.

**New Functionality:**
The answer given in #981 required that label sizes and widths needed to be adjusted to give high resolution images, as shown here.

```cs
var plt = new ScottPlot.Plot(3000, 2000);

var sig = plt.AddSignal(DataGen.Sin(51));
sig.LineWidth = 5;
sig.MarkerSize = 20;

plt.YLabel("Vertical Axis");
plt.XAxis.Label(size: 96);
plt.YAxis.Label(size: 96);

plt.XAxis.TickLabelStyle(fontSize: 36);
plt.YAxis.TickLabelStyle(fontSize: 36);

plt.XAxis.MajorGrid(lineWidth: 5);
plt.YAxis.MajorGrid(lineWidth: 5);
```

With the changes in the PR, the process is much simpler, and identical plots can be produced at different resolutions. 

```cs
var sig = plt.AddSignal(DataGen.Sin(51));
plt.YLabel("Vertical Axis");
plt.SaveFig("TestPlot_LowRes.png", 400, 300);
plt.SaveFig("TestPlot_Standard.png",800,600);
plt.SaveFig("TestPlot_HiRes.png",4000,3000);
```


PeterDavidson:
I just checked MatPlotLib to see how it deals with this, and rather than specifying an output width/height it allows a DPI setting to be given. See https://matplotlib.org/stable/api/_as_gen/matplotlib.pyplot.savefig.html. If you would prefer this I could change the SaveFig to work with a DPI setting instead.

src/ScottPlot/PlotDimensions.cs
```diff
@@ -14,6 +14,11 @@ public class PlotDimensions
         public readonly float DataHeight;
         public readonly float DataOffsetX;
         public readonly float DataOffsetY;
+        public float TransformScale { get; private set; } = 1;
+        public void SetOutputDimensions(int width, int height)
+        {
+            TransformScale = System.Math.Min(width / Width, height / Height);
```
bclehmann, line 20: I wonder if it would be more intuitive if there were a `TransformScaleX` and a `TransformScaleY`. I think if the user passes in values which do not preserve the aspect ratio we should not go out of our way to preserve the aspect ratio. Especially if doing so means we create an image of a different size than the user requested.

src/ScottPlot/Plot/Plot.Render.cs
```diff
@@ -172,10 +183,14 @@ public Bitmap RenderLegend(bool lowQuality = false)
         /// Save the plot as an image
         /// </summary>
         /// <param name="filePath">file path for the images (existing files will be overwritten)</param>
+        /// <param name="bitmapWidth">the width (in pixels) of the output bitmap</param>
```
bclehmann, line 186: `bitmapWidth` may not be the best parameter name as this method is not restricted to only saving to a bitmap image (or even bitmap-like formats such as PNG).

`imageWidth` or `outputWidth` probably makes the most sense.

src/ScottPlot/Plot/Plot.Render.cs
```diff
@@ -147,9 +157,10 @@ public void RenderUnlock()
         /// <param name="width">width (pixels) for this render</param>
         /// <param name="height">height (pixels) for this render</param>
         /// <param name="lowQuality">if true, anti-aliasing will be disabled for this render</param>
+        /// <param name="resizePlot">if true then the plot will be resized to fit the new dimensions. If false then the resolution will be changed without resizing. Default true</param>
```
bclehmann, line 160: The comment says default true, but the parameter is false by default.

src/ScottPlot/Plot/Plot.Render.cs
```diff
@@ -172,10 +183,14 @@ public Bitmap RenderLegend(bool lowQuality = false)
         /// Save the plot as an image
         /// </summary>
         /// <param name="filePath">file path for the images (existing files will be overwritten)</param>
+        /// <param name="bitmapWidth">the width (in pixels) of the output bitmap</param>
+        /// <param name="bitmapHeight">the width (in pixels) of the output bitmap</param>
+        /// <param name="lowQuality">if true, anti-aliasing will be disabled for this render. Default false</param>
+        /// <param name="resizePlot">if true then the plot will be resized to fit the new dimensions. If false then the resolution will be changed without resizing. Default true</param>
```
bclehmann, line 189: The comment says default true, but the parameter is false by default.

src/ScottPlot/Plot/Plot.Render.cs
```diff
@@ -15,14 +15,24 @@ partial class Plot
         /// </summary>
         /// <param name="bmp">an existing bitmap to render onto</param>
         /// <param name="lowQuality"></param>
+        /// <param name="resizePlot">if true then the plot will be resized to fit the new dimensions. If false then the resolution will be changed without resizing. Default true</param>
         /// <returns>the same bitmap that was passed in (but was rendered onto)</returns>
-        public Bitmap Render(Bitmap bmp, bool lowQuality = false)
+        public Bitmap Render(Bitmap bmp, bool lowQuality = false, bool resizePlot = true)
```
bclehmann, line 20: Is there a reason that `resizePlot` is true by default here, but not elsewhere?

src/ScottPlot/Drawing/GDI.cs
```diff
@@ -95,6 +95,8 @@ public static System.Drawing.Graphics Graphics(Bitmap bmp, bool lowQuality = fal
         public static System.Drawing.Graphics Graphics(Bitmap bmp, PlotDimensions dims, bool lowQuality = false, bool clipToDataArea = true)
         {
             Graphics gfx = Graphics(bmp, lowQuality);
+            gfx.ScaleTransform(dims.TransformScale, dims.TransformScale);
```
bclehmann, line 98: Very elegant, I wouldn't have thought to use a transform like this.

src/ScottPlot/Settings.cs
```diff
@@ -75,7 +78,12 @@ public PlotDimensions GetPlotDimensions(int xAxisIndex, int yAxisIndex)
             (double yMin, double yMax) = yAxis.Dims.RationalLimits();
             var limits = (xMin, xMax, yMin, yMax);
 
-            return new PlotDimensions(figureSize, dataSize, dataOffset, limits);
+            var dims = new PlotDimensions(figureSize, dataSize, dataOffset, limits);
+            if (OutputWidth.HasValue && OutputHeight.HasValue)
```
bclehmann, line 82: If we allow the user to pass in values that change the aspect ratio we don't need both of these to be non-null, only one. Might be easiest to let `SetOutputDimensions` take nullable values and have it deal with it.

bclehmann (commented):
Very clean overall, the main thing is whether we should let the user squish/stretch the image into a different aspect ratio. Other than that it's mostly boring stuff like parameter names and comments.

bclehmann:
It would also be useful to be able to set the size without rendering, for interactive plots.

bclehmann:
From #982: 

> There were still some calls to GDI.Graphics that I couldn't change to use the version that required the PlotDimensions, so these will not have the ScaleTransform applied - these seem mostly to be associated with measuring text.

Measuring text is commonly used in ScottPlot to allocate space for text so that it doesn't overflow onto the plotting area or get obscured by axes. I think the best solution is to move the resolution setting up to the `GDI` class, seeing as it's not really connected to the dimensions of any one plot. This has already been done for DPI-correcting mouse inputs, and similar things.

PeterDavidson:
Thanks for the fast response @bclehmann and for your helpful comments. I will make the naming and comment changes that you suggested. I was worried about having separate transforms in x and y, since it would stretch elements (fonts, circles etc) in a way that might be surprising to users. The way I have it there will just be whitespace on the plot if they specify and output dimension that is not compatible with the input dimension. The other alternative (as mentioned above for matplotlib) is to pass in a single DPI parameter, but this would require an understanding of what the default DPI of the base plot is. The TransformScale could then just be the ratio of the specified DPI to the default DPI. 

I wasn't sure what to do for interactive plots because I haven't looked at them yet (I'm mainly interested in generating static outputs at the moment) and I didn't want to break anything. Happy to take a suggestion on what should change here.

With respect to your comment on measuring text I'm not sure what you mean about moving the resolution setting up to the GDI class. As I understand it GDI is a static class so can't contain these sort of settings.

bclehmann:
> With respect to your comment on measuring text I'm not sure what you mean about moving the resolution setting up to the GDI class. As I understand it GDI is a static class so can't contain these sort of settings.

Yeah, I think I edited my response right as you replied. I think you could move the methods for creating `System.Drawing.Bitmap` objects into a separate class or remove the static modifier from `GDI`, although I am hesitant about such a move.

PeterDavidson:
Looking through the library one of the things that struck me was that the Render function has a passed in bitmap, even though (as far as I could see) it only uses it to pass into GDI.Graphics. It would seem clearer to me if the Render function only used a passed in Graphics, as then all of the Plottables could be used to draw directly onto a graphics context, not only a bitmap. This would make it easier to apply transforms on a Graphics context when establishing a render pipeline and then it would automatically be used by all Plottables.

PeterDavidson:
I also have found when combining the Circle code with this high DPI code described here that the HashBrush stuff doesn't work properly with the resolution scaling, since HashBrushes are pixel based. I would like to make another change to GDI to get it to return a TextureBrush rather than a HashBrush if the output resolution has changed, but this would require PlotDimensions to be passed into GDI.Brush. This would require changing lots of files so I thought I'd check before I add it to the PR.

src/ScottPlot/Drawing/GDI.cs
swharden: Same! With all my exposure to System.Drawing I previously overlooked this method. Lots of good potential here

https://docs.microsoft.com/en-us/dotnet/api/system.drawing.graphics.scaletransform?view=net-5.0

swharden:
Thanks so much for this PR @PeterDavidson! I'm already finding this useful not just for high DPI images, but also for making thumbnails. This is a fantastic improvement!

Here are a few changes I made:
* I changed the language/behavior away from resizing or not to _scaling_
* The scale factor is now passed into the Render and SaveFig methods but never saved (it's stateless)
* Argument docstrings have been rewritten to better reflect what's going on: if you pass dimensions into SaveFig(), the plot is resized by those dimensions, then the output figure is scaled by the scale factor.
* This configuration gets around the state issues I hit calling the old code multiple times, and also gets around X/Y ratio issues touched on by @bclehmann 

Future improvements:
* As @bclehmann noted, these concepts could be extended to the user controls to allow comfortable mouse-interactive plots in high DPI environments. This could be a can of worms though, so I'll save this topic for another day/PR

I'll touch this up a bit more today, but hope to merge/release this weekend.

Issue #1032, opened 2021-05-17
Link vertical draggable axis


EmanuelFeru:
### Your Environment
- ScottPlot Version: 4.1.13 Beta
- Operating System: W10
- GUI app WinForms

**Questions:**
* How to link 2 vertical lines in 2 different plots?
* How to dynamically display the X coordinate (ideally I want the current `index` in the data and display the Y data value)?

**What I want to achieve:**
![image](https://user-images.githubusercontent.com/24780745/118472074-bb01a280-b708-11eb-9d23-0505b96c1847.png)

EmanuelFeru:
I implemented the code below using `.Dragged` Handle
Positive:
* it works, the values are updated correctly

Negative:
* Because I had to use `formsPlot2.Render();` The line for the 2nd plot suffer responsiveness, which means it lags behind the 1st line that drives it.
* If I don't use `formsPlot2.Render();`, then the 2nd plot is not updated and Red line does not move until I interract with the 2nd plot.

```cs
        ScottPlot.Plottable.VLine hLine1;
        ScottPlot.Plottable.VLine hLine2;

        // Add Draggable Vertical Axis Line
        hLine1 = formsPlot1.Plot.AddVerticalLine(1000, color: Color.Red);
        hLine1.DragEnabled = true;
        hLine1.Dragged += OnAxisDragged;

        hLine2 = formsPlot2.Plot.AddVerticalLine(1000, color: Color.Red);
        hLine2.DragEnabled = true;

        private void OnAxisDragged(object sender, EventArgs e)
        {
            label1.Text = "Line position = " + hLine1.X.ToString();
            hLine2.DragTo(hLine1.X, 0, true);
            formsPlot2.Render();
        }
```

StendProg:
Hi @EmanuelFeru,

> I implemented the code below using .Dragged Event Handler

I would advise you to make the next replacement so that the 2nd Plot does not trigger an extra event when it is not needed.
```c#
    label1.Text = "Line position = " + hLine1.X.ToString();

    // hLine2.DragTo(hLine1.X, 0, true);
    hLine2.X = hLine1.X;

    formsPlot2.Render();
```

You can try `formsPlot2.Render(skipIfCurrentlyRendering: true)` to reduce lags.

> Is it possible to only Render the vertical line

Actually not, I've experimented with this, but it hasn't been updated for a long time. #687

[...]

Issue #1028, opened 2021-05-17
Triaged Tasks and Features


swharden:
**This is a collection of tasks and features I intend to focus on** in the order I plan to work on them. I desire to accomplish everything eventually, but this list helps me to arrange near-term priorities. Previous triage issues: #651, #412.

<details open>
<summary><h2>Primary Focus (June, 2022)</h2></summary>

* #1888

</details>

<details>
<summary><h2>Primary Focus (March, 2022)</h2></summary>

* ScottPlot 4.1 is stable. Focus on maintenance and bug fixes (do not aggressively add features).

* ScottPlot 5.0 is early in development. Wait for `Microsoft.Maui.Graphics` to mature and stabalize before aggressively resuming development. See comments in #1647 for considerations and next steps.

* As ScottPlot 5 matures, focus on improving the website and documentation

</details>

<details>
<summary><h2>Primary Focus (February, 2022)</h2></summary>

* I enacted a soft feature freeze to allow me to focus on ScottPlot 5
  * #1624
  * #1647

</details>

<details>
<summary><h2>Primary Focus (November, 2021)</h2></summary>

* **Maintenance:** Rapidly address or close issues/PRs to concentrate effort on these high priority goals

* **Abstract the Rendering System:** Work toward #1036 while [Microsoft.Maui.Graphics](https://github.com/dotnet/Microsoft.Maui.Graphics) continues to mature (release targeted for Q2, 2022)

* **Refactor TickCollection** to improve performance, reliability, and add support for advanced tick features:
  * #725
  * #925
  * #926 
  * #475 
  * #469 
  * #935 
  * #648 
  * #644 
  * #926 6 
  * #1070 
  * #926
</details>

<details>
<summary><h2>Triaged Tasks and Features</h2></summary>

_This list will be reviewed and re-categorized as the primary focus tasks above have are accomplished_

### Get ScottPlot 4.1 out of Beta
- [x] Modernize demo application (#887)
- [x] Fix mouse events and make enter/leave demo (#999)
- [x] Fix offset and multiplier notation (#1026)
- [x] Improve design-time appearance of WpfPlot (#1043)
- [x] Improve support and documentation for queued rendering (#1034)
- [x] set up DocFX for API documentation (#1047)
- [x] Update documents to describe ScottPlot 4.1 (#949, #966)
- [x] add FAQ page describing 4.0 to 4.1 changes ([link](https://swharden.com/scottplot/faq/version-4.1/))
[...]

PR #1034, opened 2021-05-17
Add RenderRequest() to controls for non-blocking rendering

StendProg:
**Purpose:**
It is useful enough to provide the user with the ability to manually update Plot using the RenderQueue. #1032
This PR redirects all available rendering methods from the backend to controls.
- RenderLowQuality()
- RenderHighQuality()
- RenderDelayedHighQuality()
- RenderLowThenImmediateHighQuality() - already present

Avalonia control was left without these possibilities, I'm afraid to break something there because I have no experience.

**New Functionality:**

```cs
formsPlot1.RenderHighQuality();
formsPlot1.RenderLowQuality();
formsPlot1.RenderDelayedHighQuality();
```

### DIFF
--- a/src/ScottPlot/Control/Backend.cs
+++ b/src/ScottPlot/Control/Backend.cs
@@ -344,31 +344,41 @@ namespace ScottPlot.Control
             currentlyRendering = false;
         }
 
-        public void RenderLowQuality()
-        {
-            ProcessEvent(EventFactory.CreateManualLowQualityRender());
-        }
-
-        public void RenderHighQuality()
-        {
-            ProcessEvent(EventFactory.CreateManualHighQualityRender());
-        }
-
         /// <summary>
-        /// Render the plot using low quality (fast) then immediate re-render using high quality (slower).
+        /// Request a render using the render queue. 
+        /// This method does not block the calling thread.
         /// </summary>
-        public void RenderLowThenImmediateHighQuality()
+        public void RenderRequest(RenderType renderType)
         {
-            ProcessEvent(EventFactory.CreateManualLowQualityRender());
-            ProcessEvent(EventFactory.CreateManualHighQualityRender());
-        }
+            switch (renderType)
+            {
+                case RenderType.LQOnly:
+                    ProcessEvent(EventFactory.CreateManualLowQualityRender());
+                    return;
 
-        /// <summary>
-        /// Render the plot using low quality (fast) then delayed re-render using high quality (slower).
-        /// </summary>
-        public void RenderDelayedHighQuality()
-        {
-            ProcessEvent(EventFactory.CreateManualDelayedHighQualityRender());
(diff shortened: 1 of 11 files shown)


swharden:
Hi @StendProg, 

Improving discoverability and documentation of the render queue is one of my upcoming goals in the triage list (#1028)! I really want to make this feature more discoverable and easier to understand for new users.

My concern with this PR is that new users may get confused about how these are different:
```cs
formsPlot1.Render(lowQuality: true);
formsPlot1.RenderLowQuality();
```

### Name Async() to improve queue discoverability?

What do you think about something like this? It doesn't return a `Task` and it isn't even a real `async` method (so maybe this is bad)... but having Async() in the function name makes it more obvious to the user that the render may occur later and that the call is non-blocking.  XML documentation for the method and the enum could eplain the details.

```cs
public void RenderAsync(AsyncRenderQuality quality = AsyncRenderQuality.High) {
    quality switch {
        AsyncRenderQuality.Low => Backend.Backend.RenderLowQuality(),
        AsyncRenderQuality.High=> Backend.Backend.RenderLowQuality(),
        AsyncRenderQuality.LowHigh => Backend.Backend.RenderLowQuality(),
        AsyncRenderQuality.LowDelayHigh => Backend.Backend.RenderLowQuality(),
    };
}
```

### Method to toggle render queue?

While we're here, we may want to make it easier to discover how to enable the render queue. The advantage here is that users looking for this type of behavior will immediately know what "async" means (even though they may not think to look for the word queue).

```cs
/// <summary>
/// Control whether asynchronous rendering is used for future renders.
/// Asynchronous render calls are non-blocking, but slightly less responsive to mouse events.
/// </summary>
public void RenderAsync(bool enable = true, bool renderNow = true) {
    Configuration.UseRenderQueue = true;
    if (renderNow)
        Render();
}
```

Again, I haven't thought about this in detail, and I'm not an async/await expert... but I'm very interested to hear what you think of these potential designs. Thanks for your input!

Scott

swharden:
@StendProg thanks for your input! I took your advice and arrived at a solution I feel good about:

All user controls now have a `RenderRequest()` method that accepts a `RenderType` enum.

```cs
formsPlot1.RenderRequest(RenderType.LowQualityThenHighQualityDelayed)
```
XML docs indicate this method is non-blocking, and the enum items all have XML docs that describe what they do and how they are commonly used. If you have any suggestions let me know, otherwise I'll merge this shortly 👍

StendProg:
I add default renderType value = `RenderType.LowQualityThenHighQualityDelayed`
I think it is suitable for 95% of cases and will allow to make a simple call `formsplot.RenderRequest();`

swharden:
Thank you for all your work on this @StendProg! The render queue is a very nice system, and I'm happy it is now easier to find and use 👍

Issue #827, opened 2021-02-24
Improve in-place updates for finance plots


swharden:
In a discussion raised by @ticool (#826) we found there's no way to easily edit the last OHLC's values after a candlestick plot has been made (especially if `Add()` was used to add it). This function would be extremely useful for displaying real-time price data, where the last candle continuously updates.

Perhaps the OHLC module should have a way to expand its `Low` and `High` based on new price information. I'll continue to think about how to best name this.

```cs
public Expand(double newPrice)
{
    Low = Math.Min(Low, newPrice);
    High = Math.Max(High, newPrice);
}
```

There should be an easy way for the user to be able to access the last OHLC so its values can be adjusted. Perhaps simply this:

```cs
public OHLC Last() => OHLCs.Last();
```

I'm reminded the OHLC list is private

https://github.com/ScottPlot/ScottPlot/blob/17bdbb53ea8d47c8ddd033836aeaaae5bef1838c/src/ScottPlot/Plottable/FinancePlot.cs#L9-L14

swharden:
This topic will be tracked in the latest triage issue

PR #1038, opened 2021-05-19
Allow realtime updates of OHLC plots

CalderWhite:
**Purpose:**
I added in your suggested change from #827. Have been building it locally with my own project that uses ScottPlot and it works great!

**New Functionality:**
Adds the following to `FinancePlot.cs`:
```cs
public OHLC Last() => OHLCs.Last();
```

This allows users to mutate the last item of the list, allowing for realtime plotting updates.

**Notes:**

Although, I am unable to horizontally zoom the plot. Maybe the key bindings from that type of zoom have been changed in the beta, but that is unrelated.

I don't think I am being much help here since you already proposed the change, but at the very least I am using it and it works!

Thanks for the project, it's one of the few C# graphing libraries that has what I need and doesn't make me pay for it :)

### DIFF
--- a/dev/changelog.md
+++ b/dev/changelog.md
@@ -4,6 +4,7 @@
 * Add support for custom linestyles in SignalXY plots (#1017, #1016) _Thanks @StendProg and @breakwinz_
 * Support Avalonia 0.10.4 (#1018) _Thanks @bclehmann_
 * Controls now properly process `MouseEnter` and `MouseLeave` events (#999) _Thanks @kirsan31 and @breakwinz_
+* Added Last() to finance plots to make it easier to access the final OHLC (#1038) _Thanks @CalderWhite_
 
 ## ScottPlot 4.1.13-beta
 * `Plot.Render()` and `Plot.SaveFig()` now have a `scale` argument to allow for the creation of high resolution scaled plots (#983, #982, #981) _Thanks @PeterDavidson_
--- a/src/ScottPlot/Plottable/FinancePlot.cs
+++ b/src/ScottPlot/Plottable/FinancePlot.cs
@@ -12,6 +12,10 @@ namespace ScottPlot.Plottable
     public class FinancePlot : IPlottable
     {
         private readonly List<OHLC> OHLCs = new List<OHLC>();
+        
+        // <summary>
+        // Returns the last element of OHLCs so users can modify FinancePlots in real time.
+        // </summary>
         public OHLC Last() => OHLCs.Last();
 
         /// <summary>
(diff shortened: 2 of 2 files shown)


CalderWhite:
I think I haven't touched enough files. I am going to close this because of that.

swharden:
Hi @CalderWhite,

I configured the build to fail if public properties/methods of plottables aren't documented with a `<summary>`. I recognize the error message is a bit cryptic and hard to get to. 

image

The feature in this PR looks useful! If you re-open this PR and add a summary so the build passes, I'd be happy to merge it 👍

CalderWhite:
@swharden ah! thank you so much! I'll get right on it.

swharden:
Sorry this one-liner turned-out to be so difficult! 😝

I suspect the code I wrote that asserts public methods are documented is getting stuck on this public _property_. I'll fix this shortly and merge this in 👍

swharden:
In the original PR the XML docs were preceded by `//` instead of `///` so the summary wasn't being recognized.
https://docs.microsoft.com/en-us/dotnet/csharp/programming-guide/xmldoc/

I added the extra slashes and it seems to work now. This new feature will be in the next release I hope to put on NuGet in the next few days 🤞 

Thanks for the contribution @CalderWhite!

CalderWhite:
Woo! If I do any more hacking I'll be sure to put up a PR :)

swharden:
> Woo! If I do any more hacking I'll be sure to put up a PR :)

Sounds good! Sometimes what seems like a good idea as a developer is actually awkward for the user, so if you have any more ideas for how to smooth out the rough edges I welcome your feedback! 🚀

ticool:
Hi im trying to use the  .Last();  by setting my new value (Hig, Low, etc) in this variable in my program loop !

OHLC lastPlot = CandlePlots.Last(); 
lastPlot.Low = 1234;
lastPlot.High= 4567;
.....

Im doing a  formsPlot1.Render() after, I tried formsPlot1.Refresh() too but the last candle doesn't refresh ! How do you use this ? Do you get an example plz !

swharden:
Hi @ticool, 

It looks like `Last()` is a public method in the current version (4.1.16) of ScottPlot
https://www.fuget.org/packages/ScottPlot/4.1.16/lib/netstandard2.0/ScottPlot.dll/ScottPlot.Plottable/FinancePlot

Probably confirm you're running the latest public version and upgrade the package if a newer one is available, and let me know if this doesn't do the trick 👍

swharden:
Hi @ticool, sorry about misunderstanding your question a moment ago.

After modifying the last `OHLC`, call `formsPlot1.Render()` to redraw that plot on the screen. Let me know if that doesn't do the trick! 👍

Issue #999, opened 2021-04-29
Improve mouse interactions in controls


kirsan31:
# Tasks
_This section was added by @swharden_
- [x] trigger mouse enter and leave events in user controls (#1040)
- [x] provide options for automatic axis limits to ignore axis lines and spans (#1083)
- [x] implement an axis crosshair label feature (#1093)

# Original Question

### Your Environment
- ScottPlot Version: 4.1.12 beta
- Operating System: Win10 20H2
- Is it a GUI app? What platform? WinForms
- .Net:  `net472` and `net5.0-windows` was checked.

Simple Mdi Winforms app. Child form with only formsPlot1 on it. `MouseLeave` and `MouseEnter` not working, `MouseMove` is working.

This is critical to implement some info on `MouseMove` (to hide it after we leave the plot).

P.s. Related question:
How to properly implement axes cursor on `MouseMove` with ScottPlot?
![image](https://user-images.githubusercontent.com/17767561/116544741-32a59400-a8f8-11eb-90bc-fe43cdb62cfe.png)

I was tring something like this:
```cs
private void formsPlot1_MouseMove(object sender, MouseEventArgs e)
{
    (double x, double y) = formsPlot1.GetMouseCoordinates();
    _HLine.Y = y;
    _HLine.IsVisible = true;
    _VLine.X = x;
    _VLine.IsVisible = true;
    formsPlot1.Render(skipIfCurrentlyRendering: true);
}
```
And it's worked BUT. Because of `_HLine` and `_VLine` are plots, ScottPlot autozoom and autoaxes always trying to consider their values. And this lead to whole wrong behavior. For example:

https://user-images.githubusercontent.com/17767561/116545566-47cef280-a8f9-11eb-9c29-ee21ead7fceb.mp4

Cursors must be completely excluded from plotting logic.

Also, is it possible to display Cursor coordinates on the axes (red circles on screen above)?

StendProg:
Hi @kirsan31,

1. #892 is related issue. But it doesn't solve your specific problem.
But it doesn't solve your specific problem.
I think the whole point is that the `PictureBox` rises all events correctly, including `MouseLeave` and `MouseEnter`, but they are not redirected to the Control in any way. And `PictureBox` is in private. 
The solution would be either to provide the user with access to the `PictureBox`, or to redirect all his events to the control (the problem is that there are too many of them).

[...]

PR #1040, opened 2021-05-19
Improve mouse interactions in Avalonia control

bclehmann:
**New Contributors:**

**Purpose:**
https://github.com/ScottPlot/ScottPlot/issues/999#issuecomment-843719638

**New Functionality:**
Forwards mouse moved events to the parent object in Avalonia.

### DIFF
--- a/src/controls/ScottPlot.Avalonia/AvaPlot.axaml.cs
+++ b/src/controls/ScottPlot.Avalonia/AvaPlot.axaml.cs
@@ -32,7 +32,16 @@ namespace ScottPlot.Avalonia
     {
         public Plot Plot => Backend.Plot;
         public ScottPlot.Control.Configuration Configuration => Backend.Configuration;
+
+        /// <summary>
+        /// This event is invoked any time the axis limits are modified.
+        /// </summary>
         public event EventHandler AxesChanged;
+
+        /// <summary>
+        /// This event is invoked any time the plot is right-clicked.
+        /// By default it contains DefaultRightClickEvent(), but you can remove this and add your own method.
+        /// </summary>
         public event EventHandler RightClicked;
 
         private readonly Control.ControlBackEnd Backend;
@@ -58,8 +67,6 @@ namespace ScottPlot.Avalonia
                 [ScottPlot.Cursor.Question] = new Ava.Input.Cursor(StandardCursorType.Help),
             };
 
-            // TODO: delete this?
-            //SetContextMenu(backend.DefaultRightClickMenu());
             Backend = new ScottPlot.Control.ControlBackEnd((float)this.Bounds.Width, (float)this.Bounds.Height);
             Backend.BitmapChanged += new EventHandler(OnBitmapChanged);
             Backend.BitmapUpdated += new EventHandler(OnBitmapUpdated);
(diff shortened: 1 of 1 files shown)


bclehmann:
As an aside, there is still quite a bit of functionality commented out in AvaPlot.axaml.cs. It fits mostly into these categories:
- Dead or no longer necessary code, perhaps from a previous AvaPlot implementation or debugging code
- Code for features Avalonia does not/did not support when it was written
    - Copying images to clipboard I believe is still quite difficult
    - I don't know if Avalonia supports changing the cursor yet
- Code for features that I never added but probably should've
    - I still am unsure what we're doing about context menus on the new common backend
    - Detecting designer mode was removed in order to get the very first version working and then never revisited

Let me know your opinion.

bclehmann:
It looks like Avalonia does support changing the cursor on `InputElements`, which is one of the base types of `UserControl`.

src/controls/ScottPlot.Avalonia/AvaPlot.axaml.cs
```diff
@@ -46,11 +46,24 @@ public partial class AvaPlot : UserControl
         public AvaPlot()
         {
             InitializeComponent();
+
```
bclehmann, line 58: I have never seen this syntax, I like it, very clean.

swharden:
Thanks for this PR and the notes about the commented-out blocks! I was about to list-out the changes I made but it looks like you're already on it.

Regarding right-click, controls respond to right-clicks by invoking whatever is in `RightClicked`. This lets users do things other than deploy context menus. I added XML docs to make this a little easier to figure out.

Let me know if all these changes look good to you! If so I'll merge now. Thanks again for all your Avalonia input! 👍 🚀

bclehmann:
> Let me know if all these changes look good to you! If so I'll merge now. Thanks again for all your Avalonia input! 👍 🚀

All looks good, thanks for taking this up!

bclehmann:
I made it so it does not render in the designer view:
image

It looks like this no longer works in WPF or on WinForms, was this an intentional choice or a regression?

swharden:
> was this an intentional choice or a regression

I removed it in https://github.com/ScottPlot/ScottPlot/pull/1040/commits/99c786fc7c23726f20370ce21bf86c3311f4e1a9 because `bool isDesignerMode = false;` was hard-coded so I didn't think it worked

I didn't realize until now Avalonia even had a designer mode until now (it doesn't work out of the box on my system)

I'd like to move toward the design time display I'm using in WinForms (#1043), but I'm really scared about difficult to diagnose design-time crashes. Do you think this can be done safely in a way that falls back to something simple? I'll try to implement something similar for WPF shortly

bclehmann:
> I didn't realize until now Avalonia even had a designer mode

I think it might have been this extension: https://marketplace.visualstudio.com/items?itemName=AvaloniaTeam.AvaloniaforVisualStudio

> Do you think this can be done safely in a way that falls back to something simple? I'll try to implement something similar for WPF shortly

By and large, if it can be done in WPF it can be done in Avalonia the same way (although sometimes with a bit of extra work). As for doing it in WPF, perhaps if it detects designer mode it enables a massive try-catch block that swallows everything (like how webservers stay alive and just send an HTTP 500 if there's an unhandled exception).

Now that I think about it, it may be easier to always have the massive try-catch block, but rethrow if not in the designer, and handle it in the designer. I don't think this is likely to be worth the extra effort in debugging. Perhaps when in designer mode it might be best to use an entirely different code path.

Or simply accept that designer crashes may happen, I think most of the time when these issues occur it is related to .csproj issues, so I don't know how much we're increasing the surface.

swharden:
> I think it might have been this extension

Ah, that makes sense

> massive try-catch block ... I don't think this is likely to be worth the extra effort in debugging ... or simply accept that designer crashes may happen

I'll research this topic more as it relates to the WPF control and note my findings in #1043, and I'm happy to leave things as they are now for the Avalonia control. I'll go ahead and merge this is, and if/when you're inspired to make a fancier designer view for the Avalonia control you can choose the implementation you feel best about. 

Thanks again for this PR! This one-liner turned into a solid improvement 👍

PR #1041, opened 2021-05-19
Avalonia Version 0.10.4 -> 0.10.5

bclehmann:
**New Contributors:**

**Purpose:**
Bumps Avalonia versions to 0.10.5

**New Functionality:**
N/A

### DIFF
--- a/src/controls/ScottPlot.Avalonia/ScottPlot.Avalonia.NUGET.csproj
+++ b/src/controls/ScottPlot.Avalonia/ScottPlot.Avalonia.NUGET.csproj
@@ -33,8 +33,8 @@ Release demo: http://swharden.com/scottplot/demo</PackageReleaseNotes>
     <None Include="icon.png" Pack="true" PackagePath="\" />
     <None Include="icon.ico" Pack="true" PackagePath="\" />
     <None Include="README.txt" Pack="true" PackagePath="\" />
-    <PackageReference Include="Avalonia" Version="0.10.4" />
-    <PackageReference Include="Avalonia.Desktop" Version="0.10.4" />
+    <PackageReference Include="Avalonia" Version="0.10.5" />
+    <PackageReference Include="Avalonia.Desktop" Version="0.10.5" />
   </ItemGroup>
   
 </Project>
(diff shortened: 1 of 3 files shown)


swharden:
I'm happy to continue merging these, but I'm curious - are they necessary? Note that this indicates users must have `0.10.11` as a minimum, but they can upgrade on their end at any time (as long as the new package has a non-breaking API).
```xml
<PackageReference Include="Avalonia" Version="0.10.11" />
```

### Minimum Supported Version

Every time we bump the minimum version for AvaPlot, all users are forced to update the version in their application, but it we choose not to bump we leave that option up to the user. Similar to ScottPlot deciding to support back to .NET Framework 4.6.1, would it be useful to define/maintain a minimum supported AvaPlot version and keep targeting that? 

### Wildcard / Range Versioning

... and/or, we could use range versions. This could ensure that tests/builds continue to pass using the latest version without requiring modification to the csproj files when new Avalonia NuGet packages become available.
https://docs.microsoft.com/en-us/nuget/concepts/package-versioning#references-in-project-files-packagereference
```xml
<PackageReference Include="Avalonia" Version="0.10.*" />
```

bclehmann:
> Wildcard / Range Versioning

Oooh, I hadn't known that NuGet supported this. It doesn't look like they support the caret (i.e. `^1.1.1` means 1.1.1 and after so long as the major version doesn't change), which is what we would normally use in this situation.

I think `0.10.*` is the best option. Technically a minor release isn't supposed to break backwards compatibility, but Avalonia does sometimes break backwards compatibility in minor releases. You could argue that they're not true minor releases because it's still an 0.x release, but either way, 0.11.0 or 1.0.0, whichever comes first, are both reasonably likely to break something.

swharden:
> It doesn't look like they support the caret (i.e. ^1.1.1 means 1.1.1 and after so long as the major version doesn't change), which is what we would normally use in this situation.

I think this will allow anything above `0.9.0` but disallow `1.0.0` and higher

```xml
<PackageReference Include="Avalonia" Version="[0.9, 1.0)" />
```

It will prevent us from having to bump `0.10.*` to `0.11.*` when that time comes. What do you think?

bclehmann:
> It will prevent us from having to bump 0.10.* to 0.11.* when that time comes. What do you think?

I think there's a decent chance we need to make small changes for 0.11.0, given we had to do that for 0.10.0

Issue #1043, opened 2021-05-20
Improve appearance of controls in designer mode


swharden:
I think the best thing to display is an empty plot with a title that shows the ScottPlot version. 

This is currently implemented in the WinForms control, but not the WPF or Avalonia controls

Careful though! A crash in designer mode produces bugs which are difficult to solve (#161, #356), so exception handling and a fallback design may be a good idea.

WinForms | WPF | Avalonia
---|---|---
![image](https://user-images.githubusercontent.com/4165489/118904648-4edd9500-b8e8-11eb-8c2e-932d8eb60975.png)|![image](https://user-images.githubusercontent.com/4165489/118904694-6288fb80-b8e8-11eb-83db-810c2d421e27.png)|![image](https://user-images.githubusercontent.com/8635304/118899507-1a5be000-b8cc-11eb-9a42-c74af17c0bce.png)

swharden:
FormsPlot and WpfPlot now both show plots at design time.

If there is a problem with `System.Drawing.Common`, it may crash Visual Studio ☠️ 

https://user-images.githubusercontent.com/4165489/119075793-650a5480-b9bf-11eb-85e1-65515e309f31.mp4

Issue #993, opened 2021-04-26
could not load file or assembly 'System.Drawing.Common, Version=4.0.0.1, ...'


humaneer822:
### Your Environment
- ScottPlot Version: 4.0.48
- Operating System: Windows 10 64bit, Visual Studio 2019
- Is it a GUI app? What platform?
    - Platforms include WinForms, OpenCVsharp4

Describe the bug and how we can reproduce it.
Hello, I see this error message, I have not had this error before. 
If I ignore this error, I don't see any formsplots in winform designer neither add new forms plot.
How can I solve this problem?

![image](https://user-images.githubusercontent.com/76927022/116020015-68404800-a680-11eb-8317-82248b5cb813.png)

![image](https://user-images.githubusercontent.com/76927022/116020724-da655c80-a681-11eb-9ec3-37d5a2aa1032.png)

like following screenshot, there's no error in build & debugging/execution.
![image](https://user-images.githubusercontent.com/76927022/116026891-90cf3e80-a68e-11eb-9544-4d2414697f56.png)

// a code sample may improve communication
```

StendProg:
Hi @humaneer822,

A similar problem has already been discussed #924, maybe you can find a solution there.

bclehmann:
In general, this error can come from mixed up dependencies. For example, having ScottPlot.WinForms 4.1 but ScottPlot 4.0 or vice versa.

I'd note that OpenCVSharp4 depends on System.Drawing.Common >= 5.0.2, which is a later version than what ScottPlot 4.0 requires (and a lesser version than what ScottPlot 4.1 requires). I would check that that dependency in particular makes sense.

bukkideme:
I also started to have problems. Yep, Syste.Drawing.Common wants to update itself from 4.7 to 5.0.2. 
Somehow these different versioning start to make troubles in my VS. Error msg, etc.

bukkideme:
FYI:
Uninstalled System.Drawing.Common, and ScottPlot.Winform (plus related) packages, then re-installed ScottPlot.Winform plus dependencies. Finally I let System.Drawing update itself from nuget packages. This solved the problem.

bukkideme:
Unfortunately this fix did not last long. I closed VS2019, then re-opened it. I cannot open the Form1.cs designer, I get error msg see below. I guess ScottPlot requires System.Drawing.Common 4.0.0.1, and the 5.0.2 update broke things?

[...]

PR #1048, opened 2021-05-21
WpfPlot: Show plot in design mode

swharden:
addresses #1043

### DIFF
--- a/src/ScottPlot/Drawing/GDI.cs
+++ b/src/ScottPlot/Drawing/GDI.cs
@@ -19,6 +19,34 @@ namespace ScottPlot.Drawing
         private const float yMultiplierMacOS = 27.16f / 20;
 
         /// <summary>
+        /// Test if the system can properly render graphics using System.Drawing.Common.
+        /// If an exception occurs this will return a string description, otherwise it returns null.
+        /// </summary>
+        public static string DrawingTest()
+        {
+            try
+            {
+                var bmp = new Bitmap(10, 10);
+                using var gfx = GDI.Graphics(bmp);
+                gfx.Clear(Color.Magenta);
+                gfx.DrawLine(Pens.Black, 0, 0, 10, 10);
+
+                using var stream = new System.IO.MemoryStream();
+                bmp.Save(stream, ImageFormat.Bmp);
+                byte[] bytes = stream.ToArray();
+                int expectedBitmapSize = 454;
+                if (bytes.Length != expectedBitmapSize)
+                    throw new InvalidOperationException($"test image was {bytes.Length} bytes (expected {expectedBitmapSize})");
+            }
+            catch (Exception ex)
+            {
+                return ex.ToString();
+            }
+
+            return null;
+        }
+
+        /// <summary>
         /// Return the display scale ratio being used.
         /// A scaling ratio of 1.0 means scaling is not active.
         /// </summary>
--- a/src/ScottPlot/Drawing/GDI.cs
+++ b/src/ScottPlot/Drawing/GDI.cs
@@ -40,7 +40,7 @@ namespace ScottPlot.Drawing
             }
             catch (Exception ex)
             {
-                return ex.ToString();
+                return "ScottPlot Render Error:" + Environment.NewLine + ex.ToString();
             }
 
             return null;
--- a/dev/changelog.md
+++ b/dev/changelog.md
@@ -6,6 +6,7 @@
 * Controls now properly process `MouseEnter` and `MouseLeave` events (#999) _Thanks @kirsan31 and @breakwinz_
 * Controls now have a `RenderRequest()` method that uses a render queue to facilitate non-blocking render calls (#813, #1034) _Thanks @StendProg_
 * Added Last() to finance plots to make it easier to access the final OHLC (#1038) _Thanks @CalderWhite_
+* Controls that fail to render in design mode now display the error message in a textbox to prevent Visual Studio exceptions (#1048)
 
 ## ScottPlot 4.1.13-beta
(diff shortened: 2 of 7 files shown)


swharden:
FormsPlot and WpfPlot now both show plots at design time.

If there is a problem with `System.Drawing.Common`, it may crash Visual Studio ☠️ 

https://user-images.githubusercontent.com/4165489/119075793-650a5480-b9bf-11eb-85e1-65515e309f31.mp4

swharden:
I added a render test inside a try/catch block that should identify render errors and display them as text in the control at design time
image

swharden:
WpfPlot and FormsPlot now do this as the first thing in their initializers:

```cs
string renderErrorMessage = ScottPlot.Drawing.GDI.DrawingTest();
if (renderErrorMessage != null)
{
    InitializeComponent();
    /* display the error message in a textbox */
    return;
}
```

@bclehmann do you think this is worth adding to AvaPlot?

src/ScottPlot/Drawing/GDI.cs
```diff
@@ -18,6 +18,34 @@ public static class GDI
...
+                using var gfx = GDI.Graphics(bmp);
+                gfx.Clear(Color.Magenta);
+                gfx.DrawLine(Pens.Black, 0, 0, 10, 10);
+
+                using var stream = new System.IO.MemoryStream();
+                bmp.Save(stream, ImageFormat.Bmp);
+                byte[] bytes = stream.ToArray();
+                int expectedBitmapSize = 454;
+                if (bytes.Length != expectedBitmapSize)
+                    throw new InvalidOperationException($"test image was {bytes.Length} bytes (expected {expectedBitmapSize})");
```
bclehmann, line 39: Throwing something in order to immediately catch it is something of an antipattern. It's using exception handling as flow control instead of an if/else.

This method should probably either return a boolean to indicate success/failure, or just let any exceptions bubble up to the caller.

If the actual error message is important then it should throw an exception which is caught by the caller and the error message surfaced to the user. Alternatively, you can return false to indicate failure and then set an error field somewhere (like in many C APIs), or return a struct with a success field set to false and some extra information (like in many AJAX libraries).

I don't normally like the C way of doing things because it's used everywhere to get around the lack of exception support, however it does have some benefits:
- It does not obligate the caller to handle any exceptions
    - Thus it's a pattern often used in AJAX when it doesn't matter if the request fails (or when failures are already handled by some middleware which will retry)
- Performance and compatibility with embedded systems (not really a consideration here)

bclehmann: Returning a struct like this would probably be the best option, provided you don't want to let the exception bubble up to the caller. That choice depends on whether you think the caller ought to be obligated to handle the issue or die.

```cs
struct TestResult {
    bool success;
    string errorMessage;
}
```

bclehmann:
> @bclehmann do you think this is worth adding to AvaPlot?

Yeah, I think it looks easy and helpful for a lot of users.

I don't really like how `GDI.DrawingTest` works, I think there's a better way to indicate a failure. Mostly from an API clarity perspective. And maybe issues like this must be handled by the caller under the threat of crashing due to an unhandled exception.

swharden:
> I think there's a better way to indicate a failure.

Hey @bclehmann, I'm a bit confused by your objection/suggestion. Could you rephrase your concern or proposed solution to help me understand it better?

My intent is that `GDI.DrawingTest()` will only get called in designer mode, and only once when the control is first loaded. When a user drags a control from the toolbox and drops it onto the window, a System.Drawing support or version issue may cause it to throw an exception which Visual Studio has to catch and display (e.g., #993). I think those design-time exceptions are pretty jarring and confusing for new users (especially if the program builds and runs fine when the user presses F5).

I could move `DrawingTest()` somewhere in the `Control` namespace since it's not supposed to be used for any other purpose.

Interested in your feedback!
Scott

bclehmann:
>  Could you rephrase your concern or proposed solution to help me understand it better?

I had two concerns, the first was the throw on [line 39](https://github.com/ScottPlot/ScottPlot/blob/f1d9e0851654507b5520680893aaf449cb501cdd/src/ScottPlot/Drawing/GDI.cs#L39) which looked like it was using try/catch as a control structure, but the catch can actually catch multiple issues so my conclusion was too hasty. One could argue about it as a matter of taste, but now that I have thought about it I think it's probably the better solution.

The other was how the function indicates the design-time exception. I think returning a string with an error message is counter-intuitive. The name `DrawingTest` implies it either passes or fails a test, which is generally shown through returning a boolean indicating whether it passed, or through throwing an exception in the event that it fails. Should the error message be necessary you can read it when you catch the exception. Also, normally returning `null` implies an error.

My preferred solution is just to change the return type to void and remove the catch statement. The caller catches any exceptions and displays the error message. However, I think returning a boolean (or a struct with a boolean success field) is more appropriate for a method with the word "Test" in its name.

> I could move DrawingTest() somewhere in the Control namespace since it's not supposed to be used for any other purpose.

I like this idea, but I think there is a benefit to keeping it in GDI, since moving to a different graphics library would require this to be rewritten. I like the current state where nothing (that I know of) in the Control namespace is dependent on System.Drawing.Common.

bclehmann:
You can see what I meant in #1052. I don't know how well I communicated my concerns.

swharden:
> returning a string with an error message is counter-intuitive

I'm sold! A try/catch block in the initializer of each control is a lot more straightforward, and a better design.

Thanks for expanding on your thoughts/suggestions!

I'll work in #1052 to test the AvaPlot and also add similar try/catch blocks to the other controls 👍

PR #1053, opened 2021-05-22
create docfx project

### LINKED ISSUE #1047: Reconsider DocFX for API documentation
Currently the API is lightly documented in the cookbook using custom code that lives in the test library (#727). DocFX was suggested by @bclehmann back in #617, and I didn't pursue it at that time but I'm revisiting the idea now.

I blew a lot of time tonight debugging my custom documentation generator, and I'm starting to appreciate that using a standard package could have lots of advantages. https://dotnet.github.io/docfx

I'm reminded that ImageSharp's documentation is generated with DocFX, and it looks great!
* https://docs.sixlabors.com/api/ImageSharp.Drawing/SixLabors.ImageSharp.Drawing.Path.html

I added this to the triage list #1028 so it can be further explored as this library stabalizes after 4.1 is released.

swharden:
resolves #1047

### DIFF
--- a/dev/docfx/.gitignore
+++ b/dev/docfx/.gitignore
@@ -1,0 +1,3 @@
+obj
+api
+_site
--- a/dev/docfx/.gitignore
+++ b/dev/docfx/.gitignore
@@ -1,3 +1,4 @@
 obj
-api
+api/*.yml
+api/.manifest
 _site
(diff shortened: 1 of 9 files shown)


swharden:
This is a good start! 
https://swharden.com/scottplot/cookbooks/4.1.13-beta/docfx

@bclehmann I wish I explored this further when you first suggested it in #617 ... in hindsight, it would have been way easier to just use this than to make something custom like I did. Even if I insisted on making the custom static site generator I did, the YAML intermediate files docfx produces would have been so much easier to parse than using reflection and source code parsing to get all the components and XML docs! In hindsight, that task really sucked up a lot of time. So, a bit late, but thanks! I think this will prove to be really useful.

Issue #1086, opened 2021-05-28
Provide an Update() method in RadarPlot


arthurits:
## Updating values in RadarPlot

**Feature description:** The `Update()` method in several `Plottables` is quite useful especially when plotting data continously. Unfortunately, the RadarPlot doesn't seem to implement any updating at all, which can become a nuisance for continous data plotting.

I first hoped to create a `RadarPlot` derived-control in order to implement the value updating using a simple and straightforward code:
```cs  
class ClassPlotRadar : ScottPlot.Plottable.RadarPlot
{
    public ClassPlotRadar(double[,] values, Color[] lineColors, Color[] fillColors, bool independentAxes, double[] maxValues = null)
        :base(values, lineColors, fillColors, independentAxes, maxValues)
    {
    }
    public void Update(double[,] values, double[] maxValues = null)
    {
        base.Norm = new double[values.GetLength(0), values.GetLength(1)];
        Array.Copy(values, 0, base.Norm, 0, values.Length);
        if (this.IndependentAxes)
            base.NormMaxes = base.NormalizeSeveralInPlace(base.Norm, maxValues);
        else
            base.NormMax = base.NormalizeInPlace(base.Norm, maxValues);
    }
}
```
Unfortunately, the above code doesn't work because `Norm`, `NormMaxes`, `NormMax`, `NormalizeSeveralInPlace`, and `NormalizeInPlace` are all defined as private.

So, as far as I understand, the only option is to resort to the "Clear and Replot" strategy suggested in #1042, althought the user-code becomes slightly cumbersome as the plot format has to be re-set each time new data is received and the `Clear()` method is called. Therefore, it would be useful if some updating options could be provided for `RadarPlot`.

swharden:
> it would be useful if some updating options could be provided for RadarPlot

Hi @arthurits, thanks for the suggestion! It would be nice if more plot types had an `Update()` method to manage changing data (and provide validation / error checking), so I like the direction this is going.

## Restating the Issue

There is logic in the initializer that processes the data and populates private readonly variables, so you can't implement an `Update()` method that modifies them using inheritance.

https://github.com/ScottPlot/ScottPlot/blob/b1d26dc508f15cb4ff3a0c0feeae6ca803cfe769/src/ScottPlot/Plottable/RadarPlot.cs#L16-L18

[...]

PR #1091, opened 2021-05-29
1086 Implement an update method in RadarPlot

arthurits:
**Purpose:**
Pull request for issue [#1086](https://github.com/ScottPlot/ScottPlot/issues/1086)

**New functionality:**
Implements an `Update()` method.
The class initializer calls this `Update()` method.
`Norm`, `NormMax`, and `NormMaxes` are no longer `readonly`.
Added documentation.
Suggestions (as comments) for leaving out the `Array.Copy` as well as for moving the normalization to the `Render()` method.

### DIFF
--- a/dev/changelog.md
+++ b/dev/changelog.md
@@ -4,6 +4,7 @@
 * Made it easier to use custom color palettes (see cookbook) (#1058, #1082) _Thanks @EmanuelFeru_
 * Added a `IgnoreAxisAuto` field to axis lines and spans (#999) _Thanks @kirsan31_
 * Heatmaps now have a `Smooth` field which uses bicubic interpolation to display smooth heatmaps (#1003) _Thanks @xichaoqiang_
+* Radar plots now have an `Update()` method for updating data values without clearing the plot (#1086, #1091) _Thanks @arthurits_
 
 ## ScottPlot 4.1.15
 * Hide design-time error message component at run time to reduce flicking when resizing (#1073, #1075) _Thanks @Superberti and @bclehmann_
--- a/src/ScottPlot/Plot/Plot.Add.cs
+++ b/src/ScottPlot/Plot/Plot.Add.cs
@@ -706,8 +706,13 @@ namespace ScottPlot
         }
 
         /// <summary>
-        /// Add a radar plot
+        /// Add a radar plot (a two-dimensional chart of three or more quantitative variables represented on axes starting from the same point)
         /// </summary>
+        /// <param name="values">2D array containing categories (columns) and groups (rows)</param>
+        /// <param name="independentAxes">if true, axis (category) values are scaled independently</param>
+        /// <param name="maxValues">if provided, each category (column) is normalized to these values</param>
+        /// <param name="disableFrameAndGrid">also make the plot frameless and disable its grid</param>
+        /// <returns>the radar plot that was just created and added to the plot</returns>
         public RadarPlot AddRadar(double[,] values, bool independentAxes = false, double[] maxValues = null, bool disableFrameAndGrid = true)
         {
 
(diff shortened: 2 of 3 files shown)


bclehmann:
It looks like you forgot to autoformat your code, we use dotnet-format: https://github.com/dotnet/format

I don't have time to review your code now, but thanks for contributing!

arthurits:
> It looks like you forgot to autoformat your code, we use dotnet-format: https://github.com/dotnet/format

I've followed Scott's short and comprehensive guide in issue #1092 and hope formatting is now OK.

> I don't have time to review your code now, but thanks for contributing!

Thank you!

swharden:
Thanks for this PR @arthurits, it looks great! 

I added more XML docs to try to make this plot type and its options less confusing.

I also pulled some of your code comments out of the code and am copying them here so they're easier to discuss:

## Normalize on every render?

> The passed values are copied into the internal array `Norm`, which stores the plot's data.

> @swharden and @bclehmann consider if this is needed or desirable: do we want to allow the user to update its "value" array and call Render, (similarly to SignalPlot)? After all, in `SignalPlot`, the `Update()` method is used whenever the array is redimensioned but not when the array values are modified.

> If so, then the `Array.Copy` could be left out (just keep a reference to 'values'), and move the normalization at the beginning of the Render() method.

Restating the suggestion, normalizing inside `Render()` would allow the user to manipulate values in the 2D array externally (similar to other plot types). It would also prevent accumulation of state in `Norm`, `NormMax`, and `NormMaxes`. Reducing state and moving toward static methods for calculations is almost always a good thing!

The drawback is that normalizing on every render could degrade performance, but there are probably a small number of values in these arrays so perhaps the hit is negligible. We may want to benchmark this so we can discuss it quantitatively rather than speculate about what the performance it is.

@bclehmann do you have a preference here?

swharden:
This PR solves the #1086 issue so I'm going to go ahead and merge it, but if refactoring is desired related to https://github.com/ScottPlot/ScottPlot/pull/1091#issuecomment-851041209 we can make a new issue/PR to track progress on that front.

Thanks again for your help here @arthurits, and congratulations on your successful pull request! 👍 🎉 🚀

bclehmann:
> The drawback is that normalizing on every render could degrade performance

I think any performance hit here is going to be inconsequential, we normalize heatmaps with more than a million times as many values as we expect to have here very quickly. In fact, for small heatmaps (100x100 ish) we could probably eschew precomputing and get away with it, not that I would recommend it.

I think normalizing on render gives a better API and the performance impact is likely to be negligible. There is another option, make the `values` a property and normalize in the setter. That's a good API, and good performance, although one would keep the normalized values in the plottable.

swharden:
> I think normalizing on render gives a better API and the performance impact is likely to be negligible

Thanks for your input! Next time I work on this module I'll see about refactoring toward this.

> make the values a property and normalize in the setter

This is possible, but I think `Update()` is better here because it makes it a little more obvious to the user which variables have to be changed at the same time. 

Using properties users have to _remember_ to change `values` and `maxValues` at the same time, whereas an `Update()` method makes it easier by tying related values together that are expected to change at the same time.

arthurits:
Just one more related issued. The new 'Update()' method in 'RadarPlot' requires the user to provide the 'independentAxes' parameter as the method's second argument. However, the user most likely created the plot using 'AddRadar(values)' (line 729 in Plot.Add.cs) where the 'independentAxes' is defaulted to 'false'.

Not sure if it's a good idea to force the user to set the 'independentAxes' every time the plot's data values are updated especially when he didn't set that property (the most common scenario) when he added the radar plot.

swharden:
> Not sure if it's a good idea to force the user to set the `independentAxes` every time the plot's data values are updated, especially when they didn't set that property (the most common scenario) when he added the radar plot

Good point!

That argument has a default value in `Plot.AddRadar()`:
https://github.com/ScottPlot/ScottPlot/blob/a86a0ba2a983c4046753f224097231986184df25/src/ScottPlot/Plot/Plot.Add.cs#L729

but not in `Update()`:
https://github.com/ScottPlot/ScottPlot/blob/a86a0ba2a983c4046753f224097231986184df25/src/ScottPlot/Plottable/RadarPlot.cs#L104

I'll modify `Update()` to make that argument default to `false` 👍

PR #1093, opened 2021-05-29
Crosshair

swharden:
This PR adds a new `Crosshair` plot type (suggested in #999)

crosshair


EmanuelFeru:
Quick question, does it work only with a single axis? So, only with a vertical or only with a horizontal axis? If yes, i will use this right away.

swharden:
> does it work only with a single axis?

Good idea! I just added `IsVisibleX` and `IsVisibleY` fields to control visibility of individual axes. I also created a cookbook example demonstrating this 👍

Issue #926, opened 2021-03-30
Add support for a custom tick label formatting function


damiandixon:
I'm attempting to move from LiveCharts and I have a custom label formatter to display elapsed time
in the format: DD:HH:mm:ss, where DD is elapsed number of days.

I have found how to do HH:mm:ss using the following:
```
      wpfGridGraphControl.plt.Ticks(dateTimeX: true, dateTimeFormatStringX: "HH:mm:ss");
```
I've looked at the available format strings: https://tinyurl.com/y86clj9k, but there is nothing that seems to help.

Is there any way to implement a truly custom format callback that is passed an x-axis value and returns a string to display at the tick mark?

swharden:
Hi @damiandixon, thanks for the question!

I see your problem - You want to use a custom timestamp tick label format that includes days, but the [Standard numeric format strings](https://docs.microsoft.com/en-us/dotnet/standard/base-types/standard-numeric-format-strings) provided by Microsoft can't represent day as a number of days elapsed relative to an arbitrary date.

Your suggested solution is a good one too - add a custom format callback to enhance the user's ability to customize label formats.

Unfortunately I don't think there's an easy way to do this with the code the way it is right now. After getting ScottPlot 4.1 out of beta I intend to refactor the tick module to make tasks like this easier to achieve. I've added this issue to the triage list so we can keep track of it, but it may take me a little while to get to.

### Intermediate Solution

If you want to try implementing this yourself, you may be able to achieve this effect by adding a bit of logic in the top of this function ...

https://github.com/ScottPlot/ScottPlot/blob/822b975be9d4b5f5d1b5057782bcc4f6e0d7740b/src/ScottPlot/Renderable/AxisTicksRender.cs#L78-L82

... which replaces each label in `tc.tickLabels` (a string array) with a custom string based on values described by `tc.tickPositionsMajor` (a double array).

If you don't get to it, I may try this myself in the next day or two to determine how well it works. If this is a good spot where we can insert a custom function to modify tick labels at the last minute, it might be a good stand-in option until the tick module is fully/properly refactored.

StendProg:
Hi there,

@damiandixon I'd like to clarify. Do you want to get the total number of days, which can be more than 31? (because the number of days in a month is described by a simple `dd` format specifier).
Most likely I understood correctly, and the problems cannot be solved without a custom callback.

A good place to inject a new addiction is to add field to `TickCollection`:
```c#
public class TickCollection
{
    ...
[...]

PR #1100, opened 2021-05-30
Axis: TickLabelFormat accepts Func<double, string>

swharden:
This PR adds an option for the user to use a `Func<double, string>` to define how to generate tick labels from positions. 

Unlike current methods which only control the _format_ of labels or require the user to interact with a complex inheritance system, this is a quick and simple way to control tick label format in a way that uses logic.

```cs
plt.AddSignal(ScottPlot.DataGen.Sin(51));
plt.AddSignal(ScottPlot.DataGen.Cos(51));

static string customTickFormatter(double position)
{
    if (position == 0)
        return "zero";
    else if (position > 0)
        return $"+{position:F2}";
    else
        return $"({Math.Abs(position):F2})";
}

plt.XAxis.TickLabelFormat(customTickFormatter);
plt.YAxis.TickLabelFormat(customTickFormatter);
```

image

This PR was motivated by #926 and #1070


danhi:
my Plot object doesn't have and xaxis item.  It has axes, axes.bottom, but they wont take the customtickformater.  I'm sure I'm missing something simple here.  i'm using VB.Net and the latest version of scottplot.

Issue #1002, opened 2021-04-30
implement DragEnabled on SignalPlotXY


breakwinz:
Hi. It would be very useful if there was an option to enable dragging on a signalplotXY. 
For example im drawing several signalplotsXY on a single formsplot. I am then calculating the standard deviation Y for every point of X and drawing a 1 standard deviation signalPlotXY in a thick and clear line.

It would be quite useful if it were possible to enable a drag feature on a signalplot in horizontal or vertical direction as such:

var std_line_positive = FormsPlot1.Plot.AddSignalXY(x_std[], y_positive_std[]);
var std_line_negative = FormsPlot1.Plot.AddSignalXY(x_std[], y_negative_std[])

std_line_positive.DragEnabledVertical = true; <---- doesnt exist
std_line_negative.DragEnabledVertical = true;<---- doesnt exist

this would allow the user to click and drag the two signalplot lines and "all it does" in the background is shifting the y-values up and down. Similarly if there is DragEnabledHorizontal then the x values can be shifted up and down (left and right on graph)

Here is a beautiful paint image to explain. Bottom and top brown lines are the standard deviation i have calculated for all the signalplots in the middle. I want to be able to shift them up and down graphically

![image](https://user-images.githubusercontent.com/81701546/116723947-f3e70b00-a9d7-11eb-804c-4aa9ebf24c5b.png)

Is this something that would be easy/difficuly to implement?

Many thanks

StendProg:
Hi @breakwinz,

First of all, you can solve your problem in another way. You can simply bind SignalXY.OffsetY to some slider in the interface, and move `Signal` using it.

As for the implementation of `IDraggable` for a `SignalXY`, this is not a trivial task, but a solvable one. The main difficulties will be with `IsUnderMouse()` method, especially if we take into account the `snap`.
The problem is that it would be logical to  also add `IDraggable` to many other `Plotables`, and this is already a rather voluminous task.
I'll see what can be done for `SignalXY` the other day.

StendProg:
I ran into certain difficulties, the problem is that it is impossible to move the SignalXY l using existed `DragTo()`, you also need to know from what point the dragging was.
The very architecture of dragging plottables does not allow dragging anything other than horizontal/vertical lines. It is necessary to change the approach itself, I cannot promise that I will do it in the near future, but I will try to do it.

breakwinz:
[...]

PR #1111, opened 2021-06-03
Draggable Scaatter, Signal, SignalXY (CoordinatedHeatmap)

StendProg:
Work In Progress

**Purpose:**
Implement IDraggableModern interface for Scatter, Signal, SignalXY (CoordinatedHeatmap). #1002

Early shared, I plan to add some work, but can do it only after weekend, and there is some question i want to discuss.
Questions to discuss:

1. IDraggableModern - is working title. We need to come up with a suitable name. On the other hand, We can absorb the existing IDraggable. Only Span for now use this old interface. With such a takeover, backward compatibility of users may be broken, or maybe not, they seem to not use DragTo() directly.
2. `Drag(double fromX, double toX, double fromY, double toY, ...)` can be transformed to
`Drag(double offsetX, double offsetY)`.  Initially, I thought that the starting point might be useful in the handler, but in fact it was never used.
3. If plot dimensions additionaly are passed to IsUnderMouse(), then it will be possible to carry out additional optimizations by collapsing a set of points into one column for `Signals`, as is done during rendering. Although, oddly enough, on the test signals, I did not notice any performance problems with the current implementation.
4. The scatter, oddly enough, has no offsets (OffsetX, OffsetY). So dragging happens through modifying the original (XS, YS) arrays. Which leads to funny effects. In the demo, many scatters have a common XS array, and when you drag one, all the others are dragged, but only in X. I think it would be a good idea to implement Offsets for the Scatter, and rewrite the drag to change offsets as it is done for signals. But this task is best solved in a separate PR.
5. For now all signals, scatters, DragEnabled by default for test porpouse. It would be nice to make separate demos for new  Plottable. Then we can turn off the default DragEnabled.
6. For dragging, I selected the existing hand cursor. Maybe we can think of something more interesting.

What do I plan to finish in this PR:
1.  :heavy_check_mark: implement the algorithm for intersection the cursor with the line for the scatter in the same way as for signals. I started with a scatter, and found a more efficient algorithm on the signals.
2. :heavy_check_mark: Do not search for intersections with lines if the display of lines is disabled.
3. :heavy_check_mark: Think about how to reduce the duplication of the intersection code for different Plottables.
4. :heavy_check_mark: Write tests for intersection algorithms.
5. :heavy_check_mark: Add xml documentation to methods and fields.
6. Make Demo.

**New Functionality:**
```cs
signal = Plot.AddSignalXY(xs, ys);
signal.DragEnabled = true;
```
DraggableSignals

### DIFF
--- a/src/ScottPlot/Plottable/HelperAlgorithms/CloseToSegmentChecker.cs
+++ b/src/ScottPlot/Plottable/HelperAlgorithms/CloseToSegmentChecker.cs
@@ -1,5 +1,11 @@
 ﻿namespace ScottPlot.Plottable.HelperAlgorithms
 {
+    /// <summary>
+    /// Check if point close to segment
+    /// This algorithm is not completely fair, and cannot be used separately;
+    /// a separate preliminary check for proximity to the ends of the segment (AABBChecker)
+    /// makes the results of this algorithm reliable.
+    /// </summary>
     public class CloseToSegmentChecker
     {
         public double SnapX { get; set; }
--- a/src/ScottPlot/Plottable/HelperAlgorithms/CloseToSegmentChecker.cs
+++ b/src/ScottPlot/Plottable/HelperAlgorithms/CloseToSegmentChecker.cs
@@ -1,0 +1,100 @@
+﻿namespace ScottPlot.Plottable.HelperAlgorithms
+{
+    public class CloseToSegmentChecker
+    {
+        public double SnapX { get; set; }
+        public double SnapY { get; set; }
+        public double CoordinateX { get; set; }
+        public double CoordinateY { get; set; }
+
+        private (double x, double y)[][] _boundaryPolys;
+
+        /// <summary>
+        /// Constructor
+        /// </summary>
+        /// <param name="coordinateX">point x coordinate</param>
+        /// <param name="coordinateY">point y coordinate</param>
+        /// <param name="snapX">permissible deviation for x</param>
+        /// <param name="snapY">permissible deviation for y</param>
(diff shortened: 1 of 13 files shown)


swharden:
I'm going to close this PR since it's unlikely to be merged soon, but there is active discussion about this PR in https://github.com/ScottPlot/ScottPlot/issues/1002#issuecomment-886126095

src/ScottPlot/Plottable/HelperAlgorithms/CloseToSegmentChecker.cs
```diff
@@ -0,0 +1,106 @@
...
+
+            for (int k = 0; k < _boundaryPolys.Length; k++)
+            {
+                bool inside = false;
+                for (int j = 0, j1 = _boundaryPolys[k].Length - 1; j < _boundaryPolys[k].Length; j1 = j++)
+                {
+                    if ((_boundaryPolys[k][j].y > CoordinateY) != (_boundaryPolys[k][j1].y > CoordinateY) &&
+                        (CoordinateX < (_boundaryPolys[k][j1].x - _boundaryPolys[k][j].x) * (CoordinateY - _boundaryPolys[k][j].y) / (_boundaryPolys[k][j1].y - _boundaryPolys[k][j].y) + _boundaryPolys[k][j].x))
+                    {
+                        inside = !inside;
```
swharden, line 96: I may be confused about what this section is doing... but should this be `inside |= inside`?

swharden: oops, that was a pretty silly comment. I'm going to delete it and let's pretend it didn't happen 😅

Issue #1092, opened 2021-05-29
Add "step" option for low density rendering mode in Signal plots


EmanuelFeru:
## Features Suggestion

Hi Scott, it would be great to have a function:
```cs
public SignalPlotConstStep<T> AddSignalConstStep<T>(T[] ys, double sampleRate = 1, Color? color = null, string label = null) where T : struct, IComparable;
```
similar to `AddScatterStep(...)`. I know `AddScatterStep()` can already do the job. But if I understand correctly the `AddSignalConst` is much faster in case of high number of data points. So, I was wondering if we can do exactly as `AddSignalConst` to maintain performance but plot it as step.

 I work with microcontrollers and there the signals are discrete, so it would be ideal to display them as steps. Do you think it would be much effort to achieve something like this?

`AddSignalConst`
![image](https://user-images.githubusercontent.com/24780745/120076257-eab78f80-c0a4-11eb-8756-360f27452501.png)

`AddSignalConstStep` (Suggested feature)
![image](https://user-images.githubusercontent.com/24780745/120076266-facf6f00-c0a4-11eb-9c27-4c73a6d65ed5.png)

bclehmann:
The reason `Signal`, `SignalConst` and `SignalXY` are faster is because they can take advantage of optimizations between points that are very close together. If there are 10 sequential points in an area smaller than an individual pixel, ScottPlot can simply draw from the first to the tenth and skip everything in between.

For points that are that densely packed on whatever zoom level, you wouldn't be able to tell that it was a stepped plot, it would look like a normal line plot.

EmanuelFeru:
When the plot is completely zoomed out, I agree with you @bclehmann that there is no difference step or not step. And you can do some tricks to render quickly because the details are not seen anyway.

My question is what happens when we zoom until details become visible and we can see the steps? Is there a speed difference between const and non-const plot?

swharden:
> For points that are that densely packed on whatever zoom level, you wouldn't be able to tell that it was a stepped plot, it would look like a normal line plot.

This is true, and Signal Plots call this rendering type "high density mode".

## Step Display in Low Density Mode

However in low density rendering mode the points on the screen are connected with straight lines here:

https://github.com/ScottPlot/ScottPlot/blob/b1d26dc508f15cb4ff3a0c0feeae6ca803cfe769/src/ScottPlot/Plottable/SignalPlotBase.cs#L279

[...]

PR #1128, opened 2021-06-20
Implemented Step Signal Plot

EmanuelFeru:
**Purpose:**
Step Display is very useful to show digital signals. 

**New Functionality:** This PR implements #1092
- added function for `StepDisplay` (take-over from Scatter Plot)
- added Cookbook example
- all tests are passing

image

### DIFF
--- a/src/ScottPlot/Plottable/SignalPlotBase.cs
+++ b/src/ScottPlot/Plottable/SignalPlotBase.cs
@@ -276,28 +276,42 @@ namespace ScottPlot.Plottable
                 PointF[] pointsArray = linePoints.ToArray();
                 ValidatePoints(pointsArray);
 
-                if (penLD.Width > 0)
-                    if (StepDisplay)
+                if (StepDisplay)
+                {
+                    // add necessary points for step display
+                    PointF[] pointsStep = new PointF[pointsArray.Length * 2 - 1];
+                    for (int i = 0; i < pointsArray.Length - 1; i++)
                     {
-                        PointF[] pointsStep = new PointF[pointsArray.Length * 2 - 1];
-                        for (int i = 0; i < pointsArray.Length - 1; i++)
-                        {
-                            pointsStep[i * 2] = pointsArray[i];
-                            pointsStep[i * 2 + 1] = new PointF(pointsArray[i + 1].X, pointsArray[i].Y);
-                        }
-                        pointsStep[pointsStep.Length - 1] = pointsArray[pointsArray.Length - 1];
-                        gfx.DrawLines(penLD, pointsStep);
+                        pointsStep[i * 2] = pointsArray[i];
+                        pointsStep[i * 2 + 1] = new PointF(pointsArray[i + 1].X, pointsArray[i].Y);
                     }
-                    else
+                    pointsStep[pointsStep.Length - 1] = pointsArray[pointsArray.Length - 1];
+
+                    if (penLD.Width > 0)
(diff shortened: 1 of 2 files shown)


bclehmann:
The step plot looks a little weird with dots drawn. I don't know if that's something that can/should be changed. Maybe just changing the demo to not draw marker points is all that should be done?

StendProg:
Hi @EmanuelFeru,

Good improvement, thanks for your contribution.

Unfortunately, SignalXY will inherit `StepDisplays`, but will not react to it in any way. There is no simple solution, I think it is better to leave it as it is for now.

It seems that the `StepDisplay` shows no respect for the existing signal fill above/below. At first glance, it looks easy, it is enough to transfer the already formed array to the code below, but in the process, difficulties may arise. At least it looks like it's worth at least a try. 
What do you think?

> The step plot looks a little weird with dots drawn. I don't know if that's something that can/should be changed. Maybe just changing the demo to not draw marker points is all that should be done?

It also immediately caught my eye, most likely the lack of symmetry gives such an effect.

EmanuelFeru:
Yes, it's the symmetry. I was looking strange at the beginning too 😃 the dots can be hidden by making the Markersize 0. What is your opinion for the example? I can put the Markersize to 0 to have a more natural look.

@StendProg, I did not fully understood. Do you mean it we do surface below for example then it still considers straight line between points?

StendProg:
> @StendProg, I did not fully understood. Do you mean if we do surface below for example then it still considers straight line between points?

Exactly. The fill in the background will connect the dots and will probably be ugly. Maybe we don't need it, but most likely sooner or later this question will come up, it's up to you, I just noticed that the fill uses the old array of points in `stepDisplay` mode.

EmanuelFeru:
I see... Then it's better to have it done now. I will have a look in the next couple of days.

EmanuelFeru:
I quickly checked FillBelow. Indeed it does not draw well around the steps. Still investigating how to make it plot steps. I think a change is needed on the Rectangle fill:
https://github.com/ScottPlot/ScottPlot/blob/3ead8c204116ea95baed60df6715b70cd668beab/src/ScottPlot/Plottable/SignalPlotBase.cs#L428-L432
image

Code for the above plot:
```cs
{
    double[] ys = DataGen.Sin(51);
    int sampleRate = 1;

    var sig = plt.AddSignal(ys, sampleRate);
    sig.StepDisplay = true;
    sig.MarkerSize = 0;
    sig.FillType = FillType.FillBelow;
    sig.FillColor1 = Color.Blue;

}
```
____
Also I am not sure if the fill is done correctly, it behaves strange when I move and looks diagonal. I would expect to fill similar to here:
https://swharden.com/scottplot/cookbooks/4.1.16/category/plottable-fill/#fill-under-curve
image

EmanuelFeru:
Fixed. I will do a commit.
image

EmanuelFeru:
image

swharden:
Hi @EmanuelFeru, thank you for this PR! Thanks @bclehmann and @StendProg for your input along the way.

I'm going to refine this PR and merge it in shortly despite it not looking perfect with filled signal plots. It looks like filled signal plots have slight rendering errors already, and I'll work to investigate and hopefully fix that in a separate issue (#1163)

swharden:
Thanks @EmanuelFeru, this looks great! This feature will be in the next release 👍🚀 

I'll fix the error affecting the alignment of fills shortly (#1163)

PR #1147, opened 2021-07-07
Improve multi-axis support for crosshair labels

rutkowskit:
**Purpose:**
Current Crosshair implementation doesn't work well when you want to attach it to Right or Top axis. 
After setting YAxisIndex to 1, I would expect the label with the value on the right axis. 
After setting XAxisIndex to 1, I would expect the label with the value on the top axis. 
This Pull Request contains a small fix to handle this case.

Below is the code of a sample recipe for multiple crosshairs attached to different axes.
```cs
    public class CrosshairMultipleForDifferentAxes : IRecipe
    {
        public string Category => "Plottable: Crosshair";
        public string ID => "crosshair_multiple_different_axes";
        public string Title => "Multiple crosshairs for different axes";
        public string Description => "Chrosshair for right and top axis";

        {
            plt.Title("Multiple Crosshairs for different Axes");
            plt.YLabel("Vertical Axis");
            plt.YAxis2.Label("Signal2 Value");
            plt.YAxis2.Ticks(true);
            plt.XAxis2.Ticks(true);

            var signal = plt.AddSignal(DataGen.RandomWalk(null, 100));
            var signal2 = plt.AddSignal(DataGen.RandomWalk(null, 50));
            signal2.YAxisIndex = 1;
            signal2.XAxisIndex = 1;

            var ch1 = plt.AddCrosshair(40, signal.Ys[50]);
            ch1.LabelBackgroundColor = signal.Color;

            var ch2 = plt.AddCrosshair(20, signal.Ys[25]);            
            ch2.LineColor = System.Drawing.Color.Transparent; 
            ch2.LabelBackgroundColor = signal2.Color;
            ch2.YAxisIndex = signal2.YAxisIndex;
            ch2.XAxisIndex = signal2.XAxisIndex;
        }
    }
```

### DIFF
--- a/dev/changelog.md
+++ b/dev/changelog.md
@@ -1,7 +1,10 @@
 # ScottPlot Changelog
 
 ## ScottPlot 4.1.17
-* Improved `RadarPlot.Update()` default arguments (#1097) _Thanks @arthurits_
+* Radar Plot: Improved `Update()` default arguments (#1097) _Thanks @arthurits_
+* Crosshair: Added `XLabelOnTop` and `YLabelOnRight` options to improve multi-axis support and label customization (#1147) _Thanks @rutkowskit_
+* Signal Plot: Added `StepDisplay` option to render signal plots as step plots when zoomed in (#1092, #1128) _Thanks @EmanuelFeru_
+* Testing: Improved error reporting on failed XML documentation tests (#1127) _Thanks @StendProg_
 
 ## ScottPlot 4.1.16
 * Made it easier to use custom color palettes (see cookbook) (#1058, #1082) _Thanks @EmanuelFeru_
--- a/src/ScottPlot/Plottable/Crosshair.cs
+++ b/src/ScottPlot/Plottable/Crosshair.cs
@@ -64,6 +64,16 @@ namespace ScottPlot.Plottable
         /// </summary>
         public bool IsVisibleY = true;
 
+        /// <summary>
+        /// Control whether the label for the X crosshair appears on the top or bottom edge of the plot
+        /// </summary>
+        public bool XLabelOnTop = false;
+
+        /// <summary>
+        /// Control whether the label for the Y crosshair appears on the left or right edge of the plot
+        /// </summary>
+        public bool YLabelOnRight = false;
+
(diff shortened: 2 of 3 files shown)


swharden:
Hi @rutkowskit, thanks for this PR! This looks fantastic!

image

There is an edge case where users can add more axes https://swharden.com/scottplot/cookbooks/4.1.16/category/multi-axis/#additional-y-axis so code won't function as expected since it places the labels on the right/top for any axis index other than 0

I think the solution here may be to expose an enumeration indicating which edge to render the label on and use that (instead of the axis index) to determine positioning. It's getting a bit late here so I'll put this down for today, but I hope to merge it in tomorrow 👍

rutkowskit:
Hi,
Originally I wanted to solve this by passing function that returns "Axis array" (provider) instead of adding multiple Crosshair renderers and setting X and Y axis index.
In theory it would work for additional Axes as well.
In the end I decided to make it as simple as possible to avoid possible backward compatibility issues.

My original thought was something like this:
1. In "Crosshair" class:
1a. Add Func <Axis []> AxisProvider {get; set;}
1b. Modify Render method to use dimensions, position, line color (for label color), from Axes returned by AxisProvider instead of PlotDimentions and additional configuration
1c. Render labels for all Axes (returned by provider) in one go
2. In my project (wpf) add:
2a. Crosshair = wpfPlot1.Plot.AddCrosshair (0, 0);
2b. Crosshair.AxisProvider = () => wpfPlot1.Plot.GetSettings (). Axes.ToArray ();
2c. In "OunMouseMove" pass x and y value returned from plot.GetMousePixel() to Crosshair (so it can use it to render correct label values for each Axis)

It should work in theory, but I guess it would strongly violate the original concept, because most Render function parameters would be ignored.

swharden:
I refactored this module a bit more, and came up with a left/right top/bottom system independent of axis index:

```cs
// alternate label positions can be set manually as needed
var cross2 = plt.AddCrosshair(33, 5.1);
cross2.YLabelOnRight = true;
cross2.XLabelOnTop = true;
```

Thanks again for this idea and PR @rutkowskit! I'm happy to merge this in now 🚀

Issue #1172, opened 2021-07-30
Crosshair and Axis Line: Support custom label formatters


Maoyao233:
## Features Suggestion

**Feature description:** In #1100 , a custom formatter can be used to format the tick lables, but crosshair doesn't support it, which leads to inconsistencies in the label display. It might be possible to add `CustomLabelFormatX` and `CustomLabelFormatY` to solve this problem. I can make a pull request if needed.

**Code example:** 

```cs
plt.AddSignal(ScottPlot.DataGen.Sin(51));
plt.AddSignal(ScottPlot.DataGen.Cos(51));

var crosshair = plt.AddCrosshair(10, 0);

// create a custom formatter as a static class
static string customLabelFormatter(double position)
{
    if (position == 0)
        return "zero";
    else if (position > 0)
        return $"+{position:F2}";
    else
        return $"({Math.Abs(position):F2})";
}

// use the custom formatter for horizontal and vertical tick labels
plt.XAxis.TickLabelFormat(customLabelFormatter);
plt.YAxis.TickLabelFormat(customLabelFormatter);

// use the custom formmater for crosshair labels
crosshair.CustomLabelFormatX(customLabelFormatter);
crosshair.CustomLabelFormatY(customLabelFormatter);
```

Issue #1122, opened 2021-06-15
Line Plot: Show position label on axis


EmanuelFeru:
**Feature description:** 

It would be very nice to have the posibility of tooltip for Draggable axis similar to crosshair tooltip #1093

```cs
VLine vLine = new VLine();
vLine = formsPlot.Plot.AddVerticalLine(0);
vLine.DragEnabled = true;
vLine.LabelEnabled = true;   // new

HLine hLine = new HLine();
hLine = formsPlot.Plot.AddHorizontalLine(0);
hLine.DragEnabled = true;
hLine.LabelEnabled = true;   // new
```

swharden:
Hi @EmanuelFeru, by a "draggable axis" do you mean you want an axis (the axis line, tick, tick marks, label, etc) to be draggable by click-dragging it?

![image](https://user-images.githubusercontent.com/4165489/122063388-387b0a00-cdbe-11eb-973b-1c9fb77e383f.png)

If this is what you meant, it's a pretty complex task to make the _axes_ draggable so it seems unlikely to be achieved soon.

However, a new _plot type_ could be created (nearly identical to the draggable axis lines) that draws a line across the plot but also adds ticks and labels at certain positions to achieve a nearly identical effect. 

Alternatively the existing axis lines could be extended to add optional ticks and tick labels. This may be the easiest way to achieve this effect rapidly.

swharden:
Hi @EmanuelFeru, after reading your message again an alternative interpretation is that you want the _crosshair axis label_ to be draggable such that click-dragging the label moves the crosshair line.

![image](https://user-images.githubusercontent.com/4165489/122064205-f900ed80-cdbe-11eb-8c8d-b09ebac8362f.png)

If you could clarify your suggestion it will be helpful in evaluating its feasibility 👍

EmanuelFeru:
Hi Scott,

Yes, I meant the 2nd option, showing the cross hair label also in Draggable axis Vertical or Horizontal.

To be more clear, I don't need to drag the label, just dragging the axis (as it is now) and the lable to be displayed and following, exactly same as you did for the cross hair, but now for vertical or horizontal Draggable axis.

swharden:
Thanks for the suggestion @EmanuelFeru! This one took me a while to get to, but I'm finally making progress. I'm combining this with crosshair refactoring #1173, and it looks like we'll have something pretty similar to this shortly:

```cs
var plt = new ScottPlot.Plot(400, 300);
[...]

PR #1173, opened 2021-08-02
Add custom label formatter support for crosshair

Maoyao233:
**New Contributors:**

**Purpose:**
#1172

**New Functionality:**

```cs
 var ch = plt.AddCrosshair(20, 0);

// use the custom formatter for crosshair labels
ch.ManualStringFormatterX = customTickFormatter;
ch.ManualStringFormatterY = customTickFormatter;
```


src/ScottPlot/Plottable/Crosshair.cs
```diff
@@ -43,6 +43,11 @@ public class Crosshair : IPlottable
         /// </summary>
         public bool IsDateTimeX = false;
 
+        /// <summary>
+        /// If defined, this function will be used to generate labels from X position, value of StringFormatX and IsDateTimeX will be ignored
+        /// </summary>
+        public Func<double, string> ManualStringFormatterX = null;
```
bclehmann, line 49: I think `TickLabelFormat` is ignored if it's a datetime axis, so this introduces an inconsistency.

I think it's probably better that it isn't ignored for datetime axes, but that change would mean changing the axis code as well.

src/ScottPlot/Plottable/Crosshair.cs
```diff
@@ -119,7 +129,7 @@ private void RenderHorizontalLine(PlotDimensions dims, Graphics gfx, Pen pen, Fo
             float yPixel = dims.GetPixelY(Y);
             gfx.DrawLine(pen, dims.DataOffsetX, yPixel, dims.DataOffsetX + dims.DataWidth, yPixel);
 
-            string yLabel = IsDateTimeY ? DateTime.FromOADate(Y).ToString(StringFormatY) : Y.ToString(StringFormatY);
+            string yLabel = ManualStringFormatterY != null ? ManualStringFormatterY(Y) : IsDateTimeY ? DateTime.FromOADate(Y).ToString(StringFormatY) : Y.ToString(StringFormatY);
```
bclehmann, line 132: This ternary within a ternary is pretty unwieldly.

src/ScottPlot/ScottPlot.csproj
```diff
@@ -22,9 +22,9 @@ Release cookbook: https://swharden.com/scottplot/cookbook
 Release demo: https://swharden.com/scottplot/demo</PackageReleaseNotes>
     <IncludeSymbols>true</IncludeSymbols>
     <SymbolPackageFormat>snupkg</SymbolPackageFormat>
-    <Version>4.1.16</Version>
-    <AssemblyVersion>4.1.16.0</AssemblyVersion>
-    <FileVersion>4.1.16.0</FileVersion>
+    <Version>4.1.17</Version>
```
bclehmann, line 25: I think Scott will bump the version himself once this is merged and any other PRs have been merged in.

src/cookbook/ScottPlot.Cookbook/Recipes/Plottable/Crosshair.cs
```diff
@@ -116,6 +116,49 @@ public void ExecuteRecipe(Plot plt)
...
+            // use the custom formatter for horizontal and vertical tick labels
+            plt.XAxis.TickLabelFormat(customTickFormatter);
+            plt.YAxis.TickLabelFormat(customTickFormatter);
+
+            // add a crosshair
+            var ch = plt.AddCrosshair(20, 0);
+
+            // use the custom formatter for crosshair labels
+            ch.ManualStringFormatterX = customTickFormatter;
+            ch.ManualStringFormatterY = customTickFormatter;
```
bclehmann, line 158: Again on the consistency argument, it's a little weird that for one of them you pass in a formatter delegate to a function and for the other you set a property.

I think that setting a property is probably the better API, but if we want consistency that would mean changing the axis code.

bclehmann (commented):
Looks good, just a few minor things. They're not about your code so much as maintaining a consistent API, as anyone who uses this feature probably is also using the custom formatter on the axis as well. So those features should probably expose a similar API.

Maybe even the crosshair label formatter should use the axis formatter? I don't know how easy that would be though, especially now that we have support for multiple axes. It would also introduce a lot of coupling.

bclehmann:
Also, it looks like you made this PR from your master branch, instead of a new branch. That's not a problem for us but it could make it more difficult for you to fetch new commits that get merged in.

src/ScottPlot/ScottPlot.csproj
swharden: Indeed! This can be left as `4.1.16` for now 👍

src/ScottPlot/Plottable/Crosshair.cs
swharden: > I think `TickLabelFormat` is ignored if it's a datetime axis

Maybe there should be a `NumericFormatter` and `DateTimeFormatter`?

swharden:
Hi @Maoyao233, thank you for this PR! I really like the idea of exposing custom formatters in more places throughout ScottPlot, and this is a great improvement.

Thinking out loud, an alternative to allowing formatters to be `null` we could always use these formatters and pre-populate them with generic formatters created in the constructor.

I think separate functions would be helpful, but I recognize this is super verbose:
* `XNumericFormatter` 
* `YNumericFormatter` 
* `XDateTimeFormatter`
* `YDateTimeFormatter`

Maybe this is an indication this class should be refactored as two smaller objects (because this isn't the only duplication of X and Y properties). Alternatively I wonder if these advanced label features should be added to `AxisLine`, then `Crosshair` could contain two of those. We can consider that type of refactoring in a future PR. For now this PR is going in a good direction, and I'm happy to merge it in (with its cookbook recipe) after @Maoyao233 has a chance to review some of the code comments. Thanks @bclehmann for your input!

src/ScottPlot/Plottable/Crosshair.cs
bclehmann: I think it's fine for the user to figure that out themselves, as they should know whether or not it's a DateTime axis.

Although I suppose we could make it easier for them by having `DateTimeFormatter` take in a `DateTime` instead of a `double`.

bclehmann:
> I think separate functions would be helpful, but I recognize this is super verbose:
.
XNumericFormatter
YNumericFormatter
XDateTimeFormatter
YDateTimeFormatter

One could allow the user to pass in one function (or class) which would override one or several of these? It could be done with the decorator pattern. There is a way to do this where they pass in a function, where they're passed an object, and setting fields of that object sets the formatter. The formatter should probably be allowed to be null just so it's easy for them to reset it.

Both of these are departures from the rest of the ScottPlot API however.

src/cookbook/ScottPlot.Cookbook/Recipes/Plottable/Crosshair.cs
Maoyao233: The consistency argument is also present in the original API. I think changing the axis code is a good way.

src/ScottPlot/ScottPlot.csproj
Maoyao233: I didn't intentionally change the version number, probably because I'm not familiar with ScottPlot's automated build system. I also think this small change should not change the version number.

src/ScottPlot/Plottable/Crosshair.cs
Maoyao233: Thank. I'll avoid using it in future.

swharden:
@Maoyao233 thanks again for this PR and your feedback! When you're done making changes to this PR let me know and I'll make a few small changes of my own then merge this in 👍 

> One could allow the user to pass in one function (or class) which would override one or several of these? It could be done with the decorator pattern.

> Both of these are departures from the rest of the ScottPlot API however.

@bclehmann this is a good observation! This trend could be taken to its logical conclusion by having a fully-abstracted "configuration object" or "styling object" that could be instantiated and modified by the user and passed into plots or plottables. This would help separate code for configuration, styling, and rendering. If ScottPlot were a private library that only I used, I would prefer this. Since it's a public library and I strive to make it quick and easy to learn for new users, I find myself preferring the simplicity of having users assign public fields.

bclehmann:
I wonder if we should have a demo as well, for the datetime axes? Perhaps that should wait for when axes get custom datetime formatters.

swharden:
Thanks @Maoyao233 this looks fantastic! I also appreciate your input along the way @bclehmann

I'll be refactoring the Crosshair module shortly. I intend to add support for position labels to axis lines, and when I do that I'll be sure to support this type of custom formatter support as well. I suspect the `Crosshair` will get refactored to contain one `HorizontalLine` and one `VerticalLine`, but we'll see. Progress there will be tracked in #1122

PR #1186, opened 2021-08-05
Creates bash script for autoformatting

bclehmann:
**Purpose:**
Creates an `autoformat.sh` script, just like there's an `autoformat.bat` script. Note that the bash script doesn't pause until the user presses a key, as I don't know why that was there in the first place.

**New Functionality:**
Linux and Mac users can now use an autoformat script.

We could consider creating a `ps1` script as well, but I don't know if that's necessary or not. Although PowerShell is cross-platform now.

### DIFF
--- a/src/autoformat.sh
+++ b/src/autoformat.sh
@@ -1,0 +1,7 @@
+#!/bin/sh
+
+# Install/update the dotnet 5 autoformatter
+dotnet tool update -g dotnet-format
+
+# Run the dotnet formatter
+dotnet format


swharden:
This is a good idea - thanks for the PR!

> Note that the bash script doesn't pause until the user presses a key, as I don't know why that was there in the first place.

Double-clicking the .bat file opens it in a console window. Without the pause it closes when it finishes so you can't see if changes were made. Seeing that the formatter did something may indicate a git commit (or reset) is required. I cringe to type this, but I suspect most .bat files on Windows are executed by double-clicking them rather than executing them from a terminal 😅

The .sh file doesn't need a pause because it was probably called from a terminal and the results won't disappear when it finishes running.

> We could consider creating a ps1 script as well, but I don't know if that's necessary or not. Although PowerShell is cross-platform now.

While a ps1 script could do everything, I think the `.bat` and `.sh` extensions make it more obvious to developers that the autoformat scripts are supported on their platforms.

Issue #1131, opened 2021-06-23
Polar Area Chart


bclehmann:
## Features Suggestion

I was using chartjs at work today and I noticed [this ](https://www.chartjs.org/docs/latest/charts/polar.html) in the documentation. It appears to be a mix between a pie chart and a radar chart. Note that there is only one dimension being plotted, everything else is a function of the radius. On further research, they are also called Coxcomb or Rose charts.

![image](https://user-images.githubusercontent.com/8635304/123017258-744b3c00-d389-11eb-8da4-0ec6b5087f1b.png)

If implemented this should probably borrow as much of their API as possible from either pie or radar charts so that users can switch between them quickly in testing. Perhaps this should even be a subclass of one of them.

I also wonder if this would benefit from layering multiple datasets on top of each other (like is done with Radar charts). There is a famous historical example of this here:
![image](https://user-images.githubusercontent.com/8635304/123017774-824d8c80-d38a-11eb-9181-7cef215bb148.png)

**Code example:** If applicable, use this space to demonstrate how you wish the new feature could be used in code.

```cs
double[] values = { 778, 283, 184, 76, 43 };
plt.AddPolarArea(values);
```

swharden:
Similarly I noticed interesting circular charts on https://github.com/beto-rodriguez/LiveCharts2 recently which may be worth implementing. There may be some good ideas at https://play.grafana.org too

![image](https://user-images.githubusercontent.com/4165489/128082689-290c5da7-d4c2-4dc5-8a68-7b549dbc06d0.png)

I'll probably close this issue now and focus on bugs/refactoring trying to get ready for #1036, but if you decide to make a PR for these new plot types I'd be happy to merge it in! We can re-open this issue at that time

bclehmann:
I started an implementation, but I'm a little busy so it may take some time before the PR is ready to merge.

Those circular charts may be achievable by adjusting `donutSize` on regular pie charts.

PR #1188, opened 2021-08-06
Coxcomb Chart

bclehmann:
**New Contributors:**

**Purpose:**
#1131

I have flip-flopped on the name a little bit, but for now I'm using Coxcomb.

**New Functionality:**

```cs
var coxcomb = plt.AddCoxcomb(values);
```

image


bclehmann:
Worth noting that as part of this I refactored the same axis that `RadarPlot` uses. This could offer more customization options for radar plots, for example by letting users customize the tick placement.

swharden:
Hi @bclehmann thanks for this PR!

> Worth noting that as part of this I refactored the same axis that RadarPlot uses. This could offer more customization options for radar plots, for example by letting users customize the tick placement.

I like what you did with `StarAxis` and `StarAxisTick` 👍 

I'm also not worried that `ValidateData()` isn't implemented. I suspect the validation system will get reworked in the future (#1028)

PR #1189, opened 2021-08-06
Plot.GetImageBytes()

### LINKED ISSUE #1107: Plot: GetImageBytes()
For cloud applications we need to convert a `Bitmap` to `byte[]` to save into blob storage.

This requires interacting with `System.Drawing.Common`, which means taking that dependency on in the cloud app.

The Plot module should have a `GetBitmapBytes()` that lets you specify file format type, preventing the need to pass `Bitmap` objects around in the cloud

```cs
public static byte[] GetBytesPng(System.Drawing.Bitmap bmp)
{
    using MemoryStream stream = new();
    bmp.Save(stream, System.Drawing.Imaging.ImageFormat.Png);
    return stream.ToArray();
}
```

swharden:
resolves #1107

### DIFF
--- a/src/ScottPlot/Plot/Plot.Render.cs
+++ b/src/ScottPlot/Plot/Plot.Render.cs
@@ -2,6 +2,7 @@
 using System.Diagnostics;
 using System.Drawing;
 using System.Drawing.Imaging;
+using System.IO;
 using System.Linq;
 
 namespace ScottPlot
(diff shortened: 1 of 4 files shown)


src/ScottPlot/Plot/Plot.Render.cs
```diff
@@ -166,6 +167,20 @@ public Bitmap Render(int width, int height, bool lowQuality = false, double scal
         /// </summary>
         public Bitmap GetBitmap(bool lowQuality = false, double scale = 1.0) => Render(settings.Width, settings.Height, lowQuality, scale);
 
+        /// <summary>
+        /// Render the plot and return the bytes for a PNG file.
+        /// This method is useful for rendering in stateless cloud environments that do not use a traditional filesystem.
+        /// </summary>
+        /// <returns></returns>
+        public byte[] GetImageBytes(bool lowQuality = false, double scale = 1.0)
```
bclehmann, line 175: It's a little weird that this returns a PNG but it's name makes out like it returns only the body (i.e. doesn't include the header).

swharden: > it's name makes out like...

I tried to leave this open-ended because I plan to have it accept a `Maui.Graphics.ImageFormat` enum as an argument and perhaps then it will be more obvious what this returns 🤔 

https://github.com/dotnet/Microsoft.Maui.Graphics/blob/0e9a4cdadbc077cc29103c0a12a5b24b5beb6e7e/src/Microsoft.Maui.Graphics/ImageFormat.cs#L1-L11

bclehmann: I think that would make more sense for something like `GetImage`, but you're right, taking in that enum would resolve the confusion.

PR #1190, opened 2021-08-06
ScatterPlotList: Fix GetLegendItems()

Maoyao233:
**Purpose:**
The legend item of ScatterPlotList cannot display because of the GetLegendItems() function doesn't set properties correctly.

**New Functionality:**
None


swharden:
Good catch @Maoyao233, thanks!

Issue #1171, opened 2021-07-30
Bug with EqualScaleMode.ZoomOut, behaves like ZoomIn


noob765:
**Bug Description:** 
```cs
Plot.AddScatter(x, y);
Plot.AxisAuto();
Plot.AxisScaleLock(true, EqualScaleMode.ZoomOut);
```

It's my very first day playing with this so I won't claim to be an expert here but based on everything I've seen after cloning the project and going through the testing folder, I'm expecting the previous 3x lines to:

1. plot an xy scatter line

2. resize the axes in a zoom to fit fashion so as to make everything visible (in this case, all the line)

3. lock the axes at an x:y aspect ratio of 1:1 (so that circles are not ellipses and squares are not rectangles) ***WHILE MAINTAINING EVERYTHING ZOOMED TO FIT*** (the upper case part is the bug)

(1)-(2) work just fine. (3) locks the axes at 1:1 alright, but instead of ZoomingOut as the option says, ZoomsIn... so if I plot a thin flat but long line of noise, the end result is a very narrow view of a few points with the plot spilling big time on both ends. Manually zooming out reveals the whole line.

As I said, I cloned the repo so I peeked in to settings.cs, line 268 which is where EqualScaleMode.ZoomOut is handled:

https://github.com/ScottPlot/ScottPlot/blob/cabadfd153db0d563947977afa3965061c50dde5/src/ScottPlot/Settings.cs#L268-L273

I didn't single step that code directly, but I did replicate it in my app code trying to figure out what was going on. The answer is simple (though weird): If you're expecting X/YAxis.Dims.UnitsPerPx to be units per pix (ie, **positive**), then this should work but **both UnitsPerPix are negative!** (the scale is good, but the sign is... well, 'wrong'?). I'm assuming there's a good under the hood reason for this but the end result is that the Math.Max statement as line 269 must really be Math.Min (or more robust: max(abs(x), abs(y)) with abs on both 270 and 271 too to remove all signs)

In my replication code, If I replace 269 to Math.Min to get the largest negative number, everything behaves as expected... 
(though middle mouse zoom to fit doesn't work right so the fix should really be deep inside the package and not just a superficial homebrewed patch as I've done... This is when I registered for github and wrote all this)

Hopefully there's enough details in there to get you on track (I'm in the process of learning C# and this package is outstanding).

### System Information
* **ScottPlot Version:** 4.1.16
* **Operating System:** Win10 x64
* **Application Type:** WPF

[...]

Issue #1165, opened 2021-07-26
Controls: Remove timer-based auto-render system


swharden:
This issue tracks discussion/PRs related to removal of the automatic render system.

**Problem:** After adding data to a `Plot` in a user control, that control's `Render()` must be called to display the change. Currently there is an auto-render system which periodically checks the number of plottable objects and calls `Render()` if it changes. This timer system is prone to memory leaks and may also result in unwanted renders at inopportune times (#1117). 

```cs
formsPlot1.Plot.AddScatter(xs, ys);
formsPlot1.Render(); // <-- this line is required if automatic renders are disabled
```

**Upgrading:** If the auto-render functionality is removed, users who upgrade will have to add the `Render()` line themselves. To make this more obvious an error message will be displayed on the plot when the debugger is attached if the control is displayed but the user never called `Render()` on start-up.

![image](https://user-images.githubusercontent.com/4165489/126996033-c93d47a2-75b7-4a35-921d-b7a183636d96.png)

gigios:
Sorry but for me it's not clear how update my code 🤔
First I initialize the plot by setting some properties and adding my 'signalplot' values, then I call the Render method (at the end of all the plot initialization).
Probably the problem on my side is that, when update the values data of my plot, I'm using the RenderRequest of the WpfPlot control.
If instead to use the RenderRequest I'm using the Render method the warning message disappear and the plot is shown correctly.

Is it wrong to use the method RenderRequest?

swharden:
@gigios thank you for reporting this! This is a bug and I'll track its progress in #1251

A temporary fix is to disable the warning
```cs
formsPlot1.Configuration.WarnIfRenderNotCalledManually = false;
```

PR #1191, opened 2021-08-06
Creates PlotInitialized event

bclehmann:
**New Contributors:**

**Purpose:**
Adds a `PlotInitializedEvent`, which should address #1171

**New Functionality:**
Now you can do things like this:

```cs
wpfPlot.PlotInitialized += Foo;
```

### DIFF
--- a/src/ScottPlot/Control/Backend.cs
+++ b/src/ScottPlot/Control/Backend.cs
@@ -98,6 +98,11 @@ namespace ScottPlot.Control
         public event EventHandler PlottableDropped = delegate { };
 
         /// <summary>
+        /// This event is invoked every time the plot is reset and is ready to be configured.
+        /// </summary>
+        public event EventHandler PlotInitialized = delegate { };
+
+        /// <summary>
         /// The control configuration object stores advanced customization and behavior settings
         /// for mouse-interactive plots.
         /// </summary>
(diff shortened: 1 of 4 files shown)


bclehmann:
It doesn't look like this resolves the issue, we might need to fire the event later?

src/ScottPlot/Control/Backend.cs
```diff
@@ -97,6 +97,11 @@ public class ControlBackEnd
         /// </summary>
         public event EventHandler PlottableDropped = delegate { };
 
+        /// <summary>
+        /// This event is invoked every time the plot is reset and is ready to be configured.
+        /// </summary>
+        public event EventHandler PlotInitialized = delegate { };
```
swharden, line 103: I haven't read #1171 yet, but just reading the handler name and XML summary I'm still unclear what "ready to be configured" means or what the use case is. 

After reading #1171 I have a better idea. Maybe this should be renamed to `AxesInitialized`? And/or adding more info to the XML summary may be helpful...

bclehmann: Aye, I think this can be better named, but it is vague because I think I need to invoke this event later, as it doesn't look like this resolves the issue.

swharden:
This issue will be resolved when the auto-render system is removed (#1165)

swharden:
Oops, I got confused... while not fixing #1165, this PR still has merit 😅👍

bclehmann:
Well, it only has merit if I finish it and I haven't touched it in a long time.

I think I'll close it and we can revisit it later if need be.

PR #1200, opened 2021-08-10
FinancePlot: fix out-of-range error in GetAxisLimits() when plot is empty

Maoyao233:
**New Contributors:**

**Purpose:**
Funciton `GetAxisLimits` of FinancePlot visits `OHLCs[0]` without bounds checking. It will cause an exception when plot is empty.

**New Functionality:**
None


swharden:
Good catch @Maoyao233, thanks for the PR!

PR #1205, opened 2021-08-12
Signal Plot: fix fill render error

### LINKED ISSUE #1163: SignalPlot: Filled plot render error
image

This may be related to errors reported in #1128

There is also a HARD CRASH when zooming near the transition area between low/high density when using density gradient mode:

crash2

swharden:
resolves #1163


swharden:
Wow, signal plots were deeply broken:

* hard crashes when using density mode and zooming in/out near the transition between high density and regular density
* bad horizontal fill alignment in every case
* gradient fills weren't properly aligned vertically

This PR fixes these issues and I think all is well now

Issue #1179, opened 2021-08-03
Demo: Pan and zoom one axis at a time


swharden:
### Discussed in https://github.com/ScottPlot/ScottPlot/discussions/1176 and #1158

<div type='discussions-op-text'>

<sup>Originally posted by **kkaiser41** August  3, 2021</sup>
I am trying to independently pan and zoom individual axis on a multiple axis plot. The link below shows the functionality I am trying to achieve. Any ideas how to do this? https://github.com/ScottPlot/ScottPlot/issues/162#issuecomment-704977569

![](https://user-images.githubusercontent.com/4165489/91741604-f52f1200-eb82-11ea-9361-17641d9f2b96.gif)

</div>

EmanuelFeru:
I might use multiaxes in the future. In the way I use it, I have always the legend with signals next to the plot, so what you have more or less applies to me:
* no signal selected, zoom all axes
* specific signals selected (can be multiple in my case), zoom the selected axes
![image](https://user-images.githubusercontent.com/24780745/128131311-df781575-8279-4bc9-b178-7dbaaa720d71.png)

I think the question is how to do it with just the plot... maybe then two modes:
*  Manual mode (user to provide selection in the code), this is the situation I have and what you presented.
      - This manual mode is useful because, in my case, to save space, I would prefer to only display the Y axis for the selected signal, e.g., if one signal is selected (show only one Y axis, all the other Y axes are hidden/not shown).
      - Default zoom (no axes selected): all axes zoomed together.
*  Auto mode (all Y axes shown):
      - Selection: user clicks the specific Y axis area, this selects the axis to be zoomed. With "Ctrl" key pressed, used can select more than one axis
      - Reset: With "Ctrl" key pressed, user clicks inside the plot area, then we go to Default zoom
      - Default zoom (no axes selected): all axes zoomed together.

kkaiser41:
How are you zooming only the selected Y axis?

I feel like I am overlooking how to address each Y axis separately. From what I have seen so far pan and zoom are hooked to the primary axis.

I am new to ScottPlot so I apologize if I am just missing that part.

swharden:
> How are you zooming only the selected Y axis?

The functionality shown in the screenshot is from an old beta version of ScottPlot. This functionality is not currently possible in ScottPlot, but I hope to add it shortly.

## Notes

Using panning as an example, the event is not axis aware:

[...]

Issue #1197, opened 2021-08-08
Resize: automatically zoom to fit data


noob765:
In engineering, equal axes are a big deal to the point of often being a show stopper... and while describing the functionality provided here as 'wrong' might seem like hyperbole, the user experience falls short of what is provided by other packages (and implicitly, what is expected). 

First a demonstration of the issue:  Set any plot to EqualScaleMode.ZoomOut and the axes will be equal alright (so it's sort of the correct behavior), but testing user experience demands we go further. Grab the lower right corner of your window and drag it right, then down, then left, then up, then right... Keep doing this resizing the window one axis at a time and watch your plot shrink smaller and smaller until it becomes but a blip in the center of the screen surrounded by ever growing axes... Shrinking data as I shrink the plot? Yes. Same data size as I expand it? No! It should expand too! (that's the 'wrong' behavior, which is truly annoying).

After finally figuring out what properties represent what in there, and that DataSizePx is NOT the pixel size the data needs (which would make sense considering how it's used in EnforceEqualAxisScales() ), but in fact the max pixel size that data HAS available, the source of the problem became obvious.

In EnforceEqualAxisScales, the line
```cs
double num = Math.Max(XAxis.Dims.UnitsPerPx, YAxis.Dims.UnitsPerPx);
```
can't possibly use UnitsPerPx since it's determined using DataSizePx. Equal scaling must be determined considering what the data would actually need to get plotted at a given scale, not how much it has available.

The following is a hacked EnforceEqualAxisScales that uses C&P code from AxisAutoX/Y to determine data extents. It's ugly, it doesn't respect any kind of margins, but it works as expected:

```cs
  case EqualScaleMode.ZoomOut:
        double minX = double.NaN;
        double maxX = double.NaN;
        double minY = double.NaN;
        double maxY = double.NaN;
        int xAxisIndex = 0;
        int yAxisIndex = 0;

        var plottableXLimits = Plottables.Where(x => x is IPlottable)
                                           .Select(x => (IPlottable)x)
                                           .Where(x => x.IsVisible)
                                           .Where(x => x.XAxisIndex == xAxisIndex)
                                           .Select(x => x.GetAxisLimits())
                                           .ToArray();

        foreach (var limits in plottableXLimits)
        {
            if (!double.IsNaN(limits.XMin))
                minX = double.IsNaN(minX) ? limits.XMin : Math.Min(minX, limits.XMin);
            if (!double.IsNaN(limits.XMax))
[...]

PR #1210, opened 2021-08-18
Axis: LockLimits()

swharden:
This PR adds methods to lock axes to prevent them from being panned/zoomed. Solves #1179

multi


bclehmann:
The naming here sounds a little bit too similar to `AxisScaleLock`, at least to me.

Maybe `DisableInteractivity`? Let me know what you're thinking.

swharden:
> The naming here sounds a little bit too similar to AxisScaleLock

Good point! I haven't done a deep dive into this topic yet, but according to #1197 some of the existing language results in unexpected behavior and may benefit from improvement.

I'm not totally sold on `DisableInteractivity()` because it's not clear if interactivity only applies to mouse events. I'm thinking a locked axis should remain unchanged if the axis limits are set manually or zoom/pan is called programmatically.

I agree with you `Lock()` is too ambiguous. Maybe `LockLimits()`?

swharden:
@bclehmann what do you think of the new API and demos? 

Are you up for making the Avalonia demo or do you want me to do that? I'm cool either way

bclehmann:
I don't mind doing it for Avalonia but I can't do it tonight. If you've already written a WPF version it's normally trivial to turn that into Avalonia.

bclehmann:
I like the API, it looks like you avoided splitting it up so there could be different functions for disabling zoom and pan. I wonder if disabling pan but not zoom is a common usecase, if someone had a plot that was only drawn from x1 to x2 panning outside of that range is not useful.

It seems like that usecase would be better served by something like `RestrictTo(start, end)` or`RestrictTo(x1, y1, x2, y2)` anyways.

As for the demo, it seems fine, we mostly avoid MVVM elsewhere, I don't know if that was intentional or not? It is more WPF-ey, and I don't think it should be confusing to new people.

swharden:
Thanks for your feedback @bclehmann!

> it looks like you avoided splitting it up into different functions for disabling zoom and pan

I left the door open to possibly implement more similar functions later:

* `LockSpan()` - allows min/max to change together but the delta remains fixed. Resizing the figure stretches the contents.

* `LockScale()` - keeps the units/pixel fixed. Resizing the figure does not change the scale of the contents.

> mostly avoid MVVM [in demos] ... intentional or not?

It usually is the result of making the demo first for WinForms then copy/pasting to realize the same effect in WPF as a code-behind. Admittedly demos that use designs idiomatic to each platform would be better, but are a little more work to create.

> I don't mind doing it for Avalonia but I can't do it tonight. If you've already written a WPF version it's normally trivial to turn that into Avalonia, so it's not much work either way.

No rush! If you're up for making the Avalonia demo that would be great. I can leave this PR open so you can create a PR into this branch, or I can merge this one into master and you can make a new branch/PR when you want - let me know what sounds good to you.

You're welcome to copy the WPF design (code-behind) or do something more idiomatic for Avalonia if you want to get fancy!

bclehmann:
I made a PR. It's fairly similar to the WPF version. #1215

PR #1215, opened 2021-08-20
Demo (Avalonia): Multi-Axis Lock

bclehmann:
**New Contributors:**

**Purpose:**
#1210

**New Functionality:**
Adds multi-axis lock demo for Avalonia. There are some bigger changes from the WPF version as Avalonians normally do MVVM with observables, often with `INotifyPropertyChanged`, not event handlers. This is because event handlers can be fired before the data context is updated. So generally event handlers are used only for code-behind where you're setting the value in the handler. I think most Avalonia users use something like ReactiveUI so their viewmodels are less verbose.


swharden:
This looks great @bclehmann, thanks! It's great to see an example of "proper" MVVM pattern in this demo

PR #1235, opened 2021-09-02
Microcharts colorset

arthurits:
**Purpose:**
Add a new colorset using the colors from Microcharts.

**New Functionality:**
10 colors grouped under a new Colorset


bclehmann (changes requested):
I think you need to add a line in `src/ScottPlot/Drawing/Palette.cs` so the colour set can be used easily.

bclehmann:
I also don't know if there are legal implications to this change, the default colourset is from matplotlib, but they use a different license. In fact different parts of matplotlib are under different licenses and it's complicated to find out which one you need to comply with.

arthurits:
> I think you need to add a line in `src/ScottPlot/Drawing/Palette.cs` so the colour set can be used easily.

That's very true! Thanks for letting me know. I just added a new commit modifying `Palette.cs`.

> I also don't know if there are legal implications to this change...

Me neither: I'm at a loss here. Should any legal problem could be foreseen, maybe the most sensible option would be to disregad this PR.

bclehmann (approved):

bclehmann:
The code looks good, I think legally this is fine, but we may need to add a note in the LICENSE or somewhere similar

swharden:
> The code looks good

I agree! Thanks @arthurits 

I added a cookbook recipe too which shows-off this palette

palette_microcharts

> I think legally this is fine, but we may need to add a note in the LICENSE or somewhere similar

I put a note about the MIT license and original copyright in the source code of the derived work. I'm pretty sure this is sufficient.

PR #1242, opened 2021-09-03
RadialGauge: plot and recipe

arthurits:
**Purpose:** Create a new `RadialGauge` plot type and add cookbook recipes

```cs
plt.Palette = ScottPlot.Drawing.Palette.Nord;
double[] values = { 100, 80, 65, 45, 20 };
plt.AddRadialGauge(values);
```

image


src/ScottPlot/Plottable/RadialGaugePlot.cs
```diff
@@ -0,0 +1,709 @@
...
+                int len = 32;
+                if (start + len >= text.Length) len = text.Length - start;
+                string substring = text.Substring(start, len);
+
+                // Measure the characters.
+                List<RectangleF> rects =
+                    MeasureCharactersInWord(gfx, font, clientRectangle, substring);
+
+                // Remove lead-in for the first character.
+                if (start == 0) x += rects[0].Left;
```
bclehmann, line 687: Our autoformatter was ok with this? I'm mildly surprised.

I normally prefer normal if statements, but idk how Scott feels.

src/ScottPlot/Plottable/RadialGaugePlot.cs
```diff
@@ -0,0 +1,709 @@
...
+        /// <returns></returns>
+        private List<RectangleF> MeasureCharacters(Graphics gfx, System.Drawing.Font font, RectangleF clientRectangle, string text)
+        {
+            List<RectangleF> results = new List<RectangleF>();
+
+            // The X location for the next character.
+            float x = 0;
+
+            // Get the character sizes 31 characters at a time.
+            for (int start = 0; start < text.Length; start += 32)
```
bclehmann, line 675: Why is this in batches of 32?

src/ScottPlot/Plottable/RadialGaugePlot.cs
```diff
@@ -0,0 +1,709 @@
...
+            {
+                _MaxScale = value;
+                ComputeAngularData();
+            }
+        }
+        private double _MaxScale;
+
+        /// <summary>
+        /// The minimum value for scaling the gauges.
+        /// This value is associated to <see cref="StartingAngleGauges"/> + <see cref="AngleRange"/>.
```
bclehmann, line 72: I think you copy-pasted this comment from above incorrectly

src/ScottPlot/Plottable/RadialGaugePlot.cs
```diff
@@ -0,0 +1,709 @@
...
+        /// <summary>
+        /// The maximum angular interval that the gauges will consist of.
+        /// It takes values in the range [0°-360°], default value is 360°. Outside this range, unexpected side-effects might happen.
+        /// </summary>
+        public double AngleRange
+        {
+            set
+            {
+                // Modify value to be in the range [0, 360]
+                _AngleRange = value > 360 ? 360 : (value < 0 ? 0 : value);
```
bclehmann, line 94: I think you can use `ReduceAngle` for this

src/ScottPlot/Plottable/RadialGaugePlot.cs
```diff
@@ -0,0 +1,709 @@
...
+        /// </summary>
+        public string[] GaugeLabels;
+
+        /// <summary>
+        /// Colors for each gauge. These colors are dimmed according to <see cref="DimPercentage"/> to draw the gauges' background.
+        /// Length must be equal to the length of data passed to either the constructor or the <see cref="Update(double[])"/> method.
+        /// </summary>
+        public Color[] GaugeColors;
+
+        /// <summary>
```
bclehmann, line 114: I think this should be renamed and the comments changed. This looks like its from `RadarPlot`, but there isn't a web here.

src/ScottPlot/Plottable/RadialGaugePlot.cs
```diff
@@ -0,0 +1,709 @@
...
+        /// <see langword="Color"/> of the value labels drawn inside the gauges.
+        /// </summary>
+        public Color GaugeLabelsColor { get; set; } = Color.White;
+
+        /// <summary>
+        /// Size of the gague label text as a percentage of the gauge width.
+        /// Values in the range [0-100], default value is 75 [percent]. Other values might produce unexpected side-effects.
+        /// </summary>
+        public float GaugeLabelsFontPct
```
bclehmann, line 231: I would rename this, maybe something like `GaugeLabelFontSize`?

I don't think the current name makes it clear that this changes font size, and in C# we often avoid abbreviations like `pct`. I also don't know if the fact that this is a fraction is as important as what it does (changes font size).

src/ScottPlot/Plottable/RadialGaugePlot.cs
```diff
@@ -0,0 +1,709 @@
...
+        /// </summary>
+        public Drawing.Font Font { get; set; } = new();
+
+        /// <summary>
+        /// <see langword="True"/> if value labels are shown inside the gauges.
+        /// Size of the text is set by <see cref="GaugeLabelsFontPct"/> and color by <see cref="GaugeLabelsColor"/>.
+        /// </summary>
+        public bool ShowGaugeValues { get; set; } = true;
+
+        public System.Drawing.Drawing2D.LineCap EndCap { get; set; } = System.Drawing.Drawing2D.LineCap.Triangle;
```
bclehmann, line 249: This is a nice way of doing this.

src/ScottPlot/Plottable/RadialGaugePlot.cs
```diff
@@ -0,0 +1,709 @@
...
+            $"RadialGauge with {DataRaw.Length} points.";
+
+        /// <summary>
+        /// Replace the data values with new ones. This passed data is copied and stored in <see cref="DataRaw"/>.
+        /// Implicitly calls the <see cref="ComputeAngularData"/> routine
+        /// </summary>
+        /// <param name="values">Array of values to be plotted as gauges.</param>
+        public void Update(double[] values)
+        {
+            DataRaw = new double[values.Length];
```
bclehmann, line 286: How come we need to copy this instead of just assigning it? Do we mutate it?

src/ScottPlot/Plottable/RadialGaugePlot.cs
```diff
@@ -0,0 +1,709 @@
...
+            // Compute MaxScale and MinScale values and transform DataRaw into DataAngular
+            ComputeMaxMin();
+            ComputeAngularData();
+        }
+
+        /// <summary>
+        /// Returns the original data in case it should be needed.
+        /// </summary>
+        /// <returns>Original data vector</returns>
+        public double[] GetData() => DataRaw;
```
bclehmann, line 298: I think you have a getter for this already. If not, I would just change it to something like this:

```cs
public double DataRaw { public get; private set; }
```

src/ScottPlot/Plottable/RadialGaugePlot.cs
```diff
@@ -0,0 +1,709 @@
...
+        public double[,] GetAngularData() => DataAngular;
+
+        /// <summary>
+        /// Converts <see cref="DataRaw"/> into <see cref="DataAngular"/>.
+        /// Depends on <see cref="DataRaw"/>, <see cref="GaugeMode"/>, <see cref="GaugeDirection"/>, <see cref="StartingAngleGauges"/>, <see cref="AngleRange"/>, and <see cref="MaxScale"/>
+        /// </summary>
+        private void ComputeAngularData()
+        {
+            // Check if there's data
+            if (DataRaw == null) return;
```
bclehmann, line 313: I think this should throw, not fail silently.

And maybe the null check should be in update?

src/ScottPlot/Plottable/RadialGaugePlot.cs
```diff
@@ -0,0 +1,709 @@
...
+            DataAngular = new double[DataRaw.Length, 2];
+
+            // Internal variables
+            float AngleSumPos = _StartingAngleGauges;
```
bclehmann, line 317: Do these need to be different variables? You do the same thing to them either way.

src/ScottPlot/Plottable/RadialGaugePlot.cs
```diff
@@ -0,0 +1,709 @@
...
+        /// </summary>
+        /// <param name="deep"></param>
+        public void ValidateData(bool deep = false)
+        {
+            if (GaugeLabels != null && GaugeLabels.Length != DataRaw.Length)
+                throw new InvalidOperationException("Gauge labels must match size of data values");
+        }
+
+        /// <summary>
+        /// Needed as part of IPlottable in ScottPlot.ScottForm
```
bclehmann, line 372: You don't need this comment and I don't think it achieves much.

src/ScottPlot/Plottable/RadialGaugePlot.cs
```diff
@@ -0,0 +1,709 @@
...
+            return legendItems.ToArray();
+        }
+
+        /// <summary>
+        /// Needed as part of IPlottable in ScottPlot.ScottForm
+        /// </summary>
+        /// <returns></returns>
+        public AxisLimits GetAxisLimits() =>
+            (GaugeLabels != null) ? new AxisLimits(-3.5, 3.5, -3.5, 3.5) : new AxisLimits(-2.5, 2.5, -2.5, 2.5);
+
```
bclehmann, line 402: Why does this depend on whether labels are present or not?

src/ScottPlot/Plottable/RadialGaugePlot.cs
```diff
@@ -0,0 +1,709 @@
...
+        /// <summary>
+        /// Reduces an angle into the range [0°-360°]
+        /// </summary>
+        /// <param name="angle">Angle value</param>
+        /// <returns>Return the angle whithin [0°-360°]</returns>
+        private double ReduceAngle(double angle)
```
bclehmann, line 408: This return the wrong answer for negative numbers.

i.e. `ReduceAngle(-365) == -725` 

I think a modulus based approach would be simpler.

```cs
private double ReduceAngle(double angle)
{
    double reduced = angle % 360;
    if (angle < 0) {
        angle += 360;
    }
}
```

src/ScottPlot/Plottable/RadialGaugePlot.cs
```diff
@@ -0,0 +1,709 @@
...
+        /// <summary>
+        /// This is where the drawing of the plot is done
+        /// </summary>
+        /// <param name="dims">Plot dimensions</param>
+        /// <param name="bmp">Bitmap where the drawing is done</param>
+        /// <param name="lowQuality">Image quality</param>
+        public virtual void Render(PlotDimensions dims, Bitmap bmp, bool lowQuality = false)
+        {
+            int numGroups = DataRaw.Length;
+            double minScale = new double[] { dims.GetPixelX(1), dims.GetPixelY(1) }.Min();  // Not sure why, but GetPixelX(1) returns a reasonable dimension to draw the plot
```
bclehmann, line 429: I think you should do this:

```cs
double minScale = new double[] { dims.PxPerUnitX, dims.PxPerUnitX }.Min();
```

src/ScottPlot/Plottable/RadialGaugePlot.cs
```diff
@@ -0,0 +1,709 @@
...
+        /// <summary>
+        /// Color of the axis lines and concentric circles representing ticks
+        /// </summary>
+        public Color WebColor { get; set; } = Color.Gray;
+
+        /// <summary>
+        /// Gets or sets the size (in pixels) of each gauge.
+        /// If less than 0, then it will be calculated from the available space.
+        /// </summary>
+        public float LineWidth { get; set; } = -1;
```
bclehmann, line 123: I think you should use a nullable float `float?` rather than a sentinel.

src/ScottPlot/Plottable/RadialGaugePlot.cs
```diff
@@ -0,0 +1,709 @@
...
+            float maxBackAngle = (GaugeDirection == RadialGaugeDirection.AntiClockwise ? -1 : 1) * (NormBackGauge ? (float)AngleRange : 360);
+
+            pen.Width = (float)lineWidth;
+            pen.StartCap = StartCap;
+            pen.EndCap = EndCap;
+            penCircle.Width = (float)lineWidth;
+            penCircle.StartCap = System.Drawing.Drawing2D.LineCap.Round;
+            penCircle.EndCap = System.Drawing.Drawing2D.LineCap.Round;
+
+            using System.Drawing.Font fontGauge = new(Font.Name, lineWidth * GaugeLabelsFontPct / 100, FontStyle.Bold);
```
bclehmann, line 449: I think you can just use the font object that was given rather than creating a new one. I'm pretty sure you can still modify the font size and weight.

src/ScottPlot/Plottable/RadialGaugePlot.cs
```diff
@@ -0,0 +1,709 @@
...
+
+            lock (this)
```
bclehmann, line 451: It's bad practice to lock on yourself, but why do we even need this lock at all?

src/ScottPlot/Plottable/RadialGaugePlot.cs
```diff
@@ -0,0 +1,709 @@
...
+                double x = cx + radius * Math.Cos(theta);
+                double y = cy + radius * Math.Sin(theta);
+
+                // Transform to position the character.
+                if (isBelow)
+                    gfx.RotateTransform((float)(RadToDeg * (theta - Math.PI / 2)));
+                else
+                    gfx.RotateTransform((float)(RadToDeg * (theta + Math.PI / 2)));
+
+                gfx.TranslateTransform((float)x, (float)y, System.Drawing.Drawing2D.MatrixOrder.Append);
```
bclehmann, line 609: Can you pass in `x` and `y` instead of `(0, 0)`, and then remove the translation? This would avoid issues around the noncommutativity of matrix multiplication.

bclehmann (changes requested):
It looks good, there's just a few changes around code cleanliness and the like.

A healthy chunk of them are boring things like renames and default parameters.

bclehmann:
I know that list of suggestions is pretty big, but a lot of them are trivial.

There is the big one around the section that is yoinked from somewhere and is probably not licensed.

src/cookbook/ScottPlot.Cookbook/Recipes/Plottable/RadialGauge.cs
```diff
@@ -0,0 +1,325 @@
...
+        public string Description =>
+            "The gauges' background are drawn as full circles by default. If normalization is set to true, background gauges are limited to the values of StartingAngle and AngleRange.";
+
+        public void ExecuteRecipe(Plot plt)
+        {
+            plt.Palette = ScottPlot.Drawing.Palette.Nord;
+            double[] values = { 100, 80, 65, 45, 20 };
+
+            var RadialGauge = plt.AddRadialGauge(values);
+            RadialGauge.NormBackGauge = true;
```
bclehmann, line 321: This is kind of a weird name. I think angle range should be nullable, and if it's null it's treated as if it's 360. If not, you use that number.

arthurits:
> I know that list of suggestions is pretty big, but a lot of them are trivial.

Thanks for the suggestions. I'm not sure, however, if either @swharden or anyone else is supposed to implement them.
I would say that some of them are trivial for my coding level, but many others I just simply don't know, don't really understand, or are personal preferences. I'm OK with implementing the trivial ones, but would appreciate some help for the ones I'm not qualified to carry out.

> There is the big one around the section that is yoinked from somewhere and is probably not licensed.

That's correct. The color brightness routine was taken from [here](https://gist.github.com/zihotki/09fc41d52981fb6f93a81ebf20b35cd5) and the text along circle from [here](http://csharphelper.com/blog/2018/02/draw-text-on-a-circle-in-c/).
I believe I leaved a reference note with both links at the corresponding summary function. Not sure about any legal or license issue though.

swharden:
Hi @arthurits, thanks for this PR! @bclehmann thanks for your insightful comments too. 

@arthurits and I corresponded a bit over email and I encouraged them to submit a PR even though they are somewhat new to the language and don't have much experience using GitHub for things like PRs. If they hit the edge of their knowledge, we can help fill-in the gaps and we all can benefit from the experience 👍 

## Licensing

> probably not licensed

Clarifying why this is important, ScottPlot is released under a MIT license (one of the most permissive licenses out there). If we include code whose source is a more restrictive license (e.g., GPL) those licenses trickle-up and ScottPlot must change its license too (and possibly all the applications people create that use ScottPlot). To avoid this we ensure all code in ScottPlot can be released under the MIT license.

> The color brightness routine was taken from [here](https://gist.github.com/zihotki/09fc41d52981fb6f93a81ebf20b35cd5) and the text along circle from [here](http://csharphelper.com/blog/2018/02/draw-text-on-a-circle-in-c/)

Neither source has an explicit license, so I think it's okay to include as long as the source is referenced - a code comment in the function seems OK with a link to the URL.

## Next Steps

> I'm not sure, however, if either @swharden or anyone else is supposed to implement them. I would say that some of them are trivial for my coding level, but many others I just simply don't know, don't really understand, or are personal preferences. 

@arthurits, you can make the changes you feel comfortable with and can mark those code review comments as "resolved".

Let us know when you're done and @bclehmann and I can work together to figure out what to do about all the remaining comments that are left open.

What do you think about this plan?

bclehmann:
> Neither source has an explicit license, so I think it's okay to include as long as the source is referenced - a code comment in the function seems OK with a link to the URL.

If something is not explicitly licensed it is fully protected by copyright.That would make our usage of it illegal.

This issue may go away if we decide to use HSL for lightening/darkening. Otherwise though, we would need to reimplement it or negotiate a license with the original author.

bclehmann:
As for any of my suggestions, you can tackle any of the ones that you want and I can come back and help with whatever's left.

src/ScottPlot/Plottable/RadialGaugePlot.cs
```diff
@@ -0,0 +1,709 @@
...
+        /// <param name="radius">Radius of the circle in pixels</param>
+        /// <param name="angleInit">Staring angle (in degrees) of the circle</param>
+        /// <param name="angleSwept">Angular span (in degrees) of the circle</param>
+        /// <param name="cx">The x-coordinate of the circle centre</param>
+        /// <param name="cy">The y-coordinate of the circle centre</param>
+        /// <param name="text">String to be drawn</param>
+        /// <param name="posPct">Position of text as a percentage of the angle swept</param>
+        /// <seealso href="http://csharphelper.com/blog/2018/02/draw-text-on-a-circle-in-c/"/>
+        protected virtual void DrawTextOnCircle(Graphics gfx, System.Drawing.Font font,
+            Brush brush, RectangleF clientRectangle, float radius, float angleInit, float angleSwept, float cx, float cy,
```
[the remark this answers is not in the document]
arthurits: What about this?

`angleInit` -> `start`
`angleSwept` -> `sweep` This is really an angular span: if `start` is 90 and `sweep` is 15, then the `end` or `total` angle would be 90+15
`cx` and `cy` - The function already defines and uses local variables named `x` and `y`
`posPct` -> `pos`

src/ScottPlot/Plottable/RadialGaugePlot.cs
```diff
@@ -0,0 +1,709 @@
...
+
+            // Measure the characters. Use LINQ to add up the character widths.
+            List<RectangleF> rects = MeasureCharacters(gfx, font, clientRectangle, text);
+            var width_query = from RectangleF rect in rects select rect.Width;
+            double text_width = width_query.Sum() / radius;
+
+            // Angular data
+            bool isPositive = angleSwept >= 0;
+            double angle = ReduceAngle(angleInit + angleSwept * (posPct / 100));
+            angle += (1 - 2 * (posPct / 100)) * (isPositive ? 1 : -1) * RadToDeg * text_width / 2; // Set the position to the middle of the text
```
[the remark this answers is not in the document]
arthurits: I'm not sure whether `Math.Sign` being equal 0 could be an issue or not.
At least for me, the code would be more clear using `isPositive` instead.

src/ScottPlot/Plottable/RadialGaugePlot.cs
arthurits: I believe that's because [SetMeasurableCharacterRanges](https://docs.microsoft.com/es-es/dotnet/api/system.drawing.stringformat.setmeasurablecharacterranges?view=net-5.0), which is called a few lines later inside
https://github.com/ScottPlot/ScottPlot/blob/d7f90e1346e91c60c85a45e2d2f75b36302bd19a/src/ScottPlot/Plottable/RadialGaugePlot.cs#L684
throws an `OveflowException` when the array passed is longer than 32 items.

src/ScottPlot/Plottable/RadialGaugePlot.cs
arthurits: I'm not really sure. While `ReduceAngle` should transform any angle into the [0-360] range, here we are doing a cut-off (something like a hard limiter filter in audio processing, or a king of 'angular trimming').

For instante, let's take a value of -365:
- `ReduceAngle` should return 355
- Here we want to get 0

src/ScottPlot/Plottable/RadialGaugePlot.cs
arthurits: That's true. I first used `RadarPlot` as a template for `RadialGauge` thinking they would share most of the code. It turned out it wasn't the case. This is one piede of code I forgot to delete because it was linked to the initialization of these `Pen` instances:
https://github.com/ScottPlot/ScottPlot/blob/d7f90e1346e91c60c85a45e2d2f75b36302bd19a/src/ScottPlot/Plottable/RadialGaugePlot.cs#L433-L434

I'll deleted this property and use a dummy color for the instances' initialization.

src/ScottPlot/Plottable/RadialGaugePlot.cs
arthurits: No, I believe that data is not mutated throughout the routines.
I was afraid of just keeping a reference to the data values. Since `DataRaw` is used several times in different functions, I thought there was a change the user could modify (or dispose of) the original data values; thus causing exceptions whenever that reference would be used. Copying data would avoid that.

src/ScottPlot/Plottable/RadialGaugePlot.cs
arthurits: That's true.
https://github.com/ScottPlot/ScottPlot/blob/d7f90e1346e91c60c85a45e2d2f75b36302bd19a/src/ScottPlot/Plottable/RadialGaugePlot.cs#L44-L48
However, I get [compiler error CS0274](https://docs.microsoft.com/en-us/dotnet/csharp/misc/cs0274?f1url=%3FappId%3Droslyn%26k%3Dk(CS0274)) when trying the suggested code.
I'm at a loss here.

src/ScottPlot/Plottable/RadialGaugePlot.cs
arthurits: I don't know what would be the best options here.

src/ScottPlot/Plottable/RadialGaugePlot.cs
arthurits: Yes. Although it seems they accomplish the same, they just keep the total angular count for positive and negative data values.
https://github.com/ScottPlot/ScottPlot/blob/d7f90e1346e91c60c85a45e2d2f75b36302bd19a/src/ScottPlot/Plottable/RadialGaugePlot.cs#L332-L335

src/ScottPlot/Plottable/RadialGaugePlot.cs
arthurits: This was copy-pasted from `RadarPlot`. This is needed as part of the IPlottable interface, and I don't know how to handle it here.

src/ScottPlot/Plottable/RadialGaugePlot.cs
arthurits: Sure!
By the way, I seems that using `PxPerUnit` results in the plot being too small regarding the size of the whole control. Visually there are some chucks of empty space at both the plot's top and bottom. The same thing happens with `RadarPlot`.
That's the only reason I tried `GetPixel` instead: the plot seemed to better fill the control area.

src/ScottPlot/Plottable/RadialGaugePlot.cs
arthurits: Sure, both properties can be modified in the original font object.
However that font is from `ScottPlot.Drawing`, and the `fontGauge` is from `System.Drawing` since it's used extensively in the gauge labels text drawing (used by the Graphics object `gfx` inside `DrawTextOnCircle`, `MeasureCharactersInWord`, and `MeasureCharacters`).
I'm also unsure whether modifying the original font object, just for the gauge labels, would also affect other plot text like title or legend...

src/ScottPlot/Plottable/RadialGaugePlot.cs
arthurits: It makes sense.
However, when I tested your suggestion the gauge labels didn't render.

src/cookbook/ScottPlot.Cookbook/Recipes/Plottable/RadialGauge.cs
arthurits: I'm OK with changing the name to whatever feels more appropiate.
Making it nullable is a big modification since this would affect the `RadialGauge` plot code in several places.

src/cookbook/ScottPlot.Cookbook/Recipes/Plottable/RadialGauge.cs
```diff
@@ -0,0 +1,325 @@
...
+    }
+
+    public class RadialGaugeLabelPos : IRecipe
+    {
+        public string Category => "Plottable: RadialGauge";
+        public string ID => "radialgauge_labelpos";
+        public string Title => "Gauge label position";
+        public string Description =>
+            "The position of the gauges' labels can be adjusted as a percentage of the gauges' angular length. " +
+            "A value of 100 (default) draws the labels at the gauges' end whereas a value of 0 does it at the gauge's beginning.";
```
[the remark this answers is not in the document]
arthurits: > I wonder if this should be on the interval [0, 1]? It's more common to express fractions this way.

No problem with using the [0, 1]. As a client however, I find it more useful when any percentage property is defined on a [0, 100] interval. If a [0, 1] is considered better, this implies several modifications in `RadialGauglePlot.cs`.

> Also I think the default shouldn't be zero, because the labels look very close to the end. And I think 0 should mean at the base of the gauge, not the tail.

Not sure I get the idea. I believe that a value of 0 draws the label at the beginning (I see that as the gauge's base) and not the ending.

arthurits:
Thanks @swharden and @bclehmann. I think that the following checklist might be useful to track down the state of each comment/issue.

#### Recipe suggestions
* [ ]  GaugeLabelPos could be in the interval [0, 1] instead of the interval [0, 100]
https://github.com/ScottPlot/ScottPlot/blob/d7f90e1346e91c60c85a45e2d2f75b36302bd19a/src/cookbook/ScottPlot.Cookbook/Recipes/Plottable/RadialGauge.cs#L238
* [ ]  GaugeLabelPos default value shouldn't be zero
https://github.com/ScottPlot/ScottPlot/blob/d7f90e1346e91c60c85a45e2d2f75b36302bd19a/src/cookbook/ScottPlot.Cookbook/Recipes/Plottable/RadialGauge.cs#L239
* [ ]  GaugeLabelsFontPct ¿does it need a maximum?
https://github.com/ScottPlot/ScottPlot/blob/d7f90e1346e91c60c85a45e2d2f75b36302bd19a/src/cookbook/ScottPlot.Cookbook/Recipes/Plottable/RadialGauge.cs#L258
* [x]  Typo in RadialGaugeStart (Title string: "staring")
https://github.com/ScottPlot/ScottPlot/blob/d7f90e1346e91c60c85a45e2d2f75b36302bd19a/src/cookbook/ScottPlot.Cookbook
* [x]  Titles should be in title case
https://github.com/ScottPlot/ScottPlot/blob/d7f90e1346e91c60c85a45e2d2f75b36302bd19a/src/cookbook/ScottPlot.Cookbook/Recipes/Plottable/RadialGauge.cs#L12
* [ ] `NormBackGauge`  is kind of a weird name. I think angle range should be nullable, and if it's null it's treated as if it's 360. If not, you use that number.
https://github.com/ScottPlot/ScottPlot/blob/d7f90e1346e91c60c85a45e2d2f75b36302bd19a/src/cookbook/ScottPlot.Cookbook/Recipes/Plottable/RadialGauge.cs#L321

#### Plottable suggestions
* [ ]  ChangeColorBrightness is verbatim
https://github.com/ScottPlot/ScottPlot/blob/d7f90e1346e91c60c85a45e2d2f75b36302bd19a/src/ScottPlot/Plottable/RadialGaugePlot.cs#L509
* [x]  ChangeColorBrightness can overflow
https://github.com/ScottPlot/ScottPlot/blob/d7f90e1346e91c60c85a45e2d2f75b36302bd19a/src/ScottPlot/Plottable/RadialGaugePlot.cs#L529
* [ ]  Usage of `region`
https://github.com/ScottPlot/ScottPlot/blob/d7f90e1346e91c60c85a45e2d2f75b36302bd19a/src/ScottPlot/Plottable/RadialGaugePlot.cs#L43
* [ ] Extract region Color routines
https://github.com/ScottPlot/ScottPlot/blob/d7f90e1346e91c60c85a45e2d2f75b36302bd19a/src/ScottPlot/Plottable/RadialGaugePlot.cs#L501
* [x] Change code to LINQ syntax and do the sum in one single line
https://github.com/ScottPlot/ScottPlot/blob/d7f90e1346e91c60c85a45e2d2f75b36302bd19a/src/ScottPlot/Plottable/RadialGaugePlot.cs#L581-L582
* [x] Rename arguments in `DrawTextOnCircle`
https://github.com/ScottPlot/ScottPlot/blob/d7f90e1346e91c60c85a45e2d2f75b36302bd19a/src/ScottPlot/Plottable/RadialGaugePlot.cs#L560-L562
* [ ] Use `Math.Sign` instead
https://github.com/ScottPlot/ScottPlot/blob/d7f90e1346e91c60c85a45e2d2f75b36302bd19a/src/ScottPlot/Plottable/RadialGaugePlot.cs#L584-L587
* [x] Rewrite `if` statement
https://github.com/ScottPlot/ScottPlot/blob/d7f90e1346e91c60c85a45e2d2f75b36302bd19a/src/ScottPlot/Plottable/RadialGaugePlot.cs#L686-L687
* [ ] ~~Characters in batches of 32~~
https://github.com/ScottPlot/ScottPlot/blob/d7f90e1346e91c60c85a45e2d2f75b36302bd19a/src/ScottPlot/Plottable/RadialGaugePlot.cs#L674-L675
* [ ] Correct copy-pasted comment from previous property
https://github.com/ScottPlot/ScottPlot/blob/d7f90e1346e91c60c85a45e2d2f75b36302bd19a/src/ScottPlot/Plottable/RadialGaugePlot.cs#L70-L74
* [ ] Use `ReduceAngle` instead
https://github.com/ScottPlot/ScottPlot/blob/d7f90e1346e91c60c85a45e2d2f75b36302bd19a/src/ScottPlot/Plottable/RadialGaugePlot.cs#L93-L94
* [x] Rename and change comments
https://github.com/ScottPlot/ScottPlot/blob/d7f90e1346e91c60c85a45e2d2f75b36302bd19a/src/ScottPlot/Plottable/RadialGaugePlot.cs#L114-L117
* [x] Rename `GaugeLabelsFontPct` to `GaugeLabelFontSize`
https://github.com/ScottPlot/ScottPlot/blob/d7f90e1346e91c60c85a45e2d2f75b36302bd19a/src/ScottPlot/Plottable/RadialGaugePlot.cs#L227-L231
* [ ] Other considerations regarding the name and use of `GaugeLabelsFontPct`
https://github.com/ScottPlot/ScottPlot/blob/d7f90e1346e91c60c85a45e2d2f75b36302bd19a/src/ScottPlot/Plottable/RadialGaugePlot.cs#L231
* [ ] Does data need to be copied into `DataRaw`?
https://github.com/ScottPlot/ScottPlot/blob/d7f90e1346e91c60c85a45e2d2f75b36302bd19a/src/ScottPlot/Plottable/RadialGaugePlot.cs#L284-L287
* [ ] Getter for `DataRaw`
https://github.com/ScottPlot/ScottPlot/blob/d7f90e1346e91c60c85a45e2d2f75b36302bd19a/src/ScottPlot/Plottable/RadialGaugePlot.cs#L294-L298
* [ ] `ComputeAngularData()` should throw, not fail silently
https://github.com/ScottPlot/ScottPlot/blob/d7f90e1346e91c60c85a45e2d2f75b36302bd19a/src/ScottPlot/Plottable/RadialGaugePlot.cs#L310-L313
* [ ] Should the null check be in `Update()`?
https://github.com/ScottPlot/ScottPlot/blob/d7f90e1346e91c60c85a45e2d2f75b36302bd19a/src/ScottPlot/Plottable/RadialGaugePlot.cs#L310-L313
* [ ] ~~Are both variables needed `AngleSumPos` and `AngleSumNeg`?~~
https://github.com/ScottPlot/ScottPlot/blob/d7f90e1346e91c60c85a45e2d2f75b36302bd19a/src/ScottPlot/Plottable/RadialGaugePlot.cs#L317-L318
* [ ] Why GetAxisLimits depend on labels?
https://github.com/ScottPlot/ScottPlot/blob/d7f90e1346e91c60c85a45e2d2f75b36302bd19a/src/ScottPlot/Plottable/RadialGaugePlot.cs#L400-L401
* [x] `ReduceAngle` returns wrong values for negative numbers
https://github.com/ScottPlot/ScottPlot/blob/d7f90e1346e91c60c85a45e2d2f75b36302bd19a/src/ScottPlot/Plottable/RadialGaugePlot.cs#L408
* [x] Modify `minScale` in `Render()`
https://github.com/ScottPlot/ScottPlot/blob/d7f90e1346e91c60c85a45e2d2f75b36302bd19a/src/ScottPlot/Plottable/RadialGaugePlot.cs#L426-L429
* [x] Nullable float `float?` instead of sentinel.
https://github.com/ScottPlot/ScottPlot/blob/d7f90e1346e91c60c85a45e2d2f75b36302bd19a/src/ScottPlot/Plottable/RadialGaugePlot.cs#L123
https://github.com/ScottPlot/ScottPlot/blob/d7f90e1346e91c60c85a45e2d2f75b36302bd19a/src/ScottPlot/Plottable/RadialGaugePlot.cs#L437
* [ ] Use the font object that was given rather than creating a new one
https://github.com/ScottPlot/ScottPlot/blob/d7f90e1346e91c60c85a45e2d2f75b36302bd19a/src/ScottPlot/Plottable/RadialGaugePlot.cs#L449
* [x] It's bad practice to lock on yourself, but why do we even need this lock at all?
https://github.com/ScottPlot/ScottPlot/blob/d7f90e1346e91c60c85a45e2d2f75b36302bd19a/src/ScottPlot/Plottable/RadialGaugePlot.cs#L451
* [ ] Can you pass in x and y instead of (0, 0), and then remove the translation?
https://github.com/ScottPlot/ScottPlot/blob/d7f90e1346e91c60c85a45e2d2f75b36302bd19a/src/ScottPlot/Plottable/RadialGaugePlot.cs#L607-L609

src/ScottPlot/Plottable/RadialGaugePlot.cs
```diff
@@ -0,0 +1,709 @@
...
+
+        #region Color routines
+
+        /// <summary>Creates color with corrected brightness.</summary>
+        /// <param name="color">Color to correct.</param>
+        /// <param name="correctionFactor">The brightness correction factor. Must be between -1 and 1. 
+        /// Negative values produce darker colors.</param>
+        /// <returns>Corrected <see cref="Color"/> structure.</returns>
+        /// <seealso href="http://gist.github.com/zihotki/09fc41d52981fb6f93a81ebf20b35cd5"/>
+        private Color ChangeColorBrightness(Color color, float correctionFactor)
```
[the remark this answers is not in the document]
arthurits: I was aware that the Color object as a method returning hue, saturarion, and [brightness](https://docs.microsoft.com/es-es/dotnet/api/system.drawing.color.getbrightness?view=net-5.0)
The only function I could find in the .NET API to convert between HSL and RGB can be found in 
 [Microsoft.VisualStudio.Modeling.Diagrams](https://docs.microsoft.com/es-es/dotnet/api/microsoft.visualstudio.modeling.diagrams.hslcolor?view=visualstudiosdk-2019) (Microsoft.VisualStudio.Modeling.Sdk.15.0.dll)

I'm not sure it would be a good idea to include that assembly reference into the project.

arthurits:
> _EDIT: We should be safe and not use it to avoid copyright issues (unless we get permission from the original author). These methods should probably be re-implemented by writing them from scratch._

This are the functions that were sourced from other places and their verbatim status:
* `ChangeColorBrightness()` has been already modified in order to avoid possible overflows, so it's not verbatim.
* `LightenBy()` and `DarkenBy()` where taken from one of the answers in [this StackFlow question](https://stackoverflow.com/questions/801406/c-create-a-lighter-darker-color-based-on-a-system-color). These are just one-liners calling `ChangeColorBrightness()`.
* `DrawTextOnCircle()` was modified from the [blog post here](http://csharphelper.com/blog/2018/02/draw-text-on-a-circle-in-c/). Some modifications were needed to fit the particular needs of the gauge labels (ie labels are aways drawn from left to right no matter independently of the gauge mode, direction and so on). So it's also no verbatim.
* `MeasureCharactersInWord()` and `MeasureCharacters()` are verbatim from [this blog post](http://csharphelper.com/blog/2015/02/measure-character-positions-when-drawing-long-strings-in-c/). Writing these two functions from scratch is way beyond my skills.

src/ScottPlot/Plottable/RadialGaugePlot.cs
arthurits: I'm afraid I'm failing to see your point.

bclehmann:
For the verbatim or nearly verbatim changes, those still need to be original work. For `LightenBy` and `DarkenBy` you're fine because there's no other way to write that and still use `ChangeColorBrightness`

The relevant rules with copyright are below: 
- Any modifications to a copyrighted work constitute a derivative work and are infringement unless explicitly licensed
    - This means to avoid infringement we need to rewrite the applicable function
- Any sufficiently long piece of text to be considered original and creative is protected
    - This means if a section is too short or is unoriginal it cannot be infringement 
        - e.g. there is no other way to write it that would be meaningfully different or the section is not the property of the author of the work it was taken from

I think it's stupid that copyright law is virtually identical between software and books, save for some different case law. Especially as many of the potentially infringing sections don't strike me as "creative" at all, which is supposed to be a requirement for protection. However, this is the law we have to follow. And even if things were to change in one country we'd likely have to follow it elsewhere because this is a project used internationally.

src/ScottPlot/Plottable/RadialGaugePlot.cs
```diff
@@ -0,0 +1,709 @@
...
+    /// A radial gauge chart is a graphical method of displaying scalar data in the form of 
+    /// a chart made of circular gauges so that each scalar is represented by each gauge.
+    /// 
+    /// Data is managed using a single array where each element is asigned to each gauge.
+    /// Internally this data is stored in a single array and is converted to angular paramters,
+    /// through ComputeAngularData(), which are more suitable for drawing purposes and stored in a 2D array.
+    /// </summary>
+    public class RadialGaugePlot : IPlottable
+    {
+        #region Properties & fields
```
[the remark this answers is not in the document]
swharden: This is a contentious topic! [Are #regions an antipattern or code smell?](https://softwareengineering.stackexchange.com/questions/53086/are-regions-an-antipattern-or-code-smell)

When you're tempted to reach for `region` it can be an indication it may be time to refactor something. Personally I only use regions in extremely large files (which itself is an anti-pattern). I find them useful to section-off large numbers of small functions, but partial classes (added in C# 2.0) are an alternative worth considering.

In this case the file isn't too big, so I'll probably remove them 👍

src/ScottPlot/Plottable/RadialGaugePlot.cs
swharden: > here we are doing a cut-off

This may have been overlooked because double ternary operators are pretty easy to misunderstand 😅 

```cs
_AngleRange = value > 360 ? 360 : (value < 0 ? 0 : value);
```

could be replaced by this which does the same thing but is less likely to be misunderstood

```cs
_AngleRange = value;
_AngleRange = Math.Min(360, _AngleRange);
_AngleRange = Math.Max(0, _AngleRange);
```

I'll probably simplify this further by wrapping that in a function `double TrimAngle(double angle)`

src/cookbook/ScottPlot.Cookbook/Recipes/Plottable/RadialGauge.cs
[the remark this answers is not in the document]
swharden: I'll refactor this so `GaugeLabelPosition` is a value `[0-1]` representing the position from base to tip

src/ScottPlot/Plottable/RadialGaugePlot.cs
swharden: @arthurits this is awesome! I wouldn't have thought of this 🚀

src/ScottPlot/Plottable/RadialGaugePlot.cs
swharden: I'll change this to `FontSizeFraction`

src/ScottPlot/Plottable/RadialGaugePlot.cs
swharden: > data is not mutated throughout the routines

I'll probably refactor data handling toward a functional style where `ComputeMaxMin()` and `ComputeAngularData()` are `static` and return tuples. 

This one-way type of data processing (rather than manipulating global state) is a little easier to extend and troubleshoot.

src/ScottPlot/Plottable/RadialGaugePlot.cs
swharden: I'm actually going to make this `private` to keep the API small. I will expect the end user to be responsible for maintaining their own data 😈 

@arthurits is correct that @bclehmann's code example throws a compile error. I couldn't figure out why at first, but eventually figured it out:

```cs
// this causes compile error CS0274
public double[] DataRaw { public get; private set; }
```

```cs
// this works
public double[] DataRaw { get; private set; }
```

src/ScottPlot/Plottable/RadialGaugePlot.cs
swharden: I'm refactoring these setters to throw exceptions rather than trim the data into compliance

src/ScottPlot/Plottable/RadialGaugePlot.cs
swharden: > You do the same thing to them either way

This is true. I refactored this section to be simpler:

```cs
float angleSum = StartingAngleGauges;

for (int i = 0; i < DataRaw.Length; i++)
{
    float angleSwept = (float)(AngleRange * DataRaw[i] / (MaxScale - MinScale));
    if (GaugeDirection == RadialGaugeDirection.AntiClockwise)
        angleSwept *= -1;

    double initialAngle = _GaugeMode == RadialGaugeMode.Stacked ? StartingAngleGauges : angleSum;
    angleSum += angleSwept;

    DataAngular[i, 0] = initialAngle;
    DataAngular[i, 1] = angleSwept;
}
```

src/ScottPlot/Plottable/RadialGaugePlot.cs
swharden: This method describes how large the plottable is on the coordinate system so when the user auto-scales (by calling `AxisAuto()` or middle-clicking a user control the axis limits are adjusted to include these dimensions.

I'll fix this up shortly 👍 I think the axis limits for this plot type depend on the number of groups and the line widths for each group

src/ScottPlot/Plottable/RadialGaugePlot.cs
swharden: I think the present code also breaks if angles are >720º, and the suggested code breaks if the angle is <-360º. I'll refactor this now to use something like @bclehmann's suggested method using a `while` instead of `if`... but it will break if the user passes-in `double.NegativeInfinity` or `double.NaN` 😆

It's probably more effort than I want to invest right now, but if there's shared rotational logic between multiple plot types it may be a good idea to eventually break this out into a static class somewhere and add several unit tests to verify it works as expected for many different values.

swharden: ... my bad, @bclehmann's code doesn't break for <-360º values, so that `while` I mentioned isn't needed

I'm still not going to work about `double.NegativeInfinity` or `double.NaN` 🙃

src/ScottPlot/Plottable/RadialGaugePlot.cs
swharden: > I think you can just use the font object that was given rather than creating a new one

It's a little deceptive since the `using` hides the full namespace, but @arthurits did it right because they needed to create a `System.Drawing.Font` from a `ScottPlot.Drawing.Font` 👍

swharden: ... although I will refactor it since there's a helper method for this

```cs
Font.Size = lineWidth * (float)FontSizeFraction;
using System.Drawing.Font fontGauge = GDI.Font(Font);
```

It's a can of worms so I won't discuss it here, but there's also a difference between fonts reported in Pt (System.Drawing default behavior) and Pixel size (ScottPlot default behavior). Making the code change above changes font size slightly breaking the hash tests I have in place, so I'll make this change last.

src/ScottPlot/Plottable/RadialGaugePlot.cs
swharden: > Can you pass in x and y instead of (0, 0), and then remove the translation? This would avoid issues around the noncommutativity of matrix multiplication.

No because the string still requires the rotation. I refactored this area to be a little more obvious

```cs
double halfPi = Math.PI / 2;
double rotationRadians = isBelow ? theta - halfPi : theta + halfPi;
double rotationDegrees = rotationRadians * RadToDeg;

gfx.RotateTransform((float)rotationDegrees);
gfx.TranslateTransform(x, y, System.Drawing.Drawing2D.MatrixOrder.Append);
gfx.DrawString(text[i].ToString(), font, brush, 0, 0, string_format);
gfx.ResetTransform();
```

src/ScottPlot/Plottable/RadialGaugePlot.cs
swharden: I may refactor this by moving some functionality into the `ScottPlot.Drawing` module and improve it along the way.

Overall there is a lot of code here to measure/render text character-by-character along a curve. Moving that logic out of this plot type will make it more testable and also simplify the code in this module 👍

src/ScottPlot/Plottable/RadialGaugePlot.cs
swharden: I typically go for two lines, but one line is permitted by the autoformatter. 

The recommended option is technically 4 lines (2 curly braces) but I'm not too strict about it.

I wrote some statements yesterday that I think read better as one-liners:

```cs
if (X > Width) { X -= X - Width; }
if (Y > Height) { Y -= Y - Height; }
if (X < 0) { X = -X; }
if (Y < 0) { Y = -Y; }
```

src/ScottPlot/Plottable/RadialGaugePlot.cs
swharden: Actually I went with this 😎 
```cs
double minScale = Math.Min(dims.PxPerUnitX, dims.PxPerUnitY);
```

src/cookbook/ScottPlot.Cookbook/Recipes/Plottable/RadialGauge.cs
bclehmann: > Making it nullable is a big modification since this would affect the RadialGauge plot code in several places.

Eh, not really, your IDE should be able to show you everywhere it's referenced. And I think changing it now is going to be much less painful than changing it later.

src/ScottPlot/Plottable/RadialGaugePlot.cs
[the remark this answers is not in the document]
bclehmann: That sounds good to me.

src/ScottPlot/Plottable/RadialGaugePlot.cs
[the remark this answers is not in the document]
bclehmann: Sounds good, some places I prefer `Math.Sign`, some places a ternary like you have here.

src/ScottPlot/Plottable/RadialGaugePlot.cs
bclehmann: The comment for `MinScale` was the same as the comment for `StartingAngleGauges`. It's resolved now.

src/ScottPlot/Plottable/RadialGaugePlot.cs
bclehmann: My bad, I misunderstood what the purpose of this section was.

src/cookbook/ScottPlot.Cookbook/Recipes/Plottable/RadialGauge.cs
swharden: Still a work in progress but I renamed this to `CircularBackground`. Anything is probably okay as long as it has a good cookbook example and XML docs.

Issue #1261, opened 2021-09-11
How to show the axis name for a Radar chart when independent axis is not enabled?


Rayffer:
## Question Template

**ScottPlot Version:** 4.1.20

**Operating System:** Windows

**Application Type:** WPF

**Question:** How do I show category name labels in a Radar Chart when independent axis is not enabled?

I want to user radar charts in a way where some data use dependant axis because it makes sense since it's variant types of the same data.

When I disable the independent axis feature, the labels go away while they can be seen when using independent axis, am I doing something wrong?

```cs

string[] categoryLabels= new[] { "Lightning", "Poison", "Force", "Slashing", "Bludgeoning", "Piercing" };

double[,] values = {
    { CreatePoint(random), CreatePoint(random), CreatePoint(random), CreatePoint(random), CreatePoint(random), CreatePoint(random)},
    { CreatePoint(random), CreatePoint(random), CreatePoint(random), CreatePoint(random), CreatePoint(random), CreatePoint(random)},
};

wpfPlot.Plot.Clear();
wpfPlot.Configuration.Pan = false;
wpfPlot.Configuration.Zoom = false;
wpfPlot.Plot.Title(title);
var radarPlot = wpfPlot.Plot.AddRadar(values, false, null, true);
radarPlot.AxisType = RadarAxis.Polygon;
radarPlot.WebColor = Color.FromArgb(0x80, 0, 0);
radarPlot.CategoryLabels = categoryLabels;
radarPlot.GroupLabels = groupLabels;
wpfPlot.Plot.SetAxisLimits(new AxisLimits(-2, 2, -2, 2), 0, 0);
wpfPlot.Render();
```

Rayffer:
I've looked further into the problem via the code available in github, and found the following:

![imagen](https://user-images.githubusercontent.com/34773204/132961795-976e765b-2072-44ea-a08f-74946938bd49.png)

As you can see, ShowCategoryLabels, and LabelEachSpoke are controlled by the value set on IndependentAxes, would it be possible to set this option via a property?

PR #1262, opened 2021-09-11
Radar: ShowCategoryLabels

Rayffer:
**Purpose:**
This pull request intends to give RadarPlot users the option to show category labels instead of it being shown when IndependentAxes property is set to true. Solves #1261

**New Functionality:**
There is no new functionality apart from giving the RadarPlot users direct control whether to show category labels or not.

### DIFF
--- a/dev/changelog.md
+++ b/dev/changelog.md
@@ -2,6 +2,7 @@
 
 ## ScottPlot 4.1.21
 * Legend: Throw an exception if `RenderLegend()` is called on a plot with no labeled plottables (#1257)
+* Radar: Improved support for category labels. (#1261, #1262) _Thanks @Rayffer_
 
 ## ScottPlot 4.1.20
 * Ticks: Fixed bug where corner labels would not render when multiplier or offset notation is in use (#1252, #1253) _Thanks @@DavidBergstromSWE_
--- a/src/ScottPlot/Plottable/RadarPlot.cs
+++ b/src/ScottPlot/Plottable/RadarPlot.cs
@@ -76,6 +76,11 @@ namespace ScottPlot.Plottable
         public bool ShowAxisValues { get; set; } = true;
 
         /// <summary>
+        /// If true, each category name will be written in text at every corner of the radar
+        /// </summary>
+        public bool ShowCategoryLabels { get; set; } = true;
+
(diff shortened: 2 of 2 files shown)


bclehmann:
Good catch! I thought this was a regression from #1188 but it looks like it always acted this way. I have two small questions:
- Do we need to demo this, such as with a cookbook recipe?
- Should the default for this property be true? It's a sensible default (otherwise why would the user provide category labels), but it is a breaking change
    - If the default changes to be dependent on whether `IndependentAxes` is set then we might need to allow the user to disable drawing category labels when `IndependentAxes` is enabled.

swharden:
As a reminder, this is what we're discussing

```cs
double[,] values = {
    { 5, 3, 10, 15, 3, 2, 256 },
    { 5, 2, 10, 10, 1, 4, 252 },
};
var radar = plt.AddRadar(values, independentAxes: true);
radar.CategoryLabels = new string[] { "Wins", "Poles", "Podiums", "Points Finishes", "DNFs", "Fastest Laps", "Points" };
radar.GroupLabels = new[] { "Sebastian Vettel", "Fernando Alonso" };
radar.ShowCategoryLabels = true; // <-- this is what we are discussing
plt.Title("2010 Formula One World Championship");
plt.Legend();
```

ShowCategoryLabels = true | ShowCategoryLabels= false
---|---
image|image

## Original Behavior / Issue this PR fixes

> When I disable the independent axis feature, the labels go away while they can be seen when using independent axis

This seems like a good fix! Thanks @Rayffer 

> Do we need to demo this, such as with a cookbook recipe?

@bclehmann I imagine the majority of people will want to show the labels if they go through the effort of populating them, and this is probably easy enough to figure out that an explicit example isn't required.

> Should the default for this property be true?

Is this variable actually necessary? What if we simplified the rule to "if category labels exist, show them". This property could be refactored to something like:

```cs
private bool ShowCategoryLabels => CategoryLabels is not null;
```

I'll go ahead and merge this now but if you think a change would make radar plots easier to use feel free to open-up a new PR to refine this behavior 👍 

> It's a sensible default ... but it is a breaking change

I'm pretty comfortable with small breaking changes if it makes the library less confusing for future users 😎

PR #1275, opened 2021-09-18
Add image support for categories.

Rayffer:
**Purpose:**
The aim of this pull request is provide support for Radar charts to display images at a 24x24 size for categories.

**New Functionality:**
The functionality allows to use Images in place of Labels for categories, follows the same behaviour as Category Labels, so the amount of images provided should be the same as the amount of categories intended to graph.

imagen


swharden:
Hi @Rayffer, this is looking pretty good, thanks! 

@bclehmann you've worked a lot with these radar plots - what do you think about this and the topics below?

## Should icon sizes be fixed?

It looks like this forces icons to always be 24x24. An alternative way to place icons would be to use the size of the image according to `CategoryImages[i].Width` and  `CategoryImages[i].Height`

https://github.com/ScottPlot/ScottPlot/blob/d471f2950c8446ba9f31a9d17a8e97e64a93ea06/src/ScottPlot/StarAxis.cs#L155

## Omit `ShowCategoryImages`?

It looks like it's okay to leave `CategoryImages` as `null`

https://github.com/ScottPlot/ScottPlot/blob/d471f2950c8446ba9f31a9d17a8e97e64a93ea06/src/ScottPlot/StarAxis.cs#L32-L36

so perhaps `ShowCategoryImages` is not needed?

https://github.com/ScottPlot/ScottPlot/blob/d471f2950c8446ba9f31a9d17a8e97e64a93ea06/src/ScottPlot/StarAxis.cs#L58-L64

The `null` check alone may be sufficient here

https://github.com/ScottPlot/ScottPlot/blob/d471f2950c8446ba9f31a9d17a8e97e64a93ea06/src/ScottPlot/StarAxis.cs#L145-L157

## Cookbook

A cookbook example may help demonstrate this functionality. There's already a folder of files for images that get packaged with the cookbook: https://github.com/ScottPlot/ScottPlot/tree/master/src/cookbook/ScottPlot.Cookbook/Images

Rayffer:
I can add an example to the cookbook as well, thanks for pointing that out

Rayffer:
For consistency, if we decide to remove the `ShowCategoryImages`, should we rename the `ShowCategoryLabels` to something like `ShowCategoryIdentifiers` or something like that?

bclehmann:
> Should icon sizes be fixed?

I think most icons would be squares around this size but I don't think that we should enforce that.

> Omit ShowCategoryImages?

I think that a null check here is fine instead of a separate property.

> should we rename the ShowCategoryLabels...

I think this is a breaking change, so no. If it isn't a breaking change then I still think label is a better term for what it does.

src/ScottPlot/StarAxis.cs
```diff
@@ -130,11 +142,23 @@ public void Render(PlotDimensions dims, Bitmap bmp, bool lowQuality = false)
                 }
             }
 
-            if (CategoryLabels != null && ShowCategoryLabels)
+            if (CategoryImages != null && ShowCategoryImages)
```
bclehmann: I'm more partial to `if (foo is not null)` rather than `if (foo != null)`, but it doesn't matter here.

`is null` is recommended in case someone overloads `Object.Equals`, but any overload which compares equal to null for non-null is probably bad to start with.

Rayffer:
How would you go about setting up icon size? Specify it in a property and if it is not set either draw the image as is or draw it downscaled following image's proportions?

I'll remove the new property I've created and just leave the null checking, should I keep consistency as to how category labels are shown and use the same property `ShowCategoryLabels` to control both if images or labels are shown, when set?

src/ScottPlot/StarAxis.cs
```diff
@@ -130,11 +142,23 @@ public void Render(PlotDimensions dims, Bitmap bmp, bool lowQuality = false)
...
             {
                 for (int i = 0; i < NumberOfSpokes; i++)
                 {
+                    double cosinus = Math.Cos(sweepAngle * i - Math.PI / 2);
+                    double sinus = Math.Sin(sweepAngle * i - Math.PI / 2);
```
bclehmann, line 142: Is that what I should have been calling this all along? It's a good variable name, but it does sound like the sinus cavity.

If you don't mind me asking, do you speak another language? I know that in French the functions are called sinus and cosinus.

bclehmann (approved):
Looks good to me, no complaints here.

bclehmann:
Now that I'm looking at it, some of the images in the demo aren't very clear, especially pole position, podiums, and points finishes.

For podium you could go with something like this:
image

But for the others I'm not sure. Maybe changing the cookbook demo might be easier, you might be able to find something with more clear images?

bclehmann:
Also, since coxcomb charts also use `StarAxis`, this might be a good option there. If you don't want to tackle that now that could be an easy Hacktoberfest task.

src/ScottPlot/StarAxis.cs
Rayffer: I speak spanish and I went with what I think is the most aproximate word, feel free to change to whichever suits better :)

src/ScottPlot/StarAxis.cs
Rayffer: I'll do it this way, I prefer it as well but I went with the implementation already existing in `StarAxis.cs`, thanks

src/ScottPlot/StarAxis.cs
bclehmann: Oh cool, I'm Canadian so I have a teensie bit of French, but it's not great. I don't have a problem with the variable name, I just thought it was interesting.

src/ScottPlot/StarAxis.cs
bclehmann: No worries, it was probably me that wrote `!= null` anyways, I just thought we might as well improve it if we're touching that line now anyways.

Rayffer:
> Now that I'm looking at it, some of the images in the demo aren't very clear, especially pole position, podiums, and points finishes.
> 
> For podium you could go with something like this:
> image

Maybe I can restrict the set to show to wins, points, fastest laps and did not finish, for the purpose of the demo it is relevant enough. I just didn't want to break the pattern of the cookbook examples.

src/ScottPlot/StarAxis.cs
Rayffer: I'll update both names to `cosine` and `sine`, sometimes it's worth to check proper spelling of a particular word.

bclehmann: I actually like the current name, it distances it from the sine/cosine function a little bit

Rayffer:
> 
> 
> Also, since coxcomb charts also use `StarAxis`, this might be a good option there. If you don't want to tackle that now that could be an easy Hacktoberfest task.

You mean create another cookbook example for `CoxCombPlot`? I can do it as well

bclehmann:
> You mean create another cookbook example for CoxCombPlot? I can do it as well

I think you have to add image support to Coxcomb plots first, but adding a cookbook recipe would probably be part of that.

Rayffer:
> 
> 
> > You mean create another cookbook example for CoxCombPlot? I can do it as well
> 
> I think you would have to add image support to Coxcomb plots first, but adding a cookbook recipe would probably be part of that.

No problem, adding it to cookbook should be more or less easy since it uses StarAxis as well

Rayffer:
@bclehmann Can you point me, in case it is possible, to a preview version of the cookbook? I want to check that it works as expected.

bclehmann:
> @bclehmann Can you point me, in case it is possible, to a preview version of the cookbook? I want to check that it works as expected.

You can run it locally by running any of the demo projects (i.e. WPF Demo, WinForms Demo, Avalonia Demo). You can also run something to dump all the pictures to a folder but I don't remember how.

src/ScottPlot/StarAxis.cs
Rayffer: Then I'll leave them as cosinus and sinus :)

Rayffer:
When taking a look into coxcomb, the icons appear at the arms of the chart, instead of at the middle of each pie slice, would it make sense to give the `StarAxis` class information on what is it rendering, either Radar chart or CoxComb chart, and offset the icons accordingly?

src/ScottPlot/Enums/AxisChart.cs
```diff
@@ -0,0 +1,8 @@
+﻿namespace ScottPlot
+{
+    public enum AxisChart
```
bclehmann, line 3: I would prefer if this was agnostic to what's rendering the `StarAxis` and instead just said where to put the images.

e.g.:
```cs
enum ImagePlacement {
    Outside,
    Inside
}
```

bclehmann: This avoids hiding what the field changes and it is more extensible in case we wanted to create a new plottable which used `StarAxis`, or if we wanted the image placement to be customizable.

Rayffer: So, some values like Arms or the area between the arms, how should it be called?

Your points are fair and sound, thanks!

Rayffer: Makes sense, will do it this way!

Rayffer:
Is this ready for merge as is now?

bclehmann:
It looks good to me, but only @swharden has the power to merge.

swharden:
This looks fantastic! Thanks for this PR @Rayffer, and @bclehmann for your helpful comments!

I added an example of semitransparent coxcomb plots to the cookbook, but perhaps semitransparency should be the default? Semitransparency allows the axis tick labels to be visible 🤔 

I'll go ahead and merge this now but feel free to open an issue (or create a PR) if you have any further ideas/suggestions! 👍 

image

Rayffer:
I agree on coxcomb semitransparency, it look quite a lot better actually.

PR #1284, opened 2021-09-26
refresh after loading plot viewer

### LINKED ISSUE #1264: Controls: Render warning is unclear
It's easy to confuse `formsPlot1.Render()` with `formsPlot1.Plot.Render()` as seen in #1263.

What can we do to make it more obvious how to do this correctly? 

renderfail

Related: #1245 #1165

### LINKED ISSUE #1283: Open in new window - formsPlot in new window shows the yellow Refresh() warning
Hello! I plot my graph using formsPlot1.Refresh(); and  formsPlot1.Plot.Render();.
Via the right mouse click menu, I use the "Open in New Window" option.
The data points are where they should be, but i get the yellow warning box in the middle:

image

### System Details
* **ScottPlot Version:** 4.1.21
* **Operating System:** (Windows 10)
* **Application Type:** (WinForms)
* **.NET Version:** (NET Framework 4.7.2)

swharden:
Fixes bug caused by plot viewers not calling `Refresh()` after resetting the plot.

Interestingly the WpfPlot already made this call, so the fix affects only FormsPlot and AvaPlot.

Fixes #1283 and fixes #1264

### DIFF
--- a/dev/changelog.md
+++ b/dev/changelog.md
@@ -3,6 +3,7 @@
 ## ScottPlot 4.1.22
 * Coxcomb Plots: Added support for image labels (#1265, #1275) _Thanks @Rayffer_
 * Palette: Added overloads for `GetColor()` and `GetColors()` to support transparency
+* Plot Viewer: fixed bug causing render warning to appear in WinForms and Avalonia plot viewers (#1265, #1238) _Thanks @bukkideme, @Nexus452, and @bclehmann_
 
 ## ScottPlot 4.1.21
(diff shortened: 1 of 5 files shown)


bclehmann:
Hey I think you might have forgotten about WPF

swharden:
> Hey I think you might have forgotten about WPF

> EDIT: nvm, it seems this issue never existed on WPF ¯\_(ツ)_/¯

Thanks for looking over my work though! I can certainly use the sanity checks 😄

Issue #1274, opened 2021-09-18
Hacktoberfest 2021 🎃🚀


swharden:
# ScottPlot is participating in Hacktoberfest 2021! 🎉🚀 
Hacktoberfest is an event that encourages contributions to open-source projects run by Digital Ocean. Participants who contribute to 4 merged pull requests may be eligible for a free T-shirt.

* **PR Quality Standards:** https://hacktoberfest.digitalocean.com/resources/qualitystandards

* **Information for Participants:** https://hacktoberfest.digitalocean.com/resources/participation

* **Maintainers:** https://hacktoberfest.digitalocean.com/resources/maintainers

# Recommended Projects for new Contributers

**Never contributed to open-source before?** Welcome! Hacktoberfest has a video on [making your first pull request](https://hacktoberfest.digitalocean.com/resources). We can figure this out together! 👍🚀 

## Improve color themes by submitting a `Style` or `Palette`

* ⚠️ See [**issue #1315: Hacktoberfest: Contribute a color style or palette**](https://github.com/ScottPlot/ScottPlot/issues/1315) for a step-by-step summary of how to contribute colors

![colorset2](https://user-images.githubusercontent.com/4165489/135541759-750eef9e-08bd-49bf-ae41-2b70197a26bc.gif)

## Improve the Website Documentation and FAQ
* [ScottPlot.NET](https://scottplot.net) content is on GitHub and can be edited with a PR: https://github.com/ScottPlot/ScottPlot.NET

## Improve the Cookbook
* The [ScottPlot Cookbook](https://scottplot.net/cookbook/4.1) has many short examples demonstrating how to use ScottPlot
* Cookbook recipes are here: [`src/cookbook/ScottPlot.Cookbook/Recipes`](https://github.com/ScottPlot/ScottPlot/tree/master/src/cookbook/ScottPlot.Cookbook/Recipes)
* If a cookbook recipe description contains a typo, feel free to fix it in a PR
* If you wish to create a new cookbook recipe, it will only be permitted if it adds substantial value beyond existing recipes. Because recipes are converted to web pages, there is pressure to keep the number of recipes limited. If you are unsure if your recipe idea will be merged, feel free to [open an issue](https://github.com/ScottPlot/ScottPlot/issues) and ask!

# Advanced Project Ideas

* **Syntax highlighting** would be great to have for the textboxes which show the source code in the cookbook of the demo applications (WinForms, WPF, and Avalonia, each may be separate PRs)

[...]

Issue #1079, opened 2021-05-27
Font rendering error on Fedora Linux


matachi:
## Bug Report Template

**ScottPlot Version:** 4.1.15

**Operating System:** Fedora 34, Wayland, 64-bit

**Application Type:** console

**Bug Description:** The numbers in the chart aren't rendered correctly:

![test](https://user-images.githubusercontent.com/980838/119884435-ebc7b000-bf30-11eb-9ece-884892176835.png)

I installed `libgdiplus` as described in the README. Do I need to install something in addition to that? I don't know if a font is missing or something else is going wrong?

**Reproducing:**

```cs
namespace stocks
{
    class Program
    {
        static void Main(string[] args)
        {
            var dataX = new double[] { 1, 2, 3, 4, 5 };
            var dataY = new double[] { 1, 4, 9, 16, 25 };
            var plt = new ScottPlot.Plot();
            plt.AddScatter(dataX, dataY);
            plt.SaveFig("test.png");
        }
    }
}
```

swharden:
Hi @matachi, thanks for sharing this issue! My guess is that this issue relates to not finding a font that your system can easily render.

## Is all text affected?

If you add a title and axis label, do those render correctly?
```cs
plt.Title("test");
plt.XLabel("test");
plt.YLabel("test");
```

## Does changing font help?
Try a font name definitely installed on your system
```cs
plt.XAxis.TickLabelStyle(fontName: "comic sans ms");
plt.YAxis.TickLabelStyle(fontName: "comic sans ms");
```

## Test for this

I can add a test for this that will run on Windows/MacOS/Linux as part of the CI pipeline. It will render the number 1 and the number 2 on otherwise identically bitmaps, and assert that the resulting bitmaps are different.

## Notes
Default font names are looked-up like this:

https://github.com/ScottPlot/ScottPlot/blob/efcc3ceaebf2b4af475a5dc46c6e21b696623eaa/src/ScottPlot/Drawing/InstalledFont.cs#L11-L20

https://github.com/ScottPlot/ScottPlot/blob/efcc3ceaebf2b4af475a5dc46c6e21b696623eaa/src/ScottPlot/Drawing/InstalledFont.cs#L22-L31

matachi:
I added these lines to a project:
```c#
            plt.Title("Abcdefghijklmnopqrstquvwxyz");
            plt.XLabel("test");
            plt.YLabel("test");
```
And the chart looks like this:
[...]

Issue #1004, opened 2021-05-02
Which version of System.Drawing.Common should ScottPlot depend on?


swharden:
A new version of the [System.Drawing.Common package](https://www.nuget.org/packages/System.Drawing.Common) is released with every major .NET version

***Which version should ScottPlot depend on? How often (if ever) should it be changed?***

Currently ScottPlot 4.0 depends on `4.7.0` and ScottPlot 4.1 depends on `5.0.0`, but these weren't chosen intentionally

Both could be upgraded to the latest (`5.0.2`) but I'm not sure if this would fix problems or cause more pain

Alternatively I could make it so each build target uses the version-matched System.Drawing.Common

Related: #993, #924, #655

bukkideme:
Hi!

@swharden , you wrote it is hard to recreate this issue. For me, it is very easy, please see the steps below to reproduce the issue, using a .NET Framework 4.7.2 WinForm project (VS2019 v16.9.4 Community):

1. Create a Winform .NET FrameWork 4.7.2 project from VS
2. Install ScottPlot 4.0.48 WinForm using nuget package manager
[WindowsFormsApp1_472.zip](https://github.com/ScottPlot/ScottPlot/files/6411787/WindowsFormsApp1_472.zip)
[WindowsFormsApp1_472.zip](https://github.com/ScottPlot/ScottPlot/files/6411789/WindowsFormsApp1_472.zip)


3. Pull a FormsPlot control to the UI. Everything works as it should
4. Notice that after installing ScottPlot in nuget manager, the manager tells you to update System.Drawing.Common from 4.7 to 5.0.2 latest version. Do it.
5. Close solution, close VS2019.
6. Restart VS2019, open solution
7. Notice that you cannot open the Form1.cs [Design] anymore, you get the earlier mentioned version collision error msgs.

Attempt to fix this issue:

1. Go to nuget packages, and downgrade System.Drawing.Common back to v4.7.
2. Close and reopen solution/project.
3. Now you can open the designer (Form1.cs [Design]), it works again
4. Try to run the code.
5. You get System.IO.FileLoadException ("„System.Drawing.Common, Version=4.0.0.2, Culture=neutral, .....cannot be loaded....")

I imagine if you do not update System.Drawing.Common from 4.7, we are safe... However, it looks like if you do so, there is no way back (I did not try GIT if that would be a solution to recover to working state).
[...]

PR #1310, opened 2021-10-01
Framework-dependent System.Drawing.Common version in NuGet packages

### LINKED ISSUE #1311: Central version management
@Kritner from the CodingBlocks slack (https://www.codingblocks.net/slack) has written blog posts ([2021-10-01](https://blog.kritner.com/2021/10/01/dotnet-centralized-package-versioning/), [2018-10-10](https://blog.kritner.com/2018/10/10/using-configuration-files-to-manage-nuget-package-versions-and-other-stuff/)) discussing how to use a single file to define version of multiple projects and their dependencies, making versioning easier to track and update.

## Current System
ScottPlot version and all its dependency package versions are defined in individual csproj files:
* [ScottPlot.csproj](https://github.com/ScottPlot/ScottPlot/blob/master/src/ScottPlot/ScottPlot.csproj)
* [ScottPlot.WinForms.NUGET.csproj](https://github.com/ScottPlot/ScottPlot/blob/master/src/controls/ScottPlot.WinForms/ScottPlot.WinForms.NUGET.csproj)
* [ScottPlot.WPF.NUGET.csproj](https://github.com/ScottPlot/ScottPlot/blob/master/src/controls/ScottPlot.WPF/ScottPlot.WPF.NUGET.csproj)
* [ScottPlot.Avalonia.NUGET.csproj](https://github.com/ScottPlot/ScottPlot/blob/master/src/controls/ScottPlot.Avalonia/ScottPlot.Avalonia.NUGET.csproj)

A versioning-bumping script is manually run to increment ScottPlot version project-wide:
* [Versioner/Program.cs](https://github.com/ScottPlot/ScottPlot/blob/master/dev/tools/Versioner/Program.cs)

## Proposed System
* Use a central file to store ScottPlot version
  * This should be used for `Version`, `AssemblyVersion`, and `FileVersion`
  * Notice that the first is `x.x.x` and the other two are `x.x.x.0` (that fourth number is important)
* Use a central file to store versions of all dependencies _that we want constrained_ (not necessary all packages).
  * ⚠️ Let's be careful which version of `System.Drawing.Common` is used in the core packages. We found that it this package can be _upgraded_ in old solutions, but it breaks them such that they cannot be _downgraded_ again (#1004). For this reason we probably want to keep targeting `4.6.1` for the primary library packages, although tests and demo applications may use newer versions.
* Consider whether to use GitHub Actions to automatically bump the version number before building and publishing NuGet packages in the cloud. See [`.github/workflows/ci.yaml`](https://github.com/ScottPlot/ScottPlot/blob/master/.github/workflows/ci.yaml)

## Potential Drawbacks
* Old versions of Visual Studio may be unable to use the graphical NuGet package manager. I'm not sure about how widely supported central package management is.
  * I think this is okay because it's unlikely typical users working in the source code of this project will need to add NuGet packages.

Kritner:
**Purpose:**
* As a result of https://blog.kritner.com/2021/10/01/dotnet-centralized-package-versioning/ and chatting with Scott on slack ;)
* I can't seem to build all of the projects, mostly around the pieces that require framework and/or windows based functionality.  I don't know if this is something you would be able to check off @swharden

image

#1274, Closes #1311


swharden:
Hi @Kritner, thanks for sharing this! See my notes in #1311

If you remove the draft status of this PR you can confirm the CI/tests pass and I won't merge it until you're ready 👍

Kritner:
#1274, Closes #1311

swharden:
> I can't seem to build all of the projects, mostly around the pieces that require framework and/or windows based functionality

This might make this difficult 😅

We exert a lot of effort to maintain cross-targeted builds across .NET Framework (back to 4.6.1), .NET Standard (2.0), and .NET 5 so the library runs on Windows, MacOS, and Linux.

If `ManagePackageVersionsCentrally` isn't supported with those earlier targets, a less fancy solution [like csproj variables](https://blog.kritner.com/2018/10/10/using-configuration-files-to-manage-nuget-package-versions-and-other-stuff/) is still an improvement. I'm interested in learning more about this topic!

swharden:
> a less fancy solution like csproj variables is still an improvement

If this is the case I think we really only need to use central versioning for:
* `ScottPlot` - the four csproj files linked in #1311
* `Avalonia`, `Avalonia.Desktop`, and `Avalonia.Diagnostics`
* `System.Drawing.Common` (keep at 4.6.1)
  * related: #1004
  * related: #1005

The rest can remain hard-coded in the csproj files.

Kritner:
>> I can't seem to build all of the projects, mostly around the pieces that require framework and/or windows based functionality
>
> This might make this difficult 😅
>
> We exert a lot of effort to maintain cross-targeted builds across .NET Framework (back to 4.6.1), .NET Standard (2.0), and .NET 5 so the library runs on Windows, MacOS, and Linux.

The only sandbox app I can't build is "WinFormsFrameworkApp".  Here are outputs from the other demo apps:

WPF:
image

WinForms:
image

Console:
image

Avalonia Demo:
image

-----

FWIW, I ran `dotnet format` but that seems to be failing the CI / Check Formatting step. Additionally all tests are passing locally, I'll look at which one failed in that part of the process.

-----

> * System.Drawing.Common (keep at 4.6.1)
>   * related: Which version of System.Drawing.Common should ScottPlot depend on? #1004
>   * related: System.Drawing.Common version downgrade #1005

Addressed in https://github.com/ScottPlot/ScottPlot/pull/1310/commits/2ccae4f57a68515844efdfadbd79221a73bc01c0

Puts the central version to 4.6.1, individual packages can be upgraded with `OverrideVersion` in their respective csproj file if desired.  (I'm assuming no newer features were actually required, so I just kept everything at 4.6.1)

swharden:
> @swharden should the package version be revved within this PR to *.25? or is that a part of the build process and/or rollup of several features into a new version?

Let's leave all the ScottPlot versions what they are. After merging this PR I'll automate the process of bumping them using GitHub Actions 👍

Kritner:
> > @swharden should the package version be revved within this PR to *.25? or is that a part of the build process and/or rollup of several features into a new version?
> 
> Let's leave all the ScottPlot versions what they are. After merging this PR I'll automate the process of bumping them using GitHub Actions 👍

Sounds good, current branch is still set at version 4.1.24.  I *think* that's everything, except for the CI / Check Formatting step (which I did run the formatting...) and The failing Test on Linux: https://github.com/ScottPlot/ScottPlot/pull/1310/checks?check_run_id=3773009138#step:6:23

swharden:
> The only sandbox app I can't build is "WinFormsFrameworkApp". Here are outputs from the other demo apps:

This is encouraging! As far as building/testing I'm happy to learn on the CI system. 

### Test_LettersDontRenderAsRectangles_SerifFont

Thanks again for your work on this! Sorry it got complicated by this 😝 

☠️ It looks like the Linux tests are failing due to font rendering issues. This may be because GitHub (the system that runs on the cloud) just today changed their default supported font set. To get this PR un-stuck you can either:

* Option 1: edit the [test that fails on Linux](https://github.com/ScottPlot/ScottPlot/blob/733441bc3878ba831d0d855575f8ced8515769d4/src/tests/Misc/Drawing.cs#L153-L179) and add a statement something like:

```cs
if (!RuntimeInformation.IsOSPlatform(OSPlatform.Windows)) Assert.Pass();
```

* Option 2: In [ci.yaml](https://github.com/ScottPlot/ScottPlot/blob/master/.github/workflows/ci.yaml) add an indented line after `strategy:` with `fail-fast: false` to allow the Windows tests to run even if the Linux tests fail.

swharden:
## Auto-Formatting

Sometimes the formatter requires updating. Running this in the top-level solution directory (`src`) may solve your problem. If not let me know and I can autoformat it from my end 👍 

```bash
dotnet format
```

Kritner:
> ## Auto-Formatting
> Sometimes the formatter requires updating. Running this in the top-level solution directory (`src`) may solve your problem. If not let me know and I can autoformat it from my end 👍
> 
> ```shell
> dotnet tool update -g dotnet-format
> dotnet format
> ```

Yeah might need you to give it a whirl, I did run it locally, which caused changes.

I don't know if it *was* those changes causing the step to fail or not, but `dotnet format` definitely changed those files as a part of one of my commits, specifically https://github.com/ScottPlot/ScottPlot/pull/1310/commits/1a7586116d8689ca5ea5bb44eac445d0ca9d89ba

swharden:
I think the new errors indicate the package versioning isn't working as expected 😅 

The font rendering error was also a result of this actually... if the wrong version of System.Drawing.Common is installed on Linux, characters render as squares, and that test fails: https://github.com/ScottPlot/ScottPlot/issues/1079#issuecomment-849936525

Let's ignore autoformatting for now

Directory.Build.props
```diff
@@ -0,0 +1,11 @@
+<Project>
+    <PropertyGroup>
+        <ManagePackageVersionsCentrally>true</ManagePackageVersionsCentrally>
+    </PropertyGroup>
+
+    <PropertyGroup>
+        <Avalonia>0.10.7</Avalonia>
```
bclehmann, line 7: I think this should be `0.10.*`

src/tests/Misc/Drawing.cs
```diff
@@ -156,6 +157,12 @@ public void Test_LettersDontRenderAsRectangles_SerifFont()
             // this test ensures letters don't render as rectangles
             // https://github.com/ScottPlot/ScottPlot/issues/1079
 
+            // https://github.com/ScottPlot/ScottPlot/pull/1310#issuecomment-932638916
+            // Some potential change to the default fonts provided for the linux image,
+            // will still run through the test on the windows portion of CI
+            if (!RuntimeInformation.IsOSPlatform(OSPlatform.Windows))
```
bclehmann: This is dicey, but I think if we have to not run it on Linux we should probably do this instead:

```cs
if (RuntimeInformation.IsOSPlatform(OSPlatform.Linux))
```

bclehmann (approved):
Looks good to me, although that test does bug me. Especially as Linux font rendering on GDI+ seems to be more fragile than other OSes, and we're already rather dependent on bug reports to catch Linux issues.

.github/workflows/ci.yaml
```diff
@@ -47,6 +47,7 @@ jobs:
 
   test:
     strategy:
+      fail-fast: false
```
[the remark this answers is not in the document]
bclehmann: I think now that the Linux test has been fixed this can be removed?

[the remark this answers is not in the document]
swharden: if `fail-fast` is `true` (default) then the failure of one job terminates all other jobs (so we never find out if they would have succeeded).

Setting this to `false` lets us wait-out the result of the Windows tests to see if those pass or fail, even though the Linux tests failed quickly. A failure in any case blocks the pipeline from proceeding

https://docs.github.com/en/actions/learn-github-actions/workflow-syntax-for-github-actions#jobsjob_idstrategyfail-fast

Directory.Build.props
swharden:
```
error NU1011: Centrally defined floating package versions are not allowed
```

After fighting with `ManagePackageVersionsCentrally` for a few hours now, I think this is the last straw 😅 

I'll try not to get enticed by preview features in the future. I'll rework this PR to focus on improved ScottPlot version bumping and not worry about centralizing versions of dependencies.

I bet we can achieve the `*` behavior with a shared project variable...

src/tests/Misc/Drawing.cs
swharden: Good catch! However I'll revert this test to its original form and, now that the correct version of System.Drawing.Common is loaded, the test passes again on all operating systems.

swharden: c34cd0c

swharden (approved):

swharden:
Wow, this was a rollercoaster ride! 🎢🎃 

## I did not side with centralized dependency versioning

* It couples packages more than I desire - the solution file is used for the convenience of the developer, but packages should be largely self-contained and isolated from the rest of the stuff in this repository.

* Mixed projects are not supported - I wish I could centrally version just `System.Drawing.Common` but it doesn't look like that's possible without centrally versioning everything, or requiring building from the _solution_ (I want the nuget packages to be independently buildable because it's often hard to build mixed-target solutions on cloud platforms)

* Wildcard versioning is currently unsupported. Allowing a version like `0.10.*` is pretty helpful.

I learned a lot through this experience, including not to rush to adopt preview features 😅

I moved the .props files to /dev/ so we might be able to use them later when the .NET tools mature

## The CI pipeline improved

* Tail-fast is now disabled, allowing more thorough assessments of whether failing tests fail on all operating systems or not

* The nuget packages are built from separate csproj files than the ones used by the solution file, so a new job was added to test building those and storing them as artifacts so you can download/inspect/test them. They're only retained for 24 hours to keep storage utilization low.

## Conditional dependency on `System.Drawing.Common`

Unless I goofed and haven't realized it yet, I think this is a big win of this PR. The System.Drawing.Common versioning problems (#1004) are limited to what happens when .NET Framework users _downgrade_ their package version. Before this PR we just targeted the lowest version, but that causes poor font rendering on linux (#1079). This new discussion led to the following in the nuget package csproj files, which I think solves the problem by allowing the latest version to be distributed to everyone except .NET Framework users (though they still have the ability to upgrade on their end if they wish).

https://github.com/ScottPlot/ScottPlot/blob/4895957f2507465017b3a1148066df12b1727324/src/controls/ScottPlot.WinForms/ScottPlot.WinForms.NUGET.csproj#L38-L47

PR #1313, opened 2021-10-02
Palette: Dark

arthurits:
**New Contributors:**
Please review [CONTRIBUTING.md](https://github.com/swharden/ScottPlot/blob/master/CONTRIBUTING.md)

**Hacktoberfest Participants:**
Check-out the [Hacktoberfest 2021](https://github.com/ScottPlot/ScottPlot/issues/1274) page

**Purpose:**
* Add a new qualitative 8-dark-color palette which is both LCD and print friendly but not blind nor photocopy friendly.
* This is one of the many subjects suggested in Hacktoberfest 2021 🎃🚀 #1274.

image

PR #1314, opened 2021-10-02
Palette: DarkPastel

arthurits:
**New Contributors:**

**Purpose:**
* Add a new qualitative 8-color palette which is the lighter-color version of the Dark palette in #1313.
* This is one of the many subjects suggested in Hacktoberfest 2021 🎃🚀 Hacktoberfest 2021 🎃🚀 #1274.

image


swharden (approved):

src/ScottPlot/Palette.cs
```diff
@@ -14,6 +14,7 @@ public static class Palette
         public static ScottPlot.Drawing.Palette Aurora => new(new ScottPlot.Drawing.Colorsets.Aurora());
         public static ScottPlot.Drawing.Palette Category10 => new(new ScottPlot.Drawing.Colorsets.Category10());
         public static ScottPlot.Drawing.Palette Category20 => new(new ScottPlot.Drawing.Colorsets.Category20());
+        public static ScottPlot.Drawing.Palette DarkLight => new(new ScottPlot.Drawing.Colorsets.DarkLight());
```
bclehmann, line 17: `DarkLight` sounds like an oxymoron, maybe `DarkPastel` or `DarkDesaturated` make more sense? It's a minor thing though.

src/ScottPlot/Palettes/DarkLight.cs
```diff
@@ -0,0 +1,15 @@
...
+ * https://colorbrewer2.org
+ * which is the lighter-color version of the 'Dark' palette.
+ */
+namespace ScottPlot.Drawing.Colorsets
+{
+    class DarkLight : HexColorset, IPalette
+    {
+        public override string[] hexColors => new string[]
+        {
+            "#66c2a5", "#fc8d62", "#8da0cb", "#e78ac3", "#a6d854",
```
bclehmann, line 11: I'm not sure if these are just x% lighter than `Dark`, but if so you might be able to do that programmatically? I'm fine with having constants given that every other palette is done like this, and it's not like the pallettes are likely to change. Depends if you think there's a benefit.

bclehmann (approved):

src/ScottPlot/Palettes/DarkLight.cs
swharden: > might be able to do that programmatically?

True, but I think there's advantage to explicitly writing-out the colors. It makes it easier to use these colors elsewhere, and I'm hesitant to start encouraging coupling between palettes 🤔

src/ScottPlot/Palette.cs
swharden: I like `DarkPastel`! I'll rename this and merge 👍

swharden:
## ColorBrewer Licensing

ColorBrewer's restrictive Apache2 license has been a pain point in the past 😩
* https://github.com/ScottPlot/ScottPlot/issues/420
* https://github.com/ScottPlot/ScottPlot/pull/458

### Colors from source code: Apache 2

If we take the hex color values from its source code we would have to re-license ScottPlot as Apache2
* https://github.com/axismaps/colorbrewer/blob/9a37cbbfe7cde61c060c68ecdd1fd3a5095ef4a5/export/colorbrewer.js
* https://github.com/axismaps/colorbrewer/blob/9a37cbbfe7cde61c060c68ecdd1fd3a5095ef4a5/export/LICENSE.txt

### Colors from website: Fair Game?

Something I never considered before is "what if we don't get hex values from the source code, but instead get them from the website"? The website doesn't indicate a license (just a copyright) so maybe pulling the hex values from the app on https://colorbrewer2.org is ok? 🤷

bclehmann:
> The website doesn't indicate a license (just a copyright) so maybe pulling the hex values from the app on https://colorbrewer2.org is ok? 🤷

If something doesn't specify a license then the owner has waived none of their copyright. That means we'd have to comply with even stricter requirements than we would under the Apache 2 License. Basically, it would only be legal if we could argue it constitutes fair use, which would be pretty difficult.

swharden:
I'd love to solve this issue because it keeps popping up 😝 

The website lets you download the colors directly... and there's no copyright or licensing information in what they have you download.

image

> we could argue that it's something so small it cannot be protected by copyright

I think the main reason they went with Apache2 was to prevent other people from impersonating them by using the ColorBrewer name... not so much to protect the hex values of the colors themselves (the distribution of which is the whole point of that website).

I think we can proceed using the ColorBrewer hex values available on their website, ***not*** sourcing anything directly from their source code, and not including the customized Apache2 license with ScottPlot. I'd like to include the copyright found on their website in the source code as a gesture of good faith, and if we are ever contacted about this we can communicate with them directly to figure out the best path forward.

Sound reasonable?

bclehmann:
In the US it seems that no one has successfully extended the copyright of a program to a copyright of its output, although [some have tried](https://mwhlawgroup.com/can-copyright-in-software-extend-to-its-output/). In the cases linked the court found that the plaintiff did not establish a premise of their argument (that the program did the majority of the work), therefore the court did not need to decide whether to adopt the plaintiffs reasoning as legal precedent.

This means that it's entirely possible that the courts could decide that program copyright can cover the output, so I would be wary.

However, I agree that ColorBrewer primarily exists to share colour sets, and I have seen colour sets from ColorBrewer everywhere, and I don't think ColorBrewer has ever acted to stop that.

swharden:
That case was a good read - thanks for sharing it! If someone from ColorBrewer wants to discuss this further, I welcome the conversation.

@arthurits just wanted to contribute to open source and maybe win a free T-shirt for hacktoberfest and now we are discussing lawsuits 😝

I think this discussion landed in a good place, and I'm happy that all these issues/PRs are linked together

arthurits:
It really is quite educational going through all the stuff behind the scenes for approving, discussing, and finally merging PRs. So thanks to you both.

Copyright and legal issues are out of my limited scope, but I understand the need to make sure there's nothing interfering with ScottPlot's MIT licence. As an aside note, this reminds me of the legal aspects involved in magic (♣️♥️♠️♦️), which are also a very complex issue.

PR #1317, opened 2021-10-03
Palette: SummerSplash

KanishkKhurana:
added a new fresh summer color palette . 
i hope you like it 
do tell me if there are some errors .
happy hacking !

### DIFF
--- a/src/ScottPlot/Palette.cs
+++ b/src/ScottPlot/Palette.cs
@@ -1,5 +1,8 @@
-﻿using System.Drawing;
+﻿using System;
+using System.Drawing;
 using System.Linq;
+using System.Reflection;
+using System.Runtime.Serialization;
 
 /* 
  * Palettes are collections of colors that control the default colors for new plottables added to the plot.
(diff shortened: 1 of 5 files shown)


KanishkKhurana:
@swharden please go through the new color palette

swharden:
Hey @bclehmann, I made some additional changes to @KanishkKhurana's PR and I wanted to share my thought process with you.

`GetPalettes()` now uses reflection to return all palettes in the `ScottPlot.Drawing.Palettes` namespace (not properties of the `ScottPlot.Palette` static class). I think this will be a useful tool to allow low-impact palette contributions to be added to the project, but not clutter the primary classes themselves. I'm thinking we can include only the high-impact palettes as shortcuts in the static class. Sound reasonable?

swharden (approved):

KanishkKhurana:
thank you for reviewing my code and making the necessary changes to it @swharden. 
i hope now you can merge this PR and close it as well!
happy hacktober!

bclehmann (changes requested):
> GetPalettes() now uses reflection to return all palettes in the ScottPlot.Drawing.Palettes namespace (not properties of the ScottPlot.Palette static class).

Nice, it's quite elegant.

> I'm thinking we can include only the high-impact palettes as shortcuts in the static class.

I don't know, I think if you have a list somewhere it's reasonable to expect that the list is complete, I don't think there's any merit in adding palettes but hiding them from view. I think either there should be no palettes on `ScottPlot.Palette` and everyone constructs their own palettes, or all of the palettes should be on `ScottPlot.Palette`.

I think that we should avoid inconsistencies unless there's a good reason. I don't see how separating high/low impact is important, and it isn't easily communicable to users. I also don't see the clutter argument, as palettes aren't being mixed with unrelated members, the class is almost completely devoted to holding palette objects.

I agree that it isn't great that adding a palette requires changes in two places, but I think allowing some palettes to be in `ScottPlot.Palette` while some aren't will invite mistakes. For example, it may lead to palettes that belong in `ScottPlot.Palette` but are missed because other palettes aren't present.

Let me know what you think, I went on something of a ramble, but it's looking good.

swharden:
>> reflection to return all palettes in the ScottPlot.Drawing.Palettes namespace

> Nice, it's quite elegant.

😎 👍 

> I don't see how separating high/low impact is important, and it isn't easily communicable to users. I also don't see the clutter argument.

> may lead to palettes that belong in ScottPlot.Palette but are missed

I'd like to have a system in place to quarantine palettes submitted as hacktoberfest contributions and only expose only the highest quality palettes to the user in the static class.

Considering the culture around hacktoberfest it's likely there may be a good number of palettes that aren't high enough quality to list next to `Category10` and `Nord`, but aren't low enough quality to merit tagging as spam/invalid. I've been using the term "low-impact" to refer to these contributions.

Collecting low-impact palettes as isolated class files in a namespace seems like a better alternative than rejecting the PR for a subjective reason like not liking the colors or thinking the colors are too similar to an existing palette. My strategy is to make it easy to contribute palettes non-disruptively, and we can comb over them at the end of October and figure-out which ones to expose in the static class.

@bclehmann I'm sharing my thinking with you so we can be on the same page when we review style/palette PRs in the future 👍

swharden:
PS: these colors probably came from here:

https://color.adobe.com/Newsletter-Summer-Theme-color-theme-17991162

Issue #352, opened 2020-04-17
Is it possible to change key binds


LogDogg:
Is there an option to customize the keyboard/mouse controls? For example, use left click to zoom and right click to pan, alt+left click to rectangle zoom etc.

swharden:
Aside from being able to customize the right-click menu and disable zooming and/or dragging, I don't think it's possible to re-wire events in the FormsPlot and WpfPlot controls to significantly change their interactive behavior like that.

Users who would benefit from a highly customized user control are encouraged to copy FormsPlot.cs and modify it to create their own control that meets their needs. There's relatively little code in there, and by deleting chunks that aren't useful to your application you can thin it down a lot further. Then you have total control over the mouse and keyboard interactive behavior.

https://github.com/swharden/ScottPlot/blob/master/src/ScottPlot.WinForms/FormsPlot.cs

Good question @LogDogg, I hope my answer helps somewhat!

LogDogg:
Yes, @swharden, it's very helpful, thank you!

EmanuelFeru:
Hi Scott @swharden,

I was actually also considering to customize the mouse bottons slightly. I wanted to actually:
- Move the Zoom box from Middle button drag -> Left button drag
- Move the Pan from Left button grad -> Middle button drag

Because it would feel more natural to me due to the software I used in the past. The rest can stay default because I like it. If I understand correctly, it is not possible to customize this at the moment, right? Unless I recreate the `FormsPlot.cs` and change it to my needs?

swharden:
Hi @EmanuelFeru,

This is closer to easy than it used to be, but it still requires modification of the source code. You identified the correct files! You can probably achieve your desired behavior just by modifying the logic inside the `if` statements in the `MouseMove()` and `MouseUp()` methods:

https://github.com/ScottPlot/ScottPlot/blob/b1d26dc508f15cb4ff3a0c0feeae6ca803cfe769/src/ScottPlot/Control/Backend.cs#L460-L491

https://github.com/ScottPlot/ScottPlot/blob/b1d26dc508f15cb4ff3a0c0feeae6ca803cfe769/src/ScottPlot/Control/Backend.cs#L553-L578

I added an item to the triage list #1028 to create a demo or document this eventually 👍

PR #1335, opened 2021-10-07
Implemented swap mouse Left Pan <-> Middle zoom

EFeru:
The behavior can be activated by setting `Configuration.SwapDragLeftMiddle = true;`

**New Contributors:**

**Purpose:**
* This change allows to swap left drag pan to middle drag pan and middle drag zoom to left drag zoom by `Configuration.SwapDragLeftMiddle = true;` Default behavior is `false`, so the original key bindings are not affected
* Related issue #352 
* I implemented this swap, because for users of Matlab (like me), it feels more natural with swap enabled to match Matlab's plot default behavior

**Isssue to be solved:**
For some reason when using Draggable lines, the zoom also triggers, which is not wanted of course. @swharden maybe you can help me to fix that. I tried to debug, and the `createMouseMovedToZoomRectangle` is not triggering but I still get a zoom.
https://github.com/ScottPlot/ScottPlot/blob/a39f904543dcc54b48497a7fd7ea0b35c1bf49b2/src/ScottPlot/Control/Backend.cs#L582

You can use the Cookbook to test, just set `SwapDragLeftMiddle = true` in `Configuration.cs`

### DIFF
--- a/src/ScottPlot/Control/Backend.cs
+++ b/src/ScottPlot/Control/Backend.cs
@@ -144,6 +144,9 @@ namespace ScottPlot.Control
         /// </summary>
         private bool IsZoomingWithAlt;
 
+        /// <summary>
+        /// True if a zoom rectangle is being actively drawn using left click
+        /// </summary>
         private bool IsZoomingWithLeft;
 
         /// <summary>


bclehmann:
I think if we are going to support remapping keys/mouse buttons we should allow full (or close to it) remapping instead of just one swap. Otherwise, we open ourselves up to maintaining every configuration option in one of several booleans.

bclehmann:
I wasn't intending to comment on the naming, although `UseMiddleClickPan` or something similar sounds fine to me.

I was suggesting that the implementation be more generic, supporting arbitrary bindings and then middle-click panning could be built on top of that. A good way to support changing key bindings would prevent an explosion of complexity if we decide to support 5 different keymaps.

swharden:
> It can become `protected virtual`

Good catch @bclehmann!

Changing `GetInputState()` from `private` to `protected virtual` sounds like a good non-invasive fix to let the user control what buttons do what until I get around to refactoring the control backend.

@EFeru if you try it and it works as expected I'd be happy to merge a PR that makes this change identically in all 3 controls.

EFeru:
Just tried swapping Left with Middle button in `Input state` :
```cs
private Control.InputState GetInputState(MouseEventArgs e) => 
     new() 
     { 
         X = e.X, 
         Y = e.Y, 
         LeftWasJustPressed = e.Button == MouseButtons.Middle, 
         RightWasJustPressed = e.Button == MouseButtons.Right, 
         MiddleWasJustPressed = e.Button == MouseButtons.Left, 
         ShiftDown = ModifierKeys.HasFlag(Keys.Shift), 
         CtrlDown = ModifierKeys.HasFlag(Keys.Control), 
         AltDown = ModifierKeys.HasFlag(Keys.Alt), 
         WheelScrolledUp = e.Delta > 0, 
         WheelScrolledDown = e.Delta < 0, 
     }; 
```

Which works as intended, swaps complete behavior left <-> middle. In my opinion making this mapping accesible to the user is a first step to allow some customization. 
If I undestand correctly it is just changing `private` to `protected virtual` and then any suggestion on how to override the function on the user application side?

swharden:
## Inherit-and-Override Example

> any suggestion on how to override the function on the user application side?

I worked this through on my end to try it out. It's a bit clumsy, but it works. I created a new WinForms application and I did not add a `FormsPlot`. Instead I created a new user control `UserControl` and inherited `FormsPlot`, then drag-dropped that onto my form.

### UserControl1.cs
```cs
using ScottPlot.Control;
using System.Windows.Forms;

namespace WinFormsFrameworkApp
{
    public partial class UserControl1 : ScottPlot.FormsPlot // not "UserControl"
    {
        public UserControl1()
        {
            InitializeComponent();
        }

        protected override InputState GetInputState(MouseEventArgs e)
        {
            return new InputState()
            {
                X = e.X,
                Y = e.Y,
                LeftWasJustPressed = e.Button == MouseButtons.Middle, // CUSTOM
                RightWasJustPressed = e.Button == MouseButtons.Right, // CUSTOM
                MiddleWasJustPressed = e.Button == MouseButtons.Left, // CUSTOM
                ShiftDown = ModifierKeys.HasFlag(Keys.Shift),
                CtrlDown = ModifierKeys.HasFlag(Keys.Control),
                AltDown = ModifierKeys.HasFlag(Keys.Alt),
                WheelScrolledUp = e.Delta > 0,
                WheelScrolledDown = e.Delta < 0,
            };
        }
    }
}
```

### Form1.cs
```cs
public Form1()
{
    InitializeComponent();
    userControl11.Plot.AddSignal(ScottPlot.DataGen.Sin(51));
    userControl11.Plot.AddSignal(ScottPlot.DataGen.Cos(51));
    userControl11.Refresh();
}
```

## Should we make GetInputState() overridable?

The key to this example being possible is that `GetInputState()` was changed from `private` to `protected virtual`, allowing the override. 

@bclehmann's hesitation about this is very justified - if we encourage people to inherit-and-override, we lose our ability to modify these functions later without breaking our users' code. I think decorating this method with `[Obsolete]` and including a warning would be enough to get the message across (see below).

@EFeru if this technique seems like a good intermediate solution for you, you're welcome to modify this PR or open a new one and I'd be happy to merge it in. #1274 These changes would be made in all 3 user controls:
* change `private` → `protected virtual`
* add an obsolete decorator to `GetInputState`
* add a pragma statement to ignore obsolete warnings in the file

```cs
[Obsolete("WARNING: API may change in future (override with caution)")]
protected virtual Control.InputState GetInputState(MouseEventArgs e) =>
```

```cs
#pragma warning disable CS0618 // Type or member is obsolete
```

swharden:
> I am not sure how can we handle Avalonia and WPF while keeping Configuration.cs generic

This is the fundamental problem that makes this hard... I'm guessing the best solution is to handle mapping in the control (not the backend), then eventually update the backend to use button-agnostic names (e.g., `PanButton` instead of `LeftButton`).

> better not to override since it will be more difficult to update future releases

>  Do you think it is possible to add some mapping in the Configuration?

I like the direction this is going! Maybe something like this is even simpler...

```cs
public MouseButtons MouseButtonDragPan = MouseButtons.Left;
public MouseButtons MouseButtonDragZoomRectangle = MouseButtons.Middle;
public MouseButtons MouseButtonDragZoom = MouseButtons.Right;
public Keys KeyOnlyHorizontal = Keys.Control;
public Keys KeyOnlyVertical = Keys.Shift;
public Keys KeyZoomRegion = Keys.Alt;

private Control.InputState GetInputState(MouseEventArgs e) =>
    new()
    {
        X = e.X,
        Y = e.Y,
        LeftWasJustPressed = e.Button == MouseButtonDragPan,
        RightWasJustPressed = e.Button == MouseButtonDragZoom,
        MiddleWasJustPressed = e.Button == MouseButtonDragZoomRectangle,
        ShiftDown = ModifierKeys.HasFlag(KeyOnlyVertical),
        CtrlDown = ModifierKeys.HasFlag(KeyOnlyHorizontal),
        AltDown = ModifierKeys.HasFlag(KeyZoomRegion),
        WheelScrolledUp = e.Delta > 0,
        WheelScrolledDown = e.Delta < 0,
    };
```

EFeru:
I tried the change above. It works in normal plot but not as expected with draggble lines, see below:

https://user-images.githubusercontent.com/24780745/137019999-63ab7a79-fc32-4cf5-a8e4-058d68fd9241.mp4

Also for WPF version is hard to implement this change becasuse the buttons are in the `MouseEventArgs e`:
https://github.com/ScottPlot/ScottPlot/blob/70285d42749f5fd951be8c24489e056e9e098c3d/src/controls/ScottPlot.WPF/WpfPlot.xaml.cs#L197-L202

Since this become more intricated than expected, I am thinking for now to close this PR, and maybe later if you have a better idea on how to do it we can re-address this topic.

Issue #1219, opened 2021-08-22
Scot Plot font size is so small and the performance is low


MohammadSadeghSalimian:
Scott Plot's performance decreases significantly when I plot it in WPF. The conditions are a bit better in Windows Form. However, since my monitor is a 4K the axis label is tiny, unable to be read in both WPF and Windows Form. 

```cs
Random rand = new Random(0);
int pointCount = 100_000;
double[] ys = new double[pointCount];
double[] xs = new double[pointCount];
for (int i = 1; i < ys.Length; i++)
{
    ys[i] = ys[i - 1] + rand.NextDouble() - .5;
    xs[i] = xs[i - 1] + rand.NextDouble();
}

WpfPlot.Plot.AddSignalXYConst(xs, ys);
WpfPlot.Plot.Style(ScottPlot.Style.Blue1);
WpfPlot.Plot.SaveFig("signalxyconst_quickstart.png");
```

![ScottPlot](https://user-images.githubusercontent.com/23130925/130355535-1576b555-fccc-47c0-9aee-e0fd3a1d87db.png)

bclehmann:
The WPF control is DPI aware so the font should be the same as on an HD screen, although maybe font rendering doesn't yet respect DPI. Regardless, you can adjust the font size for the axes by following this example: https://swharden.com/scottplot/cookbooks/4.1.16/category/plottable-text/#custom-fonts

As for performance, I don't think 100 000 points should be too much for signalXY, although I haven't used signalXY very much. I also wouldn't expect a noticable difference between equivalent code in WPF and WinForms. This makes me think you're accidentally calling `AddSignalXY` and reallocating that array on every render or something.

If you could show more detailed source code (with any edits you'd like to preserve anonymity or similar) that may help us fix your performance issues.

MohammadSadeghSalimian:
> 
> 
> The WPF control is DPI aware so the font should be the same as on an HD screen, although maybe font rendering doesn't yet respect DPI. Regardless, you can adjust the font size for the axes by following this example: https://swharden.com/scottplot/cookbooks/4.1.16/category/plottable-text/#custom-fonts
> 
> As for performance, I don't think 100 000 points should be too much for signalXY, although I haven't used signalXY very much. I also wouldn't expect a noticable difference between equivalent code in WPF and WinForms. This makes me think you're accidentally calling `AddSignalXY` and reallocating that array on every render or something.
> 
[...]

PR #1364, opened 2021-10-15
Controls: enable toggling display scaling

### LINKED ISSUE #1352: Text and lines are blurry when display scaling is enabled
## Bug Report

**Issue:** 
When using a display that has a scale factor other than 100% in Windows (see image) will result in bad scaling of the text.
image

**Reproducing:** 

1. Set the scale factor of a display to something other than 100% (e.g. 200% as in the screenshot).
2. Make sure that this display is the main display.
3. Start your application. (Make sure that your application has at least one normal textblock and a plot. Both text sizes should be the same.)
4. => The size on screen of the textblock will be much larger than the text in the plot.

**Expected behavior:**
* The text sizes should be the same.

### System Details
* **ScottPlot Version:** 4.1.25
* **Application Type:** WPF
* **.NET Version:** NET Framework 4.7.2

swharden:
This PR refactors display scale detection and lets the user define or disable display scaling. Resolves #1352

avascale

### DIFF
--- a/dev/changelog.md
+++ b/dev/changelog.md
@@ -4,6 +4,7 @@
 _In development / not yet on NuGet_
 * Colorbar: Exposed fields for additional tick line and tick label customization (#1360) _Thanks @Maoyao233_
 * Plot: Improved `AxisAutoY()` margins (#1363) _Thanks @Maoyao233_
+* Radar Plot: `LineWidth` may now be customized (#1277, #1369) _Thanks @bclehmann_
 * Controls: Stretching due to display scaling can be disabled with `Configuration.DpiStretch` in WPF and Avalonia controls (#1352, #1364) _Thanks @ktheijs and @bclehmann_
 
 ## ScottPlot 4.1.26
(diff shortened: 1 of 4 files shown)


swharden:
Hi @bclehmann, I'd value your input on this one!

I'd like to make display scaling toggleable, and I'm having trouble figuring out what the least astonishing default behavior should be and how to name these things...

scaling2\

## Default Behavior

Do you think plots should be stretched (but may appear blurry, #1352) or 1:1 pixel resolution so everything is sharp but text is too small (#1219)... both are astonishing 😅 

## Naming

`WpfPlot.DisplayScale.Enabled = true` may be unclear whether this means the plot is to be stretched to _respect_ display scaling (large, blurry text), or _compensated_ for display scaling to achieve native pixel resolution (sharp, but small text). Which behavior would you expect `Enabled` to do?

Do you think we should avoid the binary name and go for something more descriptive like `DisplayScale.StretchBitmap = true`?

swharden:
Hey @bclehmann, thanks again for your input! I added DPI scaling to the Avalonia control (and created a demo too). I'm pretty sure it works as expected, but if you want to look over it real quick it could probably use another set of eyes for a sanity check 😅 

avascale

Issue #1387, opened 2021-10-20
WPF Plot Stops Refreshing Until Mouse Panning Stops


jbuckmccready:
## Bug Report

**Issue:** After a brief time of continuous mouse panning (a second or so) the plot stops refreshing until the mouse stops moving.

**Reproducing:** Click and drag pan constantly (do not let the mouse stop), at first everything refreshes and works but then it stops refreshing and as long as the mouse is kept moving it never redraws, once the mouse stops (even with mouse button not released) it refreshes but has the problem again if panning starts again. The bug is consistently reproducible with a debugger attached to essentially an empty example (see code, axes stop refreshing), or when using the exact same example code in a larger application (without debugger attached). I am not sure why it doesn't happen in all cases. I have investigated and tried toggling/changing all of the configuration quality options (under `WpfPlot.Configuration.Quality` and `WpfPlot.Configuration.QualityConfiguration`) but nothing seems to fix the issue.

This issue is also always reproducible when running the WPF ScottPlot Demo or sandbox/WpfApp in the repository with debugger attached (but does not have the problem when running without debugger). This issue does not seem to be reproducible in WinForms with debugger attached.

I don't mind the problem being present with a debugger attached, but I have the problem without a debugger attached when using ScottPlot.WPF in a larger application.

```cs
public partial class MainWindow : Window
{
    public MainWindow()
    {
        InitializeComponent();
        WpfPlot.Plot.Clear();
        WpfPlot.Refresh();
    }
}
```

```xaml
<Window x:Class="GeometryScene.MainWindow"
        xmlns="http://schemas.microsoft.com/winfx/2006/xaml/presentation"
        xmlns:x="http://schemas.microsoft.com/winfx/2006/xaml"
        xmlns:d="http://schemas.microsoft.com/expression/blend/2008"
        xmlns:mc="http://schemas.openxmlformats.org/markup-compatibility/2006"
        mc:Ignorable="d"
        Title="MainWindow" Height="450" Width="800">
    <Grid x:Name="MainGrid">
        <WpfPlot x:Name="WpfPlot"></WpfPlot>
    </Grid>
</Window>

```

### System Details
* **ScottPlot Version:** 4.1.26
* **Operating System:** Windows 10
* **Application Type:** WPF
* **.NET Version:** NET Framework 4.7.1 or NET Framework 4.8.1

jbuckmccready:
[...]

Issue #1036, opened 2021-05-18
Refactor: System.Drawing → Maui.Graphics


swharden:
* September, 2021: [System.Drawing.Common only supported on Windows](https://aka.ms/systemdrawingnonwindows)
* July, 2021: [Make System.Drawing.Common only supported on Windows](https://github.com/dotnet/designs/blob/main/accepted/2021/system-drawing-win-only/system-drawing-win-only.md)

Microsoft is sunsetting cross-platform support for `System.Drawing.Common` in .NET 6. This will be highly disruptive since ScottPlot currently uses `System.Drawing` extensively for rendering graphics.

This issue will track progress toward migrating to an alternative render system. This will likely be `Microsoft.Maui.Graphics` (which works on all platforms, not just in MAUI apps). `System.Drawing` will be totally removed from ScottPlot. 

This change will be associated with a major version increase (ScottPlot 5.0)

Related: #1413, #1431

bclehmann:
> It would be nice if console applications could reference colors without having to explicitly install/use the System.Drawing.Common package.

Don't console apps already have to include System.Drawing.Common in order to use ScottPlot?

That said, I think this is a good plan, especially looking towards the future. It would be nice if it could have functions for decoding hex colour strings (i.e. `"#ff0000ff"`), as that's one of the things I want most in System.Drawing.Common, but that's just a wishlist, no pressure.

It would also be helpful to have a healthy dictionary of "known colours", whether that's just the predefined colours in System.Drawing.Common, or those in CSS, or those in Skia. For what it's worth, I think they all have the same predefined colours as it is.

swharden:
> Don't console apps already have to include System.Drawing.Common in order to use ScottPlot?

I don't think so... I think you can make a new console project and just use NuGet to install ScottPlot, and it works as long as you never pass a System.Drawing.Color as an argument.

> nice if it could have functions for decoding hex colour strings

Yes! Quite a few similar helper functions would be very useful.

> one of the things I want most in System.Drawing.Common
```cs
System.Drawing.Color myColor = System.Drawing.ColorTranslator.FromHtml("#ff0000ff");
```

> dictionary of "known colours"
* System.Drawing: http://www.flounder.com/csharp_color_table.htm
* CSS: https://css-tricks.com/snippets/css/named-colors-and-hex-equivalents/
[...]

PR #1388, opened 2021-10-20
Use WriteableBitmap to improve performance of WPF rendering

jbuckmccready:
Added use of a WriteableBitmap to avoid unneeded allocations and rebinding the image element source when the backend bitmap is only updated (not changed)

**Purpose:**
Improves performance of WPF rendering and fixes https://github.com/ScottPlot/ScottPlot/issues/1387.

### DIFF
--- a/dev/changelog.md
+++ b/dev/changelog.md
@@ -6,6 +6,8 @@ _In development / not yet on NuGet_
 * Plot: Improved `AxisAutoY()` margins (#1363) _Thanks @Maoyao233_
 * Radar Plot: `LineWidth` may now be customized (#1277, #1369) _Thanks @bclehmann_
 * Controls: Stretching due to display scaling can be disabled with `Configuration.DpiStretch` in WPF and Avalonia controls (#1352, #1364) _Thanks @ktheijs and @bclehmann_
+* Axes: Improved support for log-distributed minor tick and grid lines (#1386, #1393) _Thanks @at2software_
+* Axes: `GetTicks()` can be used to get the tick positions and labels from the previous render
 * WPF Control: Improved responsiveness while dragging with the mouse to pan or zoom (#1387, #1388) _Thanks @jbuckmccready_
 
 ## ScottPlot 4.1.26
--- a/src/ScottPlot/Renderable/Axis.cs
+++ b/src/ScottPlot/Renderable/Axis.cs
@@ -374,8 +374,12 @@ namespace ScottPlot.Renderable
         /// <summary>
         /// Sets whether minor ticks are evenly spaced or log-distributed between major tick positions
         /// </summary>
-        public void MinorLogScale(bool enable) => AxisTicks.TickCollection.MinorTickDistribution =
-            enable ? MinorTickDistribution.log : MinorTickDistribution.even;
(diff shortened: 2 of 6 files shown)


bclehmann:
> OK I cleaned it up - feel free to squash the commits before merge or let me know if you want me to.

We don't normally squash PRs into one commit, so I think that's ok.

bclehmann:
Also, it's worth noting that this enhancement could likely be made to WinForms and Avalonia. Non-Windows platforms don't have `System.Windows.Media.Imaging.WritableBitmap`, but Avalonia seems to offer an [alternative](http://reference.avaloniaui.net/api/Avalonia.Media.Imaging/WriteableBitmap/).

swharden:
> Non-Windows platforms don't have `System.Windows.Media.Imaging.WritableBitmap`, but Avalonia seems to offer an [alternative](http://reference.avaloniaui.net/api/Avalonia.Media.Imaging/WriteableBitmap/).

I don't want to reinvent the wheel here... but is this something we could implement ourselves? The [BMP file format](https://en.wikipedia.org/wiki/BMP_file_format) isn't that bad, and an obligate _writer_ could be more straightforward to implement than a _reader_ that would have to support multiple/nuanced file structures and header types.

> If you're interested that could become part of this or a later PR, or we can open an issue for someone to do that later.

I'm definitely interested in exploring this idea for the other two controls! A separate PR/issue sounds like a good idea because I see the warning sighs of a rabbit hole 😉

swharden:
I'm going to test this now and hopefully merge it in tonight... Thanks @jbuckmccready for this PR, and @bclehmann for your input! 🚀 

> @swharden: but is this something we could implement ourselves? 

> @jbuckmccready: I didn't want to affect the backend API

I should have clarified https://github.com/ScottPlot/ScottPlot/pull/1388#issuecomment-948105381 was directed toward @bclehmann, and implementing an in-memory bitmap writer could be a goal for another issue/PR. This PR is looking great, and it certainly makes sense to use `System.Windows.Media.Imaging.WritableBitmap` for the WPF control 👍 

> utilizing the GPU to rasterize directly to a frame

I'm inclined away from this because I want to minimize the dependencies (software and hardware) this project takes on. If we went into GPU land there would be some interesting mathematical operations we could optimize too. I don't intend to go there any time soon (beyond Skia/OpenGL support for hardware-accelerated 2D drawing, #1036).

> It should be possible to reuse all the bitmap buffers involved until a resize event occurs and the bitmap buffer must be increased or decreased

ScottPlot actually does this under the hood (re-using a `System.Drawing.Bitmap` and drawing on it, creating a new one only when resized). 

> double swap type of architecture so the consumer is never blocked

Double-buffering is an excellent concept to consider! I added it to the triage list #1028 to revisit in the future. I did some early experiments with this long ago, but I think it's definitely worth reconsidering in the future.

Note that the primary `System.Drawing.Bitmap` that is re-used by ScottPlot is displayed as an `Image` in a `PictureBox` in the Windows Forms control. The `PictureBox` `Image` is double-buffered out of the box, allowing the WinForms control to update that image in some situations without blocking the UI thread.

Issue #1420, opened 2021-11-02
Ticks: do not hide minor ticks when major ticks are off the screen


swharden:
As reported by @at2software, minor ticks are being hidden if there are not at least 2 major ticks present. This causes minor ticks to appear and disappear when zoomed-in to manual or custom major ticks.

![minor-tick](https://user-images.githubusercontent.com/4165489/139955820-ccdd77a1-0b43-4298-8837-74ab9453b441.gif)

https://github.com/ScottPlot/ScottPlot/blob/36c42ab14bdcf1665378bbd25b9b34875c4f2c6e/src/ScottPlot/Ticks/TickCollection.cs#L406-L407

https://github.com/ScottPlot/ScottPlot/blob/36c42ab14bdcf1665378bbd25b9b34875c4f2c6e/src/ScottPlot/Ticks/TickCollection.cs#L432-L433

PR #1421, opened 2021-11-03
Always generate at least 2 major ticks (#1420)

bclehmann:
**Purpose:** #1420 

Note that this will create minor ticks even if there are no major ticks on-screen (but the minor ticks may be above or below the drawn area).

### 1 major tick visible
image

### 0 major ticks visible
image

### 0 major or minor ticks visible (zoomed in between minor ticks)
image

### DIFF
--- a/dev/changelog.md
+++ b/dev/changelog.md
@@ -2,6 +2,9 @@
 
 ## ScottPlot 4.1.28
 _In development / not yet on NuGet_
+* Ticks: Improved minor tick and minor grid line placement (#1420, #1421) _Thanks @bclehmann and @at2software_
+* Palette: Added Amber and Nero palettes (#1411, #1412) _Thanks @gauravagrwal_
+* Style: Hazel style (#1414) _Thanks @gauravagrwal_
 
 ## ScottPlot 4.1.27
 _Published on [NuGet](https://www.nuget.org/packages?q=scottplot) on 2021-10-24_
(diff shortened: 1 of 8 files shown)


bclehmann:
I still need to fix an issue where we can draw major ticks above the plotting area:

bclehmann:
I noticed something while looking at this code: https://github.com/ScottPlot/ScottPlot/blob/1d1d8631b165b5f4549981941b048611ff753ab2/src/ScottPlot/Renderable/AxisTicks.cs#L72-L85

These function calls operate on predominantly the same data, but some of them take `double[]`, and some `TickCollection`. This could be prone to regressions or other bugs in the future, as they take different codepaths for restricting the ticks to only those that are visible.

bclehmann:
I'm now realizing that this might still be broken in the case where the user set manual ticks rather than manual tick step.

I'll take another look tomorrow.

swharden:
Hi @bclehmann, thanks for this PR! ... and sorry for my extreme delay. I moved to a new house this month and I got hit with way more disruption to my productivity than expected 😅 Things are starting to return to normal though!

I added some extra logic which adds extra integer ticks in the case that `IntegerPositionsOnly` removes too many ticks after reducing positions down to integers only (the original use case for #1420).

before | after
---|---
|1420b

swharden:
> ... some of them take double[], and some TickCollection. This could be prone to regressions or other bugs in the future, as they take different codepaths for restricting the ticks to only those that are visible.

This warning is well-received! This is one of many serious issues with the present tick system. I'm going to prioritize refactoring the tick system soon (#1028) since official Maui.Graphics development is moving more slowly than anticipated.

Issue #1277, opened 2021-09-18
Allow Radar Chart to specify a width for the Radar's border


Rayffer:
## Features Suggestion

**Feature description:** 
Give the option to specify a value for the thickness of the RadarChart border to allow for improved visibility when an arm has values while the neighbouring arms do not, which produces a polygon of the same width as the radarchart arm and difficults visibility on it.

**Code example:**

```cs

WpfPlot wpfPlot = new();
int borderThickness = 2;
RadarPlot radarPlot = wpfPlot.Plot.AddRadar(values);
radarPlot.BorderThickness = borderThickness;

wpfPlot.Render();
```

If it makes no sense, I believe it would to make the Thickness of the border at least 2 instead of the current default value, which is 1.

Rayffer:
![image](https://user-images.githubusercontent.com/34773204/134175563-5680f2d0-a72f-4690-888d-1b6684af1ce2.png)

Just to give an example of how this would look.

swharden:
I like the idea! It looks like the pen that outlines that polygon is defined here, and since it lacks a thickness argument its thickness is always 1 in the current system.

https://github.com/ScottPlot/ScottPlot/blob/e54af09a39137ce50f0e8552d5e55595a7b5fa7e/src/ScottPlot/Plottable/RadarPlot.cs#L242

`BorderThickness` is a good and literally accurate name. Elsewhere in the library this is called `LineWidth`. Maybe `OutlineThickness` could work too. I'm probably okay with any name here - it's easily discoverable with a predictive IDE.

Seems like a good hacktoberfest PR 😎  #1274

Rayffer:
Hello, after checking the changes I think this does not solve this issue, @bclehmann added support for customizing the width of the chart lines, not the outline of the polygon for each entry in the radar chart.

I might have expressed incorrectly but my intent was to allow for polygon outline's line width to be customizable

PR #1426, opened 2021-11-10
Define polygon outline border thickness

Rayffer:
**Purpose:**
This pull request aims to provide the possibility to specify a Radar Chart's polygon outline
#1277 

To specify outline, type the following code
```cs
    double[,] values = {
        { 5, 3, 10, 15, 3, 2 },
        { 5, 2, 10, 10, 1, 4 },
    };

    var radar = plt.AddRadar(values);
    radar.OutlineThickness = 3;
```


bclehmann:
Hey, it looks like this PR is based off of a pretty out of date head (from mid-september). I think you need to rebase.

Rayffer:
Done, thought I was up to date in my fork, thanks for the heads up!

src/ScottPlot/Plottable/RadarPlot.cs
```diff
@@ -106,6 +106,8 @@ public class RadarPlot : IPlottable
         public int XAxisIndex { get; set; } = 0;
         public int YAxisIndex { get; set; } = 0;
 
+        public float OutlineThickness { get; set; } = 1;
```
bclehmann, line 109: This is a confusing name, especially with `LineWidth`, but I can't think of a better one. Maybe `DataThickness`?

Maybe Scott will have an idea.

bclehmann (approved):
Looks good, I don't know why the CI failed, but I don't think it had anything to do with you

src/ScottPlot/Plottable/RadarPlot.cs
swharden: I'll split the difference and go with `OutlineWidth` 👍

swharden:
Hi @Rayffer, thanks for this PR! Sorry for the long delay on this one - I'm in the middle of moving and the process has been a lot more disruptive than I anticipated it would be.

> Looks good, I don't know why the CI failed, but I don't think it had anything to do with you

Indeed! Re-running the job it appears all the tests passed this time

Thanks for your help along the way @bclehmann!

Issue #1220, opened 2021-08-22
Default font renders incorrectly on Ubuntu Linux


Joe-Westra:
## Bug Report

**Issue:** Incorrect characters on axis tick labels (see attached picture).  I have tried changing the font, which changes the amount of boxes shown, but cannot get tick labels as shown on the website examples.
![UNfiltered](https://user-images.githubusercontent.com/20119513/130373383-295aeb6e-a186-4343-811d-c50c4f6babe7.png)

**Reproducing:** Create a .NET 5 console application, create a basic plot and save it to disk.  Open saved file and examine the x and y axis ticks.  Box characters show up instead of values.

```cs
            //signal + noise
            double fs = 1000; //sampling rate
            double fw = 5; //signal frequency in hz
            double fn = 50; //noise frequency in hz
            double n = 10; //number of periods to show
            double A = 100; //signal amplitude
            double N = 10; //noise amplitude
            int size = (int)(n * fs / fw); //sample size

            var t = Enumerable.Range(1, size).Select(p => p * 1 / fs).ToArray();
            var y = t.Select(p => (A * Math.Sin(2 * Math.PI * fw * p)) + (N * Math.Sin(2 * Math.PI * fn * p))).ToArray();

            var bandpass = OnlineFirFilter.CreateBandpass(ImpulseResponse.Finite, fs, 0, 10);

            double[] bp = bandpass.ProcessSamples(y);

            double[] dataX = t;
            var plt = new ScottPlot.Plot(640, 480);
            plt.XAxis.TickLabelStyle(fontName: "Samanata", fontSize: 12);
            plt.YAxis.TickLabelStyle(color: Color.DarkBlue, fontName: "Meera", fontSize: 16);
            plt.AddScatter(t, brwide);
            plt.SaveFig("brwide.png");
```

### System Details
* **ScottPlot Version:** 4.1.16
* **Operating System:** Ubuntu Linux 20.04
* **Application Type:** console
* **.NET Version:** .NET 5.0

bclehmann:
I can replicate this on WSL, and it looks like this is a duplicate of #1079, also on [StackOverflow](https://stackoverflow.com/a/67729839/3513171). The solution is to explicitly install `System.Drawing.Common`.

swharden:
Hi @Joe-Westra, thanks for reporting this!

It looks like System.Drawing is having a hard time properly rendering text on your system.

## Code Sample

[...]

PR #1438, opened 2021-11-16
Port to eto

rafntor:
Ref discussion here https://github.com/ScottPlot/ScottPlot/discussions/1425

This PR adds a control for [Eto.Forms](https://github.com/picoe/Eto) and also a tiny sandbox-project and a port of the extensive demo-application that shows the capabilities of the library.

Missing :
- The demo-application started as a copy of the winforms-version of the demo and there are designer-generated sections that are just commented out and should probably be deleted before merge(?)
- There is no nuget-generation project added and no CI-steps made
-


src/controls/ScottPlot.Eto/ScottPlot.Eto.csproj
```diff
@@ -0,0 +1,18 @@
...
+
+    <PropertyGroup>
+        <TargetFramework>netstandard2.0</TargetFramework>
+        <Nullable>enable</Nullable>
+        <LangVersion>latest</LangVersion>
+        <GeneratePackageOnBuild>false</GeneratePackageOnBuild>
+    </PropertyGroup>
+
+    <ItemGroup>
+        <PackageReference Include="Eto.Forms" Version="2.5.11" />
```
bclehmann, line 11: Does this mean this control is restricted to Windows Forms? I know I asked that question earlier but I wasn't sure if your answer was restricted to the demo app, not the control.

As an aside, it would probably be beneficial to have the demo target something cross-platform (or distribute multiple builds of it), but I would shy away from something like GTK as it's not installed by default on Windows. The more people can click and run the demo the better, in my opinion.

rafntor: The control is not restricted to WinForms - it is ready to be used by all Eto platforms.
The demo app is also ready to go for all Eto platforms - but needs to be built specifically for each.

In GitHub Actions it should be possible to do it with something like this ;-)
image
There are probably details to be addressed in creating those executables, but it should definitely be doable

rafntor: This is maybe preferable
image

rafntor: WinForms on the left - Wpf to the right
image
Wpf-buttons in Eto will not wrap text so looks bad (It might be doable but at a minimum requires some extra tweaking)

swharden:
Hi @rafntor, thanks again for this PR!

I made some small changes but I think this is about ready to merge. Shortly after merging I'll publish the control as a package and can add a quickstart to the website. Can you help by updating this PR to resolve the following two issues?

1) There are some warnings at build time - can you resolve them? At a glance it looks like most of them relate to `nullable`

2) Can you add a minimal quickstart and any special setup comments you have to the README.md file that now lives in the Eto control folder? That document will be the documentation displayed on the NuGet page. You could reference the [Avalonia readme file](https://github.com/ScottPlot/ScottPlot/blob/master/src/controls/ScottPlot.Avalonia/README.md) as an example. Note that file is rendered on https://www.nuget.org/packages/ScottPlot.Avalonia

After those two tasks are complete this looks ready to merge! 🚀 

PS: I've been copy/pasting this a lot lately: _Sorry for the delay! It took me an unexpectedly long time to get to this because I moved houses and it was a lot more disruptive than I predicted, then right after I got sick for several weeks. I'm doing better now and am happy to be picking this up again. Thanks for your patience along the way!_

rafntor:
Great! Sure, will attack those points, not quite sure how fast I can get it done but as soon as I can

swharden:
> will attack those points, not quite sure how fast I can get it done but as soon as I can

Sounds great! No need to rush - I'll leave this open as long as you need. I have plenty of open issues to work on, and everyone can take some time to relax over the holidays. Thanks again for this contribution!

rafntor:
I have done the issues I think, but unfortunately also detected a problem ; pls check the GTK-version picture [here](https://github.com/ScottPlot/ScottPlot/blob/cad20906038d76b2faeef97f3e39cb108054d02b/src/controls/ScottPlot.Eto/README.md).
There is something wrong with axis-labels and I dont know what the cause is, do you have any idea? Maybe something with calculation of space needed for the text somehow?

swharden:
> There is something wrong with axis-labels and I dont know what the cause is, do you have any idea?

image

I recall seeing this before somewhere... #1220

I think the solution is to manually `dotnet add package System.Drawing.Common` or edit the csproj file like this:
https://github.com/ScottPlot/ScottPlot/blob/e7ababd235acc13979fbb9d4b9b60c5914e28eba/src/ScottPlot/ScottPlot.csproj#L33

Let me know if that does the trick!

rafntor:
Yes, spot on ! I'm happy with the way it looks now, let me know if you spot any remaining things ;-)

PR #1440, opened 2021-11-16
PopulationPlot: ScatterOnBox data format

Syntaxrabbit:
Screenshot_20211117-171029_Firefox

Screenshot_20211117-171047_Firefox

This commit creates a new type of population plot with semitransparent box plots and scatters.
The DataFormat enum type "ScatterOnBox" saves some space on the diagram; often used in scientific environments for comparison of many populations.


swharden:
Hi @Syntaxrabbit, thanks for this! I'll take a closer look this evening.

If you could edit your original comment to include a screenshot of what this plot type looks like, that would be great!

Bonus points if you can add a cookbook example by adding to the bottom of [Recipes/Plottable/Population.cs](https://github.com/ScottPlot/ScottPlot/blob/master/src/cookbook/ScottPlot.Cookbook/Recipes/Plottable/Population.cs) 🚀

Syntaxrabbit:
Thank you for your help. Of course, I will do the cookbook later.

swharden:
Thanks for this feature @Syntaxrabbit! The `PopulationPlot` plot type is an intimidatingly large and complex plot type with a confusing API. It will likely get deeply refactored in the next major version of ScottPlot to make it easier to use. You did a good job wedging this feature in!

I added a cookbook example showing how to change the display format to overlapping scatter/bar (not scatter/box). I'm not really satisfied with the current options - it seems that box-and-whisker plots are _always_ drawn as half-boxes. I'll create a triaged issue to hopefully work to improve this some time in the future (#1028)

PS: I've been copy/pasting this message a lot lately: _Sorry for the delay! It took me an unexpectedly long time to get to this because I moved houses and it was a lot more disruptive than I predicted, then right after I got sick for several weeks. I'm doing better now and am happy to be picking this up again. Thanks for your patience along the way!_

PR #1471, opened 2021-12-09
 IRenderable and IPlottable.Render with Graphics object for clean printouts

CBuckel:
Hi,

I refactored the IRenderable and IPlottable Render function to work directly on a Graphics object insted of a bitmap. This allows to use the Graphics object during printing and render directly on the printer. This improves the print result a lot. 

[ScottPlotClear.pdf](https://github.com/ScottPlot/ScottPlot/files/7684122/ScottPlotClear.pdf)
[ScottPlotBlurry.pdf](https://github.com/ScottPlot/ScottPlot/files/7684124/ScottPlotBlurry.pdf)

I have implemented this for myself because I need a clean printout. Maybe it is also interesting for others. If not feel free to decline the request. Since it's my first request, I ask to be lenient with me and possibly ask for a small hint what I can do better. 

- IRenderable.Render and IPlottable.Render changed to _void Render(PlotDimensions dims, Graphics gfx)_

- OffsetX + OffsetY Added to Settings --> Adds an offset for the entire plot in Graphics object (need for printing at other coordinates than 0:0 )

- ClipToDataArea() used to clip drawing to data area of the plot, set quality and transform coordinates according to OffsetX/OffsetY

### DIFF
--- a/src/ScottPlot/Drawing/GDI.cs
+++ b/src/ScottPlot/Drawing/GDI.cs
@@ -94,6 +94,7 @@ namespace ScottPlot.Drawing
             return gfx;
         }
 
+
         public static System.Drawing.Graphics Graphics(Bitmap bmp, PlotDimensions dims, bool lowQuality = false, bool clipToDataArea = true)
         {
             Graphics gfx = Graphics(bmp, lowQuality, dims.ScaleFactor);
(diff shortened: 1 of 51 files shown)


swharden:
Hi @CBuckel, thanks for this PR! I'm finally getting time to check out out. The improvement is impressive:

Drawing on a `Bitmap` | Drawing on a `Graphics`
---|---
image|image

It sounds like you found an excellent solution to your issue! Thanks so much for sharing this PR. It seems likely it will help others who have a focus on printing and/or PDF export.

## PR Modifications

Although this PR looks large (>50 files), it changes the function signatures in `IPlottable` and `IRenderable` which require modification of numerous classes. The core modification is confined to just a few files: GDI.cs, Plot.cs, Plot.Render.cs, PlotDimensions.cs, and Settings.cs

## Concern regarding merging (backward compatibility)

My biggest hesitation with this PR (and why I'll probably close it) is because I expect many people will have made their own custom renderable or plottable objects. Changing the interface in this way could be largely disruptive for many people:

```cs
// IPlottable (original)
void Render(PlotDimensions dims, System.Drawing.Bitmap bmp, bool lowQuality = false);
```

```cs
// IPlottable (this PR)
void Render(PlotDimensions dims, Graphics gfx);
```

Alternative implementations could be explored (e.g., preserving the existing signature but _adding_ a second one that passes a graphics object), but I don't think we should go down that road because of the issue discussed in the roadmap below

## Roadmap

ScottPlot 4 uses `System.Drawing` and has largely stabilized. ScottPlot 5 is early in development and will not use `System.Drawing` at all (since Microsoft is sunsetting cross-platform support for it) and will instead use the `Microsoft.Maui.Graphics` package (#1036). Unfortunately that release has been delayed until next year, so this project is stuck between two drawing technologies.

Because I intend to move entirely away from System.Drawing soon, I don't have a lot of motivation to execute such a large refactor of the existing code base to support a System.Drawing-specific feature. However I really appreciate this PR - I can tell these modifications have already helped your application, and I am sure this code will be useful for anyone who may wish to achieve superior printing quality from the existing library in the future!

Thanks again,
Scott

PR #1492, opened 2021-12-23
Add checks to avoid OutOfMemoryException in DrawArc

arthurits:
**Purpose:**
* Add checks for small sweepAngle when calling `DrawArc` in functions `RenderBackground` and `RenderGaugeForeground`
https://github.com/ScottPlot/ScottPlot/blob/b4f881945493e0afb45d7d43f4ea54ee6369252d/src/ScottPlot/Plottable/RadialGauge.cs#L126
```cs
// This check is specific to System.Drawing since DrawArc throws an OutOfMemoryException when the sweepAngle is very small.
            if (BackEndAngle <= 0.01)
                BackEndAngle = 0;
```
https://github.com/ScottPlot/ScottPlot/blob/b4f881945493e0afb45d7d43f4ea54ee6369252d/src/ScottPlot/Plottable/RadialGauge.cs#L145
```cs
// This check is specific to System.Drawing since DrawArc throws an OutOfMemoryException when the sweepAngle is very small.
            if (SweepAngle <= 0.01)
                SweepAngle = 0;
```
* A private setter was also added to `BackEndAngle` property
```cs
public double BackEndAngle
        {
            get
            {
                double maxBackAngle = CircularBackground ? 360 : MaximumSizeAngle;
                if (!Clockwise) maxBackAngle = -maxBackAngle;
                return maxBackAngle;
            }
            private set { BackEndAngle = value; } // Added for the sweepAngle check in DrawArc due to System.Drawing throwing an OutOfMemoryException.
        }
```


arthurits:
Maybe this would better discussed in another PR, but I've found that the property name `BackEndAngle` could be modified into `BackAngleSweep`, which would be more descriptive.
https://github.com/ScottPlot/ScottPlot/blob/b4f881945493e0afb45d7d43f4ea54ee6369252d/src/ScottPlot/Plottable/RadialGauge.cs#L42

swharden:
> Maybe this would better discussed in another PR, but I've found that the property name BackEndAngle could be modified into BackAngleSweep, which would be more descriptive.

This is technically a breaking change, but a small one and on a relatively new plot type so I'm fine with it. It can be included in this PR too. It's up to you whether you want to keep the old property name too marked `Obsolete` (I'm fine either way).

Perhaps the XML docs for this property could be more descriptive though, noting what it does and when to use it rather than just reiterating how it works.

Thanks!

swharden:
Also let me know when you're done with this PR so I don't merge it too early 👍

arthurits:
> This is technically a breaking change, but a small one and on a relatively new plot type so I'm fine with it. It can be included in this PR too. It's up to you whether you want to keep the old property name too marked `Obsolete` (I'm fine either way).

I believe that `BackEndAngle` (now renamed to `BackAngleSweep`) is only used internally inside `RadialGauge.cs` to just retrieve the angular size of the gauge background when calling `gfx.DrawArc`. Although the property access is public, in fact it's used as private one. That's why I decided against keeping and marking the old property as `Obsolete`, but I'm completely fine if you think otherwise.

I also updated the XML comments. Feel free to adapt them to your preferences, and merge the PR if everything is OK.

PR #1528, opened 2022-01-02
CI: build all packages

swharden:
Something changed in GitHub packages and the Avalonia package will not build and was not able to be deployed.

This wasn't noticed previously because only the deployment action (not the CI action) built packages routinely.

This PR adds a package building step to the CI pipeline.


swharden:
replaced by #1529

PR #1529, opened 2022-01-02
CI: build and deploy packages individually

swharden:
copy of #1528 because auto-merge was failing for unknown reasons


swharden:
auto-merge was failing because the pipeline itself was modified in this PR

PR #1557, opened 2022-01-15
New Plottables: RepeatingAxisLine and AxisLineVector

BambOoxX:
New Plottables - `RepeatingAxisLine `and `AxisLineVector`
New plots proposed after the discussion in #1535

- The `RepeatingAxisLine `behaves as an `AxisLine` with additional surrounding lines
    - The first line is positioned at `Position`
    - The subsequent ` Count - 1` lines are positioned according to parameters `Offset `, `Shift `and `RelativePosition`
    - The first line is thicker by default but this can be disabled with the `HighlightReferenceLine `property
- The `AxisLineVector `behaves as an undraggable `AxisLine `and multiple lines are plotted according to the values in a `double[]` object.

PR #1556, opened 2022-01-15
Fix - Support for draggable plottables on `AxisIndex` 0 and 1

BambOoxX:
- Former behavior only worked for `AxisIndex` 0

This PR is related to the discussion on #1545 

There is still work to generalize this to other axis indexes, but I don't know to reach other axes at the moment.

PR #1558, opened 2022-01-15
New plottables - `DraggableMarkerPlot` and `DraggableMarkerPlotInVector`

BambOoxX:
- `DraggableMarkerPlot` is a `MarkerPlot` that can be dragged through a `Plot`. This can serve as a *static* way to pinpoint a location on the axes. By *static* I mean that does not move with the mouse unless dragged
- `DraggableMarkerPlotInVector` behaves somewhat similar to as `ScatterPlot` that in addition has a draggable point. When you drag the point, it goes to the coordinates in the vector closest to the mouse position.

### DIFF
--- a/dev/changelog.md
+++ b/dev/changelog.md
@@ -3,6 +3,7 @@
 ## ScottPlot 4.1.31
 _In development / not yet on NuGet_
 * MultiAxis: Improved support for draggable items placed on non-primary axes (#1556, #1545) _Thanks @BambOoxX_
+* Marker: New plot types `DraggableMarkerPlot` and `DraggableMarkerPlotInVector` give users options to add mouse-interactive markers to plots (#1558) _Thanks @BambOoxX_
 
 ## ScottPlot 4.1.30
 _Published on [NuGet](https://www.nuget.org/packages?q=scottplot) on 2022-01-15_
--- a/src/ScottPlot/Plottable/DraggableMarkerPlot.cs
+++ b/src/ScottPlot/Plottable/DraggableMarkerPlot.cs
@@ -44,16 +44,6 @@ namespace ScottPlot.Plottable
         public event EventHandler Dragged = delegate { };
 
         /// <summary>
-        /// The lower bound of the axis line.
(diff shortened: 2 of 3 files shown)


swharden:
Hi @BambOoxX, thank you for this interesting PR! I'll begin working on it after #1556 and #1557

I haven't read the code in detail yet but I'm curious - would it be possible to inherit [`MarkerPlot`](MarkerPlot.cs) and `IDraggable` to reduce a little bit of code duplication? This technique is exactly what I suggested for @anose001 in #1555 when this exact topic was raised in a discussion question earlier today.

BambOoxX:
@swharden I just trie to use some inheritance, but for `DraggableMarkerPlotInVector` the draggable marker isn't plot anymore. This feels like the new `Render` is not taken into account.

swharden:
Hi @BambOoxX, related to #1558 I got you started here by adding a cookbook example

screenshot-drag

I was confused at first because the default marker color is white and I thought it was a rendering failure but it turns out it was just white on white, but all looks good now!

I changed the marker type to `Hand` but if you want the user to be able to make it a `Crosshair` you could make it `{get;set;}` and default to `Hand`

Can you add a similar cookbook example to demonstrate DraggableMarkerPlotInVector? That use case is a little more complex but I'd like to see how you choose to showcase it. Thanks! 🚀

BambOoxX:
> Can you add a similar cookbook example to demonstrate `DraggableMarkerPlotInVector`? That use case is a little more complex but I'd like to see how you choose to showcase it. Thanks! 🚀

I will try and do an example, I have in mind something close to `ShowValueOnHover` but with something draggable.

However, I can't understand why the marker is not plotted now in `DraggableMarkerPlotInVector`. Could you have a look at the `Render` method ? Is it the correct way to inherit from `ScatterPlot` ?

swharden:
Hi @BambOoxX, thanks for this PR! It looks like you figured everything out and this is ready to merge. These new plot types will be available in the next release 👍 😎 🚀

PR #1559, opened 2022-01-16
Fix - `MouseZoomRect` affects all axes

### LINKED ISSUE #1537: Drag zoom affects first and second axes in different manners
## Bug Report

**Issue:**
I am unsure whether this is a true bug report or a wrong use of the tools on my part, so if it is (not a bug), sorry in advance for the inconvenience.

When plotting quantities on the first and second (at least Y) axes, zooming in and out using the scroll or the drag zoom affects plottables with different `AxisIndex` in different fashions, resulting in relative size between such plottables being different.

I would have expected that the zoom would not affect relative sizes between plottables, but just the axis limits (if such limits are free of course). 

Another remark, but this could take the form of a feature request, in addition/extension to the LockAxisLimits, it may be relevant to give the possibility to only lock one of the limits. This could be interesting in the case where power density or some bars are plotted and one wants to zoom to see small contributions, but not lose the fact that the lower limit has to remain 0.

See the screenshots below
- Initial plot 
image
- Drag zoom
image
- Result of zoom
image

Note that click the central mouse button returns to the correct initial plot

**Reproducing:** (What do others have to do to reproduce the bug?)

```cs
// a code sample may improve communication
namespace live_sandbox
{

    public static class Program
    {
        public static void Main()
        {
            // create a plot
            plt.AddSignal(ScottPlot.DataGen.Sin(51,5));
            plt.AddSignal(ScottPlot.DataGen.Sin(51, 5,0,0.5));
            plt.AddSignal(ScottPlot.DataGen.Sin(51, 5,0,0.25));
            plt.AddSignal(ScottPlot.DataGen.Sin(51, 5, 0, 0.125));

            var yaxis1plot = plt.AddScatterPoints(ScottPlot.DataGen.Sin(51), ScottPlot.DataGen.Cos(51));
            yaxis1plot.YAxisIndex = 1;
            plt.Add(yaxis1plot);


            // launch the plot in an interactive window
            var window = new ScottPlot.FormsPlotViewer(plt);
            window.ShowDialog();
        }
    }
}
```

### System Details
* **ScottPlot Version:** 4.1.29
* **Application Type:** WinForms
* **.NET Version:** NET Framework 4.7.2

BambOoxX:
This PR fixes #1537 to preserve the aspect ratio between axes during a Zoom.
In fact, the current behavior seems in contradiction with the fact that by default axes limits are free and should (to my understanding) be affected in the same manner by the drag zoom, regardless of the axis.

### DIFF
--- a/src/ScottPlot/Settings.cs
+++ b/src/ScottPlot/Settings.cs
@@ -495,9 +495,9 @@ namespace ScottPlot
 
             if (finalize)
             {
+                ZoomRectangle.Clear();
                 foreach (Axis axis in Axes)
                 {
-                    ZoomRectangle.Clear();
                     if (axis.IsHorizontal)
                     {
                         double x1 = axis.Dims.GetUnit(left);
--- a/src/ScottPlot/Settings.cs
+++ b/src/ScottPlot/Settings.cs
@@ -495,12 +495,22 @@ namespace ScottPlot
 
             if (finalize)
             {
-                double x1 = XAxis.Dims.GetUnit(left);
-                double x2 = XAxis.Dims.GetUnit(right);
-                double y1 = YAxis.Dims.GetUnit(bottom);
-                double y2 = YAxis.Dims.GetUnit(top);
-                ZoomRectangle.Clear();
-                AxisSet(x1, x2, y1, y2);
+                foreach (Axis axis in Axes)
+                {
+                    ZoomRectangle.Clear();
+                    if (axis.IsHorizontal)
+                    {
+                        double x1 = axis.Dims.GetUnit(left);
+                        double x2 = axis.Dims.GetUnit(right);
+                        axis.Dims.SetAxis(x1, x2);
+                    }
+                    else
+                    {
+                        double y1 = axis.Dims.GetUnit(bottom);
+                        double y2 = axis.Dims.GetUnit(top);
+                        axis.Dims.SetAxis(y1, y2);
+                    }
+                }
             }
             else
             {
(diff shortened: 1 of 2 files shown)


src/ScottPlot/Settings.cs
```diff
@@ -495,12 +495,22 @@ public void MouseZoomRect(float mouseNowX, float mouseNowY, bool finalize = fals
...
             {
-                double x1 = XAxis.Dims.GetUnit(left);
-                double x2 = XAxis.Dims.GetUnit(right);
-                double y1 = YAxis.Dims.GetUnit(bottom);
-                double y2 = YAxis.Dims.GetUnit(top);
-                ZoomRectangle.Clear();
-                AxisSet(x1, x2, y1, y2);
+                foreach (Axis axis in Axes)
+                {
+                    ZoomRectangle.Clear();
```
bclehmann, line 500: Is there a reason we clear it up here? If possible it would be nice to move it out of the loop.

bclehmann (commented):
Looks good, for whatever reason when I tackled it there was a weird bug, and I didn't have time to figure out why. Thanks for taking a stab at it yoursself.

src/ScottPlot/Settings.cs
BambOoxX: Nope, no reason, I just missed this when moving things around. Solved in 724e892ab66c3111b3dc3cdda5f66a00f7658f9a

bclehmann (approved):

src/ScottPlot/Settings.cs
bclehmann: Looks good, I think only you can resolve this thread.

swharden: > I think only you can resolve this thread

I can resolve it too 😈

Thanks @BambOoxX! I'll merge this tonight 👍

Issue #1422, opened 2021-11-03
ScatterPlotDraggable: experimental scatter plot with mouse-draggable points


EFeru:
## Features Suggestion

**Feature description:** It would be very nice if we could drag points in a plot to reposition them. The use-case that I have is to tune a look-up table and to get quickly a nice shape dragging points in plot comes very handy. At the moment I am changing values in a dataGridView next to it and Refresh plot accordigly. It works but it is time consuming.

Example use-case:
![image](https://user-images.githubusercontent.com/24780745/140175357-199ae9a7-822c-4c53-8ff1-1ab2307e48ad.png)

swharden:
This may be hard, but I'd like to explore it! I'd probably create a new plot type like `ScatterPlotDraggable` and perhaps an interface like `IHasDraggablePoints`

My use case would be to have a more elegant way to create colormaps by dragging points on R/G/B curves (rather than having to lay-out separate controls for each draggable point). See discussion in #458

![](https://user-images.githubusercontent.com/4165489/84584391-e2057500-add1-11ea-8580-56628a887766.png)

EFeru:
Are the sliders already available in ScottPlot or is something experimental?
It could be a solution for me as well. The only drawback would be:
- it occupies more space and
- probably it restricts the drag to only up and down. Of course ideally is to be able to drag the point freely within the previous point and next point X axis value.

swharden:
That screenshot is from an experimental/throwaway app I made to manually adjust RGB levels to create a custom colormap curve. Adding true draggable points _on the plot_ would have made such an app much more elegant

BambOoxX:
@swharden I just had a need for a plot with one draggable point among a vector of values, to serve as an overlay to another plot. Out of curiosity I will try and find an implementation for this issue. In the meantime I will send a PR for the aforementioned plottable.

BambOoxX:
@swharden Here is a toy implementation for the `ScatterPlotDraggable`. It is based on `IPlottable` and `IDraggable` not `IHasPoints` at the moment


![image](https://user-images.githubusercontent.com/42067365/149639843-09603dc1-e4f6-441b-b12b-7b1132f7893e.png)


<details>
<summary>The source looks like this at the moment</summary>
```cs
using System;
using System.Linq;
using System.Drawing;
using ScottPlot.Drawing;
using System.Diagnostics;

namespace ScottPlot.Plottable
{
    /// <summary>
[...]

PR #1560, opened 2022-01-16
New plottable : `ScatterPlotDraggable`

BambOoxX:
This PR implements the functionnalities described in #1422.
The code inherits `ScatterPlot` and `IDraggable` to reduce code duplication
Two additional options (both true by default) are provided in addition to `DragEnabled`
- `DragEnabledX` : allows to free/lock horizontal drag of the scatter points
- `DragEnabledY` : allows to free/lock vertical drag of the scatter points


swharden:
very cool @BambOoxX, thanks! 

You're probably sensing a pattern with this request, but can you add a cookbook example to let others learn how to use this and make it easier for me to explore as well? Thanks! (e.g., #1558 and #1557)

PR #1585, opened 2022-01-21
WIP : Attempt to constrain drag to adjacent points

BambOoxX:
- This currently does not work
- Unless I missed something, I do not understand why the points leave the plot area
- With respect to https://github.com/ScottPlot/ScottPlot/issues/1422#issuecomment-1015981553 this implementation should be able to handle the cases where `Xs` and `Ys` are totally random, at the cost of two initial sort.


swharden:
> I do not understand why the points leave the plot area

Don't feel too bad about this - I gave this a quick try and couldn't figure it out either!

I don't know if this will result in less effort/time or more... but what about modifying `ScatterPlotDraggable` to inherit `ScatterPlotList` instead of `ScatterPlot`? Maybe there's some funniness going on with arrays being referenced vs. copied, and switching to one that uses a `List` instead may help.

BambOoxX:
@swharden I think I'm gonna close this for the moment. I really do not see how I can improve this more at the moment.

PR #1586, opened 2022-01-21
1557 plottable repeatingaxisline continued

BambOoxX:
This PR implements some missing features from #1557 
- `AxisLineVector` now has an example in the cookbook and is draggable as a `ScatterPlotDraggable`
- The bug making `PositionLabel`s disappearing is fixed


swharden:
Looks great! Thanks @BambOoxX 🚀

Issue #1466, opened 2021-12-05
ErrorBar: New plot type for advanced error bar features


bclehmann:
## Features Suggestion

**Feature description:** Allow errorbars which are not necessarily centered on the marker. This could be done by reviving `plt.PlotErrorBars` from 4.0.

**Code example:**

```cs
var err1 = plt.PlotErrorBars(dataX, dataY1, errorXPositive, errorXNegative, errorYPositive, errorYNegative);
```

**More Context:**

The 4.0 cookbook demonstrating this: https://scottplot.net/cookbook/4.0/#plottypes-errorbar-scatter-plot-with-asymmetric-errorbars

A recent StackOverflow question about this: https://stackoverflow.com/questions/70200872/scottplot-errorbar-with-different-top-and-bottom-values/70230626#70230626

bclehmann:
Hey, I know I kinda abandoned this for a month, but I was wondering if you had a preference for either of these two APIs:

```cs
plt.AddErrorBars(dataX, dataY, errorX, errorY);
```

```cs
plt.AddErrorBars(xs1, xs2, ys1, ys2);
```

The first one is probably more convenient for most people, as it's most likely things they already have, their data and their error. But the second one is more flexible. I'm struggling to think of a situation where someone would prefer the second over the first, but I can't shake the feeling that there is one.

swharden:
I'm staring to avoid using the term `data` in variable names because, technically, all variables contain data 🙃 

If I were approaching this from scratch I'd probably make arguments/overloads like this:

```cs
plt.AddErrorBars(xs, ys, xErr, yErr);
plt.AddErrorBars(xs, ys, xErrNeg, xErrPos, yErrNeg, yErrPos);
```

Matplotlib's `pyplot.errorbar()` was interesting to reference here https://matplotlib.org/stable/api/_as_gen/matplotlib.pyplot.errorbar.html ... I don't think it's really idiomatic to do this with C#, but in the Python world it looks like `xErr` and `yErr` are 1D arrays if errorbars are symmetrical and 2D (with two rows) if separate positive/negative values are specified.

> I'm struggling to think of a situation where someone would prefer the second over the first, but I can't shake the feeling that there is one.

I agree! I think defining the center points separate from error magnitudes makes sense here.

PR #1561, opened 2022-01-16
Experimental rendering system using Maui.Graphics, SkiaSharp, and OpenGL

swharden:
This PR explores feasibility of cutting the dependency on `System.Drawing.Common` in favor of a rendering system using `Microsoft.Maui.Graphics` (#1036). Depending on what is discovered here, code here may evolve into the next major version of ScottPlot (version 5).

### Goals of this PR
* Support back to `.NET Standard 2.0` and `.NET Framework 4.6.2`
* no `System.Drawing` anywhere in the project
* nullable support throughout the code base #691
* hardware-accelerated rendering in controls using SkiaSharp and OpenGL
* stateless rendering system to improve support for multi-threading and testing
  * size will never be stored - it will be passed-in and the layout will be recalculated on each render
  * mouse positions or state should not be stored - mouse actions (e.g., click/drag) will be built by controls and passed in

### Major Limitations
These limitations are because Maui.Graphics is not fully mature yet, but:
* There is no `Microsoft.Maui.Graphics.ICanvas.MeasureString()`
* `Microsoft.Maui.Graphics.ICanvas.DrawString()` does not support custom fonts

### Resources
* https://maui.graphics

PR #1588, opened 2022-01-22
Creates errorbar plottable and cookbook

bclehmann:
**New Contributors:**

**Purpose:**
#1466
image

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -1,13 +1,8 @@
-**New Contributors:**
-Please review [CONTRIBUTING.md](https://github.com/swharden/ScottPlot/blob/master/CONTRIBUTING.md)
-
-**Hacktoberfest Participants:**
-Check-out the [Hacktoberfest 2021](https://github.com/ScottPlot/ScottPlot/issues/1274) page
-
 **Purpose:**
 * Use this space to describe what this pull request accomplishes
 * If this PR references an issue be sure to link to it
 * Consider including a picture if it would help clarify what this pull request does
+* New contributors please review CONTRIBUTING.md in the root folder
 
 ```cs
 // You can insert code examples like this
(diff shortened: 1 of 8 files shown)


bclehmann:
For what it's worth, it's been a while since I've made a PR that touched rendering (I think it was #1188...) and I have to say that things like `Coordinate` and `Pixel` seem like a nice common-sense addition. I remember a lot of tuple-wrangling before.

It might even be possible to claw back some performance by converting them to structs since they're small and potentially could be allocated every render.

swharden:
> things like Coordinate and Pixel seem like a nice common-sense addition

Thanks! I made them because I kept confusing myself passing `(double x, double y)` 😝 

> It might even be possible to claw back some performance by converting them to structs since they're small and potentially could be allocated every render

If I'm going to refactor ScottPlot 4, it'll probably be to favor readability and portability over performance. Switching to Maui.Graphics will be a pretty big shake-up, so it may not be worth optimizing ScottPlot 4 at this time.

BTW you might get a kick out of #1561 - It's barebones but the WinForms app is the most developed. I'm experimenting with architecture. After a few years getting to know the pain points of this library, I'm leaning toward minimizing state stored in the plot module itself. Types like `AxisLimits`, `Pixel`, and `Coordinate` are structs with a fluent interface. I don't intend to greatly change the public API (the plot methods will be essentially the same), but I'm starting to like things like this...

```cs
PlotView mouseDownView = plt.GetAxisLimits();
PlotView currentView = mouseDownLimits.WithPan(mouseDownPixel, mouseNowPixel);
Bitmap bmp = plt.Render(currentView);
```

None of this is set in stone (I'm still waiting for Maui.Graphics to mature - it's getting there though), but it's interesting to consider after working with this library and having a good understanding of its roughest edges.

swharden:
Great work as always @bclehmann, thanks so much for this PR! I'll merge this now and release this weekend.

~~This could even be extended in the future~~ by adding a `MarkerShape` and `MarkerSize` to allow people to make plots like this (although I'd probably use the same color for the error lines and the marker)

_EDIT: I'll put this in now real quick, then merge_

image

bclehmann:
Sorry, I found this after you merged, but there's a bug in the error bar rendering code. They can be drawn outside the plotting frame.

image

swharden:
> Sorry, I found this after you merged

No worries! It'll be a few hours before I publish the next package, so this fix will make it in

> I'm confused why this happens, because I don't think the other plottables check for points outside the frame, but they don't seem to suffer from this problem

My trick is to avoid calling `Graphics.FromImage(bmp)` and instead I call an internal method which clips the data area according to the `PlotDimensions` - here's the fix https://github.com/ScottPlot/ScottPlot/commit/6bb43a30e294227ef4954c96abe5be0f8bf44841

PR #1616, opened 2022-01-29
Text: implement IDraggable

BambOoxX:
This PR implements drag functionality for `Text` plottables. 

The current `isUnderMouse` test fails, but that seems internal to GDI. 

Another implementation, not pushed though, works fine using this test 
https://math.stackexchange.com/questions/190111/how-to-check-if-a-point-is-inside-a-rectangle
image

~This PR also contains a patch for `GetCoordinateY` in 650f60e29f3d405331d84a1ac344791ca68fbab7~ _EDIT: moved to #1625/#1626_

### DIFF
--- a/dev/changelog.md
+++ b/dev/changelog.md
@@ -7,6 +7,7 @@ _In development / not yet on NuGet_
 * Marker: Added an optional `Text` (and `TextFont`) for displaying a message that moves with a marker (#1599)
 * Heatmap: Heatmaps with custom X and Y sizing or positioning no longer call `AxisScaleLock()` automatically (#1145) _Thanks @bclehmann_
 * Axis: GetCoordinateY() now returns more accurate coordinate (#1625, #1616) _Thanks @BambOoxX_
+* Text: Now has `IsDraggable` field and improved mouseover detection that supports rotation (#1616) _Thanks @BambOoxX_
 
 ## ScottPlot 4.1.32
 _Published on [NuGet](https://www.nuget.org/packages?q=scottplot) on 2022-01-23_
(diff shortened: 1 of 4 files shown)


swharden:
@BambOoxX thank you so much for identifying and fixing the error `PlotDimensions.GetCoordinateY()`! I implemented your important fix in a separate issue/PR #1625/#1626, added regression tests to ensure it doesn't break in the future, and merged those changes back into this PR.

Note that I'm working toward a soft feature freeze #1624 to let me focus on some deep refactoring of this library (namely #1036), but that I'll do my best to devote good time to this PR and the other existing open issues over the next few days. Thanks again for all your work on this! 👍

swharden:
> The current isUnderMouse test fails, but that seems internal to GDI.

Since ScottPlot 5 will not use GDI or System.Drawing (#1036) I was motivated to figure-out how to implement hit detection on arbitrary rotated rectangles. 

> Another implementation, not pushed though, works fine using this test

It sounds like I may have duplicate some of your work here, but I made a little test application to demonstrate the effect

point-inside-rotated-rectangle

https://github.com/swharden/Csharp-Data-Visualization/blob/203e024253a2545fc325d1f68d2861a1b9fac74d/projects/rotated-rectangle-intersection/Form1.cs#L98-L134

... I'll put this down for the night, but work on this PR more tomorrow 👍

swharden:
After _hours_ of debugging I figured-out the problem was because the coordinate rectangle had a negative height. Swapping the Y values used for the top and bottom of the rectangle fixed the problem, and draggable functionality worked beautifully. 

Thanks again for all your help on this @BambOoxX!

I even got a blog post out of the deal 😎 https://swharden.com/blog/2022-02-01-point-in-rectangle/

BambOoxX:
@swharden Nice work finding the bug ! I saw the triangle surface method, but I think the [scalar projection method](https://math.stackexchange.com/a/190373/486921) is much more elegant. Of course for a single rectangle that should not make much difference, but this way you only compute 3 distances and at most 4 scalar products. I will try to implement this when I get some time.

PR #1626, opened 2022-02-01
PlotDimensions: fix GetCoordinateY(), add tests, and refactor

### LINKED ISSUE #1625: Axis: inaccurate GetCoordinateY() conversion
When using converting to pixel with GetPixelY() then back to coordinate with GetCoordinateY() does not result in the same number

The error is here:
https://github.com/ScottPlot/ScottPlot/blob/d8bce9076ee4b0ac9ea2541561972f7ed435cdbf/src/ScottPlot/PlotDimensions.cs#L43

@BambOoxX pointed-out in #1616 this should be:
```cs
public double GetCoordinateY(float pixel) => YMax - (pixel - DataOffsetY) / PxPerUnitY;
```

swharden:
resolves #1625


swharden:
Previously this was overlooked because the `Plot` module performs pixel/unit conversions like this

https://github.com/ScottPlot/ScottPlot/blob/41d6cbb630bf5eb6e010ece4d68c59748bc53427/src/ScottPlot/Renderable/AxisDimensions.cs#L279-L288

https://github.com/ScottPlot/ScottPlot/blob/41d6cbb630bf5eb6e010ece4d68c59748bc53427/src/ScottPlot/Renderable/AxisDimensions.cs#L290-L297

Perhaps some of this duplicated logic could be removed

PR #1632, opened 2022-02-02
Change : Change word from Secondary to Primary

NeelPatel31:
```cs
//Before
plt.YAxis.Label("Secondary Y Axis");
//After
plt.YAxis.Label("Primary Y Axis");
```


swharden:
Looks great! Thanks @NeelPatel31 🚀

PR #1660, opened 2022-02-13
Plottable: IHasLine, IHasMarker, and IHlightable

BambOoxX:
This PR implements the interfaces discussed in #1655, which has been rebased on this branch (not sure if it was the right thing to do though). 
- Three new interfaces are created `IHasLine`, `IHasMarker` and `IIsHighlightable`. 
- A new marker plotting method `DrawMarkers` is proposed to support customizeable markers in `SignalPlot`s, as well as new implementations for marker plotting avoiding `DrawString` when possible. 
- Two new `MarkerShapes` are implemented : `openTriangle` and `filledTriangle`

Hopefully this will have some use beyond just interactive plotting.

PR #1689, opened 2022-02-19
Marker

BambOoxX:
- Added new maker types from #1660 and factored methods for triangle-shaped markers
- Modified slightly `Asterisk` marker


BambOoxX:
@swharden Actually I have some other things to push here. Do I need to create another PR ?

PR #1690, opened 2022-02-19
Marker markerlinewidth

BambOoxX:
This PR implements a `MarkerLineWidth` property and subsequent modifications to allow to change the stroke width of a marker. 

This is still a WIP.

Commit cea499473b3e44cc15f3f7fd0bf5df7294654f61 breaks the legend for pie chart and maybe other plot types. 
This **might** be the occasion to create a `IHasArea` interface handling all plots that are related (visually at least) to an area / surface with respect to `IHasLine`. This would avoid to use the `isRectangle` property in `Legend`.

Here is how looks the "All Markers" example with the new capabilities and `sp.LineWidth = 2 --> sp.LineWidth = i;`

image

### DIFF
--- a/dev/changelog.md
+++ b/dev/changelog.md
@@ -13,6 +13,7 @@ _In development / not yet on NuGet ..._
 * SignalXY: Fix bug affecting the edge of the plot when step mode is active (#1703, #1699) _Thanks @PeppermintKing_
 * SignalXY: Improve appearance of filled regions when step mode is active (#1703, #1697) _Thanks @PeppermintKing_
 * Axis Span: Added options to customize fill pattern and border (#1692) _Thanks @BambOoxX_
+* Markers: Additional customization options such as `MarkerLineWidth` (#1690) _Thanks @BambOoxX_
 
 ## ScottPlot 4.1.33
 _Published on [NuGet](https://www.nuget.org/packages?q=scottplot) on 2022-02-04_
--- a/src/ScottPlot/Marker/MarkerTools.cs
+++ b/src/ScottPlot/Marker/MarkerTools.cs
@@ -26,21 +26,21 @@ namespace ScottPlot
             marker.Draw(gfx, pixelLocation, radius, brush, pen);
         }
 
-        public static void DrawMarker(Graphics gfx, PointF pixelLocation, MarkerShape shape, float size, Color color)
+        public static void DrawMarker(Graphics gfx, PointF pixelLocation, MarkerShape shape, float size, Color color, float linewidth = 1)
         {
             using Brush brush = new SolidBrush(color);
-            using Pen pen = new(color);
+            using Pen pen = new(color, linewidth);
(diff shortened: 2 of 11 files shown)


BambOoxX:
cea499473b3e44cc15f3f7fd0bf5df7294654f61 breaks pie legends because these actually draw a 10px width line instead of what should be a rectangle of some sort. 

https://github.com/ScottPlot/ScottPlot/blob/9d0b93678f065e63ea91b9f01d44baf6494f6cbe/src/ScottPlot/Plottable/PiePlot.cs#L60-L69

This seems to be hardcoded in `AxisSpan`, `BarPlot`, `CoxcombPlot` ... so an interface (or something equivalent) would really help I believe. 
@swharden What do you think ?

BambOoxX:
@swharden I just pushed a first proposal for a `IHasArea` interface in #1692. Tell me if you think it can be useful 🚀 🔬

swharden:
Hi @BambOoxX, 

I think I'm mostly done reviewing this PR but can you help me understand the intent behind this modifications to scatter plots?

```cs
public double LineWidth
{
    get => IsHighlighted ? _lineWidth * HighlightCoefficient : _lineWidth;
    set { _lineWidth = value; _markerLineWidth = (float)value / 2; _markerSize = (float)value * 2; }
}
```

It seems a little strange to me that changing the `LineWidth` (part of `IHasLine`) will secretly change `MarkerLineWidth` (part of `IHasMarker`). I'm going to remove that part of the setter, but let me know if there's a good reason for it to be there and I may put it back 👍

swharden:
This landed in an awesome spot, thanks so much @BambOoxX!

image

```cs

double[] ys1 = DataGen.Sin(30);
var cmap1 = ScottPlot.Drawing.Colormap.Viridis;

double[] ys2 = DataGen.Cos(30);
var cmap2 = ScottPlot.Drawing.Colormap.Turbo;

for (int i = 0; i < ys1.Length; i++)
{
    double frac = i / (ys1.Length - 1f);

    var circle = plt.AddMarker(i, ys1[i]);
    circle.MarkerShape = MarkerShape.openCircle;
    circle.MarkerSize = i + 5;
    circle.MarkerLineWidth = 1 + i / 2;
    circle.MarkerColor = cmap1.GetColor(1 - frac, .8);

    var triangle = plt.AddMarker(i, ys2[i]);
    triangle.MarkerShape = MarkerShape.openTriangleUp;
    triangle.MarkerSize = i + 5;
    triangle.MarkerLineWidth = 1 + i / 4;
    triangle.MarkerColor = cmap2.GetColor(frac, .8);
}
```

BambOoxX:
The idea behind changing the marker properties when changing the line width is only to preserve consistency. The main point is that if you draw a thick line to make it very visible, ans that it has markers, you probably want to see the markers pretty well too.
 But that feature is really not mandatory

In a more marker-centric context, it could be interesting that setting the marker size sets the marker line width because large markers with thin lines look correct but small markers with thick lines are very distorded

PR #1692, opened 2022-02-20
Add `IHasArea` interface

BambOoxX:
This PR implements a `IHasArea` interface to handle plots that have some sort of `Area` plotted. 
The goal is to implement a common interface for `AxisSpan`, `Pie/Donut` ...
Basically the interface comes in addition to `IHasColor` than only handles a line color or area fill color with 
- `BorderColor`
- `BorderLineWidth`
- `BorderLineStyle`
- `HatchColor`
- `HatchStyle`

Among the benefits, the current way of creating `LegendItem` for such plots is to define their `LineWidth` as 10 resulting in the setting of the `isRectangle` property to `true`, this new interface allows to draw a specific type of legend item for this kind of plottables. 

This allows to *update* some plots by adding new properties, such as in `AxisSpan` that serves as a demonstration case for this new interface.

### **At the moment the other plots that do not implement `IHasArea` have their legend items broken**

Here is a preview of the `AxisSpan` created using the modified example in 16225299c145cd0923b7a4abba9cc79d067bf34a

image

### DIFF
--- a/dev/changelog.md
+++ b/dev/changelog.md
@@ -2,7 +2,7 @@
 
 ## ScottPlot 4.1.34
 _In development / not yet on NuGet ..._
-* Bubble plot: Added methods to get the point nearest the cursor (#1657, #1652) _Thanks @BambOoxX and @adgriff2_
+* Bubble plot: Added methods to get the point nearest the cursor (#1657, #1652, #1705) _Thanks @BambOoxX, @Maoyao233, and @adgriff2_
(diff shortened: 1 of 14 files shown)


swharden:
Looks like this change broke legend construction for some plot types 😅 

image

BambOoxX:
> Looks like this change broke legend construction for some plot types 😅 

Yeah, it was not yet ready for a merge I believe but that's fine ^^.
I think that the next step for legend rendering is to react depending on the interface type of each plottable.

PR #1705, opened 2022-02-22
BubblePlot: fix GetPointNearest calculations

Maoyao233:
**Purpose:**
#1657 introduced an obvious out-of-range error. The index `i` reaches `Bubbles.Count`.

### DIFF
--- a/dev/changelog.md
+++ b/dev/changelog.md
@@ -2,7 +2,7 @@
 
 ## ScottPlot 4.1.34
 _In development / not yet on NuGet ..._
-* Bubble plot: Added methods to get the point nearest the cursor (#1657, #1652) _Thanks @BambOoxX and @adgriff2_
+* Bubble plot: Added methods to get the point nearest the cursor (#1657, #1652, #1705) _Thanks @BambOoxX, @Maoyao233, and @adgriff2_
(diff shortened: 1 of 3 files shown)


swharden:
oof, thanks for pointing this out! I'll review and merge tonight.

This is a good example why new features should always be accompanied by new tests 😅

https://github.com/ScottPlot/ScottPlot/blob/984938fea36cc719b89ea7682e06a700e91dc6b4/src/ScottPlot/Plottable/BubblePlot.cs#L150

https://github.com/ScottPlot/ScottPlot/blob/984938fea36cc719b89ea7682e06a700e91dc6b4/src/ScottPlot/Plottable/BubblePlot.cs#L174

https://github.com/ScottPlot/ScottPlot/blob/984938fea36cc719b89ea7682e06a700e91dc6b4/src/ScottPlot/Plottable/BubblePlot.cs#L204

swharden:
... while adding tests I found these were more broken than I first realized!

https://github.com/ScottPlot/ScottPlot/blob/984938fea36cc719b89ea7682e06a700e91dc6b4/src/ScottPlot/Plottable/BubblePlot.cs#L148-L161

The returned values are always of the last bubble, not the closest bubble 😅 

I'll edit this PR to fix the problem then merge shortly 👍

Issue #1721, opened 2022-03-09
Axis.PixelSnap() has no effect


Xerxes004:
## Bug Report

**Issue:** The method `PixelSnap` in class `Axis` has no effect. It sets the `SnapPx` property in `AxisTicks`, but `SnapPx` is never used anywhere. It looks like very similar functionality was added in #384, but removed at some point down the line. Forgive me if this bug report is malformed, this is my first ever bug report.

**Reproducing:** Create any plot with non-low-quality anti alias settings

### System Details
* **ScottPlot Version:** 4.1.35
* **Operating System:** Windows 10
* **Application Type:** WPF, Avalonia, Winforms
* **.NET Version:** .NET 6.0.201

Xerxes004:
I found a fix for this but I need to make a PR for it. Not sure if my solution is the correct way to do it, but I'll create a PR as soon as I get home from work.

Xerxes004:
Opened draft #1722

swharden:
Hi @Xerxes004, thanks for reporting this and getting started with a PR!

I see it's still in draft, but when you're done working on it let me know and I'll test it and merge it in 👍 🚀

Xerxes004:
Thanks @swharden! I marked the PR as ready.

swharden:
Reminding us all, this is what pixel snapping does: It draws grid lines and tick marks with anti-aliasing off so these lines always "snap" to the nearest pixel and appear as simple solid lines (and not blurry anti-aliased lines)

```cs
formsPlot1.Plot.XAxis.PixelSnap(false);
formsPlot1.Plot.YAxis.PixelSnap(false);
```

pixel snapping off | pixel snapping on (default)
---|---
![image](https://user-images.githubusercontent.com/4165489/157863427-122e839f-3a33-4b06-b0f9-8b1513076275.png)|![image](https://user-images.githubusercontent.com/4165489/157863540-d645e176-9fe8-465b-b2d4-ed81eba4b61b.png)

PS: I hate how Windows 11 screenshots of windows with rounded corners don't support transparency 😝

PR #1722, opened 2022-03-10
Fix pixel snap

Xerxes004:
**Purpose:**
Fixes usage of `AxisTick.SnapPx` so that tick lines and lines appear crisp and clean, as intended.

See #1721


Xerxes004:
Removed a commit which was causing formatting issues via hard reset. This keeps the log down to a single commit.

Issue #1763, opened 2022-04-06
SignalPlotXY: Marker size and shape is ignored


ChrisCC6:
Hi,

I was testing the SignalPlotXY as alternative plot for my app, noticing that the markers are drawn always and independent  of the the size/shape. I don't know whether this is a bug or a feature :-)

I propose changing this part of `SignalPlotXYGeneric`'s `Render()` method:
https://github.com/ScottPlot/ScottPlot/blob/3780dce1a2b2df4897379d090f9e67a026aae3ca/src/ScottPlot4/ScottPlot/Plottable/SignalPlotXYGeneric.cs#L224-L255

to:
```cs
// Draw lines
if (PointsToDraw.Length > 1)
{
    ValidatePoints(PointsToDraw);
    if (LineStyle != LineStyle.None)
        gfx.DrawLines(penHD, PointsToDraw);
}

// Draw markers
if ((PointsToDraw.Length > 1) && (MarkerSize > 0) && MarkerShape != MarkerShape.none )
{
    float dataSpanXPx = PointsToDraw[PointsToDraw.Length - 1].X - PointsToDraw[0].X;
    float markerPxRadius = .3f * dataSpanXPx / PointsToDraw.Length;
    markerPxRadius = Math.Min(markerPxRadius, MarkerSize / 2);
    if (markerPxRadius > .3)
    {
        // skip not visible before and after points
        var PointsWithMarkers = PointsToDraw
                                .Skip(PointBefore.Length)
                                .Take(PointsToDraw.Length - PointBefore.Length - PointAfter.Length);

        // adjust marker offset to improve rendering on Linux and MacOS
        float markerOffsetX = (RuntimeInformation.IsOSPlatform(OSPlatform.Windows)) ? 0 : 1;

        foreach (PointF pt in PointsWithMarkers)
        {
            PointF point = new PointF(pt.X + markerOffsetX, pt.Y);
            MarkerTools.DrawMarker(gfx, point, MarkerShape, MarkerSize, Color, MarkerLineWidth);

            //gfx.FillEllipse(brush,
            //    x: pt.X - markerPxRadius + markerOffsetX,
            //    y: pt.Y - markerPxRadius,
            //    width: markerPxRadius * 2,
            //    height: markerPxRadius * 2);

        }
    }
}
```

Best Regards,
Chris

swharden:
@ChrisCC6 thank you for reporting this and offering the fix!

[...]

PR #1764, opened 2022-04-07
SignalXY respect marker settings

bclehmann:
**Purpose:**
#1763

This also fixes a similar issue where `LineWidth` of zero was ignored, as well as `LineStyle.None`.


bclehmann:
FWIW we can probably apply the same optimization of only drawing visible points to the line as well. Doubt that line drawing is anywhere near as expensive as markers, because the lines are (hopefully) all drawn at once within GDI.

PR #1767, opened 2022-04-07
Improve detached legend

### LINKED ISSUE #1765: Index out of range exception when clicking the items in Detached Legend
Hi,

Now I have two plottables (SignalXY) in FormsPlot.Plot and try to set those invisible by using detached legend window.
image

When I clicked one of those, I can only make it invisible but cannot set it visible state again.
And, even if I clicked again on Legend '3' in invisible state, it makes Legend '4' invisible.
In the state of both Legend invisible, it resulted in index out of range exception when I clicked once more.
Screenshot_2

*Also, even if I set the location of legend as Alignment.LowerCenter, but it returns to the location LowerRight when closing the detached window.

### System Details
* **ScottPlot Version: 4.1.39
* **Operating System: Windows 10
* **Application Type: WinForms
* **.NET Version: NET Framework 4.7.2

swharden:
investigates and resolves #1765


BambOoxX:
I'm happy to see this feature is being used, but sorry if it creates additional workload for you !

Issue #1771, opened 2022-04-08
SignalXY: Only check for ascending X values between min/max render index


bernhardbreuss:
**ScottPlot Version:** 4.1.40

**Operating System:** Win 10

**Application Type:** WPF

**Question:** Describe your question here

Hi, I am implementing a fast refreshing live chart with a SignalPlotXY. As suggested in the FAQ I created a large double array and limited the visible range with `MinRenderIndex` and `MaxRenderIndex`. This works great for my use case. However, if there are already some data to show when the SignalXY is created, a argument exception `Xs must only contain ascending values` is thrown.

This code snippet illustrates the problem:
```cs
double[] xs = new double[1000];
double[] ys = new double[1000];
for (int i = 0; i < 100; i++)
{
    xs[i] = i;
    ys[i] = Math.Sin(i);
}
var signalXY = WpfPlot.Plot.AddSignalXY(xs, ys);
signalXY.MinRenderIndex = 0;
signalXY.MaxRenderIndex = 99;
WpfPlot.Refresh();
```

I totally understand that SignalXY is optimized for ascending X values and for the visible range this constraint should be still fulfilled. As I am not always able to create the chart with an empty or ascending X array I have come up with the following thoughts on this problem:
1. As the ascending X values check is only done on assigning data, there are still possibilities to get non ascending X values (by changing the data after assigning the array). As this could be seen as inconsistent behaviour, maybe the check could be removed.
2. Provide the possibility to define `MinRenderIndex` and `MaxRenderIndex` when creating the signal by `AddSignalXY`. Then the logic in the setter of `SignalPlotXYGeneric.Xs` could be changed to only check the range to render.
3. Allow the user to create a derived class which is able to handle those special situations through bypassing the check. For this `SignalPlotXYGeneric.Xs` should be made virtual.

I am not sure what is the best suitable solution. Or is there even a better solution for this problem?

swharden:
Hi @bernhardbreuss, thank you for sharing this interesting use case!

It probably does make sense for me to limit the initial range check to within the starting min/max render index. I'll leave this issue open to track progress on this change.

In the mean time, perhaps this would get the job done?

```cs
double[] xs = new double[1000];
double[] ys = new double[1000];
for (int i = 0; i < 1000; i++)
{
    if (i < 100) {
        xs[i] = i;
        ys[i] = Math.Sin(i);
    } else {
        xs[i] = double.MaxValue;
        ys[i] = double.MaxValue;
    }
}
```

PR #1777, opened 2022-04-11
SignalPlotXY: defer ascending X check to ValidateData and only validate between min/max render index

bernhardbreuss:
Hi @swharden,

I have tried to solve #1771 by validating X only between min/max render index.
Doing the check only between min/max render index in the setter didn't worked out very well:
- min/max render index must be set before
- before setting max render index Ys must be set
- min/max render index would have been added to `AddSignalXY` which then already started to look like the obsolete `PlotSignalXY`

Therefore I decided to only use the already present check in `ValidateData`.

### DIFF
--- a/src/ScottPlot4/ScottPlot.Tests/PlotTypes/SignalXY.cs
+++ b/src/ScottPlot4/ScottPlot.Tests/PlotTypes/SignalXY.cs
@@ -41,8 +41,8 @@ namespace ScottPlotTests.PlotTypes
             }
 
             var plt = new ScottPlot.Plot(500, 350);
-            var signalXY = plt.AddSignalXY(xs, ys);
-            Assert.Throws<InvalidOperationException>(() => { signalXY.ValidateData(deep: true); });
+            plt.AddSignalXY(xs, ys);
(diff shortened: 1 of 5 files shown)


bernhardbreuss:
Hi @swharden, I am afraid but the information in the provided link seems only to work for forks owned by a personal account. However I have added you as a collaborator. Hopefully this way you can also edit the PR branch.

swharden:
> added you as a collaborator

That worked, thanks! I'll make a few changes then merge this in shortly. I recommend not committing new changes on your end while I'm actively working on it 👍 

Thanks again for starting this PR! This will be a great improvement to the library 🚀

swharden:
I reached a good compromise in my last few commits.

* There's now a private flag variable which indicates whether Xs have been validated or not.

* Assigning Xs resets this flag, and the validation isn't performed until Render() is called.

* This gives the user time to assign data and set min/max indexes before validation occurs.

* This strategy also ensures Xs is always checked at least once before the first render, hopefully catching the common user of attempting to plot data with non-ascending X values.

bernhardbreuss:
Hi @swharden, thank you for your hard work. ScottPlot is a great lib, easy and straight forward to use.

Issue #1779, opened 2022-04-11
Feature: option to draw grid _above_ plottables


swharden:
originally suggested by @KATAMANENI in #1773

PR #1780, opened 2022-04-12
Plot: Allows drawing gridlines above the plottables

bclehmann:
**Purpose:**
#1779 
image

```cs
// You can insert code examples like this
plt.DrawGridAbovePlottables = true;
```

I have one small note about these changes, it doesn't technically draw just the grid lines after the plottables, it draws the entire axes after. This should be the same thing as I don't believe any other part of the axis can be drawn in the data area (or any plottables can be drawn outside the data area). Just a note in case that isn't a safe assumption.


swharden:
This looks great, thanks so much @bclehmann!

To avoid accumulating styling flags in `Plot` I hid the flag in `Plot.Settings` and let users define it like this

```cs
plt.Grid(onTop: true);
```

Issue #1791, opened 2022-04-21
FormsPlot: default right-click menu throws exception for some plot types


ShenxuanLi:
### System Details
**ScottPlot Version**:  4.1.41
**Operating System**:  Windows 10
**Application Type**:  WinForms
**NET Version**:  NET Framework 4.8

**no Crosshair is ok**
![3](https://user-images.githubusercontent.com/92280805/164397881-a616f52f-e8a5-4ca6-a166-4f0408acb0d5.png)

**if add a Crosshair, will crash**
![1](https://user-images.githubusercontent.com/92280805/164398366-e6057139-bfe2-4256-ab98-84fd3bfbd713.png)
![2](https://user-images.githubusercontent.com/92280805/164398183-6ada9b47-4219-4989-9ae5-fbc42ab246bc.png)

ChunChunLiGong:
add a line or a span is ok. but add a crosshair will crash.

ChunChunLiGong:
Thr version 4.1.30 is all right，but 4.1.41 with this bug.

levimann24:
ChunChunLiGong I can confirm that am getting the same crash.

levimann24:
@ChunChunLiGong I can confirm I am seeing the same bug.

MareMare:
I think this is because the `Crosshair.GetLegendItems()` method returns null, but that is not included in the condition.

https://github.com/ScottPlot/ScottPlot/blob/f012da1a9ecb7effc4c83ab2cdcd7a049d0ba675/src/ScottPlot4/ScottPlot/Plottable/Crosshair.cs#L135

the current code:
https://github.com/ScottPlot/ScottPlot/blob/f012da1a9ecb7effc4c83ab2cdcd7a049d0ba675/src/ScottPlot4/ScottPlot.WinForms/FormsPlot.cs#L211-L214

I propose the above code as follows:
```cs
bool legendHasItems = Plot.GetPlottables()
    .Where(x => x.GetLegendItems() != null)  // <--- Add this as a condition
    .SelectMany(x => x.GetLegendItems())
    .Where(x => !string.IsNullOrEmpty(x.label))
    .Any();
```

What do you think?

levimann24:
@MareMare Seems like a good solution!

ChunChunLiGong:
@MareMare good solution！

BigDream324:
I have the same problem and tried version 4.1.39 works.

swharden:
[...]

PR #1794, opened 2022-04-21
FormsPlot: Fix exception when right mouse clicking on a plot with Crosshair.

MareMare:
**Purpose:**
I tried solve #1791.

Fixed a `NullReferenceException` exception raised by right mouse click on a WinForms Plot with Crosshair.

### DIFF
--- a/src/ScottPlot4/ScottPlot.WinForms/FormsPlot.cs
+++ b/src/ScottPlot4/ScottPlot.WinForms/FormsPlot.cs
@@ -209,6 +209,7 @@ namespace ScottPlot
         public void DefaultRightClickEvent(object sender, EventArgs e)
         {
             bool legendHasItems = Plot.GetPlottables()
+                .Where(x => x.GetLegendItems() != null)
                 .SelectMany(x => x.GetLegendItems())
                 .Where(x => !string.IsNullOrEmpty(x.label))
                 .Any();
--- a/src/ScottPlot4/ScottPlot/Plottable/Annotation.cs
+++ b/src/ScottPlot4/ScottPlot/Plottable/Annotation.cs
@@ -41,7 +41,7 @@ namespace ScottPlot.Plottable
         public int YAxisIndex { get; set; } = 0;
 
         public override string ToString() => $"PlottableAnnotation at ({X} px, {Y} px)";
-        public LegendItem[] GetLegendItems() => null;
+        public LegendItem[] GetLegendItems() => Array.Empty<LegendItem>();
         public AxisLimits GetAxisLimits() => new AxisLimits(double.NaN, double.NaN, double.NaN, double.NaN);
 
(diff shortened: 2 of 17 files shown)


StendProg:
An alternative solution would be return a default legend object in `Сrosshair` Plottable with a null or empty `label` field. Then there would be no need to change the existing filter.

MareMare:
@StendProg, Thank you for your reply.

> An alternative solution would be return a default legend object in `Сrosshair` Plottable with a null or empty `label` field. 

So you mean, for example, the following?
```cs
public LegendItem[] GetLegendItems() => Array.Empty<LegendItem>();
```

I thought about that too, but I did the same with the `Legend.UpdateLegendItems` method as it does something similar.

Shall I change to the solution you suggested?

StendProg:
I mean something like this:
```c#
public class Crosshair
{
...
 public LegendItem[] GetLegendItems() => new LegendItem[]{ new LegendItem(this){label = null}};
...
}
```

We only have 1 Plottable which is out of concept and returns `null` instead of a `LegendItems`. Instead of adapting the concept of processing (filters that may be or will appear elsewhere) to a single exception (Crosshair), it may be better to adapt this exception to an existing concept.
I don't know which is better for sure, so the decision is yours, I just pointed out an alternative.

MareMare:
Sorry for the delay.
That's what you meant. Thank you for telling me correctly.

```cs
public interface IPlottable
{
  ...
  /// <summary>
  /// Returns items to show in the legend. Most plottables return a single item. in this array will appear in the legend.
  /// Plottables which never appear in the legend can return null.
  /// </summary>
  LegendItem[] GetLegendItems();
  ...
}
```

As stated in the documentation comments for the `IPlottable.GetLegendItems()` method, I believe that cross hairs are never put out in the legend.

Therefore, I would like to return null for the said method as it is now, and use it through a filter.

swharden:
Thank you @MareMare and @StendProg!

> We only have 1 Plottable which is out of concept and returns null instead of a LegendItems. Instead of adapting the concept of processing (filters that may be or will appear elsewhere) to a single exception (Crosshair), it may be better to adapt this exception to an existing concept.

I will implement this now to bring `Crosshair` into consistency with the rest of the plottables provided with this library

> Then there would be no need to change the existing filter

I am happy the filter was updated to protect against `null` because, even though none of ScottPlot's plottables return `null`, the user may create their own `IPlottable` which does, and this fix protects against that case. 

In ScottPlot 5 I'll use the `nullable` project feature to make it clear when functions are not supposed to return `null`, and it makes sense to me that `GetLegendItems()` should always return an array (even an empty one) to prevent this issue in the future.

swharden:
Quick update, wow, lots of plottables were returning null! 4bcf592

Returning null was even recommended in the XML documentation of `IPlottable` 😅 I fixed that

PR #1806, opened 2022-04-27
SignalXY: Fix bug where extrapolated pixel location could be infinity and crash GDI

### LINKED ISSUE #1803: SignalXY sometimes not rendered on high density data
**Issue:**
SignalXY does sometimes not render on high density data when only partial data should be rendered due to X axis limits.
grafik
When changing the X axis limits just slightly (e.g. by panning) the signal renders fine:
grafik

The interpolation of the `PointBefore` results in a infinite Y value which eventually results in a overflow exception (which is handled by ScottPlot):
```
System.OverflowException: Overflow error.
   at System.Drawing.Graphics.CheckErrorStatus(Int32 status)
   at System.Drawing.Graphics.DrawLines(Pen pen, PointF[] points)
   at ScottPlot.Plottable.SignalPlotXYGeneric`2.Render(PlotDimensions dims, Bitmap bmp, Boolean lowQuality)
   at ScottPlot.Plot.RenderPlottables(Bitmap bmp, Boolean lowQuality, Double scaleFactor)
```

**Reproducing:**
```cs
double[] ys = new double[10000];
double[] xs = ys.Select(x => double.PositiveInfinity).ToArray();
double r = 0.0;
{
    r += 0.005;
    xs[i] = r;
    ys[i] = Math.Sin(r);
}
var signalXY = WpfPlot.Plot.AddSignalXY(xs, ys);
WpfPlot.Plot.AxisAutoY();
WpfPlot.Plot.XAxis.Dims.SetAxis(min: r - 10.0, r);
WpfPlot.Plot.Validate(deep: true);
WpfPlot.Refresh();
```

### System Details
* **ScottPlot Version:** 4.1.41
* **Application Type:** WPF
* **.NET Version:** .NET 6.0

StendProg:
**Purpose:**
Fix for #1803 

SignalXY in some cases, due to rounding errors when converting to `PointF`, the `PointBefore` has the same x coordinate as the first drawn point. As a result, interpolation between these points results in an `Overflow Exception`. Although ScottPlot catches this exception and continues to work, it becomes impossible to display SignalXY.

Since interpolation between these points was introduced for only one reason - GDI does not correctly draw points with strongly negative coordinates. In our case, the error occurs when both points have a coordinate close to 0. Therefore, this fix simply skips interpolation if the x coordinates are equal, because it is definitely not needed in this case.

I decided to keep the test that I used to reproduce the error, but it can be safely removed, because it can only be useful for manual debugging.

### DIFF
--- a/src/ScottPlot4/ScottPlot.Tests/PlotTypes/SignalXY.cs
+++ b/src/ScottPlot4/ScottPlot.Tests/PlotTypes/SignalXY.cs
@@ -1,9 +1,6 @@
 ﻿using NUnit.Framework;
 using System;
-using System.Collections.Generic;
-using System.Drawing;
 using System.Linq;
-using System.Text;
 
 namespace ScottPlotTests.PlotTypes
 {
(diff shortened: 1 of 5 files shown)


swharden:
@StendProg, thank you so much for figuring this out! I **_really_** appreciate your work on this PR.

> ...interpolation between these points results in an Overflow Exception. Although ScottPlot catches this exception and continues to work, it becomes impossible to display SignalXY

I added an option to throw this exception, hopefully making debugging easier in the future.

```cs
plt.GetSettings(false).IgnoreOverflowExceptionsDuringRender = false;
```

Throwing exceptions while drawing lines with large pixel coordinate values seems like a GDI-specific issue, so a lot of this complexity may disappear when we switch to SkiaSharp and/or Maui.Graphics 🤞

PR #1807, opened 2022-04-28
Fix middle-click-drag zooming for Avalonia

kivarsen:
**Purpose**
Middle-click-drag zooming was not working consistently for the Avalonia control. It would only actually zoom in about half the time.

ScottPlot Avalonia zoom bug

**Details**
The difference in behavior between Avalonia and WinForms/WPF/Eto seems to be in the semantics of MiddleWasJustPressed. For the other platforms, MiddleWasJustPressed is essentially identical to IsMiddleDown when evaluated within Backend.MouseMove(). However, in the Avalonia implementation, MiddleWasJustPressed==true only within MouseDown. The result of MiddleWasJustPressed always being false in Backend.MouseMove() was that IsZoomingRectangle would toggle between true and false on each event when the middle button was down, causing the actual zoom to randomly fail 50% of the time when MouseUp() eventually fired.

I tried to streamline the logic a bit and only depend IsXDown within the MouseMove handler. 

**Testing**
I tested the new code in the Avalonia, Eto.Forms, WinForms, and WPF demo apps using both middle-click-drag and alt-left-drag methods of zooming, and everything seems consistent and reliable. I also tested various draggables and they seem to perform as expected.

### DIFF
--- a/src/ScottPlot4/ScottPlot/Control/Backend.cs
+++ b/src/ScottPlot4/ScottPlot/Control/Backend.cs
@@ -139,11 +139,6 @@ namespace ScottPlot.Control
         private bool IsZoomingRectangle;
 
         /// <summary>
-        /// True if a zoom rectangle is being actively drawn using ALT + left click
-        /// </summary>
-        private bool IsZoomingWithAlt;
-
-        /// <summary>
         /// The plot underlying this control.
         /// </summary>
         public Plot Plot { get; private set; }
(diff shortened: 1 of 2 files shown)


swharden:
Thanks for this @kivarsen!

@bclehmann is our resident Avalonia expert, and I'd love to get their quick impression on this PR if possible before I merge it

bclehmann:
I'm actually moving this weekend and we have a launch at work next week, so I don't anticipate I'll be able to review anything soon. If you're ok waiting several days to a week then I can take a look.

swharden:
> don't anticipate I'll be able to review anything soon

No worries! #1807, #1809, and #1810 are all Avalonia related so I thought I'd mention it, but I feel good testing and merging these in myself. Good luck with your move (and your launch)!

kivarsen:
Thanks @swharden and @bclehmann! I really appreciate all the work you've done on this project -- it's very impressive!

PR #1809, opened 2022-04-28
Fix the Avalonia control to display the context menu next to the mouse...

kivarsen:
…pointer rather than along the top/bottom edge of the control.

Supports use of Avalonia's ContextMenu property, and will not show the menu if the mouse had been dragged while the right mouse button was pressed.

Also completes the implementation of the "Custom Right-Click Menu" control demo.

**Purpose:**
Here is the original behavior:
ScottPlot Avalonia context menu

And the new behavior:
ScottPlot context menu fix

**Implementation**
The public ContextMenu.Open(Control) method in Avalonia does not take pointer position into account, so the context menu gets shown with Bottom alignment by default (switching to Top if there is not enough space).

There is a 3-argument override to Open() that includes an isRequestedByPointer argument. If true, it displays the menu at the pointer position. Unfortunately, this override is private. (I have a branch that calls this method using reflection. It works well but feels messy.)

Instead, this commit handles the context menu using Avalonia's built-in ContextMenu property and display handling. The biggest downside of the default handling is that the menu will display after a right-click-drag zoom sequence. To account for this, I made Backend.MouseDownDragged public, and automatically add an event handler to any ContextMenu assigned to AvaPlot that will inhibit display of the context menu if the mouse has been dragged.

The other problem with this scheme is that OnPointerReleased will get handled by a superclass of UserControl and AvaPlot won't get a chance to detect that the right mouse button has been released. This will cause the control to continue to be in a right-drag-zoom state after selecting an item from the context menu. To prevent this, the implementation overrides OnPointerReleased so that we can handle the pointer release first and then pass on to base classes for context menu display handling.

### DIFF
--- a/src/ScottPlot4/ScottPlot.Avalonia/AvaPlot.axaml.cs
+++ b/src/ScottPlot4/ScottPlot.Avalonia/AvaPlot.axaml.cs
@@ -6,6 +6,7 @@ using Avalonia.Threading;
 using Avalonia.VisualTree;
 using System;
 using System.Collections.Generic;
+using System.ComponentModel;
 using System.Threading.Tasks;
 using Ava = Avalonia;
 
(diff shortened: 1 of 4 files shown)


swharden:
Thank you for this PR @kivarsen! I really appreciate the animation demonstrating the issue and fix.

Refactoring toward a more idiomatic Avalonia implementation of the context menu seems like a good direction too. Thanks for updating this (and the demo)

PR #1813, opened 2022-04-29
Axis: improve support for switching between formatting strings and functions

schifazl:
### The problem
I don't know if this is a bug, a wanted behaviour or a corner case which wasn't considered before.

I have a plot which is used to display one or more ScatterPlot series coming from different measures. Each measure has a different duration (from 2 minutes to 45 minutes), so I've added a checkbox to my WPF form in order to switch the X axis from normal to logarithmic. In this way, multiple series are displayed more clearly.

This means that I have to switch from a DateTime format X axis:

`simPlot.Plot.XAxis.TickLabelFormat("HH:mm:ss", true);`

to a logarithmix X axis with custom format:

`static string logTickLabels(double x) => DateTime.FromOADate(Math.Pow(10, x)).ToString("HH:mm:ss");
simPlot.Plot.XAxis.TickLabelFormat(logTickLabels);`

But I noticed that when switching the axis from the DateTime to the logarithmic format, the logTickLabels function won't be called and the label will be wrong. Digging into the library I noticed that since the `TickLabelFormat` is set as `DateTime `, the `TickCollection.Recalculate` method calls the `RecalculatePositionsAutomaticDatetime` instead of the `RecalculatePositionsAutomaticNumeric` method.

### The solution
I fixed this by explicitly calling `TickLabelFormat(null, false)` when calling the `TickLabelFormat(Func<double, string> tickFormatter)` method.

### The workaround
If this behaviour is not a bug, but is wanted and should not be changed, the workaround is to manually call `scottPlot.Plot.XAxis.DateTimeFormat(false)` whenever going from the DateTime label format to the numeric one.

### DIFF
--- a/src/ScottPlot4/ScottPlot.Tests/Axis/TickLabel.cs
+++ b/src/ScottPlot4/ScottPlot.Tests/Axis/TickLabel.cs
@@ -73,5 +73,34 @@ namespace ScottPlotTests.Axis
 
             TestTools.SaveFig(plt);
         }
+
+        [Test]
+        public void Test_Format_CanBeReversed()
+        {
+            // Demonstrates bug described in https://github.com/ScottPlot/ScottPlot/pull/1813
+
+            var plt = new ScottPlot.Plot(800, 600);
+            var sig = plt.AddSignal(ScottPlot.DataGen.Sin(1000));
+
+            // repeat this multiple times to ensure changes can be undone/redone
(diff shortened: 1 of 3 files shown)


swharden:
Thanks for this fix @schifazl!

This definitely seems like a corner case that wasn't considered before.

I'll make a few additional changes over the next few minutes then merge it in and release a new package on NuGet shortly 👍

PR #1814, opened 2022-04-29
Ticks: support combined automatic and manual ticks

Xerxes004:
**Purpose:**
* Allow the user to add manual tick marks *alongside* automatic ones
* Clean up automatic ticks by removing the one closest to each manual tick

```cs
double[] myCustomTicks = { Math.PI, 2 * Math.PI };
string[] myCustomLabels = { "π", "2π" };
// Supply additional "union" flag to keep the automatic labels
plt.XAxis.ManualTickPositions(myCustomTicks, myCustomLabels, union: true);
```
union labels example


Xerxes004:
Before closing:

- [ ] Fix bug with reverse label logic via `TickLabelNotation`

swharden:
Thanks for this PR @Xerxes004! I'm going to edit it a bit to make these extra ticks manual supplements to the automatic tick system rather than having a third mode for union ticks. Overall this is a great idea, and I like your cookbook demo. Thanks!

PR #1826, opened 2022-05-02
Fix inverted sign handling for additional axis ticks

Xerxes004:
**Purpose:**
* Fixes a bug where manual tick labels were not reversed with the other labels

**Example:**
```c#
double[] positions = { Math.PI, 2 * Math.PI };
string[] labels = { "π", "2π" };
plt.XAxis.AutomaticTickPositions(positions, labels);
plt.XAxis.TickDensity(0.5);
plt.XAxis.TickLabelNotation(invertSign: true);
```

**Before:**
before

**After:**
after

### DIFF
--- a/src/ScottPlot4/ScottPlot/Ticks/TickCollection.cs
+++ b/src/ScottPlot4/ScottPlot/Ticks/TickCollection.cs
@@ -563,10 +563,12 @@ namespace ScottPlot.Ticks
 
             if (additionalTickPositions is not null)
             {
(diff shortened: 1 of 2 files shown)


swharden:
Good catch! I forgot about https://github.com/ScottPlot/ScottPlot/pull/1814#issuecomment-1113866989 😅

Xerxes004:
It was an easy edge case to miss, I barely caught it

PR #1830, opened 2022-05-03
Improve cookbook website generator

### LINKED ISSUE #1827: Update cookbook website generator to support Hugo
ScottPlot.NET ([source code](https://github.com/ScottPlot/ScottPlot.NET)) now uses [Hugo](https://GoHugo.io) to generate the website from Markdown files, but the cookbook generator does not currently output files/menus that can be used by this new system. This issue will track progress toward updating the cookbook page generator.

While I'm working on it I hope to improve the navigability of the cookbook. No URLs should break, but at a minimum the home page should be much less overwhelming. https://scottplot.net/cookbook/4.1/

swharden:
resolves #1827


swharden:
There is much room to simplify the category organization code for ScottPlot 5, but for now this looks better and I'm pushing it into production https://scottplot.net/cookbook/4.1/

Issue #1831, opened 2022-05-04
Plottables: getters and setters are both required for PropertyGrid


bradmartin333:
**ScottPlot Version:** 4.1.43

**Operating System:** Win10

**Application Type:** WinForms

**Question:** Users keep asking for more plot customization fields to the point where I am wrapping almost everything into a custom PropertyGrid. It is a lot easier to allow the user to select a Plottable to be assigned as the active PropertyGrid object and let them have access to everything they want and more. This works well, but some params are not exposed, [such as text rotation and background color](https://github.com/ScottPlot/ScottPlot/blob/4eb9e57ee44c428c9457bdb8176be5590f168b63/src/ScottPlot4/ScottPlot/Plottable/Text.cs#L20-L38). Any underlying reason for this?

Here is an example PropertyGrid for a Text plottable:
![image](https://user-images.githubusercontent.com/19335151/166801783-df5f43a6-fef8-478e-93af-bb79896f2a06.png)

swharden:
Hi @bradmartin333, thanks for pointing this out! There are two issues here.

## Fields vs. Properties

Many plottables use fields instead of auto-properties. Here's a random example:
https://github.com/ScottPlot/ScottPlot/blob/4eb9e57ee44c428c9457bdb8176be5590f168b63/src/ScottPlot4/ScottPlot/Plottable/RadialGaugePlot.cs#L12-L121

There is no good reason for this, and it could be improved for [all plottables](https://github.com/ScottPlot/ScottPlot/tree/4eb9e57ee44c428c9457bdb8176be5590f168b63/src/ScottPlot4/ScottPlot/Plottable).

## Properties without both Get and Set & Nested Configuration Classes

Some plottables have properties with a getter only. The example you noted is here:
https://github.com/ScottPlot/ScottPlot/blob/4eb9e57ee44c428c9457bdb8176be5590f168b63/src/ScottPlot4/ScottPlot/Plottable/Text.cs#L27-L29

These typically exist in cases where configuration used to be a top-level field or property, but was later moved into a configuration object. A good example is `Font` settings. If I add a new advanced `Font` configuration option (hinting, alignment, ClearType vs. TrueType, etc.) I don't want to go around and update code for dozens of plottables to use it. In these cases code like that above was typically added to prevent breaking existing code when the new `Font` object was added.

1. Those properties should be made private and/or marked `obsolete`
2. Those properties could be improved by ensuring all have both getters and setters
[...]

PR #1842, opened 2022-05-10
FormsPlot: Plot Object Editor

swharden:
This PR adds a plot object editor to the default right-click menu. It leans on the WinForms [PropertyGrid](https://docs.microsoft.com/en-us/dotnet/api/system.windows.forms.propertygrid) to provide a simple GUI for users to edit all public properties of plottables on the plot.

There is some danger here, as it may be possible to configure plottables to enter an unrenderable state.

On the other hand, if these properties were made safer, this may be a useful strategy to allow users to investigate what properties are available for customization.

This idea extends suggestions from @bradmartin333 and @BambOoxX in #1831

grid


BambOoxX:
Very nice! What kind of properties can make the plottable non renderable? I'm pretty sure it's easy to break things but I can't find one at the moment...

bradmartin333:
Nice! Setting a min window size for the Plot Object Editor form and using a table layout panel instead of custom anchoring could help for changes down the road. I was also unable to break it 👍

swharden:
> What kind of properties can make the plottable non renderable? 

> I was also unable to break it

I set MinRenderIndex to 50 and MaxRenderIndex to 40 and it caused a hard crash 😈

Plottable objects were originally designed with the intent that developers (not end users) would directly modify the public properties, so safeguards were not placed to protect against highly irrational inputs. For this reason I don't think I'll expose the property editor by default, but I'll definitely include it in the next release behind a feature flag

> table layout panel instead of custom anchoring

Good idea! I'll do this an merge it in today

swharden:
This opt-in feature can be enabled with:

```cs
formsPlot1.Configuration.EnablePlotObjectEditor = true;
```

BambOoxX:
@swharden Would there be an easy way to filter properties? For instance the ones related to appearance (color, visibility... ) which should be unbreakable (or not?) and the ones related to data (renderindex, axis index... ) which are likely to break things?

I also belive that having the possibility to sort things (per plot type, per axis index, per label) would be of huge help when dealing with multiple objects

bradmartin333:
`System.ComponentModel` allows for categories, descriptions and much more

```cs
private int _ColorRounding = 2;
[Category("User Parameters")]
[Description("The higher the number, the more colors are shown in the Z axis and the longer it will take to create the plot. Default = 2, Min = 0, Max = 10")]
public int ColorRounding
{
    get => _ColorRounding;
    set
    {
        if (value < 0)
            _ColorRounding = 0;
        else if (value > 10)
            _ColorRounding = 10;
        else
            _ColorRounding = value;
        MakePlots();
    }
}
```

The viewable attributes can be selected like this

```cs
PropertyGrid.BrowsableAttributes = new AttributeCollection(new CategoryAttribute("User Parameters"));
```

BambOoxX:
Thanks for the tip @bradmartin333!

swharden:
> categories, descriptions and much more

This is true! I disabled the description box because I don't currently utilize these attributes, but if added to properties they would appear in the property grid

* [CategoryAttribute](https://docs.microsoft.com/en-us/dotnet/api/system.componentmodel.categoryattribute?view=net-6.0&viewFallbackFrom=windowsdesktop-6.0)
* [DescriptionAttribute](https://docs.microsoft.com/en-us/dotnet/api/system.componentmodel.descriptionattribute?view=net-6.0&viewFallbackFrom=windowsdesktop-6.0)

At first my thought was to omit these because I assumed they were in the `System.Windows.Forms`, but looking closer I now see they are in the [`System.ComponentModel` Namespace](https://docs.microsoft.com/en-us/dotnet/api/system.componentmodel?view=net-6.0) so it would be possible to sprinkle these in the core ScottPlot library.

Presently most properties have XML documentation to indicate what they do and how to use them, and I think this is most useful as it appears in the IDE when interacting with properties. Coping that text into a bunch of attributes would be a lot of duplication, and contents of those attributes don't seem to be available as hints in code editors.

swharden:
> legend item in the plot object list 

Good idea! I started #1846 to remind me to make that change. If you want to create a PR to implement this you're welcome to, otherwise I'll leave it open and get to it in a few days 👍

BublilDaniel:
Hi
any way to add this into WpfPlot as well?

PR #1845, opened 2022-05-16
IHittable

### LINKED ISSUE #1844: Plottable: IHittable
It seems like there should be an `IClickable` though which would be similar to [`IDraggable`](https://github.com/ScottPlot/ScottPlot/blob/main/src/ScottPlot4/ScottPlot/Plottable/IDraggable.cs) and have an `EventHandler` you could subscribe to that would fire when the object is clicked. 

This idea originated from #1843

swharden:
This PR creates an `IHittable` interface that plottable objects can implement so when they are clicked the control's `OnLeftClickedPlottable` event is fired. resolves #1844

### DIFF
--- a/src/ScottPlot4/ScottPlot.Avalonia/AvaPlot.axaml.cs
+++ b/src/ScottPlot4/ScottPlot.Avalonia/AvaPlot.axaml.cs
@@ -44,6 +44,11 @@ namespace ScottPlot.Avalonia
         public event EventHandler LeftClicked;
 
         /// <summary>
+        /// This event is invoked when a <seealso cref="Plottable.IHittable"/> plottable is left-clicked.
+        /// </summary>
+        public event EventHandler LeftClickedPlottable;
+
+        /// <summary>
         /// This event is invoked after the mouse moves while dragging a draggable plottable.
         /// The object passed is the plottable being dragged.
         /// </summary>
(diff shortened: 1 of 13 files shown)


StendProg:
Hi @swharden,

I looked at your work and want to leave a few comments.
1. Do we really need to support a unique cursor for each Plottable on the MouseOver event? It's hard for me to imagine such scenarios. Everything would be greatly simplified if this cursor were set the same in the `Control` itself.
2. `IHittable` may contain the ability to disable clicks by analogy with `IDraggable`.
```c#
public interface IHittable
{
  bool ClickEnabled {get; set};
  //or
  bool HitTestEnabled{get;set;}
...
}
```
3. You will not need to save the state in the plottable if you also transfer the current `Plot.Dims` to `HitTest()`, which are already saved. The hit calculation will be done every time HitTest() is called and it needs to be fast to be able to bind it to OnMouseMove events.
```c#
public bool HitTest(Coords c, PlotDimensions dims);
```

swharden:
Hi @StendProg, thank you for your feedback! I really value your input. I found this problem to be surprisingly difficult, and it I put it down for a few days and am now coming back to it.

## 1. Do plottables require unique cursors?

I think so. Here are some potential use cases.

Plottable | Cursor
---|---
draggable vertical line | image
draggable horizontal line | image
draggable crosshair center | image
selectable point | image
draggable point | image

## 2. HitTestEnabled

```cs
public interface IHittable
{
  bool HitTestEnabled { get; set; }
...
}
```

Great suggestion! I will implement this.

## 3&4. Instead of storing state, can you pass `dims` into `HitTest()`?

> The hit calculation will be done every time HitTest() is called and it needs to be fast to be able to bind it to OnMouseMove events.

Unfortunately the size of bounding box requires **_string measurement_**, and that is costly enough that I don't think it is worth performing on every mouse move event. Not finding a better solution this problem is why I put this PR down and haven't picked it up in two weeks 😅 It is much more difficult than I initially thought it would be. I can't think of a better way to achieve this, but after I merge this I welcome additional thoughts you have or suggestions for improvements here or in #1844.

PR #1847, opened 2022-05-16
Image: Add axis dependent sizing and bugfixes

wtywtykk:
**Purpose:**
* Add axis dependent sizing. This can be used to draw a background.
* The alignment was ignored with non-zero rotation. I'm not sure if that's on purpose, but seems strange. So I fixed it.
* The border was not scaled with the image. Fixed in this PR.
* Also added tests for scale and sizing.
* Fixed some typos.
Drawing1

### DIFF
--- a/src/ScottPlot4/ScottPlot/Plottable/Image.cs
+++ b/src/ScottPlot4/ScottPlot/Plottable/Image.cs
@@ -14,18 +14,28 @@ namespace ScottPlot.Plottable
         public bool IsVisible { get; set; } = true;
 
         /// <summary>
-        /// Position of the primary corner (based on Alginment)
+        /// Position of the primary corner (based on Alignment)
         /// </summary>
         public double X { get; set; }
 
         /// <summary>
-        /// Position of the primary corner (based on Alginment)
+        /// Position of the primary corner (based on Alignment)
         /// </summary>
         public double Y { get; set; }
 
         /// <summary>
+        /// Width of the image, in axis units. If null, the image will use pixel units.
+        /// </summary>
+        public double? WidthInAxisUnits { get; set; } = null;
+
+        /// <summary>
+        /// Height of the image, in axis units. If null, the image will use pixel units.
+        /// </summary>
+        public double? HeightInAxisUnits { get; set; } = null;
+
+        /// <summary>
         /// Multiply the size of the image (in pixel units) by this scale factor.
-        /// The primary corner (based on Alginment) will remain anchored.
+        /// The primary corner (based on Alignment) will remain anchored.
         /// </summary>
         public float Scale = 1.0f;
 
@@ -51,9 +61,26 @@ namespace ScottPlot.Plottable
         public int XAxisIndex { get; set; } = 0;
         public int YAxisIndex { get; set; } = 0;
 
-        public override string ToString() => $"PlottableImage Size(\"{Bitmap.Size}\") at ({X}, {Y})";
-        public AxisLimits GetAxisLimits() => new(X, X, Y, Y);
+        public AxisLimits GetAxisLimits() => new(X, X + WidthInAxisUnits ?? 0, Y, Y + HeightInAxisUnits ?? 0);
         public LegendItem[] GetLegendItems() => Array.Empty<LegendItem>();
+
+        public override string ToString()
+        {
+            if (WidthInAxisUnits is double axisWidth)
+            {
+                if (HeightInAxisUnits is double axisHeight)
+                    return $"PlottableImage Axis Size({axisWidth}, {axisHeight}) at ({X}, {Y})";
+                else
+                    return $"PlottableImage Axis Width {axisWidth}, Pixel Height {Bitmap.Width} at ({X}, {Y})";
+            }
+            else
+            {
+                if (HeightInAxisUnits is double axisHeight)
+                    return $"PlottableImage Pixel Width {Bitmap.Width}, Axis Height {axisHeight} at ({X}, {Y})";
+                else
+                    return $"PlottableImage Size(\"{Bitmap.Size}\") at ({X}, {Y})";
+            }
+        }
 
         public void ValidateData(bool deep = false)
         {
@@ -63,6 +90,17 @@ namespace ScottPlot.Plottable
             if (double.IsNaN(Y) || double.IsInfinity(Y))
                 throw new InvalidOperationException("y must be a real value");
 
+            if (WidthInAxisUnits is double axisWidth)
+                if (double.IsNaN(axisWidth) || double.IsInfinity(axisWidth))
+                    throw new InvalidOperationException("width must be a real value");
+
+            if (HeightInAxisUnits is double axisHeight)
+                if (double.IsNaN(axisHeight) || double.IsInfinity(axisHeight))
+                    throw new InvalidOperationException("height must be a real value");
+
+            if (double.IsNaN(Scale) || double.IsInfinity(Scale))
+                throw new InvalidOperationException("scale must be a real value");
+
             if (double.IsNaN(Rotation) || double.IsInfinity(Rotation))
                 throw new InvalidOperationException("rotation must be a real value");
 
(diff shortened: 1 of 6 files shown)


src/ScottPlot4/ScottPlot/Plottable/Image.cs
```diff
@@ -23,6 +23,16 @@ public class Image : IPlottable
         /// </summary>
         public double Y { get; set; }
 
+        /// <summary>
+        /// Width of the image, (0 means no scaling)
+        /// </summary>
+        public double Width { get; set; } = 0;
```
bclehmann, line 29: I don't know if I like this, for a few reasons:
- 0 is a weird sentinel value, to me width of zero means width of zero, not the width of the underlying bitmap
    - Could change it to `double?` and null would be your sentinel value. A width of null removes ambiguity, it's not a number so it cannot be a normal width.
- As a user, I expect setting `img.Width` to change the image width, but instead, the image starts growing/shrinking as I zoom in or out
    - This could be fixed by a rename, for example `img.WidthInAxisUnits`
- As a user, I (might) expect reading `img.Width` to return the size in pixels
    - This could also be fixed by a rename

I don't particularly like the name `WidthInAxisUnits`, so that part may still need some work. It's still not intuitive that nulling out that property would cause the width to fallback to the width of the underlying bitmap, so maybe there's some merit to something like this:

```cs
public enum LengthUnit
{
    Pixel,
    Axis
}
```

Then you could keep the name as `Width`, and depending on the selected unit it would mean either width in pixels or in the axis space. This seems clunky to me, and I actually think it's less clear if a user doesn't see the `LengthUnit` property. Maybe there's an option in the middle?

bclehmann: There is also an option for the user to have the width be axis dependent and the height be independent. I assume there's no use-case for this, so my question is should we prevent it?

If we don't, there is an opportunity for us to introduce bugs where we check only that the width is axis dependent, and then assume that the height is as well. For example:

```cs
if (Width is null) {
    width = CalculateWidth();
    height = CalculateHeight();
}
```

When in reality it should be:

```cs
if (Width is null) {
    width = CalculateWidth();
}

if (Height is null) {
    height = CalculateHeight();
}
```

We could prevent this by checking for it and throwing, but there's actually a rather elegant alternative that doesn't rely on run-time checks. If we put the width and height into a struct we can require them be either both null, or both non-null:

```cs
public struct Dimensions
{
    public Width { get; set; }
    public Height { get; set; }
}

public class Image : IPlottable
{
   .
   .
   .
   public Dimensions? DimensionsInAxisSpace { get; set; }
}
```

Here the user is not even capable of nulling width without nulling out the whole struct. This prevents one axis from being axis dependent while the other is not.

bclehmann:
Looks nice, just had a few comments on the API to make it a little more explicit and clear.

src/ScottPlot4/ScottPlot/Plottable/Image.cs
```diff
@@ -51,10 +56,32 @@ public class Image : IPlottable
...
+            }
+            else
+            {
+                return $"PlottableImage Size(\"{Bitmap.Size}\") at ({X}, {Y})";
+            }
+        }
+
+        public AxisLimits GetAxisLimits()
+        {
+            if (SizeInAxisUnits is SizeF axisSize)
```
bclehmann, line 75: I didn't know that the `is` operator to did a null check when the right-hand operand isn't `null`, very cool.

My first guess was that this worked because this was casting from `SizeF?` to `SizeF`, but I checked and it works with reference types as well (with nullable reference types disabled), meaning it's not a cast.

src/ScottPlot4/ScottPlot/Plottable/Image.cs
```diff
@@ -63,45 +90,71 @@ public void ValidateData(bool deep = false)
             if (double.IsNaN(Y) || double.IsInfinity(Y))
                 throw new InvalidOperationException("y must be a real value");
 
+            if (SizeInAxisUnits is SizeF axisSize)
+            {
+                if (double.IsNaN(axisSize.Width) || double.IsInfinity(axisSize.Width))
```
bclehmann, line 95: `double.IsNaN(foo) || double.IsInfinity(foo)` can be replaced by `!double.IsFinite(foo)`

src/ScottPlot4/ScottPlot/Plottable/Image.cs
```diff
@@ -14,18 +14,23 @@ public class Image : IPlottable
...
         /// <summary>
-        /// Position of the primary corner (based on Alginment)
+        /// Position of the primary corner (based on Alignment)
         /// </summary>
         public double Y { get; set; }
 
+        /// <summary>
+        /// Size of the image, in axis units. If null, the image will use pixel units.
+        /// </summary>
+        public SizeF? SizeInAxisUnits { get; set; } = null;
```
bclehmann, line 29: Nice, I hadn't remembered that System.Drawing had this struct when I made my comment.

Although this may be a downside when we transition off of System.Drawing (see #1036 if you're interested). There are already a lot of places where we expose System.Drawing structs directly to the user, so unless @swharden disagrees, I would leave this using `SizeF` for simplicity's sake. Just thought I would note it.

bclehmann (approved):
Looks good, nice work

src/ScottPlot4/ScottPlot/Plottable/Image.cs
bclehmann: I am recalling that we cannot use `double.IsFinite` on .NET Framework, which ScottPlot supports.

src/ScottPlot4/ScottPlot/Plottable/Image.cs
wtywtykk: There is still a problem, the user can't use img.SizeInAxisUnits.Width to set the width, they must set the whole img.SizeInAxisUnits. I don't like that, and that's why I converted this PR to draft. I'm thinking about adding a class for size, but it seems to be too heavy.

bclehmann: Ah, that is annoying. And more importantly I think that it would be confusing to most users, who wouldn't know nor care that `SizeF` is a struct. Perhaps a class is the way to go.

src/ScottPlot4/ScottPlot/Plottable/Image.cs
swharden: > we cannot use `double.IsFinite` on .NET Framework, which ScottPlot supports

Hopefully ScottPlot won't have to support .NET Framework forever 😜

src/ScottPlot4/ScottPlot/Plottable/Image.cs
swharden: > using System.Drawing structs ... may be a downside when we transition off of System.Drawing

I'm fine with this for now - it's littered throughout the existing code base and it will be straightforward for me to make my replacements all at once if/when these plottables are translated to a new rendering system 👍

swharden:
Thanks so much for this PR @wtywtykk! Sorry it took me so long to get to this one. I really like the direction this is going.

Also thanks @bclehmann for your review so far

## Nullable System.Drawing Structs or Custom Objects?

> There is still a problem, the user can't use img.SizeInAxisUnits.Width to set the width, they must set the whole img.SizeInAxisUnits. I don't like that, and that's why I converted this PR to draft. I'm thinking about adding a class for size, but it seems to be too heavy.

Good timing! I started creating my own structs to represent these primitive measurements, so it's a great time to add one more if it would help simplify this PR. The primary reason I am adding these is so I can wean myself off `System.Drawing` primitives like `Point` and `Rect`, but a secondary reason is to leverage the type system to distinguish between _coordinates_ and _pixels_ by using separate objects to represent each. I started creating a few of these, but perhaps you could create more as needed to finish this PR?

* Coordinate units (on the axis)
  * `Coordinate`
  * `CoordinateRect`
  * `CoordinateSize`
* Pixel units (on the screen)
  * `Pixel`
  * `PixelRect`
  * `PixelSize`
* Fractional units that (always range 0-1) to represent things like visible area
  * `FractionalPoint`
  * `FractionalRect`
  * `FractionalSize`

I made-up that last one just now based on the next comment:

## Background

> Add axis dependent sizing. This can be used to draw a background.

I like! Maybe we can expand this with a new plot type. Outside the scope of this PR, but an idea worth considering #1867

## Fixed Alignment

> The alignment was ignored with non-zero rotation. I'm not sure if that's on purpose, but seems strange. So I fixed it.

Thank you! 🙌 It was not on purpose, but it was on the low-priority fix list for a long time. I appreciate it!

## Scaled Border

> The border was not scaled with the image. Fixed in this PR.

Can we put this behind a flag like `bool AutomaticallyResizeBorder` and make it an opt-in feature? I'm thinking of users making plots in headless console/server applications who may be confused why their borer isn't the pixel size they intended.

## Next Steps

Thanks again for your work on this! I won't make any edits to this PR until you indicate you're done with it. When you are, move it from _draft_ to _ready to review_ and let me know so I can pick it up, make some small edits, then merge it 🚀

src/ScottPlot4/ScottPlot/Plottable/Image.cs
wtywtykk: I finally chosen separate width and height. Because both nullable struct and class requires the user to create new object first, this is not stright forward.

wtywtykk:
Thanks for your suggestions!

> Nullable System.Drawing Structs or Custom Objects

Those structures seems be in ScottPlot5, so I guess I can't use them in this PR? But I'll try to write the new background plot types with them.

> Can we put this behind a flag like bool AutomaticallyResizeBorder and make it an opt-in feature? I'm thinking of users making plots in headless console/server applications who may be confused why their borer isn't the pixel size they intended.

I didn't change the "line width" of the border. You can see what I've changed in the last row of the picture.

swharden:
Hey @wtywtykk, thanks again for this PR! I made a few small adjustments and will merge it in shortly.

Regarding System.Drawing primitives and custom objects, I recognize the topic isn't super clear, and it isn't specific to this plot type so I'll deal with it in a different PR and not invite that complexity here. 

Regarding scaled border and "line width" topic, I confused myself because of inconsistent naming that pre-dated this PR. I'm now realizing that some plottables use `BorderSize` and others use `BorderLineWidth` which is what threw me off. Sorry for the confusion!

Issue #1869, opened 2022-06-03
Bracket: sub-ideal angles for non-square coordinate space


swharden:
I added a test to demonstrate this. I would expect the tips of the edges to be evenly spaced around a circle centered at 0. Nitpicky, but worth mentioning. Out of scope for #1863 which is about to get merged as it is now

```cs
ScottPlot.Plot plt = new();

for (int i = 0; i < 10; i++)
{
    double t = i / 10.0 * 2 * Math.PI;
    double x = Math.Sin(t);
    double y = Math.Cos(t);
    var b = plt.AddBracket(0, 0, x, y, i.ToString());
    b.EdgeLength = 20;
    b.LineWidth = 2;
}
```

![image](https://user-images.githubusercontent.com/4165489/171781268-c736f26e-f1f8-4162-9098-5b3e450cafc3.png)

![angles](https://user-images.githubusercontent.com/4165489/171781381-3eddd8be-35e6-40b4-bec9-8b86db8b0359.gif)

PR #1870, opened 2022-06-03
Bracket: Use pixels for calculating normal vector

bclehmann:
**Purpose:**
#1869

The same snippet as in #1869:

image


swharden:
That was simpler than I though it would be - thanks for figuring this one out! 🚀

PR #1910, opened 2022-06-26
Reduce resources used to draw multiple markers

AbeniMatteo:
`.DrawMarkers()` was allocating a `SolidBrush` and a `Pen` without using them, this bug is fixed and now the `IMarker` is also reused.

Plots affected:
- Scatter
- Signal
- Error bar

Repro:
```cs
var rnd = new Random(Seed: 123);
var ys = Enumerable.Range(0, 500).Select(_ => rnd.NextDouble()).ToArray();
var xs = Enumerable.Range(0, 500).Select(_ => rnd.NextDouble()).ToArray();

var ctrl = new ScottPlot.FormsPlot { Dock = DockStyle.Fill };
var sPlot = ctrl.Plot.AddScatter(xs, ys);

for (int i = 0; i < 1; i++)
    ctrl.Refresh();
```

|Points|Iters|Type|Alloc. before|Alloc. after|
|-|-|-|-|-|
|500|1|ScottPlot.MarkerShapes.FilledCircle|500|1|
|500|1|System.Drawing.Pen|525|25|
|500|1|System.Drawing.SolidBrush|509|9|
|500|100|ScottPlot.MarkerShapes.FilledCircle|51.500|103|
|500|100|System.Drawing.Pen|52.953|1.453|
|500|100|System.Drawing.SolidBrush|52.019|519|

### DIFF
--- a/src/ScottPlot4/ScottPlot/Marker/MarkerTools.cs
+++ b/src/ScottPlot4/ScottPlot/Marker/MarkerTools.cs
@@ -12,6 +12,15 @@ namespace ScottPlot
             if (size == 0 || shape == MarkerShape.none)
                 return;
 
+            IMarker marker = Marker.Create(shape);
+            DrawMarker(gfx, pixelLocation, marker, size, brush, pen);
+        }
+
+        public static void DrawMarker(Graphics gfx, PointF pixelLocation, IMarker marker, float size, Brush brush, Pen pen)
+        {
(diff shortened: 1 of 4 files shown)


swharden:
Thanks @AbeniMatteo, 

When it comes to GDI+ we can use all the help we can get minimizing allocations, and this is a big improvement!

Issue #978, opened 2021-04-14
Use ObjectPool for GDI+ Objects


bclehmann:
**What can we do to make ScottPlot better?**
We could potentially improve performance with [`ObjectPool<T>`](https://docs.microsoft.com/en-us/dotnet/api/microsoft.extensions.objectpool.objectpool-1?view=dotnet-plat-ext-5.0) by reducing allocations and collections. This does mildly complicate things in two ways:
- We cannot assume that any object has any of its members set to its default because it comes from the pool, it was not newly constructed
    - One can create a new `PooledObjectPolicy<T>` class to run something when `Return()` is called, which could reset the state to sensible defaults instead of having to do it manually everywhere
- We need to manually call `ObjectPool.Return()` to give the object back
    - We could do this automatically by creating a wrapper class that implements `IDisposable` and then call `Return()` in the dispose method. This allows using `using` blocks

There is a `LeakTrackingObjectPool<T>` class which I assume returns objects to which there are no references back to the pool, but it is undocumented.

Lastly, all this referenced an API in the .NET Platform Extensions, i.e. it's in a NuGet package. I know you're not keen on adding dependencies but this does kind of blur the line between being part of the standard library. In any event, it's not difficult to create your own `ObjectPool` class, there's even an article on it in the docs: https://docs.microsoft.com/en-us/dotnet/standard/collections/thread-safe/how-to-create-an-object-pool

The [ScottPlot Roadmap](https://github.com/swharden/ScottPlot/blob/master/dev/roadmap.md) may provide additional context regarding current goals and future directions.

bclehmann:
```

**Comment by bclehmann** [comment]
With N = 20:

```
Includes startup time because ¯\_(ツ)_/¯

Before:
12.137002229690552
11.26367974281311
10.335201263427734
10.450330018997192
11.072436094284058
10.582040548324585
10.643038511276245
11.391994953155518
10.816736698150635
10.530665636062622
11.259594440460205
10.657943725585938
10.763400554656982
11.562257528305054
10.662458419799805
10.695979833602905
11.650984287261963
11.00367021560669
10.795217752456665
11.226184606552124

After:

10.620888710021973
11.259387254714966
10.504619121551514
10.339058637619019
10.833833932876587
10.36064076423645
10.415626525878906
11.372222185134888
10.455325365066528
10.344213962554932
11.021204471588135
10.396629095077515
10.364850282669067
11.316555500030518
10.422507286071777
10.335057258605957
[...]

PR #1915, opened 2022-06-27
Reuse FontFamily instances

AbeniMatteo:
Big reduction in allocations by caching FontFamily instances.

Things you should consider:
1. Shouldn't be a problem but fonts installed are cached and not updated until the execution is restarted (a method can be implemented to request an update at runtime)
2. Some API have changed, check if you're ok with it (mostly `string name` -> `FontFamily family`)

Repro:
```cs
var ctrl = new ScottPlot.FormsPlot();
ctrl.Plot.AddSignal(ScottPlot.DataGen.Random(System.Random.Shared, 500));

for (int i = 0; i < 100; i++)
    ctrl.Refresh();
```

Result:
|Type|Alloc. before|Alloc. after|Notes|
|-|-|-|-|
|System.String|175.106|17.343|`fontFamily.Name` is inherently reused|
|System.Drawing.FontFamily|216.840|4.400||
|System.Drawing.FontFamily[]|1.881|1|`FontFamily.Families`|

### DIFF
--- a/src/ScottPlot4/ScottPlot/Drawing/GDI.cs
+++ b/src/ScottPlot4/ScottPlot/Drawing/GDI.cs
@@ -36,14 +36,15 @@ namespace ScottPlot.Drawing
             using (Bitmap bmp = new Bitmap(1, 1))
             using (Graphics gfx = Graphics(bmp, lowQuality: true))
             {
-                return MeasureString(gfx, text, font.Name, font.Size, font.Bold);
+                return MeasureString(gfx, text, null, font.Size, font.Bold, font.Family);
             }
         }
 
-        public static SizeF MeasureString(Graphics gfx, string text, string fontName, double fontSize, bool bold = false)
+        public static SizeF MeasureString(Graphics gfx, string text, string fontName, double fontSize, bool bold = false, FontFamily fontFamily = null)
         {
+            fontFamily ??= InstalledFont.ValidFontFamily(fontName);
             var fontStyle = (bold) ? FontStyle.Bold : FontStyle.Regular;
-            using (var font = new System.Drawing.Font(fontName, (float)fontSize, fontStyle, GraphicsUnit.Pixel))
+            using (var font = new System.Drawing.Font(fontFamily, (float)fontSize, fontStyle, GraphicsUnit.Pixel))
             {
                 return MeasureString(gfx, text, font);
             }
@@ -255,13 +256,13 @@ namespace ScottPlot.Drawing
         }
 
         public static System.Drawing.Font Font(ScottPlot.Drawing.Font font) =>
-            Font(font.Name, font.Size, font.Bold);
+            Font(null, font.Size, font.Bold, font.Family);
 
-        public static System.Drawing.Font Font(string fontName = null, float fontSize = 12, bool bold = false)
+        public static System.Drawing.Font Font(string fontName = null, float fontSize = 12, bool bold = false, FontFamily fontFamily = null)
         {
-            string validFontName = InstalledFont.ValidFontName(fontName);
+            fontFamily ??= InstalledFont.ValidFontFamily(fontName);
             FontStyle fontStyle = bold ? FontStyle.Bold : FontStyle.Regular;
-            return new System.Drawing.Font(validFontName, fontSize, fontStyle, GraphicsUnit.Pixel);
+            return new System.Drawing.Font(fontFamily, fontSize, fontStyle, GraphicsUnit.Pixel);
         }
 
         public static StringFormat StringFormat(Alignment algnment)
--- a/src/ScottPlot4/ScottPlot/Drawing/GDI.cs
+++ b/src/ScottPlot4/ScottPlot/Drawing/GDI.cs
@@ -31,7 +31,11 @@ namespace ScottPlot.Drawing
             return gfx.DpiX / DEFAULT_DPI;
         }
 
-        public static SizeF MeasureString(string text, Font font)
+        /// <summary>
+        /// Create a Bitmap and Graphics and use it to measure a string.
+        /// Only use this function if an existing Graphics does not exist.
+        /// </summary>
+        public static SizeF MeasureStringUsingTemporaryGraphics(string text, Font font)
         {
             using (Bitmap bmp = new Bitmap(1, 1))
             using (Graphics gfx = Graphics(bmp, lowQuality: true))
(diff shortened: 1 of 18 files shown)


swharden:
Hi @AbeniMatteo, thanks for your work on this! 

> Fonts installed are cached and not updated until the execution is restarted

This is a big savings in allocations, improves speed, and will be a great improvement to the library. I like the design!

> Some API have changed, check if you're ok with it (mostly string name -> FontFamily family)

If we were only working with private code, I would be okay with this. The proposed API is better too: It makes more sense to store and pass around a `FontFamily` than `string`.

However, users may have written their own classes (plottables, renderables, styles, etc) which rely on the original function signatures. Changing `ScottPlot.Drawing.GDI.Font()` and `ScottPlot.Drawing.GDI.MeasureString()` to require a `FontFamily` could break these users' code. Same with changing the return type of things like `ScottPlot.Drawing.InstalledFont.Monospace()`

I think an ideal solution for this PR is one that does not require modification of any plottable/renderable/style, and minimizes the number of files that have to be edited to achieve the primary goal.

AbeniMatteo:
New approach:
- Same allocations gains
- No changes to existing public APIs
- New overloads\properties accepting FontFamily

There are new internal members on InstalledFont, let me know if you want to make them public.

AbeniMatteo:
TODO: `GDI.Font(null, 12, true)` is now ambigous.

src/ScottPlot4/ScottPlot/Drawing/GDI.cs
```diff
@@ -36,14 +36,20 @@ public static SizeF MeasureString(string text, Font font)
...
+        {
+            var fontFamily = InstalledFont.ValidFontFamily(fontName);
+            return MeasureString(gfx, text, fontFamily, fontSize, bold);
+        }
+
+        public static SizeF MeasureString(Graphics gfx, string text, FontFamily fontFamily, double fontSize, bool bold = false)
         {
             var fontStyle = (bold) ? FontStyle.Bold : FontStyle.Regular;
-            using (var font = new System.Drawing.Font(fontName, (float)fontSize, fontStyle, GraphicsUnit.Pixel))
+            using (var font = new System.Drawing.Font(fontFamily, (float)fontSize, fontStyle, GraphicsUnit.Pixel))
```
bclehmann, line 52: I'm tempted to explore if it would be worth it to cache the combination of font $(family, size, style)$ tuples, rather than creating a new GDI font every time, in my (anecdotal, not empirical) experience string measurement is a relatively common occurrence. In theory, this could even be a global cache, not just for string measurement.

Unfortunately, GDI objects are not thread-safe (discussed in #978), and while there are solutions to that, they're undoubtedly not worth it. Especially when we're supposed to be ditching System.Drawing.

So I'm mostly just saying this for a later time when we're using a different graphics library where they either are thread-safe, or we're willing to invest in the effort to use them safely.

AbeniMatteo: I'm focusing on allocations and in my case there are bigger problems before `System.Drawing.Font`.
This is the summary of the repro in the first post.

|Type|Allocations|% of total|
|-|-|-|
|System.Drawing.FontFamily|216840|36.23|
|System.String|175106|29.26|
|System.Double[]|35915|6.00|
|System.Double|15607|2.61|
|System.Func<,>|15028|2.51|
|System.Collections.Generic.List<>|10453|1.75|
|System.Linq.Enumerable.WhereListIterator<>|9986|1.67|
|ScottPlot.Plottable.SignalPlotBase<System.Double>.IntervalMinMax.<GetPoints>d__4|7700|1.29|
|ScottPlot.Plottable.SignalPlotBase<System.Double>.IntervalMinMax|7700|1.29|
|System.SByte[]|6753|1.13|
|ScottPlot.Renderable.Axis[]|5707|0.95|
|System.Drawing.Graphics|4475|0.75|
|System.Drawing.Font|4287|0.72|
|...|...|...|
|Total|~598k|100|

AbeniMatteo: `Double` and `Double[]` are my next target.

AbeniMatteo:
> TODO: `GDI.Font(null, 12, true)` is now ambigous.

```cs
public static System.Drawing.Font Font(string fontName, float fontSize, bool bold)
public static System.Drawing.Font Font(FontFamily fontFamily, float fontSize, bool bold)
```

When the first argument is `null` and not a typed variable these 2 calls are ambiguous.

How do you want me to proceed?
- Ignore the problem (the error can be suppress by casting the null `GDI.Font((string)null, 12, true)`)
- Expose publicly the new overload only in a new major and breaking version
- Use a different method name

swharden:
> How do you want me to proceed?

Hi @AbeniMatteo, given the 3 options I feel best about using a different method name.

There's technically another non-breaking option. I think the separately named function is the way to go, but it's worth noting that an optional `FontFamily` could achieve this too. If provided, it would override the `string` name.

```cs
public static System.Drawing.Font Font(FontFamily fontFamily, float fontSize, bool bold, FontFamily? fontFamily = null)
```

> Especially when we're supposed to be ditching System.Drawing.

I'll echo this for context. Although I intend to continue support for ScottPlot 4 as long as .NET Framework 4.8 is officially supported (a really long time), I've also been racing to cut System.Drawing from this library with ScottPlot 5, so I have limited motivation to pursue GDI-specific micro-optimizations. PRs like this one are simple and seem like a big win so I'm all for it, but there will be a point of diminishing returns where I will prefer to focus my development effort on cutting System.Drawing rather than further optimizing it

AbeniMatteo:
> I think the separately named function is the way to go, but it's worth noting that an optional FontFamily could achieve this too. If provided, it would override the string name.

Done, ready for review.

swharden:
> Done, ready for review.

Thanks @AbeniMatteo!

It wasn't super clear to me until now, but **the primary performance gain here is because `MeasureString()` now uses cached `FontFamily` objects** making it much faster. `MeasureString()` is called _so_ many times as the layout and ticks are recalculated for every frame.

I missed this point at first because I was distracted at first by the various changed code related to legends, text, marker plots, etc.

> Shouldn't be a problem but fonts installed are cached and not updated until the execution is restarted (a method can be implemented to request an update at runtime)

Manually updating fonts mid-run seems uncommon and I'm happy to omit that function.

However all those `FontFamily` objects implement `IDisposible` but are never disposed because they're cached in the top level of a static class. Is this really OK? One could write a method to iterate and dispose all those, but it would have to be called manually.

bclehmann:
> all those FontFamily objects implement IDisposible but are never disposed because they're cached in the top level of a static class. Is this really OK?

I think it's fine, although some people do disagree on whether you should explicitly handle disposal of eternal objects. The most convincing argument in favour is it could become incorrect after refactoring (for example the lifetime of `InstalledFont` no longer matches the lifetime of the application, or `InstalledFont` becomes a non-static class where multiple instances are created). I don't think that's very likely in this case, the point of the class is to be static and eternal.

FWIW, there is an automated way to do this, you can create your own non-static class which has a destructor to do all the cleanup. The static class should hold this non-static class as a field:

From: https://stackoverflow.com/questions/4364665/static-destructor
```cs
public static class Foo
{
    private static readonly Destructor Finalise = new Destructor();

    static Foo()
    {
        // One time only constructor.
    }

    private sealed class Destructor
    {
        ~Destructor()
        {
            // One time only destructor.
        }
    }
}
```

You'll notice this is pretty hacky, and if you want to explicitly handle the lifetime, I would probably change `InstalledFont` (or at least the section of it that caches `IDisposables`) into a non-static singleton, so you can just use a normal destructor. A cache is generally one of the better uses for singletons in my view.

PR #1927, opened 2022-06-30
Signal: create platform-specific generic conversion methods to avoid boxing doubles

AbeniMatteo:
**Purpose:**
Generic implementations such as `SegmentedTree<T>` fall back to `Convert.ToDouble` requiring the values to be boxed.
We can implement a fast path for well known and frequently used types (ex: doubles, floats).

Take this example: 
```cs
double M<T>(T value)
{
    // same as typeof(T) == typeof(double)
    if (value is double d) 
        return d;
    else if (value is float f)
        return f;
    else
        return Convert.ToDouble((object)value); // implicit cast to object
}
```
is jitted to:
```cs
double M(double value) => value;
double M(float value) => (double)value;
double M<T>(T value) => Convert.ToDouble((object)value); // implicit cast to object
```

Notes:
- `Convert.ToDouble` is overkill for a simple cast
- `Convert.ToDouble` requires boxing the value when the type is not known
- `#if NET` to avoid regression on old frameworks
- We already optimize some path using `System.Linq.Expressions.Expression` but i prefer this solution

Repro:
```cs
var rnd = new Random(123);
var x = 0d;
var ys = ScottPlot.DataGen.RandomWalk(rnd, 10_000_000);
var xs = Enumerable.Range(0, ys.Length).Select(_ => x += rnd.NextDouble() * 1000).ToArray();

var plot = new ScottPlot.Plot(1000, 1000);
plot.AddSignalXYConst(xs, ys);
plot.Render(lowQuality: true);
```

Result:
||Type|Alloc|Time|Notes
|-|-|-|-|-|
|Now|double|19.964.478|480ms|10M validation + (10M * 1 refresh)|
|After|double|7.154|415ms||


swharden:
> This is an incomplete implementation, i want some feedback before proceeding.

Hi @AbeniMatteo, thanks for looking into this!

We can get performance improvements by adding version-specific optimized code blocks that optimally handle generic values. Enhanced performance is achieved at the cost of code that gets more and more complex and harder to maintain (with duplicated logic for different platforms and data types), so these are the two sides I'm trying to balance between.

The small `double`-specific optimizations so far in this PR seem worth it, and I'd be happy to merge something like this in when you're done.

> We can implement a fast path for well known and frequently used types

Rather than go down that road now, maybe we can get a huge win for a small code change if we use generic math in .NET 7? I've been waiting for this feature forever, and it solves this exact problem. Between `Span<T>` and this, signal plots could become very performant with simple code that doesn't duplicate logic

https://devblogs.microsoft.com/dotnet/dotnet-7-generic-math/

What do you think?

swharden:
Sounds good! 

My only suggestion is to not throw it in `Tools` because big catch-all static libraries that collect random stuff are a bit of an anti-pattern and I wish I hand't created that module.

Maybe a small static class with a better name would be more expressive and portable. Maybe something like ´NumericConversion.DoubeToGeneric()´?

Thanks again for your work on this!

AbeniMatteo:
The implementation is complete. 
I still have to test and profile the results.

___

I also tried generic math, you can play with it [here](https://github.com/AbeniMatteo/ScottPlot/tree/ab-generic-math).
Some notes:
- Conversion from `T` to `double` still requires a `Convert.ToDouble` or this PR
- Supporting mixed compilation for >NET7 and <NET7 isn't trivial
- I'm currently using `INumber `as generic restraint even tho we need less operations (ex: `IAdditionOperators`+`IComparisonOperators`)
- `T.Max()` and `T.Min()` are only implemented on `INumber`
- Support for TimeSpan and DateTime was dropped
https://github.com/dotnet/runtime/pull/69331
- A solution the previous point is that it should be possible to write a simple struct wrapper around non INumber-types the user want to use
```cs
struct Wrapper : IAdditionOperators<Wrapper, Wrapper, Wrapper>, IConvertible
{
    public Wrapper(TimeSpan span) { }

    public static Wrapper op_CheckedAddition(Wrapper left, Wrapper right) { throw null; }
    public static Wrapper operator +(Wrapper left, Wrapper right) { throw null; }
    public double ToDouble(IFormatProvider provider) { throw null; }
}
```

AbeniMatteo:
And of course, `NumericConversion` has a typo.

src/ScottPlot4/ScottPlot/NumeriConversion.cs
```diff
@@ -0,0 +1,34 @@
+﻿using System;
+using System.Runtime.CompilerServices;
+
+namespace ScottPlot
+{
+    public static class NumeriConversion
```
bclehmann, line 6: typo

src/ScottPlot4/ScottPlot/NumeriConversion.cs
```diff
@@ -0,0 +1,34 @@
...
+    {
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
```
bclehmann, line 8: I'm normally pretty hesitant when it comes to `AggressiveInlining`, I find that my intuition as to where I should apply it is not as good as I'd like it to be.

I'd be interested to see a benchmark with and without this attribute. Or indeed checking with a disassembler to see if it gets inlined anyways, but iirc that gets difficult because it's inlined by the JITer, not the AOT compiler.

sharplab should let you see the JITted assembly: https://sharplab.io/#v2:C4LghgzgtgPgAgJgIwFgBQcDMACR2DC2A3utmbjnACzYCyAFAJTGnkC+6bQA

bclehmann: At any rate, this is a minor quibble, and feel free to overrule me.

Especially since writing benchmarks + examining assembly is perhaps not worth the effort.

src/ScottPlot4/ScottPlot/NumeriConversion.cs
```diff
@@ -0,0 +1,34 @@
...
+        public static double GenericToDouble<T>(T value)
+        {
+            return value switch
+            {
+#if NET
```
bclehmann, line 13: Just for my ignorance, `NET` corresponds to .NET 5 or greater?

It may make sense to replace it with `NET5_0_OR_GREATER`, because to most people .NET could reasonably encompass .NET Framework, Core, and 5+, and potentially even Mono.

Again, this is a matter of taste, anyone who has to modify this code will figure it out, it just saves them a few swearwords at Microsoft's naming scheme.

bclehmann (approved):
LGTM, a few comments but one of them's a typo, and the other two are minor.

bclehmann:
Don't take this as a suggestion for this PR, rather it's a request for input on something I've been considering for a while:

It might be worth creating a benchmarks project, just so that further performance work can rely on less ad hoc testing?

I still expect that most performance-related PRs would require additional benchmarks to be written for validation, it would just streamline the process. For example, we could hook it up to [BenchmarkDotNet ](https://github.com/dotnet/BenchmarkDotNet) and not have to write our own throwaway code to track timing and allocations (or rely on an external profiler, which is less convenient and reproducible, at least for me).

There is also the option to make benchmarks and certain minimum performance characteristics part of CI. This is probably not even worth considering until we have a large number of benchmarks written (and perhaps not even then), but I thought I'd bring it up.

AbeniMatteo:
Micro benchs results are really good:
- Gains for `net5` and `net6` are massive
- `netcore2` shows no improvements (i'm using `#if NETCOREAPP`), i think this optimization is available from `netcore3`
- no regression on `.NET framework` for small values like `float` and `double`
- small regression on <`net20` for big types like `decimal`
```cs
public double New(T value)
{
    var a = Utils.ConvertGenericToDouble(value);
    a *= 999;
    a -= 5.6;
    return a;
}

public double Old(T value)
{
    var a = Convert.ToDouble(value);
    a *= 999;
    a -= 5.6;
    return a;
}
```

|Type    | Method |       Jit |              Runtime |     Mean | Code Size |
|-------|------- |---------- |--------------------- |---------:|----------:|
|double|    New |    RyuJit |             .NET 5.0 |  0.2427 ns |      20 B |
|double|    Old |    RyuJit |             .NET 5.0 |  9.5874 ns |     136 B |
|double|    New |    RyuJit |             .NET 6.0 |  0.0041 ns |      20 B |
|double|    Old |    RyuJit |             .NET 6.0 | 10.2222 ns |     136 B |
|double|    New |    RyuJit |        .NET Core 2.0 |  9.0733 ns |     420 B |
|double|    Old |    RyuJit |        .NET Core 2.0 | 10.2520 ns |     123 B |
|double|    New |    RyuJit | .NET Framework 4.7.2 |  9.9730 ns |     135 B |
|double|    Old |    RyuJit | .NET Framework 4.7.2 |  9.9856 ns |     135 B |
|double|    New | LegacyJit | .NET Framework 4.7.2 |  9.9936 ns |     135 B |
|double|    Old | LegacyJit | .NET Framework 4.7.2 | 10.1067 ns |     135 B |
|-||||||||
|float|    New |    RyuJit |             .NET 5.0 |  0.4692 ns |      24 B |
|float|    Old |    RyuJit |             .NET 5.0 | 10.0318 ns |     136 B |
|float|    New |    RyuJit |             .NET 6.0 |  0.4512 ns |      24 B |
|float|    Old |    RyuJit |             .NET 6.0 | 10.5184 ns |     136 B |
|float|    New |    RyuJit |        .NET Core 2.0 | 12.4097 ns |     420 B |
|float|    Old |    RyuJit |        .NET Core 2.0 | 10.1572 ns |     123 B |
|float|    New |    RyuJit | .NET Framework 4.7.2 |  9.9812 ns |     135 B |
|float|    Old |    RyuJit | .NET Framework 4.7.2 |  9.9965 ns |     135 B |
|float|    New | LegacyJit | .NET Framework 4.7.2 | 10.0919 ns |     135 B |
|float|    Old | LegacyJit | .NET Framework 4.7.2 | 10.2973 ns |     135 B |
|-||||||||
|decimal|    New |    RyuJit |             .NET 5.0 | 13.96 ns |    147 B |
|decimal|    Old |    RyuJit |             .NET 5.0 | 14.42 ns |    133 B |
|decimal|    New |    RyuJit |             .NET 6.0 | 13.65 ns |    147 B |
|decimal|    Old |    RyuJit |             .NET 6.0 | 14.92 ns |    133 B |
|decimal|    New |    RyuJit |        .NET Core 2.0 | 25.17 ns |    462 B |
|decimal|    Old |    RyuJit |        .NET Core 2.0 | 17.91 ns |    123 B |
|decimal|    New |    RyuJit | .NET Framework 4.7.2 | 16.87 ns |    143 B |
|decimal|    Old |    RyuJit | .NET Framework 4.7.2 | 15.90 ns |    134 B |
|decimal|    New | LegacyJit | .NET Framework 4.7.2 | 17.02 ns |    143 B |
|decimal|    Old | LegacyJit | .NET Framework 4.7.2 | 15.47 ns |    134 B |

src/ScottPlot4/ScottPlot/NumeriConversion.cs
AbeniMatteo: > I'm normally pretty hesitant when it comes to AggressiveInlining, I find that my intuition as to where I should apply it is not as good as I'd like it to be.

I'm inlining a single cast or a single call, using this attribute can only help.

> I'd be interested to see a benchmark with and without this attribute.

(Full plot rendering benchs are coming)

|           Type |    Method |       Mean |     Error |    StdDev | Code Size |
|--------------- |---------- |-----------:|----------:|----------:|----------:|
| Decimal |       New | 15.2704 ns | 1.3047 ns | 0.0715 ns |      86 B |
|  Double |       New |  1.4432 ns | 0.2544 ns | 0.0139 ns |      41 B |
|   Int32 |       New |  1.6816 ns | 0.1801 ns | 0.0099 ns |      47 B |
|  Single |       New |  1.4643 ns | 0.6144 ns | 0.0337 ns |      45 B |
| Decimal | NewAggInl | 13.4416 ns | 1.2293 ns | 0.0674 ns |     147 B |
|  Double | NewAggInl |  0.2426 ns | 0.0863 ns | 0.0047 ns |      20 B |
|   Int32 | NewAggInl |  0.2606 ns | 0.2102 ns | 0.0115 ns |      28 B |
|  Single | NewAggInl |  0.4824 ns | 0.7316 ns | 0.0401 ns |      24 B |
| Decimal |       Old | 14.1851 ns | 0.3508 ns | 0.0192 ns |     133 B |
|  Double |       Old | 10.7178 ns | 5.1088 ns | 0.2800 ns |     136 B |
|   Int32 |       Old | 10.3100 ns | 1.0246 ns | 0.0562 ns |     126 B |
|  Single |       Old | 10.5590 ns | 0.6282 ns | 0.0344 ns |     136 B |

swharden:
> The implementation is complete.
> I still have to test and profile the results.

@AbeniMatteo I notice the PR is still marked "draft", and I don't want to merge it before it's ready.

Let me know if/when you're done making edits and I'll make commits as needed to resolve any outstanding topics and merge it 👍

AbeniMatteo:
@swharden Now it's ready for review, i was looking for a problem, fixed [051a8f1](https://github.com/ScottPlot/ScottPlot/pull/1927/commits/051a8f1d4e175b20b0b368e517e220e91a68cffe)🙃.

I need your opinion on:
- which type needs this optimization (ATM double, float, int, long, decimal).
- all the `ref` and `out` shenanigans

Notes:
- this optimization is available only for >=netcoreapp1.0
- netcoreapp2-3 are not affected because ScottPlot is compiled for net5, netstandard2 and net46 and we have to reference netstandard2
- netcoreapp2 and net461 are tested to check for regressions
- FirstRender is heavily impacted because of AssertDoesNotDescend
- net5 is slightly better the net6 (possible TODO)

Benchs for 10M points, 1000x1000px plot (mean values are in milliseconds, allocated in KB):
```cs
public void AssertDoesNotDescend() 
    => ScottPlot.Validate.AssertDoesNotDescend<T>(_xs);

public void FirstRender()
    => _plot.Render(lowQuality: true);

public void MultRenders()
{
    // First render is not measured
    for (int i = 0; i < 100; i++)
        _plot.Render(lowQuality: true);
}
```

Results:
image

src/ScottPlot4/ScottPlot/NumeriConversion.cs
AbeniMatteo: Switched to NETCOREAPP

swharden:
## Optimized Types
> Which types need this optimization? Currently `double`, `float`, `int`, `long`, and `decimal`.

**This looks like a good list to me!** The next type I would probably reach for would be `byte`, but type-specific fast paths are easy to add later, and it is easy to fall into the _just one more_ trap until we've implemented the entire native type system.

I invite @StendProg to give feedback here if they are available. @StendProg is an early contributor to this library and they greatly improved signal plots in the past, implementing the strategy pattern, creating the segmented tree implementations, and also adding generics to signal plots in the first place. I welcome their input about what types are worth optimizing, and also their overall feedback on this PR if they are available!

_I'll let this PR sit here for a few days before merging to give @StendProg an opportunity to weigh-in if they want to_

## `ref` and `out`

> I need your opinion on: ... all the `ref` and `out` shenanigans

**short answer: this PR looks good**

I typically avoid using `ref` and `out` now that the modern language rarely requires it - I find returning and deconstructing tuples is more readable. However, since we're aiming for performance optimization in non-user-facing code, I'm okay with whatever is fastest. Lots of the existing functionality in already in these strategy classes use `ref` and `out`, so I'm inclined to keep additions to these classes consistent with what's there already (and tested/working). 

I would be somewhat interested to evaluate performance (outside of this PR) of these classes if they were refactored to avoid using `ref` and `out`, but this code doesn't change often and is pretty locked-down and well-tested, so optimizing for readability probably isn't a high-yield effort.

## Target Framework
I recognize this is near bikeshedding territory, but I find `#if NETCOREAPP` to be a little unclear if this includes modern .NET targets because of how Microsoft stopped naming modern .NET versions "core". I may change these statements to `NETCOREAPP1_0_OR_GREATER` to make it more explicit. By the end of the year Microsoft will end-of-life .NET Core 3.1, so I'm even tempted to use `NET5_0_OR_GREATER` to make this extremely clear. The only downside would be soon-EOL'd targets wouldn't get this new micro-optimization. 

* [C# preprocessor directives](https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/preprocessor-directives)
* [.NET End of Life Dates](https://dotnet.microsoft.com/en-us/platform/support/policy/dotnet-core#lifecycle)

StendProg:
Hi all,

@swharden thank you for inviting me. This topic really interested me.

Yes, indeed, the conversion to double is done carelessly and can be improved. This conversions actually take up less than a percent of all other calculations, which is why they haven't been given enough attention.

@AbeniMatteo Your solution is clear. You did a great job, especially all those performance tests. The only thing I don't like is the static class, it could be replaced with an injected dependency. Then you could easily write tests that check the identity of your solution to the previous version, and automatic benchmarks too.

Since this solution is intended to improve performance, why don't we go further.
You are doing a type check every time you call a method. While in reality the type of `Generic` can only be checked in the constructor of an object, it will not change in the future for the entire lifetime of the object.

1. You could create a family of converters, each for a specific type, initialize only the one you need in the constructor.
```c#
public interface IGenericToDoubleConvertor<T>
{
  double ConvertToDouble(T value);
}

public class DoubleToDoubleConvertor : IGenericToDoubleConvertor<double>
{
  public double ConvertToDouble(double value)
  {
      return value;
  }
}

public class FloatToDoubleConvertor : IGenericToDoubleConvertor<float>
{
   public double ConvertToDouble( float value)
   {
     return (double)value;
   }
}

....


IGenericToDoubleConvertor _convertor;

public SegmentedTree()
{
  T testValue = default(T);
  _convertor = testValue swich 
  {
     double d => new DoubleToDoubleConvertor() as IGenericToDoubleConvertor(),
     float => new FloatToDoubleConvertor() as IGenericToDoubleConvertor(),
     _ => new UsingConvertToDoubleConvertor() as IGenericToDoubleConvertor, 
  };
}

//use it then need
 lowestValue = _convertor.ConvertToDouble(lowestValueT);
 ...
```

2. We can assemble such a converter based on a lambda expression, by analogy with many of the existing ones. `Convert` Expression  looks like what we need. It will be worse readable, but most likely a little faster. We already use a lot of lambda expressions and it is unlikely that one superfluous one will greatly spoil readability.

AbeniMatteo:
From @swharden:
> This looks like a good list to me! The next type I would probably reach for would be byte, but type-specific fast paths are easy to add later, and it is easy to fall into the just one more trap until we've implemented the entire native type system.

Note for the future: fast\no alloc paths can be implemented for types accepted by the various `Convert.ToXXX(double value)` and `Convert.ToDouble(XXX value)` overloads.

> I would be somewhat interested to evaluate performance (outside of this PR) of these classes if they were refactored to avoid using ref and out

You lose some perfs and have a slightly higher regression on older versions (but i'm starting to think is not worth using ref\out):
- columns 1-2 for no ref\out
- columns 3-4 for the current implementation
- columns 5-6 for the diffs between the two
image

AbeniMatteo:
From @StendProg:
> You are doing a type check every time you call a method

Type checks are eliminated by the JIT compiler and then the methods are also inlined resulting in a `noop` for types like `double`.

```cs
public static class C 
{
    [SharpLab.Runtime.JitGeneric(typeof(Double))]
    [SharpLab.Runtime.JitGeneric(typeof(float))]
    public static T DoubleToGeneric<T>(double value)
    {
        if (typeof(T) == typeof(double))
            return (T)(object)value;
        else if (typeof(T) == typeof(float))
            return (T)(object)Convert.ToSingle(value);
        else if (typeof(T) == typeof(decimal))
            return (T)(object)Convert.ToDecimal(value);
        else
            return (T)Convert.ChangeType(value, typeof(T));
    }
    
    [SharpLab.Runtime.JitGeneric(typeof(Double))]
    [SharpLab.Runtime.JitGeneric(typeof(float))]
    public static double GenericToDouble<T>(T value)
    {
        if (value is double d)
            return d;
        else if (value is float f)
            return Convert.ToDouble(f);
        else if (value is decimal dec)
            return Convert.ToDouble(dec);
        else
            return Convert.ToDouble(value);
    }
}
```
is jitted to:
```asm
; Core CLR 6.0.322.12309 on amd64

C.DoubleToGeneric[[System.Double, System.Private.CoreLib]](Double)
    L0000: vzeroupper
    L0003: ret

C.DoubleToGeneric[[System.Single, System.Private.CoreLib]](Double)
    L0000: vzeroupper
    L0003: vcvtsd2ss xmm0, xmm0, xmm0
    L0007: ret

C.GenericToDouble[[System.Double, System.Private.CoreLib]](Double)
    L0000: vzeroupper
    L0003: ret

C.GenericToDouble[[System.Single, System.Private.CoreLib]](Single)
    L0000: vzeroupper
    L0003: vcvtss2sd xmm0, xmm0, xmm0
    L0007: ret
```

https://sharplab.io/#v2:EYLgxg9gTgpgtADwGwBYA0AXEBDAzgWwB8ABAJgEYBYAKBuIGYACY8pZ0xgYUYG8bGBjANoBlABbYoABwAy2YADoASgFcAdhgCW+GAoBSmjAHEYamFE1gAFBgCeUmBABmVgCIQVwADYwAlL4BdfkFRCWk5RVUNbV0DY1NzSxt7RxcnLwhsDH8g6kFmJhY2ABVGd08fYogTMwswAB5igD4rABMPbxhGADdsLxU/YIE+PPzBTSdGZIdnK2LfRgBeRcY7GZd2ir9fIbH84gB2KfmrCGAAKxgwbN7+mABuXfyYL1wuiam11LmF5dWU2bpTLZHajPb7I4/U4XK7ZTgQNTdcwYBRVESaNQAcx8VluA18jzBYxeb0YH2m33mSxWX1mrSu2j6/ie4MOx180Mu1188MRyNREFcDPwfVxfXxhPBjBJMBZezZP15SKgKM4EixMGKKTFdzQ/3WPwJuwAvrtdqFJLJ5Mp1FodPpDDVEtZaS5yp0cubxJaIjbova4k66hTARksp6iQxmKxGJtOowg5Yqu6fI0WqU8YMiSMpeTM2TcLGOj5Y6CpRDY5LwTKyZMdQMC4wgVkm2XywI2Ur+cnizArE4jUTnq93nX85pC/SwIyvLGrm3252EcqUT2tm151W9jK5WMl3yVQKU33M4P8qbqBegA=

StendProg:
Compilers are getting smarter that you can't keep up. However, the .NET Framework is not that smart, and gave me dozens of lines of code for `GenericToDouble<T>`. Formally, your solution will degrade performance on this platform. But this difference is barely visible in your benchmarks.

AbeniMatteo:
> Compilers are getting smarter that you can't keep up. However, the .NET Framework is not that smart, and gave me dozens of lines of code for `GenericToDouble<T>`. Formally, your solution will degrade performance on this platform. But this difference is barely visible in your benchmarks.

It only calls `Convert.ToXXX` when compiled for netfx and netstandard. 

https://github.com/ScottPlot/ScottPlot/blob/6d1b85518528746b01f773536ed1a081946b4f42/src/ScottPlot4/ScottPlot/NumericConversion.cs#L16-L29

AbeniMatteo:
I'm implementing `short` and the various unsigned since i'm already in need of them.

swharden:
Thanks everyone for your collaboration on this! It seems to have landed in a good place (no major updates in about a week) so I'm going to merge this and make a release tonight. 

I'm happy `short` and `byte` made it in, and it will be easy to add support for additional types in the future 👍

AbeniMatteo:
@swharden `byte` was not implemented because `Linq.Expressions` doesn't handle `byte+byte`.

swharden:
> Linq.Expressions doesn't handle byte+byte

I didn't appreciate the impact of this until today. This PR resulted in the loss of support for plotting `byte[]` in SignalConst and SignalXY plots. An `InvalidOperationException` exception is now thrown at instantiation.

I'm motivated to figure out a workaround here... do you have any suggestions? Even if it's not an efficient solution for `byte`, it seems better than a hard crash at runtime.

https://github.com/ScottPlot/ScottPlot/blob/63e2fe139747ebf478090ee4363be92424d3d3a4/src/ScottPlot4/ScottPlot/Plottable/SignalPlotBase.cs#L266-L272

AbeniMatteo:
> This PR resulted in the loss of support for plotting byte[] in SignalConst
Are you sure? we fall back to Convert.ToXXX for types without an explicit fast path implementation.

swharden:
> Are you sure?

Yeah, I demonstrate it with a new test.

https://github.com/ScottPlot/ScottPlot/blob/63e2fe139747ebf478090ee4363be92424d3d3a4/src/ScottPlot4/ScottPlot.Tests/NumericConversionTests.cs#L38-L41

I think I found a solution though. What do you think about this? It seems to work, but I'm a bit over my head here with the advanced use of generics and reflection here.

```cs
public static byte AddBytes(byte a, byte b) => (byte)(a + b);
```

```cs
public SignalPlotBase()
{
    ParameterExpression paramA = Expression.Parameter(typeof(T), "a");
    ParameterExpression paramB = Expression.Parameter(typeof(T), "b");
    BinaryExpression bodyAdd = Type.GetTypeCode(typeof(T)) switch
    {
        TypeCode.Byte => Expression.Add(paramA, paramB, typeof(NumericConversion).GetMethod(nameof(NumericConversion.AddBytes))),
        _ => Expression.Add(paramA, paramB),
    };
    AddYsGenericExpression = Expression.Lambda<Func<T, T, T>>(bodyAdd, paramA, paramB).Compile();
}
```

swharden:
I created #1945 which I think resolves this issue, but I appreciate feedback!

I moved the complex type logic into the `NumericConversion` class so adding support for new types (or improving performance) will not require modifying plottables. It also keeps their constructors empty.

PR #1930, opened 2022-07-02
Creates Benchmark project using BenchmarkDotNet

bclehmann:
**Purpose:**
https://github.com/ScottPlot/ScottPlot/pull/1927#issuecomment-1171821418

A benchmark class looks like this:
```cs
[SimpleJob(launchCount: 2, warmupCount: 1, targetCount: 3)]
[IterationTime(200)]
[MemoryDiagnoser]
public class GdiObjectCreation
{
    [Benchmark]
    public void CreateAndDestroy_Brush()
    {
        var item = new SolidBrush(Color.Black);
        item.Dispose();
    }
    
    .
    .
    .
}
```

I overrode the `launchCount`, `warmupCount`, `targetCount`, and `IterationTime` because by default the benchmarks run for several minutes, and for what I was benchmarking that doesn't change the results much (unless you want more detailed GC stats.

Output:
```
|                        Method |       Mean |     Error |    StdDev |  Gen 0 |  Gen 1 | Allocated |
|------------------------------ |-----------:|----------:|----------:|-------:|-------:|----------:|
|        CreateAndDestroy_Brush |   254.3 ns |  36.76 ns |  13.11 ns | 0.0294 |      - |      64 B |
|          CreateAndDestroy_Pen |   212.8 ns |   8.16 ns |   2.91 ns | 0.0304 |      - |      64 B |
|         CreateAndDestroy_Font | 2,097.9 ns | 748.34 ns | 266.87 ns | 0.1892 | 0.0568 |     409 B |
|   CreateAndDestroy_FontFamily |   214.1 ns |   5.01 ns |   1.79 ns | 0.0181 |      - |      40 B |
| CreateAndDestroy_StringFormat |   282.7 ns |   9.47 ns |   3.38 ns | 0.0142 |      - |      32 B |
```

Note that Gen 0 and Gen 1 are the number of garbage collections at that generation number per 1000 iterations.


bclehmann:
Also, this removes `ScottPlot.Tests/Performance`, because those tests were moved into the Benchmarks project.

And lastly, I'm not sure how official and rigorous you want this. It could be set up to run benchmarks on every .NET version, generate detailed logs, charts, etc, but I don't know which of those are actually useful. The impetus for this was to make validating individual performance changes easier, not to track long-term performance (that would require way more benchmarks anyways), so I think it can probably stay pretty bare-bones.

swharden:
I modified the project to add a command argument to run all benchmarks without entering the menu, then added a CI job to run the benchmarks. The output is useful (especially for future discussions related to performance), so this seems worth keeping around!

Note that this output....

```
|                        Method |        Mean |    Error |   StdDev |  Gen 0 | Allocated |
|------------------------------ |------------:|---------:|---------:|-------:|----------:|
|        CreateAndDestroy_Brush |    166.1 ns |  2.34 ns |  0.83 ns | 0.0025 |      56 B |
|          CreateAndDestroy_Pen |    181.0 ns |  4.02 ns |  1.43 ns | 0.0027 |      64 B |
|         CreateAndDestroy_Font | 11,122.4 ns | 89.63 ns | 31.96 ns |      - |     104 B |
|   CreateAndDestroy_FontFamily |    280.8 ns | 16.01 ns |  5.71 ns | 0.0014 |      32 B |
| CreateAndDestroy_StringFormat |    178.1 ns |  9.68 ns |  3.45 ns | 0.0009 |      24 B |
```

is valid GitHub flavored Markdown:

|                        Method |        Mean |    Error |   StdDev |  Gen 0 | Allocated |
|------------------------------ |------------:|---------:|---------:|-------:|----------:|
|        CreateAndDestroy_Brush |    166.1 ns |  2.34 ns |  0.83 ns | 0.0025 |      56 B |
|          CreateAndDestroy_Pen |    181.0 ns |  4.02 ns |  1.43 ns | 0.0027 |      64 B |
|         CreateAndDestroy_Font | 11,122.4 ns | 89.63 ns | 31.96 ns |      - |     104 B |
|   CreateAndDestroy_FontFamily |    280.8 ns | 16.01 ns |  5.71 ns | 0.0014 |      32 B |
| CreateAndDestroy_StringFormat |    178.1 ns |  9.68 ns |  3.45 ns | 0.0009 |      24 B |

> I'm not sure how official and rigorous you want this. It could be set up to run benchmarks on every .NET version, generate detailed logs, charts, etc, but I don't know which of those are actually useful.

That's probably not needed (or at least we can add it if/when we need it), but it may be worth considering which platform would be the best representative one to test. I went with Ubuntu here because it's fast, but I see where Windows may also make sense (and would probably last further in the future as System.Drawing.Common loses Linux support)

PR #1931, opened 2022-07-03
Signal: optimization to draw fewer overlapping vertical lines

AbeniMatteo:
While we finalize the other PR, we can start the conversation for this one.

**Purpose:**
Signals render 3 lines where only 2 or 1 are needed, this PR reduces the points\lines used to draw an interval.
For simplicity, i'm not doing anything fancy with the first and the last point.

This image shows an exploded view of what is happening on the same X pixel (and the end result on the right).

image

### DIFF
--- a/src/ScottPlot4/ScottPlot/Plottable/SignalPlotXYGeneric.cs
+++ b/src/ScottPlot4/ScottPlot/Plottable/SignalPlotXYGeneric.cs
@@ -97,12 +97,19 @@ namespace ScottPlot.Plottable
 
             var pointsCount = endIndex - startIndex;
 
-            yield return new PointF(x + dims.DataOffsetX, dims.GetPixelY(Strategy.SourceElement(startIndex) + Convert.ToDouble(OffsetY)));
+            var startValue = Strategy.SourceElement(startIndex);
+            yield return new PointF(x + dims.DataOffsetX, dims.GetPixelY(startValue + Convert.ToDouble(OffsetY)));
             if (pointsCount > 1)
             {
-                yield return new PointF(x + dims.DataOffsetX, dims.GetPixelY(min + Convert.ToDouble(OffsetY)));
-                yield return new PointF(x + dims.DataOffsetX, dims.GetPixelY(max + Convert.ToDouble(OffsetY)));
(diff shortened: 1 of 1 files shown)


swharden:
> Signals render 3 lines where only 2 or 1 are needed

Hi @AbeniMatteo, thanks for spearheading this conversation! I'll share two of my thoughts which may be worth considering when exploring this topic.

1) This duplication may have been added on purpose to improve **GDI-specific rendering quirks** when anti-aliasing is enabled or disabled. I don't recall the details, and if this were the case I should have documented justification using code comments. Exploration of this topic should include comparing high density signal plots in both anti-aliased and non-anti-aliased modes to ensure the quality of the graphics don't unintentionally get worse through this optimization. Let's test it and see, and if this is the case we can improve the code by documenting why that duplication is necessary.

2) Signal plots draw a number of lines linearly related to the width of the data area (in pixels). This means that there should never be an extremely large number of lines (or duplicated points), so optimizing rendering after the data reduction has already occurred **may not yield a strong performance improvement**. Benchmarks are better than gut feelings though, so I'm definitely in favor of exploring this further!

StendProg:
Great improvement, drawing a line is billion times more expensive than just checking. But I agree with @swharden  skepticism, and I also think that in real signals the number of saved lines will always tend to zero. 
In any case, checks are very cheap, why not add them.

AbeniMatteo:
> Exploration of this topic should include comparing high density signal plots in both anti-aliased and non-anti-aliased modes 

I will setup an automated test that compares pixel by pixel.

>Benchmarks are better than gut feelings though

Quick test with a random signal and one more stable (with `lowQuality: true`).

|Values|Fix|Points|Time|
|-|-|-|-|
|Random|Before|5436|71ms|
|Random|After|4696|64ms|
|-||||
|Stable|Before|5452|69ms|
|Stable|After|3626|62ms|

Random

Stable

swharden:
This PR has been an open draft without new activity in a few weeks so I'll close it for now, but I'm happy to revisit it if again if it's opened with screenshots/data to better evaluate the change in quality (if there is any) and performance 👍

Issue #1925, opened 2022-06-30
Cookbook: show generation date on home page


swharden:
Category pages show generation date:
https://scottplot.net/cookbook/4.1/category/quickstart/

but the home page does not:
https://scottplot.net/cookbook/4.1/

PR #1939, opened 2022-07-08
Cookbook Website: Adds date generated

bclehmann:
**Purpose:**
#1925

It looks like this:
image


bclehmann:
I don't know if you want me to open a PR in [ScottPlot/ScottPlot.NET](https://github.com/ScottPlot/ScottPlot.NET) as well, or just leave it be. I personally think it's fine if we only regenerate it once a new release comes out.

swharden:
Whoops, I forgot to check open PRs before duplicated some of this effort in another PR. Your formatting is superior though, the language is more internally consistent, and the screenshot helps a ton - Thanks!

PR #1945, opened 2022-07-10
Generic Signal plots: add support for byte arrays

swharden:
The goal of this PR is to add support for `byte[]` back into generic signal (SignalConst and SignalXY) types

* extends #1927

### DIFF
--- a/src/ScottPlot4/ScottPlot.Avalonia/ScottPlot.Avalonia.NUGET.csproj
+++ b/src/ScottPlot4/ScottPlot.Avalonia/ScottPlot.Avalonia.NUGET.csproj
@@ -15,9 +15,9 @@
     <PackageReleaseNotes>https://ScottPlot.NET/changelog</PackageReleaseNotes>
     <RepositoryUrl>https://github.com/ScottPlot/ScottPlot</RepositoryUrl>
     <GeneratePackageOnBuild>true</GeneratePackageOnBuild>
-    <Version>4.1.52</Version>
-    <AssemblyVersion>4.1.52.0</AssemblyVersion>
-    <FileVersion>4.1.52.0</FileVersion>
+    <Version>4.1.53</Version>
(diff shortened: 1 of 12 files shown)


AbeniMatteo:
This problem was introduced long ago, nothing to do with #1927.

Introducing a specialized path for bytes doesn't fix the entire problem, since byte can easily overflow during an addition.

swharden:
> This problem was introduced long ago, nothing to do with https://github.com/ScottPlot/ScottPlot/pull/1927.

Thanks for pointing this out! This fails on many old versions of ScottPlot, and the failure was not introduced by #1927
```cs
ScottPlot.Plot plt = new();
plt.AddSignalConst(new byte[] { 1, 2, 3 });
```

```
Exception thrown: 'System.InvalidOperationException' in System.Linq.Expressions.dll
```

> byte can easily overflow during an addition

This is a good point, but in practice it seems okay.

```cs
public Form1()
{
    InitializeComponent();
    byte[] ys = DataGen.Sin(100_000, 5_000, 100, 100).Select(x => (byte)x).ToArray();
    formsPlot1.Plot.AddSignalConst(ys);
    formsPlot1.Refresh();
}
```

byte4

AbeniMatteo:
I'm confused, the exception was just replaced by an incorrect implementation.

> This is a good point, but in practice it seems okay.

It seems okay because you're only showing the values and not doing anything else with them.
For example, this is the result of using `OffsetY`:
```cs
var plot = new ScottPlot.FormsPlot { Dock = DockStyle.Fill };
var ys = ScottPlot.DataGen.Sin(100, 100, 100, 100).Select(x => (byte)x).ToArray();
var signal = plot.Plot.AddSignalConst(ys);
signal.OffsetY = 100;
plot.Refresh();

var form = new Form { Width = 500, Height = 500 };
form.Controls.Add(plot);
form.ShowDialog();
```

Issue #1942, opened 2022-07-10
YDataRange for ScatterPlots, SignalPlots


bclehmann:
## Feature Suggestion

**Feature description:** Gives the min and max y within an interval of x values. See also #1929

**Code example:**

```cs
(double yMin, double yMax) = sig.YDataRange(myPlot.GetAxisLimits());
myPlot.AutoAxisY(yMin, yMax);
```

PR #1946, opened 2022-07-10
Signal and Scatter: GetYDataRange()

bclehmann:
**Purpose:**
Adds `GetYDataRange` to `IHasPointsGenericX`. This interface is implemented by ScatterPlots and SignalPlots (and their derivatives). #1942, #1929

```cs
(double yMin, double yMax) = scat.GetYDataRange(xStart, xEnd);
```

I wrote some tests, though they might be a little sparse. I'm not sure how this should be put in the cookbook? I suppose we could just draw a line at the bounds like we did for nth-order statistics and quantiles: https://scottplot.net/cookbook/4.1/category/statistics/#quantiles

### DIFF
--- a/src/ScottPlot4/ScottPlot.Tests/GetYDataRange.cs
+++ b/src/ScottPlot4/ScottPlot.Tests/GetYDataRange.cs
@@ -1,0 +1,56 @@
+﻿using NUnit.Framework;
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+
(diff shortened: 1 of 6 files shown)


swharden:
Looks great, thanks @bclehmann! 🚀 

> should be put in the cookbook?

Behavior like this can be pretty difficult to document effectively in the cookbook (especially as it affects multiple - but not all - plottables). I added XML docs to these functions which I think will make it a little easier to find, but I think tactically mentioning them by name in answers to questions is the most practical way to reference them.

Issue #1943, opened 2022-07-10
Hatched fills for Pie, Coxcomb, Radar Chart


bclehmann:
## Feature Suggestion

**Feature description:** 

Hatched fills can be easier to read, especially for colourblind users or when printed in black and white. See https://github.com/ScottPlot/ScottPlot/issues/1940#issuecomment-1179654214, and also #555

**Code example:** If applicable, use this space to demonstrate how you wish the new feature could be used in code.

```cs
pie.SliceHatches = new[] {Drawing.HatchStyle.StripedWideUpwardDiagonal, Drawing.HatchStyle.StripedWideDownwardDiagonal, Drawing.HatchStyle.LargeCheckerBoard }
```

This would also need a `SliceHatchColor` field, which makes me wonder if this should be replace with a `HatchOptions` struct:

```cs
struct HatchOptions {
  public HatchStyle HatchStyle { get; set; }
  public Color HatchColor { get; set; }
}
```

I will probably defer to what Bar plots do (keeping them as separate fields) for consistency's sake, but I thought I'd bring it up in case someone had an opinion.

PR #1948, opened 2022-07-13
Hatched fills for Pie, Coxcomb, and Radar

bclehmann:
**Purpose:**
#1943 

```cs
pie.HatchOptions = new HatchOptions[] {
    new () { HatchStyle = HatchStyle.StripedUpwardDiagonal, HatchColor = Color.FromArgb(100, Color.Gray) },
    new () { HatchStyle = HatchStyle.StripedDownwardDiagonal, HatchColor = Color.FromArgb(100, Color.Gray) },
    new () { HatchStyle = HatchStyle.LargeCheckerBoard, HatchColor = Color.FromArgb(100, Color.Gray) },
    new () { HatchStyle = HatchStyle.SmallCheckerBoard, HatchColor = Color.FromArgb(100, Color.Gray) },
    new () { HatchStyle = HatchStyle.LargeGrid, HatchColor = Color.FromArgb(100, Color.Gray) },
};
```

Pie: 
image

Cocxomb:
image

Radar:
image

I promise they look better than in the screenshots, I used right-click and copy to paste the images in, and that disables anti-aliasing, which might be worth looking into separately.

### DIFF
--- a/src/ScottPlot4/ScottPlot.Cookbook/Recipes/Plottable/Coxcomb.cs
+++ b/src/ScottPlot4/ScottPlot.Cookbook/Recipes/Plottable/Coxcomb.cs
@@ -36,11 +36,11 @@ namespace ScottPlot.Cookbook.Recipes.Plottable
             double[] values = { 11, 16, 7, 3, 14 };
             var coxcomb = plt.AddCoxcomb(values);
             coxcomb.HatchOptions = new HatchOptions[] {
-                new () { HatchStyle = HatchStyle.StripedUpwardDiagonal, HatchColor = Color.FromArgb(100, Color.Gray) },
-                new () { HatchStyle = HatchStyle.StripedDownwardDiagonal, HatchColor = Color.FromArgb(100, Color.Gray) },
-                new () { HatchStyle = HatchStyle.LargeCheckerBoard, HatchColor = Color.FromArgb(100, Color.Gray) },
-                new () { HatchStyle = HatchStyle.SmallCheckerBoard, HatchColor = Color.FromArgb(100, Color.Gray) },
-                new () { HatchStyle = HatchStyle.LargeGrid, HatchColor = Color.FromArgb(100, Color.Gray) },
+                new () { Pattern = HatchStyle.StripedUpwardDiagonal, Color = Color.FromArgb(100, Color.Gray) },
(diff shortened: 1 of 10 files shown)


bclehmann:
I ended up going with a `HatchOptions` struct, and I think when ScottPlot 5 comes around we should change `BarPlot` to do the same thing. I think we ought to have a place to put changes we want to make with the next major version, because I ran into a few:

- `HatchOptions` rather than separate properties for BarPlot
- `WebColor` is used in Coxcomb and Radar plots as the name for the axis colour, and it's a confusing name. Especially as it can be confused with the other [web colours](https://en.wikipedia.org/wiki/Web_colors)
- `LineColor` and `LineWidth` can probably be usefully combined into `LineOptions` or similar.

Some of these may be too far, and make updating to ScottPlot 5 unnecessarily difficult, but I think they're at least worth considering.

bclehmann:
My implementation also has more brush allocations (1 per pie slice per render rather than 1 per render), I figured this was cleaner and the performance impact was negligible.

I am happy to change it to a more efficient version however, if you decide it's worth it.

swharden:
Naming is hard! There's a `ScottPlot.Drawing.HatchStyle` enumeration which in hindsight should have been named `ScottPlot.HatchPattern`

I'm not really worried about performance or nailing naming at this point. This library evolved so much and is strung together in weird ways due to a years-long effort to minimize breaking API changes.

Same thing with WebColor 😅 

Regarding ScottPlot 5, `MarkerOptions`, `LineOptions`, `FillOptions`, `LabelOptions` sound like fantastic structs!

Thanks for this PR btw - it looks fantastic!

swharden:
> Some of these may be too far, and make updating to ScottPlot 5 unnecessarily difficult, but I think they're at least worth considering.

FYI in ScottPlot 5 I'll be keeping the basic API (users create a `Plot` and `AddSomething()` to it) so it'll be easy for old users to figure out how to use it, but I'm giving myself total liberty to choose any names/namespaces that seem best now that I have a few years of experience maintaining a charting library. 

Answering 100s of questions and closing 100s of issues gave me a lot of insight into what works and what users (and contributors) find confusing, so I'm looking forward to some of the changes! I hope an improved simpler API will make features more discoverable and the code more maintainable 🤞 

If you're interested in being more involved in the new API design, I'd be happy to work together on it!

PR #1964, opened 2022-07-17
ScottPlot5 Backend for Controls

bclehmann:
**Purpose:**
Creates a backend for controls on ScottPlot 5, and creates a new WpfPlot user control that makes use of it.


bclehmann:
This does nothing about handling keyboard events, this just aims for parity with the existing WinForms app.

bclehmann:
I don't really like how `onDragRelease` works. We have logic to tell if a mouseMove is also a mouseDrag, and then we call a separate event.

Perhaps we should have logic to tell if a mouseUp is also a mouseDragEnd?

bclehmann:
I also don't think it handles resizing yet?

bclehmann:
There's a little quirk here about what we should do if the user drags and then sends a MouseDown using another button, without releasing the button that caused the drag.

On this PR we let the user click while dragging with another button, both events go through, neither are cancelled or modified.

On ScottPlot 4 (may differ based on UserControl) it looks like we cancel the drag if it gets a MouseUp from another button. If  the user attempts to drag with multiple mouse buttons only one drag will go through according to the following precedence: Mouse1 > Mouse2 > Mouse3. It does not matter in which order the buttons are pressed. If the user drags with multiple mouse buttons and releases then the results are weird. It seemingly will apply the MouseUp and DragEnd for all mouse buttons (regardless of whether those buttons are still pressed). This is really strange because if the user drags with Mouse3 and then presses Mouse1, the zoom rectangle will not update. When they release Mouse1, it will zoom to where the zoom rectangle was when they first pressed Mouse1.

Suffice it to say, ScottPlot 4 doesn't really handle multiple mouse buttons at once. So I think the only correct options are to neither cancel or modify drag events when multiple buttons are pressed, or to stop dragging when another button is pressed. I think the latter may feel weird to the user and the former is already implemented.

bclehmann:
I think this is ready, unless you want me to go after keyboard interaction as well?

bclehmann:
I went back and forth about whether drag events should use the keys that were pressed when the mouse was pressed, or the keys that are currently pressed. The latter sometimes feels faster, but it often feels kinda janky, as pressing Ctrl will immediately snap the axis.

The former is closer to a traditional Ctrl+Drag, and it doesn't invite bugs or jank, so it's what I went with. This is (I believe) a departure from ScottPlot 4.

src/ScottPlot5/ScottPlot/Control/Key.cs
```diff
@@ -0,0 +1,13 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Text;
+
+namespace ScottPlot.Control
+{
+    public enum Key // Handling all possible keys is non-trivial and probably unnecessary
```
swharden, line 7: Let's not change the design now, but...

Rather than handling all keys I wonder if there's a good way to abstract this (rather than using an enum) so users can extend it themselves to add new or custom keys.

I use OriginLab (scientific software similar to MatLab) at work, and their zooming controls are nuts. Shift + X + drag to pan vertically 😝 

* https://www.originlab.com/doc/Origin-Help/Pan-GraphPage
* https://www.originlab.com/doc/Origin-Help/Zoom-GraphPage

src/ScottPlot5/ScottPlot/Pixel.cs
```diff
@@ -20,4 +20,14 @@ public override string ToString()
     {
         return $"Pixel: X={X}, Y={Y}";
     }
+
+    public static Pixel operator +(Pixel a, Pixel b)
```
swharden, line 24: Nice! This is exactly the spirit of these new primitive types, and I'm excited to build functionality like this out as ScottPlot 5 matures.

After merging this PR I'll move this to Common and add regression tests and XML docs. Let's not do that in this PR, because it's already complex enough as it is.

src/ScottPlot5/controls/ScottPlot.WPF/WpfPlot.xaml.cs
```diff
@@ -0,0 +1,160 @@
...
+using System.Windows.Controls;
+using System.Windows.Input;
+using System.Windows.Media;
+using ScottPlot.Control;
+using SkiaSharp;
+using SkiaSharp.Views.Desktop;
+using Key = ScottPlot.Control.Key;
+using MouseButton = ScottPlot.Control.MouseButton;
+
+namespace ScottPlot.WPF
```
swharden, line 14: Do you think there's utility in creating an `IPlotControl`? It could contain things like `Plot`, `MouseCoordinates`, `Backend`, `GetInputState()`, `OnPaintSurface()`, `GetScottPlotKey()`, and might make it a little easier to implement new controls.

Not sure how much shared logic there is though, but a `ScottPlotControlBase` with virtual functions could be another way to guide new controls into implementing the necessary functionality

Not required for this PR - just thinking out loud

swharden:
Thanks @bclehmann, this looks great! I'll merge it tomorrow afternoon after looking it over one more time 👍

swharden:
> On this PR we let the user click while dragging with another button, both events go through, neither are cancelled or modified.

I think this is fine for now. We can always refine it in the future as needed, and users who do weird things deserve to get weird results 😜

src/ScottPlot5/ScottPlot/Control/Key.cs
bclehmann: Simplest option is probably just to pass in a numeric ID (for example a scan code). That's a little low level, but I think if the issue is that we don't want to make a good abstraction, well then maybe we shouldn't cram the user into a bad one.

So then we do something like this:
```cs
public enum KnownKeys {
  Ctrl,
  Alt,
  Shift
};

public struct Key {
  public KnownKey? KnownKey { get; }
  public int ScanCode { get; }
  
  public Key(int scanCode) {
    ScanCode = scanCode;
    KnownKey = GetKeyFromScanCode(scanCode)
  }
}
```

FWIW Avalonia (and probably more) have lookup-tables for scancodes, if we wanted to provide something deeper in the future.

src/ScottPlot5/controls/ScottPlot.WPF/WpfPlot.xaml.cs
bclehmann: I don't know, `Backend`, `GetInputState`, `OnPaintSurface`, and `GetScottPlotKey` are all `private`, so they can't be in the interface. Not to mention they're implementation details, they probably shouldn't be relied on as part of an interface's contract.

As for `Plot`, `Refresh`, and `MouseCoordinates`, I don't think they're bad things to guarantee across all controls, but frankly there's no reason for a control to implement the interface except for peer pressure. An interface is a great way to make code indifferent to the implementation of something, but if someone is on WPF they _need_ the WPF implementation.

Although, an interface for the `Backend` may be useful for testing, but I think we need to decide what testing will look like first.

src/ScottPlot5/ScottPlot/Control/Backend.cs
```diff
@@ -0,0 +1,187 @@
...
+        public delegate void MouseMoveHandler(object sender, MouseMoveEventArgs e);
+        public delegate void MouseUpHandler(object sender, MouseUpEventArgs e);
+        public delegate void MouseDragHandler(object sender, MouseDragEventArgs e);
+        public delegate void MouseDragEndHandler(object sender, MouseDragEventArgs e);
+        public delegate void DoubleClickHandler(object sender, MouseDownEventArgs e);
+        public delegate void MouseWheelHandler(object sender, MouseWheelEventArgs e);
+        public delegate void KeyDownHandler(object sender, KeyDownEventArgs e);
+        public delegate void KeyUpHandler(object sender, KeyUpEventArgs e);
+
+        public event MouseDownHandler MouseDown = delegate { };
```
swharden, line 32: To differentiate this big list of ScottPlot events from the typical events controls may have, do you think we should name them things like `SPMouseDown`? Or group them all in a class? A class would actually make it easier to duplicate all these across controls 🤔

swharden: My bad - ignore that comment. I got confused and thought I was reading user control code (not backend code) 🤪

src/ScottPlot5/ScottPlot/Control/Backend.cs
```diff
@@ -0,0 +1,187 @@
...
+        private MouseDownEventArgs? CurrentMouse1Down = null;
+        private MouseDownEventArgs? CurrentMouse2Down = null;
+        private MouseDownEventArgs? CurrentMouse3Down = null;
+
+        private readonly HashSet<Key> CurrentlyPressedKeys = new();
+        private Pixel? LastMousePosition = null;
+
+        private readonly Plot plot;
+        private readonly object EventSender;
+        private readonly Action requestRender;
```
swharden, line 20: I'm a little unclear what this is to be used for. It's private but not called. Is it here so the backend can eventually request its own renders (e.g., after a delay)?

src/ScottPlot5/ScottPlot/Control/EventArgs/BaseEventArgs.cs
```diff
@@ -0,0 +1,11 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Text;
+
+namespace ScottPlot.Control.EventArgs
+{
+    public class BaseEventArgs : System.EventArgs
```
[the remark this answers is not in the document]
bclehmann: I don't think it precludes us from using `Base` as a suffix since that class isn't really intended for use in public APIs, at least not directly. Users aren't going to make a new event args, they may interact with (for example) `MouseDownEventArgs` if we open up the backend's event handlers at some point.

src/ScottPlot5/ScottPlot/Control/Backend.cs
bclehmann: It's here because the events handlers can request renders, but since the event handlers are set up in the constructor we don't actually need to save this in a field.

src/ScottPlot5/ScottPlot/Control/Backend.cs
```diff
@@ -42,6 +42,8 @@ public class Backend
         public event KeyDownHandler KeyDown = delegate { };
         public event KeyUpHandler KeyUp = delegate { };
 
+        public float MinimumDragDistance = 5;
```
bclehmann, line 45: This should be `const`. And possibly `private`.

src/ScottPlot5/ScottPlot/Control/Backend.cs
```diff
@@ -114,6 +114,10 @@ public void TriggerMouseMove(Pixel position)
                     {
                         TriggerMouseDrag(lastMouseDown, position, button);
                     }
+                    else if (Plot.ZoomRectangle.IsVisible)
+                    {
+                        Plot.MouseZoomRectangleClear(applyZoom: false);
```
bclehmann, line 119: I think this belongs more in `DefaultEventHandlers.OnMouseDown`, I'm trying to keep the logic in this class strictly to delegating to updating state and routing events wherever they ought to go. The actual handling of the events belongs in the event handlers.

src/ScottPlot5/ScottPlot/Control/Backend.cs
```diff
@@ -32,27 +32,52 @@ public class Backend<T> where T : IPlotControl
...
+            DoubleClick += (T sender, MouseDownEventArgs e) => interactions.DoubleClick(sender, e);
+            MouseWheel += (T sender, MouseWheelEventArgs e) => interactions.MouseWheel(sender, e);
+            KeyDown += (T sender, KeyDownEventArgs e) => interactions.KeyDown(sender, e);
+            KeyUp += (T sender, KeyUpEventArgs e) => interactions.KeyUp(sender, e);
+        }
+
+        /// <summary>
+        /// Clear all existing interaction event handlers and replace them with a custom set
+        /// </summary>
+        public void ReplaceInteractions(IPlotInteractions interactions)
```
bclehmann, line 70: Doing it this way makes me wonder if we should have `AppendInteractions`, `PrependInteractions`, etc. This is something that native events kinda do in a clunky way, compared to a list of observers which can be reordered to your heart's content.

I think it's fine (even preferable) to expect the user to use a decorator if they want to preserve some of the default behaviours, it's just if that's the expectation then why do we initially give them the default event handlers, and then replace them? Why not let them just pass in their own event handlers as a parameter? (I believe this is also what you did in an earlier commit?)

src/ScottPlot5/ScottPlot/Control/Backend.cs
swharden: I think this is resolved now using generic event handlers

src/ScottPlot5/ScottPlot/Control/EventArgs/BaseEventArgs.cs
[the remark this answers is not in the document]
swharden: This question will come up again: SignalPlotBase or BaseSignalPlot? We have some time to decide

src/ScottPlot5/ScottPlot/Control/Backend.cs
swharden: I think there's real utility for the user to disable this. Personally I don't like it because it makes the plot feel low low FPS when you move it in small circles, when really you're just getting near the mouse origin. 

Maybe there needs to be some logic to track total distance traveled or something? We can pick that up later.

src/ScottPlot5/ScottPlot/Control/Backend.cs
swharden: The reason I did this is because there was a bug where when you re-entered the dead zone the red box would still be drawn. I agree there is probably a better solution than this though.

I'll merge this PR but open a new issue to track this.

src/ScottPlot5/ScottPlot/Control/Backend.cs
swharden: I'll pick this topic up in a new issue 👍

Issue #1968, opened 2022-07-19
ScottPlot5: Improve user control API


swharden:
After merging #1964 the `IPlotControl` API and related code could still use some review and fine tuning.

* **Dead zone:** Let's figure out how to better handle the "dead zone" pixel area (where click-drag is not respected) and to ensure the zoom rectangle isn't drawn when returning back to the dead zone.
  * https://github.com/ScottPlot/ScottPlot/pull/1964#discussion_r923976640
  * https://github.com/ScottPlot/ScottPlot/pull/1964#discussion_r923977581

* **Custom interactions:** There's a new `IPlotInteractions` which lets users load a whole set of alternative plot bindings (which mouse buttons do what), but there's probably a more convenient API for this. Let's explore some options. 
  * https://github.com/ScottPlot/ScottPlot/pull/1964#pullrequestreview-1042809240

Overall, you can see I'm going for absolute simplicity when it comes to implementing user controls on new frameworks. Here's the entire WpfPlot control for ScottPlot 5:

https://github.com/ScottPlot/ScottPlot/blob/84ec8c9bd7617d54d7999fc79322d9dd0650b37f/src/ScottPlot5/controls/ScottPlot.WPF/WpfPlot.xaml.cs#L13-L127

BambOoxX:
Hi Scott, just a quick reaction to your comment, 
> to ensure the zoom rectangle isn't drawn when returning back to the dead zone.

Some other plotting libraries tend to provide a zoom rectangle that only zooms in horizontal or vertical directions if the mouse is dragged along the horizontal or vertical direction only (within a given tolerance). At the moment, I think this is something that `ScottPlot` does not allow, unless you lock the limits of the axis, but that's not very flexible.
This could be another nice improvement for v5.

bclehmann:
A good example of that is this jsfiddle in plotly: https://jsfiddle.net/Lb476dan/

I find that it switching between types of zooms can be a little fiddly and jarring. They also have a similar thing for only the X axis that I've seen in their financial charts: https://jsfiddle.net/2c8e7qnm/ (might have to scroll down, by default the preview window is pretty small on jsfiddle).

swharden:
> This could be another nice improvement for v5

It would be nice to make this easier! I agree with @bclehmann the zoom example demonstrated in plotly is a bit jarring. 

[...]

PR #1969, opened 2022-07-19
ScottPlot5: WinForms control

swharden:
This PR extends #1964 and adds an interactive Windows Forms control


swharden:
WinForms control is now working, but the backend API can continue to be refined (#1968)

PR #1970, opened 2022-07-20
Draggable marker: fix drag limits

xmln17:
This is my first PR so if I am doing somethign wrong please correct me. Today i noticed an bug while using the DraggableMarkerPlot with DragLimits. The limits for X-Axis works fine but for Y-Axis it does not work as intendet. The conditions are checking the X-Coordinate in all four cases, which is obviously wrong. As this is no big change i hope that my description meets the requirements for an PR.


swharden:
Thank you @xmln17, this looks great! The fix will be in the next release.

> This is my first PR

GitHub can be a little intimidating to get started with, but you did great! Congratulations! 🚀

> meets the requirements for an PR

This library has improved thanks to many small contributions from developers like you - thanks for the bug fix! 
* https://scottplot.net/changelog/
* https://scottplot.net/contributors/

PR #1979, opened 2022-07-24
ScottPlot5 controls: InputMap

swharden:
Controls now have an InputMap class which defines which mouse buttons and keyboard keys are connected to which actions.

Users who wish to achieve superior UI customization can still implement a custom IInteraction that the backend uses.

### DIFF
--- a/src/ScottPlot5/Sandbox/ScottPlot.Sandbox.WPF/MainWindow.xaml.cs
+++ b/src/ScottPlot5/Sandbox/ScottPlot.Sandbox.WPF/MainWindow.xaml.cs
@@ -25,6 +25,10 @@ namespace WPFDemo
             WpfPlot.Plot.AddScatter(Generate.Consecutive(N), Generate.Cos(N), Colors.Red);
             WpfPlot.Plot.AddScatter(Generate.Consecutive(N), Generate.Sin(N, 0.5), Colors.Green);
 
+            // swap left and right buttons
+            WpfPlot.ButtonMap.ClickDragPan = ScottPlot.Control.MouseButton.Right;
+            WpfPlot.ButtonMap.ClickDragZoom = ScottPlot.Control.MouseButton.Left;
+
             WpfPlot.MouseMove += WpfPlot_MouseMove;
         }
 
--- a/src/ScottPlot5/Sandbox/ScottPlot.Sandbox.WPF/MainWindow.xaml.cs
+++ b/src/ScottPlot5/Sandbox/ScottPlot.Sandbox.WPF/MainWindow.xaml.cs
@@ -26,8 +26,8 @@ namespace WPFDemo
             WpfPlot.Plot.AddScatter(Generate.Consecutive(N), Generate.Sin(N, 0.5), Colors.Green);
 
             // swap left and right buttons
-            WpfPlot.ButtonMap.ClickDragPan = ScottPlot.Control.MouseButton.Right;
-            WpfPlot.ButtonMap.ClickDragZoom = ScottPlot.Control.MouseButton.Left;
+            WpfPlot.InputMap.ClickDragPan = ScottPlot.Control.MouseButton.Right;
+            WpfPlot.InputMap.ClickDragZoom = ScottPlot.Control.MouseButton.Left;
 
             WpfPlot.MouseMove += WpfPlot_MouseMove;
         }
--- a/src/ScottPlot5/ScottPlot/Control/InputMap.cs
+++ b/src/ScottPlot5/ScottPlot/Control/InputMap.cs
@@ -1,0 +1,76 @@
+﻿namespace ScottPlot.Control;
+
+/// <summary>
+/// This class maps common plot actions with specific mouse and keyboard inputs
+/// </summary>
+public class InputMap
+{
+    public MouseButton? ClickDragPan = null;
+    public MouseButton? ClickDragZoom = null;
+    public MouseButton? ClickDragZoomRectangle = null;
+    public MouseButton? DoubleClick = null;
+    public MouseButton? ZoomIn = null;
+    public MouseButton? ZoomOut = null;
+
+    public Key? LockHorizontalAxis = null;
+    public Key? LockVerticalAxis = null;
+    public Key? PanZoomRectangle = null;
+
+    public bool IsLockedX(IEnumerable<Key> keys)
+    {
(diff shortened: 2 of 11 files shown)


src/ScottPlot5/Sandbox/ScottPlot.Sandbox.WPF/MainWindow.xaml.cs
```diff
@@ -25,6 +25,10 @@ public MainWindow()
             WpfPlot.Plot.AddScatter(Generate.Consecutive(N), Generate.Cos(N), Colors.Red);
             WpfPlot.Plot.AddScatter(Generate.Consecutive(N), Generate.Sin(N, 0.5), Colors.Green);
 
+            // swap left and right buttons
```
bclehmann, line 28: Not sure if you want to undo this?

src/ScottPlot5/ScottPlot/Control/InputMap.cs
```diff
@@ -0,0 +1,76 @@
...
+    public MouseButton? ClickDragZoomRectangle = null;
+    public MouseButton? DoubleClick = null;
+    public MouseButton? ZoomIn = null;
+    public MouseButton? ZoomOut = null;
+
+    public Key? LockHorizontalAxis = null;
+    public Key? LockVerticalAxis = null;
+    public Key? PanZoomRectangle = null;
+
+    public bool IsLockedX(IEnumerable<Key> keys)
```
bclehmann, line 19: This method (and `IsLockedY` and `IsZoomingRectangle`) sound like they're stateful. i.e. They're storing whether or not the axis should be locked or the rectangle should be zooming.

Perhaps it should be `ShouldLockX`, `ShouldLockY` and `ShouldZoomRectangle`?

src/ScottPlot5/ScottPlot/Control/Interactions/StandardInteractions.cs
```diff
@@ -18,27 +25,51 @@ public StandardInteractions(IPlotControl control)
...
+        }
+        else
+        {
+            return;
+        }
+    }
+
+    private void ZoomIn(Pixel pixel, IEnumerable<Key> keys)
     {
+        double xFrac = InputMap.IsLockedX(keys) ? 1 : ZoomInFraction;
```
bclehmann, line 46: This is nice

src/ScottPlot5/ScottPlot/Control/Interactions/StandardInteractions.cs
```diff
@@ -18,27 +25,51 @@ public StandardInteractions(IPlotControl control)
...
+        double yFrac = InputMap.IsLockedY(keys) ? 1 : ZoomOutFraction;
+
+        Plot.MouseZoom(xFrac, yFrac, pixel);
+        Control.Refresh();
     }
 
     public virtual void MouseUp(Pixel pixel, MouseButton button, bool endDrag)
     {
-        switch (button)
+        if (button == InputMap.ClickDragZoomRectangle)
```
bclehmann, line 64: Can this be `InputMap.IsZoomingRectangle`?

src/ScottPlot5/ScottPlot/Control/MouseButton.cs
```diff
@@ -11,5 +11,9 @@ public enum MouseButton
         Left,
         Right,
         Middle,
+        WheelUp,
```
bclehmann, line 14: I don't think I like considering mouse wheel movements to `MouseButton`s.

I know a `MouseWheelDirection` enum is a little verbose, but I think it's a better option.

bclehmann:
I also went back and forth on whether I like `InputMap` as a name or not, so far I've settled on apathy, but if you'd like to change it I think `InputBindings` or `InputOptions` are alright alternatives.

src/ScottPlot5/Sandbox/ScottPlot.Sandbox.WPF/MainWindow.xaml.cs
swharden: It's in there for now to demo how to use it. I'll keep it a little longer while I keep refining this backend. Thanks for pointing it out though!

src/ScottPlot5/ScottPlot/Control/InputMap.cs
swharden: Good point! I'll do this, and improve understanding using XML docs too 👍

src/ScottPlot5/ScottPlot/Control/Interactions/StandardInteractions.cs
swharden: It makes a little more sense when you see it in context

```cs
public class InputMap
{
    public MouseButton? ClickDragPan = null;
    public MouseButton? ClickDragZoom = null;
    public MouseButton? ClickDragZoomRectangle = null;
    public MouseButton? DoubleClick = null;
    public MouseButton? ZoomIn = null;
    public MouseButton? ZoomOut = null;

    public Key? LockHorizontalAxis = null;
    public Key? LockVerticalAxis = null;
    public Key? PanZoomRectangle = null;
```

It would be more verbose, but maybe this could be better communicated by renaming to

```cs
public MouseButton? ClickDragPanButton = null;
public MouseButton? ClickDragZoomButton = null;
public MouseButton? ClickDragZoomRectangleButton = null;
public MouseButton? DoubleClickButton = null;
public MouseButton? ZoomInButton = null;
public MouseButton? ZoomOutButton = null;

public Key? LockHorizontalAxisKey = null;
public Key? LockVerticalAxisKey = null;
public Key? PanZoomRectangleKey = null;
```

src/ScottPlot5/ScottPlot/Control/MouseButton.cs
swharden: The gotcha here is that a wheel event is being treated like a button press

https://github.com/ScottPlot/ScottPlot/blob/d79f1898eda169daf884077ef9c6fb1e0c46a269/src/ScottPlot5/ScottPlot/Control/Backend.cs#L97-L101

I think you're right though that wheel events should be distinct, and probably not passed into the MouseDown function.

However this gets redesigned, it should be easy to re-write mouse wheel events to do different things (like pan)

swharden: In #1982 I refactored this to be separate enumerations, `MouseButton` and `MouseWheelDirection`

Issue #1966, opened 2022-07-17
ScottPlot is on Discord


swharden:
ScottPlot is on Discord!

* https://ScottPlot.NET/discord

I'm starting to work on ScottPlot 5 every day (#1955) and I have been sharing progress, getting advice, and coordinating development efforts with @bclehmann in the ScottPlot Discord server. Anyone interested in ScottPlot is invited to join!

PR #1982, opened 2022-07-25
ScottPlot5: refactor backend

swharden:
The primary goal of this PR is to simplify custom handling of mouse events in the control backend


src/ScottPlot5/ScottPlot/Control/Backend.cs
```diff
@@ -9,107 +7,211 @@ namespace ScottPlot.Control
...
 
-        public void MouseUp(Pixel position, MouseButton button)
+        public virtual void MouseUp(Pixel position, MouseButton button)
         {
-            bool drag = MouseButtonStates.IsDragging(position);
+            bool drag = Mouse.IsDragging(position);
+
+            // TODO: when should the rectangle be cleared?
 
             if (drag)
```
bclehmann, line 58: I think we should keep `MouseDragEnd` as a separate function, that way we don't mix the interpretation of events (e.g. was this a drag, was this key held, etc) with acting upon them. If we discover (or introduce) a problem we have a lot more places to check when it can read state wherever it wants rather than in a single function call. Especially if we start supporting more complicated behaviour.

I'd also note that it would be much easier for users to customize behaviour if there's a distinct `MouseDragEnd` event to override, rather than them having to to override (and clobber) `MouseUp` and reimplement drag-detection. Conceptually I think `MouseUp` and `MouseDragEnd` are very different events which deserve distinct handlers, even if every `MouseDragEnd` has a `MouseUp`. Similar to how we have `MouseMove` and `MouseDrag`

src/ScottPlot5/ScottPlot/Control/Backend.cs
```diff
@@ -9,107 +7,211 @@ namespace ScottPlot.Control
...
+                }
+            }
+
+            Mouse.Up(button);
 
-            MouseButtonStates.Clear();
-            Interactions.MouseUp(position, button, drag);
+            if (button == InputBindings.DragZoomRectangleButton)
+            {
+                if (!drag)
```
bclehmann, line 71: This can probably be merged into the above `if` statements?

bclehmann (commented):
I already said in the discord that I don't like the axing of `IInteractions`, and that this current implementation brings things like drag detection and drag handling too close together, but there is a middle ground.

I assume that `IInteractions` was axed due to its verbosity, but we can still get most of the separation of concerns from the old model without much of the verbosity. For example, we could have `SendMouseUp` and `HandleMouseUp` be separate functions, so that its easy for users to modify behaviour and we keep things clean and prevent complexity from ballooning. 

After all, one of the main problems in the ScottPlot 4 backend was the constant checks to see if a mouse event was really a drag event, and whether it was really a drag event with the `Alt` key held down. And that code ought to be minimized, if it can't be minimized it should be centralized, and if it can't be centralized it should be homogenized, or else it will cause bugs and complexity.

swharden:
Hey @bclehmann I split this up like we discussed. I'm running out of steam today and stopped short of finishing everything: 
* I think more actions should be tracked
* The commander doesn't have an interface yet
* There's not a `Replace()` function

However, I'd love your input on the direction it's going. I'm going to merge this because it's getting pretty big, but we can refine this structure in a future PR.

Unless it's helpful to refer to specific line numbers, let's discuss next time you're free in Discord #1966 👍 

Thanks again for your help on this today! (and #1984)

swharden:
https://github.com/ScottPlot/ScottPlot/blob/87d48e7652791b262755f2ddf3562d68a4a36b49/src/ScottPlot5/ScottPlot/Control/Backend.cs#L3-L70

https://github.com/ScottPlot/ScottPlot/blob/87d48e7652791b262755f2ddf3562d68a4a36b49/src/ScottPlot5/ScottPlot/Control/PlotActionCommander.cs#L4-L125

https://github.com/ScottPlot/ScottPlot/blob/87d48e7652791b262755f2ddf3562d68a4a36b49/src/ScottPlot5/ScottPlot/Control/PlotActions.cs#L6-L120

Issue #1955, opened 2022-07-15
ScottPlot 5.0 alpha


swharden:
## ScottPlot 5 is being actively developed 🚀 
* ScottPlot 5 uses SkiaSharp instead of System.Drawing for improved performance and cross-platform support
* ⚠️ ScottPlot 5 is a work in progress and is not ready for use
* Contributions to ScottPlot 5 are not invited at this time
* I am working in `/src/ScottPlot5/` of the `main` branch so code and tests can be shared with ScottPlot 4

## ScottPlot 4 will continue to be supported ✔️
* ScottPlot 4 issues and bug fixes take priority over ScottPlot 5 development
* I will avoid implementing new features in ScottPlot 4 until ScottPlot 5 is released

## Join the Discussion on Discord!
* https://ScottPlot.NET/discord (all are welcome)

## Resources
* #1888
* #1036
* #1647

swharden:
## About The ScottPlot.Common Project

The goal of this project is to facilitate sharing logic and tests between ScottPlot 4 and ScottPlot 5

* The `ScottPlot.Common` contains code which is useful to all versions of ScottPlot 
  * (e.g., data generation, numeric conversion, etc)
  * It targets .NET Standard 2.0 and has no dependencies
  * It will probably not become its own NuGet package, but it could
  * objects are in the `ScottPlot` namespace (not `ScottPlot.Common`)
  * code is tested in a `ScottPlot.Common.Tests` project
* ScottPlot 5 will reference the _project_ so all types it contains are available
* ScottPlot 4 will reference _individual source files_ from this project 
  * Types like `ScottPlot.Color` are not imported (they would conflict with `System.Drawing.Color`)
  * New features (like the new `Generate` module, alternative to `DataGen`) developed for ScottPlot 5 can be used in new versions of ScottPlot 4

gy-ge:
Fantastic! Does this mean MAUI can be supported in ScottPlot 5?

swharden:
That's the plan!

Maui, WinUI, and Blazor are high priorities 🤞

swharden:
Let's move ScottPlot 5 to beta and publish a package ASAP #2245

PR #1990, opened 2022-07-30
ScottPlot5: use actions for UI logic

swharden:
This PR strives to simplify how UI bindings/actions can be customized.

#1955, #1966


bclehmann (approved):
Looks good.

About whether you remove `Backend`, in it's current form I don't think it achieves anything, it's just a class that forwards a bunch of things to `Interaction`. If you were to pull some things out of `Interaction` and put them in `Backend` it might make more sense. Depends whether you want to or not.

swharden:
> About whether you remove Backend, in it's current form I don't think it achieves anything, it's just a class that forwards a bunch of things to Interaction

I tried removing `Backend` and it looks pretty clean now.

Note that controls now have a function to allow interaction replacement. I'm on the fence whether this should be simplified to a property 🤔 

```cs
public void Replace(Interaction customInteraction)
```

PR #1991, opened 2022-07-30
ScottPlot 5: Eto Control

bclehmann:
Limitations:

- When targeting WPF (Eto.Platform.WPF) keyboard events don't work. Apparently `Control.Focus()` doesn't work on every platform so that might be the issue.
- Performance is a little poor for large windows on WinForms. It's ok, though not great, on WPF.


swharden:
I'm going to merge this now, but I think there's room for improvement: click-drag-pan freezes if you continuously move the mouse. This may be a limitation of `ImageView`, but forcing the window to redraw somehow would fix the remaining issue. I tried `this.Invalidate()` but it did not help

PR #2000, opened 2022-08-04
EtoPlot: Usse Drawable.OnPaint directly rather than setting ImageView.Image

bclehmann:
**Purpose:**
Eto allows us to render directly to the control, rather than setting an image that gets rendered. This is simpler and yields massive performance boosts. It also seems to be more consistent between WPF and WinForms


bclehmann:
As I noted in the discord this is likely also applicable to several other platforms. There are benefits from double-buffering when using image controls, but they shouldn't matter now that we already need to copy whole images.

PR #2001, opened 2022-08-05
Axis Line: Label Positioning

### LINKED ISSUE #1972: Axis line position labels should move out of the way instead of overlapping
## Feature Suggestion

**Feature description:** 
When using multiple axis lines, especially when draggable by the user, their position labels can end up overlapping each other. The suggested feature would be an option that could be turned on in the axis line that would cause its label to move out of the way when overlapping with another axis label.
**Code example:** 

```cs
plt.AddSignal(DataGen.Sin(51));

var vline1 = plt.AddVerticalLine(23);
vline1.LineWidth = 2;
vline1.PositionLabel = true;
vline1.PositionLabelBackground = vline.Color;
vline1.DragEnabled = true;
vline1.PositionLabelOverlap = false;

var vline2 = plt.AddVerticalLine(24);
vline2.LineWidth = 2;
vline2.PositionLabel = true;
vline2.PositionLabelBackground = vline.Color;
vline2.DragEnabled = true;
vline2.PositionLabelOverlap = false;
```
The resulting plot would have two labels sitting side-by-side instead of partially overlapped

swharden:
Adds properties to axis lines for customizing label positions. This allows labels to be moved out of the way to prevent overlapping. See code in #1972 for details.

overlap3

Resolves #1972

### DIFF
--- a/src/ScottPlot4/ScottPlot.Sandbox/WinFormsApp/Form1.Designer.cs
+++ b/src/ScottPlot4/ScottPlot.Sandbox/WinFormsApp/Form1.Designer.cs
@@ -29,33 +29,50 @@
         private void InitializeComponent()
         {
             this.formsPlot1 = new ScottPlot.FormsPlot();
+            this.checkBox1 = new System.Windows.Forms.CheckBox();
             this.SuspendLayout();
             // 
             // formsPlot1
             // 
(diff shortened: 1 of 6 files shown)


swharden:
tested/working on multi-axis plots too
overlap4

PR #2004, opened 2022-08-06
ScottPlot5: grid line and tick measurement optimizations

bclehmann:
**Purpose:**
Ticks are measured (recursively) rather than guessing at their length

This also brings a significant performance uplift by drawing the gridlines as a single path, rather than in multiple operations. Weirdly drawing the vertical gridlines is still about 30% of our time (with the window fullscreened)? There are a lot of gridlines I suppose, and if the plottable gets more complicated I would expect this to change.

image

Note that with the window in fullscreen you'll have far more gridlines than normal.


swharden:
For context, I'm throwing these things together to push the ScottPlot5 API toward something that works, then will come back and improve memory and performance. Lots of these classes (e.g., grid and tick system) are placeholders just to get this thing off the ground, and I intend to rewrite them when all of the main components are working. It may not be worth spending much time optimizing these parts of the code base until the whole thing stabilizes a bit more.

Maybe this is a good reminder that a good design would put common drawing tasks in methods of a static class, then future performance optimizations can be isolated there, and those small code optimizations will improve many functions across the whole code base.

## DrawLine vs. DrawPath performance

I think DrawLine is hardware-accelerated but DrawPath is not, so relative performance by differ by platform

## Real String Measurement

This gets really thorny because proper support requires supporting font families, sizes, bold, etc., and that's a lot of baggage to carry around these placeholder classes. I'll merge this now because it's an improvement, but I'd probably recommend not investing much time into improving the tick/axis/layout system right now since it's still being so actively worked on at a higher API level

Adding new plot types or static methods for the Generate module may be safer sites for contribution over the next couple weeks until this thing stabilizes a bit more

PR #2009, opened 2022-08-06
ScottPlot 5: Palettes

bclehmann:
```cs
avaPlot.Plot.AddScatter(Generate.Consecutive(N), Generate.Sin(N));
avaPlot.Plot.AddScatter(Generate.Consecutive(N), Generate.Cos(N));
avaPlot.Plot.AddScatter(Generate.Consecutive(N), Generate.Sin(N, 0.5));
avaPlot.Plot.Add(DebugPoint);
```

image

```cs
avaPlot.Plot.Palette = new ScottPlot.Palettes.Tsitsulin();

avaPlot.Plot.AddScatter(Generate.Consecutive(N), Generate.Sin(N));
avaPlot.Plot.AddScatter(Generate.Consecutive(N), Generate.Cos(N));
avaPlot.Plot.AddScatter(Generate.Consecutive(N), Generate.Sin(N, 0.5));
avaPlot.Plot.Add(DebugPoint);
```

image

PR #2028, opened 2022-08-10
Common: refactor shared source files

swharden:
This PR dissolves the ScottPlot.Common package and instead tactically imports specific .cs files directly from ScottPlot4 into ScottPlot5.

ScottPlot5 now uses all ScottPlot4's palettes.

Extends recent work:
* #2027 
* #2018
* #2009
* #1967
* #1963
* #1956


swharden:
this comment is for testing the Discord web hook

PR #2030, opened 2022-08-10
Use Monitor for RenderLock

anprevost:
**Purpose:**
The current way of handling multithreading is not thread safe.
If one thread change the value of `IsRenderLocked` from `false` to `true`, there is a chance that another thread reading the same field will see `false`. Therefore, a Render might occur even if `RenderLock()` has been previously called.

### DIFF
--- a/src/ScottPlot4/ScottPlot.Tests/Concurrency/ConcurrencyTests.cs
+++ b/src/ScottPlot4/ScottPlot.Tests/Concurrency/ConcurrencyTests.cs
@@ -27,11 +27,9 @@ namespace ScottPlotTests.Concurrency
                 plt.AddSignal(new double[] { 1, 2, 3 });
                 plt.RenderUnlock();
             }
-            Debug.WriteLine("Modification thread shutting down...");
-            Thread.Sleep(100);
         }
 
-        // This test takes 5 seconds to complete so it is not run routinely
+        // This test takes a long time to run so they are not performed routinely
         //[Test]
         public void Test_Concurrent_ModifyPlottablesWhileRendering()
         {
@@ -44,7 +42,8 @@ namespace ScottPlotTests.Concurrency
 
             // render repeatedly while the plottalbles are being modified
             var sw = Stopwatch.StartNew();
-            while (sw.ElapsedMilliseconds < 5000)
+            int runTimeSeconds = 30;
+            while (sw.ElapsedMilliseconds < runTimeSeconds * 1000)
             {
                 plt.Render(bmp);
             }
--- a/src/ScottPlot4/ScottPlot/Plot/Plot.Render.cs
+++ b/src/ScottPlot4/ScottPlot/Plot/Plot.Render.cs
@@ -4,6 +4,7 @@ using System.Drawing;
 using System.Drawing.Imaging;
 using System.IO;
 using System.Linq;
+using System.Threading;
 
 namespace ScottPlot
 {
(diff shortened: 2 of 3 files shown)


swharden:
Hi @anprevost, thank you for pointing this out!

I'll take a closer look and merge this in tonight 👍 

## Demonstrating the Problem

Do you have any code that can help me efficiently demonstrate this problem? It would be helpful to know if/how you encountered this issue.

## Changes in this PR

Summarizing what this PR does, a reference type is created to serve as a lock object

```cs
private readonly object _renderLocker = new();
```

Then `Render()` locks it for the duration of the render loop with a [lock statement](https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/statements/lock).

`RenderLock()` and `RenderUnlock()` call [System.Threading.Monitor](https://docs.microsoft.com/en-us/dotnet/api/system.threading.monitor?view=net-6.0)'s [Enter()](https://docs.microsoft.com/en-us/dotnet/api/system.threading.monitor.enter?view=net-6.0) and [Exit()](https://docs.microsoft.com/en-us/dotnet/api/system.threading.monitor.exit?view=net-6.0) methods to hold/release an exclusive lock on the object.

anprevost:
## Demonstrating the Problem
You can use `ScottPlotTests.Concurrency.ConcurrencyTests` just remove the `Thread.Sleep(100)` inside `ContinuouslyModifyPlottables`. Increase the running time from 5 seconds to 30 and it should be enough to provoke an exception. `System.InvalidOperationException: Collection was modified; enumeration operation may not execute`.

It should look like that:
```cs
class ConcurrencyTests
    {
        public static bool ContinueModifyingPlottables = true;
        public static void ContinuouslyModifyPlottables(object data)
        {
            ScottPlot.Plot plt = (ScottPlot.Plot)data;
            while (ContinueModifyingPlottables)
            {
                plt.RenderLock();
                plt.Clear();
                plt.AddSignal(new double[] { 1, 2, 3 });
                plt.RenderUnlock();
            }
            Debug.WriteLine("Modification thread shutting down...");
        }

        // This test takes 5 seconds to complete so it is not run routinely
        //[Test]
        public void Test_Concurrent_ModifyPlottablesWhileRendering()
        {
            var bmp = new System.Drawing.Bitmap(300, 200);

            // update plottables continuously in the background
            Thread updateThread = new Thread(new ParameterizedThreadStart(ContinuouslyModifyPlottables));
            updateThread.Start(plt);

            // render repeatedly while the plottalbles are being modified
            var sw = Stopwatch.StartNew();
            while (sw.ElapsedMilliseconds < 30000)
            {
                plt.Render(bmp);
            }

            ContinueModifyingPlottables = false;
            updateThread.Join();

            Console.WriteLine("Test complete");
            Assert.Pass();
        }
    }
```

## Changes in this PR
It's exactly as you described.
About the note using RenderLock() and RenderUnlock() from within Render(). It would require to use a Try/Finally , but apart from that it's exactly the same.
```cs
        public Bitmap Render(Bitmap bmp, bool lowQuality = false, double scale = 1.0)
        {
            RenderLock();
            try
            {
                ...
                return bmp;
            }
            finally
            {
                RenderUnlock();
            }
        }
```

alexandrekow:
Hi @swharden,
Thanks for merging the pull request so quickly. 👍 
Do you have any idea when version **4.1.53** will be published on NuGet?
We are waiting for it as we prefer referencing the package directly in our codebase. 🙏 
Best regards,

swharden:
> Do you have any idea when version 4.1.53 will be published on NuGet?

It should be published tonight! There were some big changes related to the ScottPlot4/5 transition that held this one back longer than typical 👍

Issue #2040, opened 2022-08-13
Scatter: opt-in feature to support plotting data with gaps using NaN


swharden:
Currently, Scatter plots may not contain `double.NaN`. This maximizes performance because:
* NaN-checking every data point when getting axis limits is costly
* NaN-checking every data point at each render is costly
* Drawing a single line is more performant than drawing many broken lines

However, many users are confused by this behavior. Also, many users with to create scatter plots with missing data appearing as a broken line (a task which is difficult to do with the current implementation).
* #1994
* #546
* #1161
* #976
* #947
* #601
* #553
* #618
* #396
* #203
* https://scottplot.net/faq/data-gaps/
* https://scottplot.net/faq/nan/

This issue will explore whether it is possible to add NaN support for plotting data with gaps as an opt-in feature.

swharden:
I'm adding an `OnNaN` field to `ScatterPlot` which customizes this behavior. The default action is `Throw`, but `Ignore` skips NaN points and `Gap` renders the line as multiple lines with gaps where data is NaN.

![image](https://user-images.githubusercontent.com/4165489/184508071-a5042bf9-6f5a-43a4-ab2b-20d30dfe8f92.png)

![image](https://user-images.githubusercontent.com/4165489/184508104-ee824839-0cbc-430f-b71a-d60a46be7faf.png)

Issue #1161, opened 2021-07-20
ScottPlot not able to handle `NaN` data points


EmanuelFeru:
**ScottPlot Version:** 4.1.16
**Operating System:** Win10
**Application Type:** WinForms

**Bug Description:** At the moment it appears that ScottPlot is not able to handle data in which some data points are `NaN`. Expected behavior is that Scottplot will not display those points.

I have seen #546 to enable the diagnostic for every render. Is there a way to disable only the diagnostic for `NaN`? So, that the data is still displayed with missing points. I am using `AddSignalConst(..)` to plot my data.
If ScottPlot needs continuos line of data for speed, then is ok and I will use the Workaround below.

**Workaroung:** For now I check for `NaN` in my data and set those points to `0`. However, I would like to keep those `NaN` because it could be a question mark for the user to see this "anomaly" and take action to fix the system that generates the data.

Screenshoot with the error:
![image](https://user-images.githubusercontent.com/24780745/126298484-a262e619-d1ac-43ee-83a0-2a1f6fb23933.png)

Call stack:
![image](https://user-images.githubusercontent.com/24780745/126299626-a1d4733f-665c-4f6d-ae77-8882212d8795.png)

StendProg:
Hi @EmanuelFeru,

There is no way to work with NaN in a `Signals` (this will require a bunch of additional checks and kill all performance), and I think that it is not needed. The best we can do is throw out the exeption when such data is found in diagnostic mode, and this has already been implemented.

As for your problem, you can solve it using `PlottableSignalXYConst`, by simply throwing out points containing NaN:
```
var XYPairs = YData.Select( (y, i) => (i, y)).Where( p => !Double.IsNaN(p.y));
var xs = XYPairs.Select( p => p.i).ToArray();
var ys = XYPairs.Select( p => p.y).ToArray();
plt.AddSignalXYConst(xs, ys);
```
Of course you will need to add the sampling rate and offset to the index (i) to X coordinate projection, but I think you get the point.

EmanuelFeru:
Ok, I understand the background. Then we can close this one. Thank you.

greatawesome:
There is some utility to handling NaN rather than tossing points out. In ZedGraph, using NaN as a data point causes a break/gap in the line, whereas just tossing out the point would mean you connect the previous and next points when drawing the line. Having a gap is a nice visible way to see that a data point is missing.

swharden:
[...]

Issue #976, opened 2021-04-14
Scatter plot - support for plotting data with gaps?


swharden:
This is an extremely common task/question, and was most recently brought-up in #947. There isn't a great way to do this with ScottPlot right now. This topic has its own FAQ page https://swharden.com/scottplot/faq/data-gaps

![image](https://user-images.githubusercontent.com/4165489/114712148-c292ed80-9cfd-11eb-995e-00477d91da9d.png)

I know we go back and forth on how to handle `double.NaN`, but perhaps scatter plots could be modified in some way to recognize when lines start and stop (separated by `NaN`), and call `System.Drawing.DrawLines()` multiple times. Implementing this properly may also solve other problems related to `NaN`, and be an alternative to adjusting min and max render index 🤔

jsr1611:
#985

swharden:
I'm going to close this since I'm not actively working on it, but it's triaged high priority so I hope to support this soon

Issue #618, opened 2020-11-18
How can I NOT draw lines between time gaps in PlotSignalXY


Wyrus:
(Or maybe a different plot type would make more sense - suggestions?)

I am pulling data from a table to be plotted. Each record has a timestamp and associated values. I am able to successful pull this into a PlotSignalXY and have been having all sorts of fun with all the various ScottPlot options - well done!

However, in some cases, I'll have a jump in record times. The plot does draw everything correctly, and the time axis is behaving correctly. However, is there some way to NOT connect data points that are more than X seconds apart?
For instance, if there's a gap of 1 hour in the data, can that appear as a gap in the plotted line?

If not, do you have any guidance as to where to start my search if I wanted to attempt a solution?

Thanks!

swharden:
Hey @Wyrus, 

This is a somewhat common question, so I'll add it to the FAQ (#572) so it's a little easier to figure out.

The short answer is that right now scatter plots and signal plots can only plot a single connected line. Values like `double.PositiveInfinity` and `doube.NaN` are not supported. There is active work (#578) which seeks to detect these invalid values and throw exceptions or write error messages to make this more obvious.

If you have a "disconnected" dataset, the best option is to determine individual lines on your end, then place multiple scatter plots as necessary. If you are in a graphical environment where the data changes a lot, every time there is a new set of data you can call `plt.Clear()` and plot everything again. It's not the most satisfying answer, but it's the best solution given the current state of the project!

Let me know if you have other questions or hit any snags along the way 👍 
Scott

johnfoll:
> 

I figured out already that I could do this. However, doing this causes duplicates to show up in the Legend. Every time the same Label is plotted it adds another instance of the Label to the Legend.

swharden:
Following-up, anyone interested in making scatter plots with gaps can now do so:

* https://scottplot.net/faq/data-gaps/
* https://scottplot.net/cookbook/4.1/category/plottable-scatter-plot/#nan-values-break-the-line

![image](https://user-images.githubusercontent.com/4165489/187461598-e54172c5-0980-4856-a973-aada3fe0b07c.png)

johnfoll:
Thanks! I will try to use double.NaN

PR #2041, opened 2022-08-13
Scatter.OnNaN: customizes NaN behavior

swharden:
This PR adds an `OnNaN` field to `ScatterPlot` which customizes this behavior. 

The default action is `Throw`, but `Ignore` skips NaN points and `Gap` renders the line as multiple lines with gaps where data is NaN.

Related: 
* #203
* #396
* #546 
* #618
* #976
* #1161

Resolves:
* #2040

image

image

### DIFF
--- a/src/ScottPlot4/ScottPlot.Tests/PlotTypes/Scatter.cs
+++ b/src/ScottPlot4/ScottPlot.Tests/PlotTypes/Scatter.cs
@@ -121,5 +121,67 @@ namespace ScottPlotTests.PlotTypes
 
             TestTools.SaveFig(plt);
         }
+
+
+        [Test]
+        public void Test_Scatter_AllNan()
+        {
+            double[] xs = ScottPlot.DataGen.Full(51, double.NaN);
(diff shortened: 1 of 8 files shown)


src/ScottPlot4/ScottPlot.Tests/PlotTypes/Scatter.cs
```diff
@@ -5,6 +5,7 @@
 using System.Text;
 using System.Threading.Tasks;
 using ScottPlot;
+using System.Drawing;
```
bclehmann, line 8: I would recommend a test for plots where once NaNs are removed there are segments of length 1 or of length 0. Judging from the code I don't think these should cause a problem (there's a check in `DrawLines` to make sure the array has length > 1), but this is the sort of thing that could break were the implementation to change.

bclehmann: I say because I have done something similar where I was splitting on null, and not handling segments that were too small caused some shenanigans that often regressed when we changed how we handled the splitted (splat? or splut?) data.

PR #2047, opened 2022-08-16
2022 08 16

### LINKED ISSUE #2046: Palette: capitalization inconsistency and enum discussion
## Feature Suggestion

The recent palette refactor is great. It is almost a complete replacement for the wrapper I had in place so that users could select a palette. All that is missing is an enum to chose palettes from.

**Code example**

I took the [public Palettes from here](https://github.com/ScottPlot/ScottPlot/blob/fff45bd0a020cf55e1523f98307f7f980e91e872/src/ScottPlot4/ScottPlot/Palette.cs#L14) and used regex to make the enum.

Capture

Users can select their palette with a dropdown and the enum value gets saved/loaded from settings

```cs
private PaletteHost.Palettes _Palette = (PaletteHost.Palettes)Properties.Settings.Default.SEYR_Palette;
[
    Category("User Parameters"),
    Description("Choose from ScottPlot palettes"),
    DisplayName("Plot Window Palette")
]
public PaletteHost.Palettes Palette 
{ 
    get => _Palette;
    set
    {
        _Palette = value;
        Properties.Settings.Default.SEYR_Palette = (int)_Palette;
        Properties.Settings.Default.Save();
    }
}
```

**Mini-bug**

The name matching method in my screenshot works (might be a performance hit, but I haven't noticed so far) - but, I did notice that [SnowStorm != Snowstorm](https://github.com/ScottPlot/ScottPlot/blob/fff45bd0a020cf55e1523f98307f7f980e91e872/src/ScottPlot4/ScottPlot/Palette.cs#L29). So, I just changed it to a lowercase `s` in the enum.

Similarly, spaces get in the way with `Colorblind Friendly`, but this works:
`IPalette thisPalette = ScottPlotPalettes.Where(x => x.Name.Replace(" ","") == palette.ToString()).First();`

**Thoughts**

I know this is all driven by old WinForms stuff, so no big deal if this is not addressed. I have been moving towards newer frameworks and SkiaSharp, so I am looking forward to ScottPlot 5!

swharden:
* Palettes
  * Snowstorm → SnowStorm resolves #2046
  * improved palette names
* ScottPlot5
  * improved layout system
  * experimental data management system for plottables


src/ScottPlot5/ScottPlot/DataSource/XYArrays.cs
```diff
@@ -0,0 +1,50 @@
...
+/// Changes made to the contents of the arrays will appear when the plot is rendered.
+/// </summary>
+public class XYArrays : IDataSource
+{
+    private readonly double[] Xs;
+    private readonly double[] Ys;
+
+    public int Count => Xs.Length;
+
+    public XYArrays(double[] xs, double[] ys)
```
bclehmann, line 14: Should probably also have a constructor that takes in a `Coordinate[]`

src/ScottPlot5/ScottPlot/DataSource/IDataSource.cs
```diff
@@ -0,0 +1,10 @@
+﻿namespace ScottPlot.DataSource;
+
+public interface IDataSource
```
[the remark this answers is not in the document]
swharden: Good point! I'll call it `ICoordinateSource`

src/ScottPlot5/ScottPlot/DataSource/IDataSource.cs
```diff
@@ -0,0 +1,10 @@
...
+{
+    int Count { get; }
+    Coordinates this[int index] { get; set; }
```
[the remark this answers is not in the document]
swharden: Nope, gone 👍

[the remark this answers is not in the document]
swharden: Great suggestion! Done

src/ScottPlot5/ScottPlot/DataSource/SignalArray.cs
```diff
@@ -0,0 +1,65 @@
+﻿namespace ScottPlot.DataSource;
+
+/// <summary>
+/// This data source has evenly-spaced Xs and Ys stored in a fixed-length array.
+/// Changes made to the contents of the array will appear when the plot is rendered.
+/// </summary>
+public class SignalArray : IDataSource, ISignalSource
```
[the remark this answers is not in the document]
swharden: Okay, killing two birds with one stone, this exposes `Ys` now

```cs
IReadOnlyList<double> Ys { get; }
```

src/ScottPlot5/ScottPlot/DataSource/SignalArray.cs
```diff
@@ -0,0 +1,65 @@
...
+
+    public double GetX(int index)
+    {
+        return (index * Period) + XOffset;
+    }
+
+    public Coordinates this[int index]
+    {
+        get => new(GetX(index), Ys[index]);
+        set { Ys[index] = value.Y; }
```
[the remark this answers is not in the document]
swharden: no, but also this seems like a weird way to iterate over signal values... I'm thinking of removing `List<Coordinates>` because it's pretty inefficient to loop this way. 

Maybe we should totally break apart the two main types of data sources: scatter-type and signal-type

I'll keep going and see how this evolves

src/ScottPlot5/ScottPlot/DataSource/XYArrays.cs
swharden: I think this would require copying values into Xs and Ys, at which point the link is broken for users to modify values in the original arrays.

Probably the fix here is a `CoordinateArray` class

src/ScottPlot5/ScottPlot/DataSource/SignalList.cs
```diff
@@ -0,0 +1,100 @@
...
+{
+    public double Period { get; set; }
+    public double XOffset { get; set; }
+    public List<double> Ys { get; }
+
+    public int Count => Ys.Count;
+
+    double IReadOnlyList<double>.this[int index] => Ys[index];
+
+    public SignalList(List<double> ys, double period)
```
bclehmann, line 17: I think this should be `IList<double> ys`. The current signature disallows arrays.

src/ScottPlot5/ScottPlot/DataSource/SignalList.cs
```diff
@@ -0,0 +1,100 @@
+﻿namespace ScottPlot.DataSource;
+
+/// <summary>
+/// This data source has evenly-spaced Xs and Ys stored in a fixed-length array.
```
bclehmann, line 4: This comment is incorrect

src/ScottPlot5/ScottPlot/DataSource/SignalList.cs
bclehmann: Oh I see, we have a `SignalArrary` and a `SignalList`. I'm not sure why? Is the only difference the type of the backing collection? If there is a larger reason why they're separate, it may still be worth generalizing the code into something like this:

```cs
public abstract class SignalCollection<T> {}
public class SignalList : SignalCollection<List<double>> {}
public class SignalArray : SignalCollection<double[]> {}
```

src/ScottPlot5/ScottPlot/DataSource/XYLists.cs
```diff
@@ -0,0 +1,62 @@
+﻿namespace ScottPlot.DataSource;
+
+/// <summary>
+/// This data source has Xs and Ys defined in fixed-length arrays.
+/// Changes made to the contents of the arrays will appear when the plot is rendered.
+/// </summary>
+public class XYLists : ICoordinateSource
```
bclehmann, line 7: Similar to `SignalList`, I don't know why this is separate from `XYArray`

src/ScottPlot5/ScottPlot/DataSource/XYArrays.cs
```diff
@@ -4,7 +4,7 @@
 /// This data source has Xs and Ys defined in fixed-length arrays.
 /// Changes made to the contents of the arrays will appear when the plot is rendered.
 /// </summary>
-public class XYArrays : IDataSource
+public class XYArrays : ICoordinateSource
```
bclehmann, line 7: Weird name, perhaps `CoordinateArray` would be better?

swharden: For the interfaces I went with `IScatterSource` and `ISignalSource`

Example scatter types are
* `ScatterCoordinateArray`
* `ScatterCoordinateList`
* `ScatterXYArrays`
* `ScatterXYLists`

src/ScottPlot5/ScottPlot/DataSource/SignalList.cs
swharden: > we have a SignalArrary and a SignalList. I'm not sure why?

This is my line of thinking...

* Users who want to do special stuff instantiate a concrete class of their choice and pass it into a plottable
  * If users want to plot an existing array in memory without copying, they can
  * If users want to plot an existing list they can add to, they can

* Plottables only access data via the interface. The interface describes data as read only.

swharden:
I'm going to merge this in for today but it'll probably evolve over the next few days.

Where it landed today:
* `IScatterSource` and `ISignalSource` define functionality that **_plottables_** use to render data
* The concrete classes can be whatever. They don't have to be consistent, and users can make their own if they want
* The concrete classes no longer inherit `IReadOnlyList<T>` or have/need iterators
* Avoiding the iterators means users can pass in their own arrays/lists and continue to use them

PR #2053, opened 2022-08-17
GDI: do not crop when frameless

### LINKED ISSUE #2051: AddHorizontalSpan: off by 1px on frameless plots
There's a 1px white upper border when using horizontal axis spans

I suspect this is typically covered-up when plots have black spines around the frame

```cs
ScottPlot.Plot plt = new(pbGraph.Width, pbGraph.Height);
plt.Frameless();
plt.AddSignal(ratioImage.GreenData.AverageByColumn(), 1, Color.Green);
plt.AddSignal(ratioImage.RedData.AverageByColumn(), 1, Color.Red);
plt.AddHorizontalSpan(structure.Min, structure.Max, Color.FromArgb(20, Color.Blue));
pbGraph.Image = plt.GetBitmap();
```
image

swharden:
Cropping results in anti-aliasing artifacts near the edges of the bitmap

resolves #2051

### DIFF
--- a/src/ScottPlot4/ScottPlot/Drawing/GDI.cs
+++ b/src/ScottPlot4/ScottPlot/Drawing/GDI.cs
@@ -159,9 +159,9 @@ namespace ScottPlot.Drawing
         {
             Graphics gfx = Graphics(bmp, lowQuality, dims.ScaleFactor);
 
-            bool dataIsSmallerThanFigure = dims.DataWidth < dims.Width && dims.DataHeight < dims.Height;
+            bool isFrameless = (dims.DataWidth == dims.Width) && (dims.DataHeight == dims.Height);
 
-            if (dataIsSmallerThanFigure && clipToDataArea)
+            if (clipToDataArea && !isFrameless)
             {
                 /* These dimensions are withdrawn by 1 pixel to leave room for a 1px wide data frame.
                  * Rounding is intended to exactly match rounding used when frame placement is determined.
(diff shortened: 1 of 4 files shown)


src/ScottPlot4/ScottPlot/Drawing/GDI.cs
```diff
@@ -159,7 +159,9 @@ public static System.Drawing.Graphics Graphics(Bitmap bmp, PlotDimensions dims,
         {
             Graphics gfx = Graphics(bmp, lowQuality, dims.ScaleFactor);
 
-            if (clipToDataArea)
+            bool dataIsSmallerThanFigure = dims.DataWidth < dims.Width && dims.DataHeight < dims.Height;
```
bclehmann, line 162: Should this not be `dims.DataWidth < dims.Width || dims.DataHeight < dims.Height`?

We need to clip if at least one dimension is smaller, not just if both are smaller.

swharden: Thank you for the careful review @bclehmann! 

I updated the code to more clearly communicate what the intent was (only skip clipping for frameless plots)

swharden:
For added context, most of what we're fighting here is anti-aliasing errors near edges.

The clipping area is automatically expanded by 1 pixel to reduce these artifacts, and typically the "blended" edges are covered by the black line that outlines the data area

https://github.com/ScottPlot/ScottPlot/blob/0c84190f424fcb39e3868326272df4acb7f97aa6/src/ScottPlot4/ScottPlot/Drawing/GDI.cs#L166-L173

PR #2054, opened 2022-08-18
ScottPlot5: basic Signal plot implementation

swharden:
not optimized for speed - mostly just exploring the API, interfaces, and primitives...


swharden:
sig

PR #2057, opened 2022-08-19
ScottPlot5: refactor IAxis, IPlottable, and ITickGenerator

swharden:
Plottables no longer store IXAxis and IYAxis and get passed-in a DataRect through Render() stack... Instead, Plottables now store a single IAxis2D (which stores both axes but also DataRect) and provides simpler methods for pixel/unit conversion inside Render().


BambOoxX:
Hey Scott, about this `IAxis2D` is it a subtle reference to the fact that you count on implementing `IAxis3D` at some point ? :)

swharden:
haha I can see how it gave that impression, but no... an IAxis2D contained two IAxis1D objects (X and Y) and a rectangle (the region of the image to draw on) and contained logic for translating between coordinate and pixel space

Maybe one day we will get to making 3d plots, but that seems to be a good distance away

BambOoxX:
That's unfortunate, but expected. It seems it is quite an open space in the .net ecosystem. All the 3d plotting package I've found seem to be closed source and payware.

PR #2067, opened 2022-08-25
RadarPlot: Add `Smooth` parameter

theelderwand:
**Purpose:**
This pull request adds a new property Smooth (default false) to RadarPlot. When enabled the radar plot will render as a closed curve instead of a polygon.

image

```cs
// You can insert code examples like this
var radar = plot.AddRadar(values);
radar.Smooth = true;
plot.Render();


swharden:
This looks fantastic @theelderwand! Thanks for your contribution 🚀

PR #2079, opened 2022-08-30
VectorFieldList

### LINKED ISSUE #2064: VectorField: support for arbitrary vector positions
## Feature Suggestion

**Overload VectorField() signature to allow for a List of vectors** 

Current VectorField signature is:

```cs
public VectorField(
    Vector2[,] vectors, 
    double[] xs, 
    double[] ys, 
    Colormap colormap, 
    double scaleFactor, 
    Color defaultColor)
```

which assumes that data are arrayed in a uniformly spaced 2D array, i.e., *Vector2[,]*.

We are working with more or less randomly positioned vector data, which converts into a very sparse Vector2[,] matrix, which the current VectorField() function cannot process.  Being able to send a 1D array or <List> of doubleXY vectors (see code below) to the VectorField() function would allow the plotting of randomly positioned vectors.  The overloaded VectorField() signature would look like this:

```cs
public VectorField(
    doubleXY[] vectors, 
    double[] xs, 
    double[] ys, 
    Colormap colormap, 
    double scaleFactor, 
    Color defaultColor)
```

where; the "vectors", "xs" and "ys" arrays are linked lists of vectors, xPositions and yPositions respectively; and *doubleXY* is defined as follows:

### class doubleXY
```cs
public class doubleXY
{
    public double X { get; set; }
    public double Y { get; set; }

    public doubleXY()
    {
        X = 0;
        Y = 0;
    }

    public doubleXY(double x, double y)
    {
        X = x;
        Y = y;
    }

    public doubleXY(string x, string y)
    {
        X = 0;
        Y = 0;
        double tempX;
        if (double.TryParse(x, out tempX))
        {
            X = tempX;
        }
        double tempY;
        if (double.TryParse(y, out tempY))
        {
            Y = tempY;
        }
    }

    public override string ToString()
    {
        return String.Format("X: {0}, Y:{1}", X, Y);
    }

    public override bool Equals(Object obj)
    {
        //Check for null and compare run-time types.
        if ((obj == null) || !this.GetType().Equals(obj.GetType()))
        {
            return false;
        }
        else
        {
            doubleXY p = (doubleXY)obj;
            return (X == p.X) && (Y == p.Y);
        }
    }
}
```

swharden:
The goal of this PR is to create a new plot type which makes it simple to place vector arrows anywhere on the plot (not constrained to a grid like the current `VectorField`)

* Resolves #2064

image

```cs
ScottPlot.Plot plt = new();

// Create a vector field list and add it to the plot
var field = plt.AddVectorFieldList();

// Optionally color arrows using a colormap and custom scaling function
field.Colormap = Colormap.Turbo;
field.ColormapScaler = (magnitude) => magnitude * 5;

// Additional arrow styling options are here
field.ArrowStyle.ScaledArrowheads = true;
field.ArrowStyle.Anchor = ArrowAnchor.Base;
field.ArrowStyle.LineWidth = 3;
field.ArrowStyle.MarkerSize = 10;
field.ArrowStyle.MarkerShape = MarkerShape.filledDiamond;

// Generate random vectors
Random rand = new(0);
int pointCount = 20;
for (int i = 0; i < pointCount + 1; i++)
{
    double x = i / (double)pointCount;
    double y = Math.Sin(i * 2 * Math.PI / pointCount);
    Coordinate coordinate = new(x, y);

    double dX = i / (double)pointCount * .15;
    double dY = i / (double)pointCount * .15;
    CoordinateVector vector = new(dX, dY);

    var rootedVector = (coordinate, vector);
    field.RootedVectors.Add(rootedVector);
}

plt.SaveFig("vectors.png");
```

### DIFF
--- a/src/ScottPlot4/ScottPlot.Cookbook/Recipes/Plottable/VectorField.cs
+++ b/src/ScottPlot4/ScottPlot.Cookbook/Recipes/Plottable/VectorField.cs
@@ -1,4 +1,5 @@
-﻿using ScottPlot.Statistics;
+﻿using ScottPlot.Drawing;
+using ScottPlot.Statistics;
 using System;
 using System.Collections.Generic;
 
@@ -155,23 +156,33 @@ namespace ScottPlot.Cookbook.Recipes.Plottable
 
         public void ExecuteRecipe(Plot plt)
         {
-            Random rand = new(0);
-
             // Create a vector field list and add it to the plot
             var field = plt.AddVectorFieldList();
 
+            // Optionally color arrows using a colormap and custom scaling function
+            field.Colormap = Colormap.Turbo;
+            field.ColormapScaler = (magnitude) => magnitude * 5;
+
+            // Additional arrow styling options are here
+            field.ArrowStyle.ScaledArrowheads = true;
+            field.ArrowStyle.Anchor = ArrowAnchor.Base;
+            field.ArrowStyle.LineWidth = 3;
+            field.ArrowStyle.MarkerSize = 10;
+            field.ArrowStyle.MarkerShape = MarkerShape.filledDiamond;
+
             // Generate random vectors
-            for (int i = 0; i < 50; i++)
+            Random rand = new(0);
+            int pointCount = 20;
+            for (int i = 0; i < pointCount + 1; i++)
             {
-                Coordinate coordinate = new(
-                    x: 20 * rand.NextDouble() - 10,
-                    y: 20 * rand.NextDouble() - 10);
+                double x = i / (double)pointCount;
+                double y = Math.Sin(i * 2 * Math.PI / pointCount);
+                Coordinate coordinate = new(x, y);
 
-                CoordinateVector vector = new(
-                    x: 4 * rand.NextDouble() - 2,
-                    y: 4 * rand.NextDouble() - 2);
+                double dX = i / (double)pointCount * .15;
+                double dY = i / (double)pointCount * .15;
+                CoordinateVector vector = new(dX, dY);
 
-                // Create a tuple that pairs a location and vector and add it to the list
                 var rootedVector = (coordinate, vector);
                 field.RootedVectors.Add(rootedVector);
             }
--- a/src/ScottPlot4/ScottPlot.Cookbook/Recipes/Plottable/VectorField.cs
+++ b/src/ScottPlot4/ScottPlot.Cookbook/Recipes/Plottable/VectorField.cs
@@ -1,5 +1,6 @@
 ﻿using ScottPlot.Statistics;
 using System;
+using System.Collections.Generic;
 
 namespace ScottPlot.Cookbook.Recipes.Plottable
 {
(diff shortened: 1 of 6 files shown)


src/ScottPlot4/ScottPlot.Cookbook/Recipes/Plottable/VectorField.cs
```diff
@@ -143,4 +145,47 @@ public void ExecuteRecipe(Plot plt)
...
+        public ICategory Category => new Categories.PlotTypes.VectorField();
+        public string ID => "vectorField_arbitrary";
+        public string Title => "Arbitrary Vectors";
+        public string Description => "Vectors can be placed arbitrarily in coordiante space " +
+            "(they are not required to be in a grid)";
+
+        public void ExecuteRecipe(Plot plt)
+        {
+            // Create a vector field list and add it to the plot
+            var field = plt.AddVectorFieldList();
```
bclehmann, line 160: I think there should be (and if there is, we should show it off) an option to just put the vectors into the list at construction

src/ScottPlot4/ScottPlot/Plottable/VectorFieldList.cs
```diff
@@ -0,0 +1,87 @@
...
+        /// If defined, this colormap is used to color each arrow based on its magnitude
+        /// </summary>
+        public Colormap Colormap { get; set; } = null;
+
+        /// <summary>
+        /// If <see cref="Colormap"/> is defined, each arrow's magnitude 
+        /// is run through this function to get the fraction (from 0 to 1) 
+        /// along the colormap to sample from.
+        /// </summary>
+        public Func<double, double> ColormapScaler = (magnitude) => magnitude;
```
bclehmann, line 45: This seems niche, this is one of the only feature requests we've ever had for vector fields after all.

src/ScottPlot4/ScottPlot.Cookbook/Recipes/Plottable/VectorField.cs
swharden: I mostly agree... I'll add it, but FYI here's my hesitation:

Since this is a List plot type, one advantage of not accepting data in the constructor is that it forces people to interact with the List in the plottable. It simplifies the documentation in that there's only one way to add arrows.

If people only know how to pass in arrows, it could be less obvious how to add more or modify existing ones.

src/ScottPlot4/ScottPlot/Plottable/VectorFieldList.cs
swharden: It makes sense if your vectors are 100s of units in size, but colormaps except a number from 0-1 😅

bclehmann:
I'd also note that it's a little confusing that `VectorField` and `VectorFieldList` use different vector objects, and I'm not sure that it's necessary to introduce a new one instead of using `Vector2`. But I think we discussed this though my COVID-addled brain cannot recall, so feel free to ignore that if you're satisfied we've exhausted that conversation.

swharden:
> a little confusing that VectorField and VectorFieldList use different vector objects

I think the API is different enough (a 2D array vs. part of a tuple) that it's unlikely to cause too much of a problem, but your point is a good one! 

I like the custom type (especially the overloads that allow it to add/subtract from `Coordinate`) that if I were to change anything it would probably be to strangle the `Vector2[,]` from the original vector field. Actually, maybe we should strangle the whole plot type and create a `VectorFieldGrid` (giving us one more shot at a simpler API).

Ideas for the future! 🚀 Thanks again for the review 👍

PR #2058, opened 2022-08-21
ScottPlot5: multi-axis support

swharden:
* Added deep multi-axis support (including updating all control interactions)
* Created a new `MultiAxisLimits` to compliment `AxisLimits`
* Sandbox examples are now located/launched dynamically (a preview of how the demo app could work)

multiaxis

PR #2080, opened 2022-08-30
ScottPlot5: Heatmap and Colormap

bclehmann:
**Purpose:**
Adds heatmaps to ScottPlot 5

```cs
int width = 100;
int height = 100;

double[,] intensities = new double[width, height];

for (int x = 0; x < width; x++)
    for (int y = 0; y < height; y++)
        intensities[x, y] = (Math.Sin(x * .2) + Math.Cos(y * .2)) * 100;


WpfPlot.Plot.Plottables.AddHeatmap(intensities);
```

image


bclehmann:
They now support transparency (using `double.NaN`):
```cs
double[,] intensities = new double[width, height];

for (int x = 0; x < width; x++)
    for (int y = 0; y < height; y++)
    {
        intensities[x, y] = (Math.Sin(x * .2) + Math.Cos(y * .2)) * 100;

        if (Math.Abs(x - y) < 5)
        {
            intensities[x, y] = double.NaN;
        }
    }
```

image

Still to do: 

- [ ] Customizable extent
- [ ] Cell Alignment
- [ ] Flip vertically
- [ ] Customize interpolation mode
- [ ] `ClippingPoints`? Is this API needed?
- [ ] Colorbar (perhaps this is a separate plottable/renderable/somethingelseable?)

src/ScottPlot5/ScottPlot/Plottables/Heatmap.cs
```diff
@@ -0,0 +1,86 @@
...
+            bitmap?.Dispose();
+        }
+
+        public void Update()
+        {
+            bitmap?.Dispose();
+            SKImageInfo imageInfo = new(Intensities.GetLength(0), Intensities.GetLength(1));
+
+            bitmap = new(imageInfo);
+            var intensitiesFlat = Intensities.Cast<double>();
```
bclehmann, line 48: The cast is likely unnecessary, `double[,]` implements `IEnumerable<double>`, not `IEnumerable<double[]>` or similar.

bclehmann: Sike, it implements `IEnumerable` but not `IEnumerable<double>` even though each element it returns is of type `double`

src/ScottPlot5/ScottPlot/Colormaps/ArgbColormapBase.cs
```diff
@@ -0,0 +1,28 @@
...
+        public Color GetColor(double intensity, Range? domain)
+        {
+            if (double.IsNaN(intensity))
+            {
+                return Colors.Transparent;
+            }
+
+            domain ??= Range.UnitRange;
+
+            double normalized = domain.Value.NormalizeAndClampToUnitRange(intensity);
```
swharden, line 22: Do you find it confusing that "domain" is of type "range"? This is the "domain range"?

swharden: Maybe a solution is to name the argument `intensityRange` or something

I'm not finding `domain` particularly intuitive as to what it's supposed to hold

swharden: Also, such a small nit it's not worth changing, but I'd probably call `normalized` something more descriptive like `colormapFraction`

src/ScottPlot5/ScottPlot/Colormaps/ByteColormapBase.cs
```diff
@@ -0,0 +1,28 @@
...
+        public Color GetColor(double intensity, Range? domain)
+        {
+            if (double.IsNaN(intensity))
+            {
+                return Colors.Transparent;
+            }
+
+            domain ??= Range.UnitRange;
+
+            double normalized = domain.Value.NormalizeAndClampToUnitRange(intensity);
```
swharden, line 22: Same comments as above

src/ScottPlot5/ScottPlot/Colormaps/Grayscale.cs
```diff
@@ -0,0 +1,22 @@
...
+        public Color GetColor(double intensity, Range? domain)
+        {
+            if (double.IsNaN(intensity))
+            {
+                return Colors.Transparent;
+            }
+
+            domain ??= Range.UnitRange;
+
+            double normalized = domain.Value.NormalizeAndClampToUnitRange(intensity);
```
swharden, line 16: Same comments as above

src/ScottPlot5/ScottPlot/Colormaps/GrayscaleReversed.cs
```diff
@@ -0,0 +1,22 @@
...
+        public Color GetColor(double intensity, Range? domain)
+        {
+            if (double.IsNaN(intensity))
+            {
+                return Colors.Transparent;
+            }
+
+            domain ??= Range.UnitRange;
+
+            double normalized = domain.Value.NormalizeAndClampToUnitRange(intensity);
```
swharden, line 16: Same comment as above (wow, lots of duplicated code I guess)

src/ScottPlot5/ScottPlot/Colormaps/IColormap.cs
```diff
@@ -0,0 +1,14 @@
...
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace ScottPlot.Colormaps
+{
+    public interface IColormap
+    {
+        string Name { get; }
+        Color GetColor(double intensity, Range? domain);
```
swharden, line 12: Same comment as above. Also I think the XML docs for GetColor should be pretty clear about how it works and what the corner cases are (e.g., transparency)

I admit things are flying around and I haven't set-up a meaningful/rigid test project yet (what's there is really thrown together), but the special NaN behavior would be a good candidate for the Beyonce rule: Regarding special features (and protecting from them getting removed by accident), _if you like it then you shoulda put a test on it_ 😜

src/ScottPlot5/ScottPlot/CoordinateRect.cs
```diff
@@ -39,6 +41,8 @@ public CoordinateRect(double xMin, double xMax, double yMin, double yMax)
 
     public CoordinateRange YRange => new(YMin, YMax);
 
+    public CoordinateRect WithTranslation(Coordinates p) => new(XMin + p.X, XMax + p.X, YMin + p.Y, YMax + p.Y);
```
swharden, line 44: We don't have to implement this now, but an alternative way to do this would be an operator overload with `CoordinateVector`

Actually, should the augment here be a `CoordinateVector` anyway?

src/ScottPlot5/ScottPlot/Plottables/Heatmap.cs
```diff
@@ -0,0 +1,123 @@
...
+            set
+            {
+                _intensities = value;
+                Update();
+            }
+        }
+
+        private SKBitmap? bitmap = null;
+
+        public Heatmap(double[,] intensities)
```
swharden, line 56: Let's do what we have to so we kill all the warnings.

I may enable `TreatWarningsAsErrors` in the CI pipeline to help with this 😈

src/ScottPlot5/ScottPlot/Range.cs
```diff
@@ -0,0 +1,45 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace ScottPlot
+{
+    public struct Range
```
swharden, line 9: Let's add XML docs indicating what this is. It's not a pixel range or a coordinate rage ... it's any arbitrary range. Maybe something like

```cs
///<summary>
///Represents a range between any two finite values (inclusive)
///</summary>
```

Should we force `min` to be < `max` in the constructor?

Should we throw if the user passes in NaN or infinity?

Also let's add XML docs to functions. I'll help:

### Normalize

```cs
///<summary>
///Returns the given value as a fraction of the range between min and max.
///Returns values <0 or >1 if the given value is outside the range.
///</summary>
```

### Clamp

```cs
///<summary>
///Returns the given value clamped to the range (inclusive).
///Attempting to clamp NaN will return NaN.
///</summary>
```

Rather than `NormalizeAndClampToUnitRange()`, maybe Normalize should just have a `bool clamp`?

swharden (changes requested):
This for this PR! It's looking great, and I appreciate this was a lot of work.

None of my recommended changes are absolutely required. Modify based on your best judgement, and let me know when you're done and I'll merge it in whatever state it's in 👍

src/ScottPlot5/ScottPlot/Colormaps/GrayscaleReversed.cs
bclehmann: Yeah I introduced a `ColormapBase` to avoid some of that, I like the final result but I'm worried we're reaching Java-levels of OOP.

src/ScottPlot5/ScottPlot/Plottables/Heatmap.cs
bclehmann: There were only nullability warnings because `Intensities` had a non-trivial setter, so the compiler couldn't tell that it set the backing field. I ended up removing the setter on `Intensities` anyways, but in the future this might be the sort of place where I would explicitly disable the warning, as it was incorrect, the backing field was definitively set to non-null (provided the constructor receives a non-null value).

src/ScottPlot5/ScottPlot/Range.cs
bclehmann: Clamping NaN to NaN is not the default behaviour of a naive implementation, and I don't know if we want to take a stance on what should happen when you clamp NaN? There is currently no codepath that (intentionally) attempts to clamp NaN and so I think it's fine for the behaviour to be "did you really think that was going to work?"

bclehmann: Also, requiring $min \neq max$ does prevent a possible divide by zero, but I think it's going to be inconvenient in the future.

For example, if someone passes in a heatmap with a single element $x$, when we calculate the extent of it we're going to get the degenerate interval $[x, x]$. I'd rather try to make sure that `Range` behaves properly on degenerate intervals than walk on eggshells around it everywhere trying to avoid such intervals.

bclehmann:
I've added a `ColormapBase` class to handle repeated logic within the colormaps.

I'm also unsure about how we handle updating, currently the user has to manually call `Update()` to update the cached image. With the current implementation (and likely any future implementation) getting rid of the cache is pretty unperformant.

We could use setters to update the cache automatically, in an earlier implementation of this class I used this to update the cache when the `Intensities` array changed. However, there are a lot of properties that could impact rendering:
- Colormap
- CellAlignment
- Extent
- FlipVertically
- Smooth
- Intensities

More perniciously however, modifying an element in the `Intensities` array without reassigning it will not call the setter. So what we probably need to do is create a data class which alerts us when the array is modified in place. I didn't want to start on this without checking in first, because in v4 we were happy to just make the user call `Update`

swharden:
> getting rid of the cache is pretty unperforming.

> in v4 we were happy to just make the user call Update

Cashing _is_ more performant, but it's not what the rest of the plottables do... For most plot types, users expect to be able to change the data they pass in and have the changes reflected on the plot automatically. 

_Not_ caching data would be more in-line with the principle of least astonishment. Maybe caching could be an optional opt-in feature for users who actually need/want the performance enhancement?

src/ScottPlot5/ScottPlot/Range.cs
swharden: These are good points! There's a lot flying around right now. I'm going to add a TODO so we can revisit this topic some day in the future.

Same with the coordinate vs. vector addition thing.

I'm just trying to get off the ground, but we can define these things carefully later and add tests to carefully lock-in all these behavior decisions. (cringy to write that - code now and test later is such a trope - but I don't see SP5 getting off the ground if we jump into all these little minutia corner cases this early in the game)

swharden (approved):

swharden:
Tying-up a couple loose ends,

> ClippingPoints? Is this API needed?

It seems okay without it

> Colorbar (perhaps this is a separate plottable/renderable/somethingelseable?)

Let's do this a lot later... I'm still deeply changing how plot layouts work (e.g., #2058) so I definitely want to push this off.

src/ScottPlot5/ScottPlot/Colormaps/IColormap.cs
```diff
@@ -10,11 +10,11 @@ public interface IColormap
...
         string Name { get; }
         /// <summary>
-        /// Returns a color for the given intensity and intensityRange. Returns Colors.Transparent for an intensity of NaN.
+        /// Returns a color for the given <paramref name="intensity"/> and <paramref name="intensityRange"/>. 
+        /// Returns <see cref="Colors.Transparent"/> for an intensity of <see cref="double.NaN"/>.
         /// </summary>
-        /// <param name="intensity">The intensity</param>
-        /// <param name="intensityRange">The range of the scale from which the intensity is drawn, or Range.UnitRange if null.</param>
-        /// <returns>Color</returns>
+        /// <param name="intensity">Intensity fraction from 0 to 1 (unless <paramref name="intensityRange"/> is specified)</param>
```
bclehmann: This isn't quite true, they can pass in null for `intensityRange` and still pass in an `intensity` < 0 or > 1. It will simply be clamped.

src/ScottPlot5/ScottPlot/Range.cs
```diff
@@ -73,5 +73,36 @@ public double Clamp(double value)
 
             return value;
         }
+
+        public static Range GetRange(double[,] input)
```
bclehmann, line 82: This can just be implemented simpler:
```cs
public static Range GetRange(double[,] input) => GetRange(input.Cast<double>());
```

`T[,]` implements `IEnumerable` but not `IEnumerable<T>`, hence the cast. However, since the overload which takes `IEnumerable<double>` doesn't create a concrete version of the stream (i.e. it doesn't call `ToArray()` or similar) I don't think the performance difference should be noticeable above the (minor) performance drop that comes from using `IEnumerable` compared to direct subscripting.

If you think that whatever performance difference there is is too high then fair enough, this isn't that much code to duplicate, I just thought I'd bring up a simpler option.

src/ScottPlot5/ScottPlot/Plottables/Heatmap.cs
```diff
@@ -22,27 +22,104 @@ public class Heatmap : IPlottable
         public bool IsVisible { get; set; } = true;
         public IAxes Axes { get; set; } = Axis.Axes.Default;
         public IColormap Colormap { get; set; } = new Viridis();
+
+        /// <summary>
+        /// Indicates position of the data point relative to the rectangle used to represent it.
+        /// An alignment of upper right means the rectangle will appear to the lower left of the point itself.
```
bclehmann, line 28: `UpperRight` doesn't exist on the enum

bclehmann: Also if we want to support having independent alignments for each dimension (which I would caution against), I'd rather put them as two separate properties than have a god enum like this:

```cs
public enum CellAlignment {
  BottomLeft
  BottomCenter
  BottomRight
  MiddleLeft,
  .
  .
  .
}
```

src/ScottPlot5/ScottPlot/Plottables/Heatmap.cs
```diff
@@ -22,27 +22,104 @@ public class Heatmap : IPlottable
...
+        /// </summary>
         public CellAlignment CellAlignment { get; set; } = CellAlignment.Center;
+
+        /// <summary>
+        /// If defined, the this rectangle sets the boundaries of colormap data.
+        /// Note that the actual colormap area is 1 cell larger than this rectangle.
```
bclehmann, line 34: I think this is confusing, it could be interpreted as a mechanism to clip the heatmap, rather then one to rescale the heatmap to the axes.

src/ScottPlot5/ScottPlot/Plottables/Heatmap.cs
```diff
@@ -22,27 +22,104 @@ public class Heatmap : IPlottable
...
+        /// </summary>
         public CoordinateRect? Extent { get; set; }
+
+        /// <summary>
+        /// By default heatmaps are drawn in the same vertical orientation as vectors: row 0 is at the top.
```
bclehmann, line 39: I don't think comparing it to the orientation of vectors makes sense, unless you meant to say matrix?

src/ScottPlot5/ScottPlot/Plottables/Heatmap.cs
```diff
@@ -22,27 +22,104 @@ public class Heatmap : IPlottable
...
+        /// If this is true, the vertical orientation will be flipped so row 0 is on the bottom.
+        /// </summary>
         public bool FlipVertically { get; set; } = false;
+
+        /// <summary>
+        /// If true, the heatmap will be blurred.
+        /// If false, the heatmap will appear as individual rectangles with sharp edges.
+        /// </summary>
```
bclehmann, line 47: I wouldn't call this blurring, it's interpolation. We're making up data rather than obscuring it.

src/ScottPlot5/ScottPlot/Range.cs
```diff
@@ -30,6 +30,11 @@ public Range(double min, double max)
                 throw new ArgumentException($"Argument ${nameof(max)} must be finite and non-NaN");
             }
 
+            if (double.IsNaN(min) || double.IsNaN(max))
```
bclehmann, line 33: This check already exists above.

FWIW, there's a `double.IsFinite` which ensures it's non-NaN and not ±infinity, but we can't use it on all the things we target. Since we use this logic a lot it might make sense to create our own.

src/ScottPlot5/ScottPlot/Range.cs
```diff
@@ -47,9 +52,9 @@ public Range(double min, double max)
         /// <returns>The normalized value</returns>
         public double Normalize(double value, bool clamp = false)
         {
-            if (Max == Min) // When Min == Max we would divide by zero without this check.
+            if (Max == Min)
             {
-                return Min;
+                throw new ArgumentException("Cannot normalize the value to a range of zero");
```
bclehmann, line 57: If we allow this to throw we need to check for `Min == Max` in `GetArgbValues` (or potentially in Colormap.GetColor), or we're going to crash on some heatmaps.

PR #2127, opened 2022-09-16
ScottPlot 5 Legends

bclehmann:
**Purpose:**
Adds legends to ScottPlot 5

```cs
var legend = new Legend(new[]
{
    new LegendItem { Label = "Blue", Fill = new Fill(Colors.Blue) },
    new LegendItem { Label = "Gray", Fill = new Fill(Colors.Gray) },
    new LegendItem { Label = "Line", Line = new Stroke(Colors.Sienna, 1)},
    new LegendItem {
        Label = "Marker and Line", Line = new Stroke(Colors.Blue, 1),
        Marker = new Marker(MarkerShape.Circle, Colors.Blue, 5),
        Children = new[] {
            new LegendItem { Label = "Red", Fill = new Fill(Colors.Red), Children = new[] {
                new LegendItem { Label = "Blue", Fill = new Fill(Colors.Blue) },
                new LegendItem { Label = "Cyan", Fill = new Fill(Colors.Cyan) },
            }},
            new LegendItem { Label = "Green", Fill = new Fill(Colors.Green) },
        }
    },
    new LegendItem { Label = "Marker", Marker = new Marker(MarkerShape.Circle, Colors.Blue, 5)},
    new LegendItem { Label = "Blue", Fill = new Fill(Colors.Blue) },
    new LegendItem { Label = "Nothing" },
    new LegendItem { Label = "Blue", Fill = new Fill(Colors.Blue) },
});

avaPlot.Plot.Plottables.Add(legend);
```

image

### DIFF
--- a/src/ScottPlot5/ScottPlot5/IPlottable.cs
+++ b/src/ScottPlot5/ScottPlot5/IPlottable.cs
@@ -31,5 +31,5 @@ public interface IPlottable
     /// By default the surface is already clipped to the data area, but this can be cleared inside the plottable.
     /// </summary>
     void Render(SKSurface surface);
-    IList<LegendItem> GetLegendItems();
+    IEnumerable<LegendItem> LegendItems { get; }
 }
--- a/src/ScottPlot5/ScottPlot5/IPlottable.cs
+++ b/src/ScottPlot5/ScottPlot5/IPlottable.cs
@@ -31,4 +31,5 @@ public interface IPlottable
     /// By default the surface is already clipped to the data area, but this can be cleared inside the plottable.
     /// </summary>
     void Render(SKSurface surface);
+    IList<LegendItem> GetLegendItems();
 }
--- a/src/ScottPlot5/ScottPlot5/Plot.cs
+++ b/src/ScottPlot5/ScottPlot5/Plot.cs
@@ -44,7 +44,7 @@ public class Plot
     /// The primary vertical axis (the first one in the list of <see cref="YAxes"/>)
     /// </summary>
     public IYAxis YAxis => YAxes.First();
-    
+
     public Plot()
     {
         var xAxis = new Axis.StandardAxes.BottomAxis();
@@ -262,10 +262,10 @@ public class Plot
     #region Legend
     public IEnumerable<LegendItem> LegendItems()
     {
-        foreach(var curr in Plottables)
+        foreach (var curr in Plottables)
             foreach (var item in curr.GetLegendItems())
                 yield return item;
     }
-    
+
     #endregion
 }
--- a/src/ScottPlot5/ScottPlot5/Plot.cs
+++ b/src/ScottPlot5/ScottPlot5/Plot.cs
@@ -3,6 +3,7 @@ using ScottPlot.Axis;
 using ScottPlot.Rendering;
 using ScottPlot.LayoutSystem;
 using SkiaSharp;
+using ScottPlot.Plottables;
 
 namespace ScottPlot;
 
@@ -43,7 +44,7 @@ public class Plot
     /// The primary vertical axis (the first one in the list of <see cref="YAxes"/>)
     /// </summary>
     public IYAxis YAxis => YAxes.First();
-
+    
     public Plot()
     {
         var xAxis = new Axis.StandardAxes.BottomAxis();
@@ -256,5 +257,15 @@ public class Plot
         File.WriteAllBytes(path, bytes);
     }
 
+    #endregion
+
+    #region Legend
+    public IEnumerable<LegendItem> LegendItems()
+    {
+        foreach(var curr in Plottables)
+            foreach (var item in curr.GetLegendItems())
+                yield return item;
+    }
+    
     #endregion
 }
--- a/src/ScottPlot5/ScottPlot5/Plot.cs
+++ b/src/ScottPlot5/ScottPlot5/Plot.cs
@@ -263,7 +263,7 @@ public class Plot
     public IEnumerable<LegendItem> LegendItems()
     {
         foreach (var curr in Plottables)
-            foreach (var item in curr.GetLegendItems())
+            foreach (var item in curr.LegendItems)
                 yield return item;
     }
 
--- a/src/ScottPlot5/ScottPlot5/PlottableFactory.cs
+++ b/src/ScottPlot5/ScottPlot5/PlottableFactory.cs
@@ -24,6 +24,13 @@ public class PlottableFactory
         return heatmap;
     }
 
+    public Legend Legend(IList<LegendItem>? legendItems = null)
+    {
+        Legend legend = new(legendItems ?? Plot.LegendItems());
+        Plot.Plottables.Add(legend);
+        return legend;
+    }
+
     public Pie Pie(IList<PieSlice> slices)
     {
         Pie pie = new(slices);
--- a/src/ScottPlot5/ScottPlot5/Plottables/Legend.cs
+++ b/src/ScottPlot5/ScottPlot5/Plottables/Legend.cs
@@ -128,7 +128,7 @@ namespace ScottPlot.Plottables
             {
                 return PixelSize.Zero;
             }
-            
+
             using SKPaint paint = new(Font);
 
             PixelSize labelRect = Drawing.MeasureString(item.Label ?? "", paint);
--- a/src/ScottPlot5/ScottPlot5/Plottables/Legend.cs
+++ b/src/ScottPlot5/ScottPlot5/Plottables/Legend.cs
@@ -17,15 +17,16 @@ namespace ScottPlot.Plottables
         public bool IsVisible { get; set; } = true;
         public IAxes Axes { get; set; } = Axis.Axes.Default;
         public AxisLimits GetAxisLimits() => AxisLimits.NoLimits;
+        public IList<LegendItem> GetLegendItems() => Array.Empty<LegendItem>();
 
         public Corner Position { get; set; } = Corner.BottomRight; // TODO: Should we support positions that aren't corners?
-        public IList<LegendItem> LegendItems { get; set; }
+        public IEnumerable<LegendItem> LegendItems { get; set; }
         public SKFont Font { get; set; } = new(); // TODO: Do we want our own abstraction for this? It wraps a native object and implements IDisposable
         public Stroke Border { get; set; } = new(Colors.DarkGray, 1);
         public Fill Background { get; set; } = new Fill(Colors.White);
 
 
-        public Legend(IList<LegendItem> legendItems)
+        public Legend(IEnumerable<LegendItem> legendItems)
         {
             LegendItems = legendItems;
         }
@@ -60,6 +61,11 @@ namespace ScottPlot.Plottables
 
         private void RenderLegendItem(SKSurface surface, LegendItem item, float y)
         {
+            if (ShouldHide(item))
+            {
+                return;
+            }
+
             using ScopedTransform transform = new(surface.Canvas);
             using SKPaint paint = new(Font);
             paint.SetFill(new(Colors.Black));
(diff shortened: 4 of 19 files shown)


bclehmann:
New testcase:

```cs
const int N = 51;

avaPlot.Plot.Plottables.Add(DebugPoint);
avaPlot.Plot.Add.Scatter(Generate.Consecutive(N), Generate.Sin(N), Colors.Blue).Label = "Scatter";
avaPlot.Plot.Add.Scatter(Generate.Consecutive(N), Generate.Cos(N), Colors.Red).Label = "Scatter 2";
avaPlot.Plot.Add.Scatter(Generate.Consecutive(N), Generate.Sin(N, 0.5), Colors.Green).Label = "Scatter 3";

avaPlot.Plot.Add.Pie(new PieSlice[]
{
    new() { Value = 6, Fill = new() { Color = Colors.Red}, Label = "Red" },
    new() { Value = 4, Fill = new() { Color = Colors.Blue }, Label = "Blue" },
    new() { Value = 3, Fill = new() { Color = Colors.Green }, Label = "Green" },
    new() { Value = 1, Fill = new() { Color = Colors.DarkCyan }, Label = "Dark Cyan" },
}).Label = "Pie Chart";

avaPlot.Plot.Add.Signal(Generate.NoisySin(new(), N, 0.1)).Label = "Signal";

avaPlot.Plot.Add.Legend();
```

src/ScottPlot5/ScottPlot5/Plot.cs
```diff
@@ -257,4 +258,14 @@ public void SaveImage(int width, int height, string path, int quality = 100)
     }
 
     #endregion
+
+    #region Legend
+    public IEnumerable<LegendItem> LegendItems()
```
swharden, line 263: Does this need to be public? Interacting with items individually seems like pretty advanced stuff, so maybe users wishing to do this should just iterate the plottables and get the legend items themselves.

src/ScottPlot5/ScottPlot5/PlottableFactory.cs
```diff
@@ -24,6 +24,13 @@ public Heatmap Heatmap(double[,] intensities)
         return heatmap;
     }
 
+    public Legend Legend(IList<LegendItem>? legendItems = null)
```
swharden, line 27: Maybe corner should be the first optional argument... Customizing placement when turning the legend on is a pretty common use case.

src/ScottPlot5/ScottPlot5/IPlottable.cs
```diff
@@ -31,4 +31,5 @@ public interface IPlottable
     /// By default the surface is already clipped to the data area, but this can be cleared inside the plottable.
     /// </summary>
     void Render(SKSurface surface);
+    IList<LegendItem> GetLegendItems();
```
swharden, line 34: I recognize that this function mimics the ScottPlot4 design, but I wonder if there's merit in making it a property?

```cs
IList<LegendItem> LegendItems { get; }
```

I guess most plottables would still need to create the legend items every time it's called (because instantiating the list once and keeping things like colors in sync would be hard), but the API may read a little smoother this way.

Interested in your thoughts

src/ScottPlot5/ScottPlot5/Plottables/Legend.cs
```diff
@@ -0,0 +1,188 @@
...
+            }
+
+            return new(width, height);
+        }
+
+        // I'm a Javascript developer, this was bound to happen someday
+        public PixelSize Measure() =>
+            LegendItems
+                .Select((LegendItem item) => Measure(item))
+                .Aggregate(
```
swharden, line 162: lol, this indeed would be the first `Aggregate` in the code base

src/ScottPlot5/ScottPlot5/Plottables/Legend.cs
```diff
@@ -0,0 +1,188 @@
...
+                        new(
+                            Math.Max(a.Width, b.Width),
+                            a.Height + b.Height)
+                        );
+
+        private Pixel GetTopLeftCorner(PixelSize size)
+        {
+            const int margin = 5;
+
+            return Position switch
```
swharden, line 174: Maybe this logic could be moved inside `Corner` because it seems like a common task

src/ScottPlot5/ScottPlot5/Plottables/Signal.cs
```diff
@@ -19,13 +19,22 @@ public class Signal : IPlottable
 
     public Color Color = new(0, 0, 255);
     public float LineWidth = 1;
+    public string? Label { get; set; }
```
swharden, line 22: General API question, should we favor nullable strings or `string.Empty` for plottables without labels?

Are there strong pros/cons for either strategy?

src/ScottPlot5/ScottPlot5/ScopedTransform.cs
```diff
@@ -0,0 +1,26 @@
+﻿using SkiaSharp;
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace ScottPlot
+{
+    internal class ScopedTransform : IDisposable // TODO: I'm a little uneasy on this name?
```
swharden, line 10: I don't mind the name. As long as the XML docs make it clear what it does, it should be fine. I would consider moving tools like this into a `Drawing` namespace though. I imagine we will start collecting stuff like this (e.g., font stuff).

---

I see there's already a `Drawing.cs`, but it looks pretty small, so I'm thinking that should get broken apart and moved into better-named classes in the new drawing namespace too. After doing this we could have something like:

* `ScottPlot.Drawing.ScopedTransform`
* `ScottPlot.Drawing.Text.MeasureString()`
* `ScottPlot.Drawing.Text.MeasureLargestString()`
* `ScottPlot.Drawing.Shapes.DrawLines()`

It's probably not too big a change to go ahead and do in this PR, but if it gets more complex than expected we can open an issue so we don't forget and address it after it's merged

... also I'm not 100% those methods are even being called anymore, so if they're not being used we can just delete them

src/ScottPlot5/ScottPlot5/Style/Marker.cs
```diff
@@ -0,0 +1,33 @@
...
+using System.Threading.Tasks;
+
+namespace ScottPlot.Style
+{
+    public enum MarkerShape
+    {
+        Circle
+    }
+
+    public struct Marker
```
swharden, line 14: I think we should build-out Marker a little more. We can do it outside this PR if it gets complex though. 

Basically I think markers should be classes that implement `IMarker`. Marker classes could live in `ScottPlot.Markers` namespace, and `ScottPlot.Marker` would be a static class that gives quick access to common markers (kind of like colors). This would make it easier for users to add their own markers, and also for new marker shapes to be added without hunting around for switch statements in the code base. Something like:

```cs
public interface IMarker {
    void Color Color { get; set; }
    void double Size { get; set; }
    void Render(SKSurface surface, IEnumerable<Pixel> points);
}
```

...and I guess some markers would have additional fields like `LineWidth` that live in the concrete classes and not the interface? I'm open to suggestions, but this is the basic idea

src/ScottPlot5/ScottPlot5/Plot.cs
bclehmann: I don't think it's trivial enough that it wouldn't be much of a loss if it was private (provided that you moved it to the `PlottableFactory` class so that it could access it), but I think there's a usecase for people who want to filter or split up the legend into multiple legends (for example by plot type).

src/ScottPlot5/ScottPlot5/Plottables/Signal.cs
bclehmann: There isn't a great way to note that a string can or cannot be empty through the type system, unless of course you create your own type for it.

Now arguably there isn't a much better way to note that a string can or cannot be null through the type system, given that non-nullable strings can still be null.

I personally prefer null, because trying to work with a null string is going to fail much faster and more obviously than trying to work with an empty string. You might consider an empty string invalid, but most operations on empty strings are going to work fine and make bugs much trickier to track down.

src/ScottPlot5/ScottPlot5/Style/Marker.cs
bclehmann: Yeah I like having an `IMarker`, though I don't mind an enum-based approach either. However, if we were to stick with enums I'd like if we had a central `DrawMarker` that handles it so we only have one switch statement. Kind of like `MarkerTools.DrawMarker`, though I believe that actually creates an `IMarker` which knows how to draw itself.

src/ScottPlot5/ScottPlot5/IPlottable.cs
bclehmann: I like this, I wonder if it could be turned into an `IEnumerable<LegendItems>` as well, as I waffled on which interface to use for a bit. I used `IEnumerable` in the `Legend` plottable because although it eats the (negligible) performance cost of reifying the enumerable it lets a legend react to its source IEnumerable changing. This means that the default legend will automatically update if/when new plottables are added.

I think caching the legend items is fine with either API, but I frankly wouldn't bother unless somehow there's a clear need. I'd actually go the opposite and turn them all into `IEnumerable`s that aren't turned into concrete lists unless necessary.

src/ScottPlot5/ScottPlot5/PlottableFactory.cs
bclehmann: Aye, although I'm not particularly happy with `Position` being a corner enum, I feel like that's going to paint us into a corner.

Perhaps we have an `IPosition` that returns a `Pixel` for a given screen size and legend size, and we only implement it for the four corners for now? That seems like an overkill solution but given that layout code always seems like it should be easier than it is it might be our best bet.

src/ScottPlot5/ScottPlot5/IPlottable.cs
bclehmann: I turned them all into `IEnumerable<LegendItem> LegendItems { get; }`

src/ScottPlot5/ScottPlot5/Style/Marker.cs
bclehmann: I'm also recalling that `LegendItem` holds an `SKFont`, but I really don't think that's something we want to expose to the user. The `SKFont` class is pretty intense, and it also needs to be disposed.

Perhaps this can just be a `Font` struct with a `GetSKFont()` though. Not quite as much thought required as with markers.

src/ScottPlot5/ScottPlot5/ScopedTransform.cs
bclehmann: I think XML docs are unnecessary, as this is `internal`. I do like the idea of a drawing namespace but that might be better handled in a separate PR.

> also I'm not 100% those methods are even being called anymore
Can't speak for all of them, but in this PR I'm using `MeasureString`, I actually had to modify it because it wasn't quite correct.

src/ScottPlot5/ScottPlot5/PlottableFactory.cs
swharden: Interesting idea!

> That seems like an overkill solution

I do like it, but I'll go ahead and merge this as it is and we can extend this functionality in the future

src/ScottPlot5/ScottPlot5/Plottables/Signal.cs
swharden: > There isn't a great way to note that a string can or cannot be empty through the type system

> I personally prefer null, because trying to work with a null string is going to fail much faster and more obviously than trying to work with an empty string

I'm sold! I'll lean toward nullable strings to signal they're optional

> You might consider an empty string invalid

I like this suggestion too... I'll think about it, but it's on my radar now.

Thanks for your input!

Issue #2130, opened 2022-09-21
ScottPlot 4: Plot.HasAxis method


bclehmann:
## Feature Suggestion
**Feature description:** Describe your idea here
A `Plot.HasAxis(int axisIndex, bool isVertical)` which checks if the plot has a matching axis.

The motivation for this is currently users are responsible for keeping track of axis indices and their validity, which could be made easier by a method which determines their validity. Perhaps another option is returning `null` from `GetXAxis`/`GetYAxis` rather than throwing, but that would be a breaking change (and arguably worse, though some may disagree).

**Code example:**
```cs
bool hasAxis = plt.HasAxis(axisIndex, vertical);
```

bclehmann:
I'm realizing that `GetXAxis` and `GetYAxis` are not publicly exposed, so maybe they should be as well.

Also discussed in the discord was automatically assigning axis indices, so that users do not need to keep track of which indexes are in use at a given time.

PR #2133, opened 2022-09-22
Adds GetAxesMatching function

bclehmann:
**Purpose:**
#2130, though I elected not to create a `HasAxisMatching` because you can just do `GetAxesMatching(criteria).Any()`. This PR also makes the `axisIndex` parameter optional on `AddAxis`, opting to auto-increment them. This should simplify multi-axis code and the creation of valid axes without breaking existing code.

```cs
var axes = plt.GetAxesMatching(null, false); // IEnumerable holding all horizontal axes.
```


bclehmann:
If this is merged in some changelog-credit should probably go to Nick in the discord, provided they have a github username.

Issue #2132, opened 2022-09-22
DateTimeTickUnits: The X-axis of `DateTime` mixture 24-hour and 12-hour notation.


MareMare:
**ScottPlot Version:** 4.1.58.0

**Operating System:** Windows10

**Application Type:** WinForms

**Question:** The X-axis of DateTime mixture 24-hour and 12-hour notation.

Zoom in on a graph displaying `DateTime` data on the X-axis with the mouse wheel or right-click.

~~In this case, the 24-hour and 12-hour displays show different times, depending on whether they zoom to the nearest second or include milliseconds.~~

It appears that 24-hour notation is used for zooms that display up to seconds, and 12-hour notation is used for zooms that also display milliseconds.

Is this as expected?

|pattern|Actual output|comment|
|--|:--:|--|
|(A) Zooms that display up to seconds.|1985/09/23<br>23:59:59| As expected.|
|(B) Zooms that also display milliseconds.|1985/09/23<br>11:59:59.5|The time format expects `23:59:59.5`.|

* In case (A):
  https://github.com/ScottPlot/ScottPlot/blob/07ab190973cbf8697ff8bdc98b4e6475ce40ea03/src/ScottPlot4/ScottPlot/Ticks/DateTimeTickUnits/DateTimeTickUnitBase.cs#L34-L39

* In case (B):
  https://github.com/ScottPlot/ScottPlot/blob/07ab190973cbf8697ff8bdc98b4e6475ce40ea03/src/ScottPlot4/ScottPlot/Ticks/DateTimeTickUnits/SecondsAndLess/DateTimeTickDecisecond.cs#L25-L30

Maybe `HH` is correct in (B) instead of `hh`?

![Animation](https://user-images.githubusercontent.com/807378/191654716-0ee5dd4e-900f-49fa-84c8-f2c204cbcd90.gif)

<details>
<summary>The reproduced code is as follows:</summary>
<div>

```cs
public partial class Form1 : Form
{
    public Form1()
    {
        InitializeComponent();

        ExecuteRecipe(formsPlot1.Plot);
    }

    private void ExecuteRecipe(Plot plt)
    {
        // original source: https://github.com/ScottPlot/ScottPlot/blob/07ab190973cbf8697ff8bdc98b4e6475ce40ea03/src/ScottPlot4/ScottPlot.Cookbook/Recipes/Axis.cs#L208-L222

        // create data sample data
[...]

PR #2135, opened 2022-09-24
Consistent 12/24 hour clock in DateTime ticks

bclehmann:
**Purpose:**
#2132

This solution detects if a locale uses a 24-hour clock and adjusts some hard-coded time formats to generate 24-hour times.


bclehmann:
Also discussed (but not acted upon) in #2132 was how the format of times involving milliseconds is not locale aware.

For example, under the locale `fr-CA` (Canadian French), we would generate the time `23:59:59.999`, but we should generate `23h59m59.999` (I think, depends how well I remember from French class).

PR #2152, opened 2022-10-05
BarSeries rectangle calculation for negative values fixed.

fe-c:
Before:
Before

After:
After


bclehmann (approved):
Looks good to me, this is unfortunately a bug I've seen a lot.

Perhaps we should move this logic into `Bar`, but I'm unsure. I don't anticipate that we'd transplant `Bar` into `BarPlot` anyways, so it can probably stay here.

PR #2155, opened 2022-10-06
Add DataPointList for Scatterplot, to add a label for each marker.

Cubikon:
#2141

### DIFF
--- a/src/ScottPlot4/ScottPlot/Plottable/ScatterPlot.cs
+++ b/src/ScottPlot4/ScottPlot/Plottable/ScatterPlot.cs
@@ -17,6 +17,7 @@ namespace ScottPlot.Plottable
         public double[] Ys { get; private set; }
         public double[] XError { get; set; }
         public double[] YError { get; set; }
+        public string[] DataPointLabels { get; set; }
 
         public enum NanBehavior
         {
@@ -426,6 +427,27 @@ namespace ScottPlot.Plottable
                 if ((MarkerSize > 0) && (MarkerShape != MarkerShape.none))
                 {
                     MarkerTools.DrawMarkers(gfx, points, MarkerShape, MarkerSize, MarkerColor, MarkerLineWidth);
+                    if (DataPointLabels != null)
+                        for (var i = 0; i < DataPointLabels.Length; i++)
+                        {
+                            var label = DataPointLabels[i];
+                            if (!string.IsNullOrEmpty(label))
+                            {
+                                Drawing.Font textFont = new();
+                                var stringSize = GDI.MeasureString(gfx, label, textFont);
+                                gfx.TranslateTransform(points[i].X, points[i].Y);
+                                gfx.RotateTransform(textFont.Rotation);
+
+                                (var dX, var dY) = GDI.TranslateString(gfx, label, textFont);
+                                gfx.TranslateTransform(-dX, -dY);
+
+                                using var font = GDI.Font(textFont);
+                                using var fontBrush = new SolidBrush(textFont.Color);
+                                gfx.DrawString(label, font, fontBrush, new PointF(0, 0));
+
+                                GDI.ResetTransformPreservingScale(gfx, dims);
+                            }
+                        }
                 }
             }
         }
--- a/src/ScottPlot4/ScottPlot/Plottable/ScatterPlot.cs
+++ b/src/ScottPlot4/ScottPlot/Plottable/ScatterPlot.cs
@@ -18,6 +18,7 @@ namespace ScottPlot.Plottable
         public double[] XError { get; set; }
         public double[] YError { get; set; }
         public string[] DataPointLabels { get; set; }
+        public Drawing.Font DataPointLabelFont { get; set; } = new();
 
         public enum NanBehavior
         {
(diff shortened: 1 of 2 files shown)


src/ScottPlot4/ScottPlot/Plottable/ScatterPlot.cs
```diff
@@ -426,6 +427,27 @@ public void Render(PlotDimensions dims, Bitmap bmp, bool lowQuality = false)
                 if ((MarkerSize > 0) && (MarkerShape != MarkerShape.none))
                 {
                     MarkerTools.DrawMarkers(gfx, points, MarkerShape, MarkerSize, MarkerColor, MarkerLineWidth);
+                    if (DataPointLabels != null)
```
swharden, line 430: Let's use { } to wrap the full contents of this if statement

swharden: Also, this if statement should be _outside_ the if statement two lines up about markers. I think it makes sense to always show the labels if they are populated (not only if marker size is >0)

swharden (changes requested):

swharden:
This looks great @Cubikon, thanks! If you make the change I requested I'll merge this in good time 👍 🚀

src/ScottPlot4/ScottPlot/Plottable/ScatterPlot.cs
```diff
@@ -422,6 +423,30 @@ public void Render(PlotDimensions dims, Bitmap bmp, bool lowQuality = false)
                     DrawLinesWithGaps(points, gfx, penLine);
                 }
 
+                if (DataPointLabels != null)
```
bclehmann, line 426: I think this should be `DataPointLabels is not null`.

In this case they're the same, but theoretically if we swapped out the type of `DataPointLabels` it is possible we could get different results. And while overloading `==` in such a way as to change the correctness is a little absurd, we could still get (marginally) better performance. If `==` is overloaded the compiler will have to call the overload when we use `==`, rather than just emitting a `ceq` opcode when we use `is`.

src/ScottPlot4/ScottPlot/Plottable/ScatterPlot.cs
```diff
@@ -422,6 +423,30 @@ public void Render(PlotDimensions dims, Bitmap bmp, bool lowQuality = false)
...
+                {
+                    for (var i = 0; i < DataPointLabels.Length; i++)
+                    {
+                        var label = DataPointLabels[i];
+                        if (!string.IsNullOrEmpty(label))
+                        {
+                            Drawing.Font textFont = new();
```
bclehmann, line 433: Perhaps this should be customizable? Although I'm happy leaving this simple and adding customization if and when it's necessary.

src/ScottPlot4/ScottPlot/Plottable/ScatterPlot.cs
```diff
@@ -422,6 +423,30 @@ public void Render(PlotDimensions dims, Bitmap bmp, bool lowQuality = false)
...
+                            var stringSize = GDI.MeasureString(gfx, label, textFont);
```
bclehmann, line 434: This variable is unused

src/ScottPlot4/ScottPlot/Plottable/ScatterPlot.cs
```diff
@@ -422,6 +423,30 @@ public void Render(PlotDimensions dims, Bitmap bmp, bool lowQuality = false)
...
+                            gfx.TranslateTransform(points[i].X, points[i].Y);
+                            gfx.RotateTransform(textFont.Rotation);
```
bclehmann, line 436: I believe `textFont.Rotation` is always zero, so this line is redundant (unless we decide to add customization later)

bclehmann (changes requested):
Looks pretty good, mostly just some minor cleanup around unnecessary code.

bclehmann:
@swharden If you think my change requests are too minor to block merging over than feel free to push it through and I can make a small PR later with them.

src/ScottPlot4/ScottPlot/Plottable/ScatterPlot.cs
swharden: We might as well expose it as a public field `DataPointLabelFont`

I'd make the change myself and merge it now but I don't have write access to the branch 😅

src/ScottPlot4/ScottPlot/Plottable/ScatterPlot.cs
swharden: If we expose `DataPointLabelFont` as described earlier, rotation could be customized, so let's leave this like it is 👍

swharden:
> @swharden If you think my change requests are too minor to block merging over than feel free to push it through and I can make a small PR later with them.

I'd make the changes myself and merge but I don't have write access to the branch...

If we don't hear back from @Cubikon by tomorrow night I'll merge it as it is 👍

Cubikon:
i did the new requested changes. And you have now access to the branch, if you want to change something else.

swharden (approved):

swharden:
Looks great, thanks! 🚀

Issue #2153, opened 2022-10-06
WinForms: Allow detached legend to be redrawn programmatically


swharden:
I think the issue described in #2131 can be resolved by making this function `public`

https://github.com/ScottPlot/ScottPlot/blob/4008bdf2c569277fcf839a5761d40d5c272b60bb/src/ScottPlot4/ScottPlot.WinForms/FormsPlotLegendViewer.cs#L40-L51

### Discussed in https://github.com/ScottPlot/ScottPlot/discussions/2131

<div type='discussions-op-text'>

<sup>Originally posted by **Z0ld3n** September 21, 2022</sup>
Hi, thanks for this amazing library.

I have a question. If I hide series programmatically (by changing .IsVisible in IPlottable) when the legend is detached it is not picking up state change automatically.  Is there a way to force refresh/update the legend when it is in a detached state other than force closing FormsPlotLegendViewer and opening a new one?</div>

rosdyana:
@swharden Thanks, I created this PR https://github.com/ScottPlot/ScottPlot/pull/2157

PR #2157, opened 2022-10-07
Making RefreshLegendImage become a public function

rosdyana:
**Purpose:**
* Fixes: #2153
* Changing RefreshLegendImage from a private to a public function
* Re-ordering private and public functions

```cs
// From private to public function 
public void RefreshLegendImage()
```


bclehmann (approved):
Looks good to me

swharden (approved):

Issue #2156, opened 2022-10-06
FunctionPlot: Optional X Limits


swharden:
Some functions should only be plotted between a range of X values.

Functions which return `null` throw exceptions and make rendering slow.

Public fields should be added so users can restrict function rendering to a range.

```cs
public double XMin = double.NegativeInfinity;
public double XMax = double.PositiveInfinity;
```

Then in the render function those values should be checked and rendering should be restricted to that range.

Presently the render function calculates X/Y pairs for every point from the full visible range of the plot:

https://github.com/ScottPlot/ScottPlot/blob/88bfe26255b22fdc795e1f18c903964eac3592be/src/ScottPlot4/ScottPlot/Plottable/FunctionPlot.cs#L37-L45

This feature may resolve the issue described by @phil100vol in #2138

bclehmann:
I actually think we can implement this without throwing, simply replacing the throws with `continue` should be much faster and it wouldn't impact correctness.

PR #2158, opened 2022-10-07
FunctionPlot: Massive performance boosts for functions with restricted domains

bclehmann:
**Purpose:**
#2156 

Most of the performance benefit comes from removing the exception-based code (that's my bad, it was my idea in the first place). `XMin` and `XMax` are just gravy, and they also are potentially useful in and of themselves, if someone only wants to consider a function on a narrow domain, this is probably an easier way to restrict that domain then manually modifying the function.

```cs
double? Normalised_Function_F(double z)
{
    return 0 <= z && z <= 1 ? z : null;
}
var f = plt.AddFunction(Normalised_Function_F);

f.XMin = 0.3;
f.XMax = 0.7;
```

### DIFF
--- a/src/ScottPlot4/ScottPlot/DoubleExtensions.cs
+++ b/src/ScottPlot4/ScottPlot/DoubleExtensions.cs
@@ -1,0 +1,13 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+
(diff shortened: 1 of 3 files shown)


bclehmann:
I would make a benchmark, but the difference is so stark that before this PR I didn't have the patience to wait for it to render (admittedly in debug mode), and after it's borderline instant

swharden (approved):

Issue #2148, opened 2022-10-04
BarPlot: Tick labels should be displayed below negative bars


swharden:
Presently negative labels are displayed on top of bars

https://scottplot.net/cookbook/4.1/category/plottable-bar-graph/#values-above-bars

![test](https://user-images.githubusercontent.com/4165489/193726363-6eb7d95c-fb5f-48ff-a4d4-182e10c7c3bd.png)

... but it kind of makes sense when the flag is "show values above bars"

The API should extended to allow negative bars to show labels _below_ them

New functionality could be added here:

https://github.com/ScottPlot/ScottPlot/blob/4008bdf2c569277fcf839a5761d40d5c272b60bb/src/ScottPlot4/ScottPlot/Plottable/BarPlot.cs#L83-L87

bclehmann:
Note that after #2152 this applies to `BarSeries` as well

PR #2159, opened 2022-10-07
Bar Plot: Show negative values below bar

bclehmann:
**Purpose:**
#2148 

~~Note that this PR doesn't yet add this for `BarSeries`, I'm writing this between classes so I'll add it later.~~

The other question is whether this should be customizable, is drawing the text above negative bars always wrong (i.e. this is a bug fix), or should the user be able to choose that behaviour (i.e. this is a new feature). If it's the latter, I think we should preserve the previous behaviour of `ShowValuesAboveBars` and add `ShowValuesAtTipOfBar` or something similar.

```cs
double[] values = { 27.3, 23.1, -21.2, 16.1, 6.4, 19.2, 18.7, 17.3, 20.3, 13.1 };

var bar = plt.AddBar(values);
bar.ShowValuesAboveBars = true;
bar.Orientation = Orientation.Horizontal;
```
image

### DIFF
--- a/src/ScottPlot4/ScottPlot/Plottable/Bar.cs
+++ b/src/ScottPlot4/ScottPlot/Plottable/Bar.cs
@@ -69,6 +69,14 @@ public class Bar
         double left = Position - Thickness / 2;
         double right = Position + Thickness / 2;
 
+        // Make room for labels, if present
+        if (Label is not null)
+        {
+            double span = top - bottom;
+            top += 0.1 * span;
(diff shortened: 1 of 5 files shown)


bclehmann:
When adding this to `BarSeries` I also made the following fixes:
- #2152 did not fix the issue for horizontally oriented bars
- `BarPlot`s expand the automatic axis limits when they are showing labels so that the labels don't get cut off. I added that to `BarSeries`

Issue #2146, opened 2022-10-04
Plot.AddBarSeries: accept array too


swharden:
presently it only accepts `List<T>` but it would be nice to optionally pass in a traditional array too

https://github.com/ScottPlot/ScottPlot/blob/4008bdf2c569277fcf839a5761d40d5c272b60bb/src/ScottPlot4/ScottPlot/Plot/Plot.Add.cs#L236-L241

spillover from #2145

KonH:
@swharden Hello! Please take a look at PR - https://github.com/ScottPlot/ScottPlot/pull/2160

PR #2160, opened 2022-10-07
Plot.AddBarSeries: new Bar[] overload

KonH:
**Purpose:**
Related to https://github.com/ScottPlot/ScottPlot/issues/2146  

Now you can use not only List<Bar> but also Bar[] argument.  
I'm not sure about corner cases related to .ToList() call - in the original approach you potentially can change the passed list after, but while converting an array to a list you can't do it. But in my opinion, we can skip this detail to not introduce additional complexity.

### DIFF
--- a/src/ScottPlot4/ScottPlot/Plot/Plot.Add.cs
+++ b/src/ScottPlot4/ScottPlot/Plot/Plot.Add.cs
@@ -241,6 +241,16 @@ namespace ScottPlot
         }
 
         /// <summary>
+        /// Create a BarSeries filled with the given bars, add it to the plot, and return it.
+        /// </summary>
+        public BarSeries AddBarSeries(Bar[] bars)
+        {
+            BarSeries barSeries = new(bars.ToList());
+            Add(barSeries);
+            return barSeries;
+        }
+
+        /// <summary>
         /// Add an empty bubble plot. Call it's Add() method to add bubbles with custom position and styling.
         /// </summary>
         public BubblePlot AddBubblePlot()


swharden (approved):

src/ScottPlot4/ScottPlot/Plot/Plot.Add.cs
```diff
@@ -240,6 +240,16 @@ public BarSeries AddBarSeries(List<Bar> bars)
             return barSeries;
         }
 
+        /// <summary>
+        /// Create a BarSeries filled with the given bars, add it to the plot, and return it.
+        /// </summary>
+        public BarSeries AddBarSeries(Bar[] bars)
```
bclehmann, line 246: I think we can merge the two into `AddBarSeries(IList<Bar> bars)`, as arrays implement `IList<T>`.

There is however a difference in behaviour between the two overloads, the `Bar[]` overload you've written copies into a list, and the existing `List<Bar>` implementation passes the list in directly, meaning mutating the list will be reflected in `BarSeries`.

I think we have to maintain that existing behaviour (probably with a type guard), as we can't change the type of `BarSeries.Bar` because that would be a breaking change. It's potentially confusing that passing in different implementers of `IList<T>` would have different behaviour, so maybe we should separate these into two separate functions with separate names.

swharden:
> I'm not sure about corner cases related to .ToList() call - in the original approach you potentially can change the passed list after, but while converting an array to a list you can't do it. But in my opinion, we can skip this detail to not introduce additional complexity.

I think this is a fine and useful addition for now! We can always add more complexity under the hood later, but this makes it a little easier for people to get off the ground quickly who are working with an array instead of a list.

Thanks for the PR! I'm happy to merge this after @bclehmann approves too 👍 🚀

bclehmann (changes requested):
I like this idea, and I want to expand it to `IList<T>`, not just arrays. Unfortunately that introduces a can of worms because the existing behaviour with `List<Bar>` allows users to mutate the plottable through the argument they pass in, and we can't replicate that with `IList<T>` without breaking changes.

So I think we need two functions with different names, `AddBarSeries(List<Bar>)`, and `AddBarSeriesWithSomeNameToBeDetermined(IList<Bar>)`, where the second function calls `IList.ToList()` and passes it in.

bclehmann:
If @swharden is ok with the differing behaviour (and not supporting IList) I can just approve this now

bclehmann (approved):

swharden:
> If @swharden is ok with the differing behaviour (and not supporting IList) I can just approve this now

I'm okay with this. Your points are very good and well-received - thank you for your thorough review! I agree this has differing behavior (and has the same name), but I think this is a good enough improvement for now that it's worth merging as it is.

I do want to do a better job figuring out what the most elegant bar plot API should be for ScottPlot5 though!

Issue #2010, opened 2022-08-06
Hacktoberfest 2022 🎃🚀


swharden:
### ScottPlot is participating in Hacktoberfest 2022! 🚀 

Our community is friendly to new developers and first-time contributors!

If you have questions about contributing, feel free to ask here the ScottPlot Discord at https://scottplot.net/discord

To get started visit: https://hacktoberfest.com/participation

### Contribute a New Color Palette

Color palettes are a collection of several colors. ScottPlot can be configured to use a custom palette to color new items as they are added to the plot. See [Cookbook: how to use a custom palette](https://scottplot.net/cookbook/4.1/category/palette/#nord) and the [full list of color palettes](https://scottplot.net/cookbook/4.1/colors/) for more information.

To create a new color palette:
* Create a new file in `src/ScottPlot4/ScottPlot/Palettes/`
* Mimic code from the Dark palette in [ScottPlot/Palettes/Dark.cs](https://github.com/ScottPlot/ScottPlot/blob/main/src/ScottPlot4/ScottPlot/Palettes/Dark.cs)
* A code comment at the top can indicate the contributor and/or the source of the palette
* See your palette in action by running the demo application

Ideas for new color palettes may be sourced from: 
* https://color.adobe.com/explore
* http://tsitsul.in/blog/coloropt/ (adding `xgfs_tarnish6` would be great)

### Contribute a New Style

Styles are like themes that describe colors of the plot background, grid lines, labels, etc. A light theme is the default, but users can customize themes.

* Create a new file in `src/ScottPlot4/ScottPlot/Styles/`
* Mimic code from [Blue1.cs](https://github.com/ScottPlot/ScottPlot/blob/main/src/ScottPlot4/ScottPlot/Styles/Blue1.cs)
* See your palette in action by running the demo application

### Add a Cookbook Recipe

The ScottPlot Cookbook (https://scottplot.net/cookbook/4.1/) is generated from code examples in `src/ScottPlot4/ScottPlot.Cookbook/Recipes`. To add a new cookbook recipe, edit one of the files and add a class.

Example cookbook recipe: [Scatter Plot Quickstart](https://scottplot.net/cookbook/4.1/category/quickstart/#scatter-plot)

[...]

PR #2161, opened 2022-10-08
Add pallete based on xgfs_normal6 from Anton Tsitsulin

martinkleppe:
This commit will add the normal pallette based on xgfs_normal6 from Anton Tsitsulin as described in issue #2010.


swharden (approved):

swharden:
Thanks @martinkleppe, this looks great! I'll merge this after @bclehmann approves too 👍

bclehmann (approved):
We should probably add this to ScottPlot 5 as well, but that can be in a later PR.

PR #2174, opened 2022-10-11
Signal: add "smooth" option for low density mode

### LINKED ISSUE #2137: Signal: add "smooth" option for low density mode
The website says you can add a smooth property to a scatter plot, so how do you set a smooth property to a signal plot?

rosdyana:
**Purpose:**
* Fixes #2137 
* Add a smooth option for the Signal plot

from
image
become
image


src/ScottPlot4/ScottPlot/Plottable/SignalPlotBase.cs
```diff
@@ -21,6 +21,8 @@ public abstract class SignalPlotBase<T> : IPlottable, IHasLine, IHasMarker, IHig
         public int YAxisIndex { get; set; } = 0;
         public bool IsVisible { get; set; } = true;
         public bool StepDisplay { get; set; } = false;
+        public bool Smooth { get; set; } = false;
+        public double SmoothTension { get; set; } = 0.5f;
```
bclehmann, line 25: This is always going to be converted to a float before use, so we can probably just expose a float rather than exposing a double and casting it each time.

bclehmann (approved):

bclehmann:
I suppose we should also add this to `SignalXY`, since it uses a different `IPlottable.Render` implementation users will see the new `Smooth` property, but changing it will have no effect.

src/ScottPlot4/ScottPlot/Plottable/SignalPlotBase.cs
swharden: I'm okay exposing the double and later casting to float. 

It's frustrating to always require users to type `.5f` instead of just `.5`

swharden:
> I suppose we should also add this to SignalXY, since it uses a different IPlottable.Render implementation users will see the new Smooth property, but changing it will have no effect.

Perhaps one way to address this is to throw an exception immediately if this is set to `true` in ` SignalXY` 🤔 I'd be happy to review/discuss this in a separate issue or PR though

Thanks so much for this PR @rosdyana 👍 🚀

swharden (approved):

Issue #2183, opened 2022-10-15
AvaPlot doesn't work with Avalonia 11.0.0-preview2


Xerxes004:
## Bug Report

**Issue:** Avalonia 11 has API differences with 0.10.*

**Reproducing:** Create an Avalonia 11 app and try to use `AvaPlot`.

Upon view construction, observe the following exception:

```c#
System.IO.FileNotFoundException: Could not load file or assembly 'Avalonia.Input, Version=0.10.18.0, Culture=neutral, PublicKeyToken=c8d484a7012f9a8b'. The system cannot find the file specified.
File name: 'Avalonia.Input, Version=0.10.18.0, Culture=neutral, PublicKeyToken=c8d484a7012f9a8b'
   at MyView.InitializeComponent(Boolean loadXaml)
   at MyView..ctor() in MyView.axaml.cs:line 22
   at System.RuntimeType.CreateInstanceDefaultCtor(Boolean publicOnly, Boolean wrapExceptions)
```

Avalonia deprecated part of their API, and changed the name of a few events. I'm going to make a PR to fix this rather than discuss what's broken here.

### System Details
* **ScottPlot Version:** 4.1.58
* **Operating System:** Windows 10
* **Application Type:** Avalonia 11.0.0-preview2
* **.NET Version:** .NET 7

PR #2184, opened 2022-10-15
Support Avalonia 11.0.0 preview2

Xerxes004:
**Purpose:**
- Updates AvaPlot and demo to use Avalonia 11.0.0-preview2
- Fixes event names breaking changes
- Uses `IStorageProvider` for save dialog

**See also:**
- https://github.com/ScottPlot/ScottPlot/issues/2183
- https://github.com/AvaloniaUI/Avalonia/discussions/9004

### DIFF
--- a/src/ScottPlot4/ScottPlot.Avalonia/AvaPlot.axaml.cs
+++ b/src/ScottPlot4/ScottPlot.Avalonia/AvaPlot.axaml.cs
@@ -318,7 +318,7 @@ namespace ScottPlot.Avalonia
             var saveFile = await window.StorageProvider.SaveFilePickerAsync(new FilePickerSaveOptions()
             {
                 SuggestedFileName = "ScottPlot.png",
-                FileTypeChoices = new []{
+                FileTypeChoices = new[]{
                     new FilePickerFileType("PNG Files")
                 {
                     Patterns = new[]{"*.png"}
--- a/src/ScottPlot4/ScottPlot.Avalonia/AvaPlot.axaml.cs
+++ b/src/ScottPlot4/ScottPlot.Avalonia/AvaPlot.axaml.cs
@@ -9,6 +9,7 @@ using System;
 using System.Collections.Generic;
 using System.ComponentModel;
 using System.Threading.Tasks;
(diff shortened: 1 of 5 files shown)


Xerxes004:
@swharden 

A note about `IStorageProvider`: The provider favors streams over filenames, probably in an effort to make the API completely agnostic to file path strings. It would be a good idea, in my opinion, to have a `SaveFigure` overload that took a stream instead of a filename to support this. Currently, I just use `Uri` to get the path.

Xerxes004:
I also just ran into a pretty major bug with `System.Drawing.Common`. It has to do with [this breaking change](https://learn.microsoft.com/en-us/dotnet/core/compatibility/core-libraries/6.0/system-drawing-common-windows-only). Microsoft has decided to offload the cross-platform work onto Skia and MAUI implementations, rather than support this "subpar" option.

On my M1 Mac I had to add the recommended *`runtimeconfig.template.json`* file with the following contents:

```json
{
   "configProperties": {
      "System.Drawing.EnableUnixSupport": true
   }
}
```

I added this file to the root of my own project which references ScottPlot. The plot shows up just fine.

I'm looking forward to ScottPlot 5!

Xerxes004:
For anyone who finds this PR, the error I received was:

```
System.TypeInitializationException: The type initializer for 'Gdip' threw an exception.
 ---> System.PlatformNotSupportedException: System.Drawing.Common is not supported on non-Windows platforms. See https://aka.ms/systemdrawingnonwindows for more information.
```

swharden:
HI @Xerxes004, sorry it took so long for me to properly address this.

We had a quick chat in the Discord #1966 and decided that ScottPlot's controls should only target Avalonia releases (not previews). I'll close this PR for now, but when Avalonia 11 is fully released we can open it up again and pick this up from there 👍

bclehmann:
I think another good alternative is merging this into a separate branch, which we can rebase or fix up whenever Avalonia 11 becomes stable.

Xerxes004:
Hey @swharden @bclehmann no problems, I totally understand. For now I'm going to target released Avalonia in a separate project and just use that separately from our main project.

Issue #2196, opened 2022-10-26
Cookbook: Signal plot with date axis


swharden:
How to enable a DateTime axis is demonstrated with scatter plots https://scottplot.net/cookbook/4.1/category/axis-and-ticks/#plotting-datetime-data

A similar cookbook recipe should be added to demonstrate this for signal plots.

(idea came from #2192)

zachesposito:
Hello happy Hacktoberfest! I can take care of this

swharden:
@zachesposito sure thing! I'll keep an eye out for a PR and am prepared to move quickly if it looks good! Let us know if you hit any snags along the way and we can help out. We are on Discord too https://scottplot.net/discord/

PR #2200, opened 2022-10-28
#2196 - add cookbook recipe for datetime axis on a signal plot

zachesposito:
**Purpose:**
Addresses #2196 
New recipe when viewed locally: 
image

### DIFF
--- a/src/ScottPlot4/ScottPlot.Cookbook/Recipes/Axis.cs
+++ b/src/ScottPlot4/ScottPlot.Cookbook/Recipes/Axis.cs
@@ -221,4 +221,30 @@ namespace ScottPlot.Cookbook.Recipes.Ticks
             plt.XAxis.DateTimeFormat(true);
         }
     }
+    class TicksDateTimeSignal : IRecipe
+    {
+        public ICategory Category => new Categories.Axis();
+        public string ID => "ticks_dateTime_signal";
+        public string Title => "Plotting DateTime Data on a Signal Plot";
+        public string Description =>
+            "DateTime can be displayed on the horizontal axis of a signal plot by " +
+            "using the sample rate to set the time interval per data point, and then " +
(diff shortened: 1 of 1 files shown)


src/ScottPlot4/ScottPlot.Cookbook/Recipes/Axis.cs
```diff
@@ -221,4 +221,31 @@ public void ExecuteRecipe(Plot plt)
...
+
+        public void ExecuteRecipe(Plot plt)
+        {
+            // create data sample data
+            double[] ys = DataGen.RandomWalk(100);
+
+            // Since the integral part of an OA date represents days,
+            // need to sample at 24 hours * 60 minutes * 60 seconds to
+            // treat data points as 1 per second
+            var signalPlot = plt.AddSignal(ys, 86400);
```
swharden, line 242: Can we break this into a few lines? Rather than a big comment, it may be more instructive to start with a DateTime or TimeSpan object... something like


```cs
TimeSpan ts = TimeSpan.FromSeconds(1); // time between data points
double sampleRate = (double)ts.Ticks / TimeSpan.TicksPerDay;
var signalPlot = plt.AddSignal(ys, sampleRate);
```

I think it's worth being verbose here because this is the confusing bit

swharden (changes requested):

bclehmann (approved):

src/ScottPlot4/ScottPlot.Cookbook/Recipes/Axis.cs
zachesposito: Yes even better

zachesposito: Corrected 
image

swharden (approved):

Issue #2189, opened 2022-10-22
Normal distribution: generation algorithm


arthurits:
I've been using functions `RandomNormal` from `DataGen.cs` and `Population` from `Population.cs` in order to visualize the distribution of a given variable
https://github.com/ScottPlot/ScottPlot/blob/0be0921e508656e19323642974a8c5d441bc686a/src/ScottPlot4/ScottPlot/Statistics/Population.cs#L40-L44

I'm no expert whatsoever in statistical algorithms, but it seems that both functions make use of `RandomNormalValue` where some kind of Box-Muller algorithm is implemented to generate gaussian numbers:

https://github.com/ScottPlot/ScottPlot/blob/0be0921e508656e19323642974a8c5d441bc686a/src/ScottPlot4/ScottPlot/DataGen.cs#L195-L200

However, the gaussian data generated seems to be sistematically underestimated as shown below, where the grey vertical line represents the mean and the bell curve appears to be slightly offsetted to the left.
![image](https://user-images.githubusercontent.com/63915486/197342246-e8dea69c-9597-475d-ac22-f2da03238428.png)

There's an [article in Wikipedia](https://en.wikipedia.org/wiki/Box%E2%80%93Muller_transform#Basic_form) where the Box-Mueller algorithm is explained. It states that 
_Suppose U1 and U2 are independent samples chosen from the uniform distribution on the unit interval (0, 1)._

Since `Random.NextDouble()` does in fact return the value 0, the current computation of both `u1` and `u2` will eventually return some 1s.

Therefore, I've been using the following modification of `RandomNormalValue`:

```cs
public double SampleGaussian(Random random, double mean, double stdDev)
    {
        double u1 = NextDouble(random);
        double u2 = NextDouble(random);

        double y1 = Math.Sqrt(-2.0 * Math.Log(u1)) * Math.Sin(2.0 * Math.PI * u2);    // Math.Cos is also fine
        return mean + stdDev * y1;

        double NextDouble(Random random)
        {
            return ((double)random.Next(1, Int32.MaxValue)) / Int32.MaxValue;    // random.Next includes 1 and exludes Int32MaxValue
        }
    }
```
This modification provides a more visually mean-centered gaussian curve:
![image](https://user-images.githubusercontent.com/63915486/197345539-21483fe2-499e-4fb9-89ad-5b7da2fb174b.png)

bclehmann:
[...]

PR #2206, opened 2022-10-30
Ensure RandomNormalValue uses u1, u2 from (0,1), not [0,1)

bclehmann:
**Purpose:**
#2189


bclehmann:
I'm noticing that this test is quite fragile, it creates noisy linear data where the standard deviation of the residual is 50, but it expects a fairly tight match.

It's seeded, so it'll continue to pass provided that .NET doesn't change the behaviour of `System.Random`, which they do fairly often, and provided that we don't change our implementation. But I think it's probably a bad test.

Issue #2199, opened 2022-10-27
Dynamic scatterline (SMA, Bollinger) issue


xenedia:
Hi,

I used the scottplot on my own project. there was dynamic candlestick and SMA lines. dynamic candlesticks working good but SMA lines does not calculated and showed as expected.  SMA lines working good on first load but when the new candlesticks was added SMA graphics are incorrect like below. I couldn't find a solution to fix this SMA or Bollinger graphics. could this be an issue or  am i doing something wrong?

```
//definition
private List <OHLC> grafik;
private FinancePlot mcp;
private ScatterPlot scp20;


// first load
mcp = formsPlot1.Plot.AddCandlesticks(grafik.ToArray());
var (xs, ys) = mcp.GetSMA(20);
scp20 = formsPlot1.Plot.AddScatterLines(xs, ys, Color.LightCoral, 1);

// timer
grafik.Add(ohlc);
if (scp20 != null) formsPlot1.Plot.Remove(scp20);
var (xs, ys) = mcp.GetSMA(20);
scp20 = formsPlot1.Plot.AddScatterLines(xs, ys, Color.LightCoral, 1);
formsPlot1.Refresh();
```

<img width="566" alt="image" src="https://user-images.githubusercontent.com/373347/198233776-2dab7cd5-4343-47f0-b494-e391aff4d13d.png">

### System Details
* **ScottPlot Version:** 4.1.58
* **Operating System:** Windows 10
* **Application Type:** WinForms
* **.NET Version:**  .NET 6.0

bclehmann:
It looks like both times you calculate the SMA you're using the same candlestick:
```cs
var (xs, ys) = mcp.GetSMA(20);
```

But you don't actually mutate `mcp`, at least not in the code I can see, noting that calling `grafik.ToArray()` will create a copy, so `mcp` won't see the changes you made to `grafik`.

I will note that the `GetSMA` function assumes the coordinates in `FinancePlot.OHLCs` are sorted, so if they're not in strictly ascending order you could have the SMA graph change directions like it appears to here, as well as calculate incorrect Y values. It appears that we don't ever indicate or enforce that unsorted OHLCs are invalid, so this looks like something we should fix.

PR #2207, opened 2022-10-30
#2199 - make FinancePlot sort OHLCs when getting SMA or Bollinger bands

zachesposito:
**Purpose:**
#2199 highlighted that `FinancePlot`'s assumption that its `OHLCs` are always sorted by datetime can be confusing to users. This change makes sure `OHLCs` gets sorted before trying to calculate SMAs or BollingerBands, so that those methods work even if the user doesn't provide sorted OHLCs.

### DIFF
--- a/src/ScottPlot4/ScottPlot/Plottable/FinancePlot.cs
+++ b/src/ScottPlot4/ScottPlot/Plottable/FinancePlot.cs
@@ -277,8 +277,9 @@ namespace ScottPlot.Plottable
             if (N >= OHLCs.Count)
                 throw new ArgumentException("can not analyze more points than are available in the OHLCs");
 
-            double[] xs = OHLCs.Skip(N).Select(x => x.DateTime.ToOADate()).ToArray();
-            double[] ys = Statistics.Finance.SMA(OHLCs.ToArray(), N);
+            var sortedOHLCs = GetSortedOHLCs();
+            double[] xs = sortedOHLCs.Skip(N).Select(x => x.DateTime.ToOADate()).ToArray();
+            double[] ys = Statistics.Finance.SMA(sortedOHLCs.ToArray(), N);
             return (xs, ys);
         }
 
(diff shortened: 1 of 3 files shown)


src/ScottPlot4/ScottPlot/Plottable/FinancePlot.cs
```diff
@@ -294,9 +295,33 @@ private void RenderOhlc(PlotDimensions dims, Bitmap bmp, bool lowQuality)
...
+            double[] xs = sortedOHLCs.Skip(N).Select(x => x.DateTime.ToOADate()).ToArray();
+            (var sma, var lower, var upper) = Statistics.Finance.Bollinger(sortedOHLCs, N);
             return (xs, sma, lower, upper);
         }
+
+        private OHLC[] GetSortedOHLCs()
+        {
+            if (OHLCsAreSorted())
+            {
+                return OHLCs.ToArray();
```
bclehmann, line 308: If they're already sorted I don't think we need to copy it? We can just return `OHLCs` and we'll save a copy?

src/ScottPlot4/ScottPlot/Plottable/FinancePlot.cs
```diff
@@ -294,9 +295,33 @@ private void RenderOhlc(PlotDimensions dims, Bitmap bmp, bool lowQuality)
             if (N >= OHLCs.Count)
                 throw new ArgumentException("can not analyze more points than are available in the OHLCs");
 
-            double[] xs = OHLCs.Skip(N).Select(x => x.DateTime.ToOADate()).ToArray();
-            (var sma, var lower, var upper) = Statistics.Finance.Bollinger(OHLCs.ToArray(), N);
+            var sortedOHLCs = GetSortedOHLCs();
```
bclehmann, line 298: Nice, I didn't realize that this function was affected as well

bclehmann (approved):

src/ScottPlot4/ScottPlot/Plottable/FinancePlot.cs
zachesposito: Good call that is simpler, fixed

swharden (approved):

PR #2215, opened 2022-11-02
AvaPlot: fix subscription to ContexMenu property changes

DmitryZhelnin:
**Purpose:**
* This PR fixes subscription to changes of `ContextMenu` property of `AvaPlot` control
* Previous implementations subscribes to `ContextMenu` changes of **ALL** controls which is wrong. As a result custom logic of `AvaPlot` `ContextMenu` opening applies to another controls and breaks their behavior.

### DIFF
--- a/src/ScottPlot4/ScottPlot.Avalonia/AvaPlot.axaml.cs
+++ b/src/ScottPlot4/ScottPlot.Avalonia/AvaPlot.axaml.cs
@@ -75,6 +75,11 @@ namespace ScottPlot.Avalonia
         [Obsolete("Reference Plot instead of plt")]
         public ScottPlot.Plot plt => Plot;
 
+        static AvaPlot()
+        {
+            ContextMenuProperty.Changed.AddClassHandler<AvaPlot>((plot, args) => plot.ContextMenuChanged(args));
+        }
+
         public AvaPlot()
         {
             InitializeComponent();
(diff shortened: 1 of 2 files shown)


src/ScottPlot4/ScottPlot.Avalonia/AvaPlot.axaml.cs
```diff
@@ -75,6 +75,11 @@ public partial class AvaPlot : UserControl
         [Obsolete("Reference Plot instead of plt")]
         public ScottPlot.Plot plt => Plot;
 
+        static AvaPlot()
```
bclehmann, line 78: I'm not quite familiar, but do all `ContextMenu`s forward their events to this instance?

In any case, I don't think this (or the existing subscription) will fire if someone swaps out the `ContextMenu` instance rather than mutate it.

e.g. `avaPlot.ContextMenu = foo` rather than `avaPlot.ContextMenu.Items = foo`

This is probably incorrect, and I think we might to add some logic to `AvaPlot_PropertyChanged` to handle that case. In principle we could add a handler for `this.ContextMenu` whenever it changes, but I do like how you used the static constructor for it instead.

bclehmann: That was a convoluted comment, so I think I'll summarize:
- I don't think `avaPlot.ContextMenu = foo` will fire the event
- I think this can be solved by modifying `AvaPlot_PropertyChanged`, but I'm open to more elegant approaches

bclehmann (changes requested):
Looks good, unfortunately I think there was another problem with our implementation. If you'd like to take a look at fixing it that'd be great, otherwise I can change it up after the fact.

bclehmann (approved):

Issue #580, opened 2020-10-05
Horizontal label draw


Bajkman:
Hi,

Is it possible to have labels drawn horizontally, so that they don't overdraw the plot?

Best regards,
Greg
![Capture1](https://user-images.githubusercontent.com/72062127/95045558-91988700-06e2-11eb-80d2-d0bb86d6f42a.PNG)

swharden:
Hi @Bajkman, currently this is not possible. You can call `plt.GetLegendBitmap()` to get the legend as a `Bitmap` which you can apply to a picturebox elsewhere on your Form. This is the best way to ensure your legend does not interfere with your data.

Supporting horizontal legend items (or even column configuration) is a good idea, and I've added it to the list of features I'd like to eventually support in #412. Thank you for the suggestion!

Scott

PR #2216, opened 2022-11-03
[#580] Added possibility to choose legend orientation (vertical/horizontal). 

lucabat:
**Purpose:**
Added support for horizontal legend orientation.
It was needed in one of my projects, so I tried to implement it for everyone. 
I tried to keep it as simple as possible.

This solves issue https://github.com/ScottPlot/ScottPlot/issues/580: Arrange legend items horizontally (instead of stacked vertically)

```cs
// Usage example
var legend = wpfPlot1.Plot.Legend();
legend.Orientation = LegendOrientation.Horizontal; //legend items placed horizontally
legend.Orientation = LegendOrientation.Vertical; //legend items placed vertically
```


src/ScottPlot4/ScottPlot/Renderable/Legend.cs
```diff
@@ -98,22 +100,50 @@ public Bitmap GetBitmap(bool lowQuality = false, double scale = 1.0)
             // determine maximum label size and use it to define legend size
             float maxLabelWidth = 0;
             float maxLabelHeight = 0;
+            float totLegendWidth = 0;
```
bclehmann, line 103: This is confusing, because this variable only holds the total legend width _if_ the legend is horizontal. If it's vertical, it doesn't hold anything of meaning.

Not sure what to rename it to, `legendWidthHorizontal` isn't really my favourite, but it's perhaps an idea.

src/ScottPlot4/ScottPlot/Renderable/Legend.cs
```diff
@@ -191,6 +220,88 @@ public Bitmap GetBitmap(bool lowQuality = false, double scale = 1.0)
             }
         }
 
+        private void RenderOnBitmapHorizontal(Graphics gfx, LegendItem[] items, System.Drawing.Font font,
```
bclehmann, line 223: This function and `RenderOnBitmapVertical` have a lot of commonality, I think they can be refactored so that we're not repeating ourselves quite so much. All of the code for actual drawing is the complete same, the only difference is where we do that drawing.

We could refactor this to have functions like `DrawLegendItem`, which could be used by both horizontal and vertical legends, they would just call them with different coordinates.

For example, in ScottPlot 5, we have [`RenderLegendItem`](https://github.com/ScottPlot/ScottPlot/blob/main/src/ScottPlot5/ScottPlot5/Plottables/Legend.cs#L64-L90), which could be reused for horizontal legends (though they aren't supported in v5).

bclehmann: FWIW, we probably should've refactored it to separate rendering the individual legend items from the legend as a whole, so this is more of a preexisting issue that became more important because of your PR, rather than anything that your PR introduced.

bclehmann (changes requested):
Looks really good, just a few suggestions mostly around duplicated code.

src/ScottPlot4/ScottPlot/Renderable/Legend.cs
```diff
@@ -244,4 +355,10 @@ public void UpdateLegendItems(Plot plot, bool includeHidden = false)
             }
         }
     }
+
+    public enum LegendOrientation
```
bclehmann, line 359: I forgot to mention in the main body of my review, but we already have an `Orientation` enum here: https://github.com/ScottPlot/ScottPlot/blob/81735684448d35d0939712e1f5389d4e4e2c3e35/src/ScottPlot4/ScottPlot/Renderable/Orientation.cs

bclehmann:
For anyone else who wants to review, it looks like this: 
image

If you want to clean up this code, it could probably be a good contribution to the cookbook
```cs
// sample data
double[] xs = DataGen.Consecutive(51);
double[] sin = DataGen.Sin(51);
double[] cos = DataGen.Cos(51);

// plot the data
plt.AddScatter(xs, sin).Label = "Sine";
plt.AddScatter(xs, cos).Label = "Cosine";

// customize the axis labels
plt.Title("ScottPlot Quickstart");
plt.YLabel("Vertical Axis");

plt.Legend().Orientation = Renderable.LegendOrientation.Horizontal;
```

src/ScottPlot4/ScottPlot/Renderable/Legend.cs
lucabat: I found two Orientation enums, the second is here: https://github.com/ScottPlot/ScottPlot/blob/81735684448d35d0939712e1f5389d4e4e2c3e35/src/ScottPlot4/ScottPlot/Enums/Orientation.cs

I deleted the one in Renderable/Orientation and kept the one in Enums/Orientation

src/ScottPlot4/ScottPlot/Renderable/Legend.cs
lucabat: Called `totalLegendWidthWhenHorizontal`

src/ScottPlot4/ScottPlot/Renderable/Legend.cs
lucabat: Refactored on commit https://github.com/ScottPlot/ScottPlot/pull/2216/commits/437a273ddd6184cdd33138009e33e2e608c53a10.

Now RenderOnBitmap sets the offsets properly and calls the common method DrawLegendItem

(LucaBattEL it's always me, I committed with the wrong account, sorry)

src/ScottPlot4/ScottPlot/Renderable/Legend.cs
bclehmann: I don't think you can delete the one in Renderable/Orientation, that would be a breaking change.

It looks like that enum is unused now, and the last reference I can find to it comes from ScottPlot 4.0, the last release to which was in August 2021, having been made obsolete by 4.1 in May 2021. Even so, since it's public, it could be used by consumers of the library, and I don't want to break their code.

PR #2217, opened 2022-11-03
Added Generic SignalPlot

codecrafty:
Implemented a SignalPlotGeneric class that allows types such as Short to be used, versus only Double.

SignalPlot is meant to be used to enhance performance, so I didn't like the fact that I have to convert my int arrays to double arrays when using SignalPlot.  The SignalPlotGeneric class now allows different types to be used, e.g.:

```cs
var plot = SignalPlotGeneric<ushort>(myShortsArray);
```


swharden:
This is a fantastic improvement @codecrafty! I added a cookbook recipe to demonstrate this and intend to release the next version of this package on NuGet today 👍 🚀

PR #2218, opened 2022-11-03
ScottPlot 5: Mirrored Axes

bclehmann:
**Purpose:**
In ScottPlot 5 there isn't a way to associate multiple x-axes or multiple y-axes with a single plottable, and this makes it difficult to have (for example) a bottom and a top axis showing the same ticks.

My solution was to simply create a `MirroredXAxis` and `MirroredYAxis`, which mirror the given axis onto the given edge. It derives its ticks and range entirely from the axis it is mirroring.

```cs
var xaxis2 = new MirroredXAxis(avaPlot.Plot.XAxis, Edge.Top);
avaPlot.Plot.XAxes.Add(xaxis2);

var yaxis2 = new MirroredYAxis(avaPlot.Plot.YAxis, Edge.Right);
avaPlot.Plot.YAxes.Add(yaxis2);
```
image

### DIFF
--- a/src/ScottPlot5/ScottPlot5/Axis/StandardAxes/MirroredXAxis.cs
+++ b/src/ScottPlot5/ScottPlot5/Axis/StandardAxes/MirroredXAxis.cs
@@ -1,0 +1,16 @@
+﻿
+namespace ScottPlot.Axis.StandardAxes;
+public sealed class MirroredXAxis : XAxisBase, IXAxis
+{
+    private readonly Edge _edge;
+    private readonly IXAxis _axis;
+    public override Edge Edge => _edge;
+    public override CoordinateRange Range => _axis.Range;
+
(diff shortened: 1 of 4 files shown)


bclehmann:
It may even be worth defining `Plot.XAxis2 IXAxis => new MirroredXAxis(this.XAxis, Edge.Top)`, so that the user can toggle thiss on more easily.

However, this may complicate things should the user swap out the default `XAxis`.

swharden:
Thanks for this @bclehmann!

> It may even be worth defining Plot.XAxis2 IXAxis => new MirroredXAxis(this.XAxis, Edge.Top), so that the user can toggle this on more easily.

I'm happy to cross that bridge when we get a little closer to it... perhaps after I make the cookbook and start adding recipes

Issue #2245, opened 2022-11-08
ScottPlot 5 Beta


swharden:
This issue tracks work toward creating a ScottPlot 5 preview package

- [ ] publish a preview package to NuGet
- [ ] documentation in this repo
- [ ] documentation on scottplot.net ([repo](https://github.com/ScottPlot/ScottPlot.NET))
- [x] start a cookbook
  * #2246
- [ ] start a demo app
- [x] create a triage/todo list for refinements to add later
- * [`ScottPlot5/todo.md`](https://github.com/ScottPlot/ScottPlot/blob/main/src/ScottPlot5/todo.md)

ckcks12:
Hello @swharden 
I really like ScottPlot and now yearning for its newer version. the 5.
I want to use it without dependencies (System.Drawing)
Also I don't wanna bother nor hurry you up so 

is there anything I can help you for publishing version 5?

swharden:
closing this issue, tracking progress in #2304

swharden:
Hi @ckcks12, thank you for your kind comments and offer to help! I'll probably keep my head down and focus pretty intensely for the next few days on ScottPlot 5, but if you're interested in participating or helping out you could check out what's going on in the Discord #1966 or offer to help answer peoples questions as they post new issues over the next several days and weeks! Any help that comes from the community is time that I get to save and devote back into making ScottPlot better, so helping others is a big help to me too 🚀

PR #2247, opened 2022-11-09
Extend ScottPlot 5 cookbook

swharden:
#2245


swharden:
I got stuck because in order for recipes to save themselves to the correct folders they must be aware of the page they're on

the test runner should look this up somehow but do it efficiently to prevent excessive reflection

Issue #2248, opened 2022-11-09
Pie Chart: 100% pie is missing a sliver


swharden:
Noted by _This Is Not Me_ in the Discord (#1966)

![image](https://user-images.githubusercontent.com/4165489/200962756-efe720bf-4405-41d4-b056-edcd5cd3ebd9.png)

swharden:
I think the problem may be in [`System.Drawing.Graphics.FillPie()`](https://learn.microsoft.com/en-us/dotnet/api/system.drawing.graphics.fillpie)

If this is a case, a fix is to detect a 100% slice and draw a circle instead of a pie 💁 

https://github.com/ScottPlot/ScottPlot/blob/2aa8a16ae52dddbe249ecbab6824ba7af1195d43/src/ScottPlot4/ScottPlot/Plottable/PiePlot.cs#L156-L162

swharden:
resolved by #2251

PR #2252, opened 2022-11-10
ScottPlot 5: Fix pies of 100%

bclehmann:
**Purpose:**
Pie slices taking up 100% were broken:
image

They are now fixed:
image

```cs
avaPlot.Plot.Add.Pie(new double[] { 1 });
```

Unlike #2248, they were broken even for non-exploded pies.


swharden:
> Pie slices taking up 100% were broken

🤣🤣🤣

PR #2258, opened 2022-11-12
Plot.AddFill(): Improve styling and low-point-count polygon support

Fruchtzwerg94:
This PR targets adding a fill is not consistend and not working for data sets of length 1.

Setting a baseline [Plot.Add.cs#L346](https://github.com/ScottPlot/ScottPlot/blob/main/src/ScottPlot4/ScottPlot/Plot/Plot.Add.cs#L346) like
```
WpfPlot1.Plot.AddSignal(new double[] { 1 });
WpfPlot1.Plot.AddFill(new double[] { 0 }, new double[] { 2 }, 0);
```

generates a vertical line like expected:
grafik

The second method [Plot.Add.cs#L381](https://github.com/ScottPlot/ScottPlot/blob/main/src/ScottPlot4/ScottPlot/Plot/Plot.Add.cs#L381)
```
WpfPlot1.Plot.AddSignal(new double[] { 1 });
WpfPlot1.Plot.AddFill(new double[] { 0 }, new double[] { 0 }, new double[] { 2 });
```

fails and throws an exception due to missing points [https://github.com/ScottPlot/ScottPlot/blob/main/src/ScottPlot4/ScottPlot/Plottable/Polygon.cs#L84](Polygon.cs#L84):

> polygons must contain at least 3 points

This fix targets this by ensuring at least 3 points are available which results in the expected plot here as well:
grafik

Additionally the PR aligns styling the lines for all methods. So note that `lineWidth: 1.0` needs tob e added as parameter to ensure the lines is visible.


swharden:
Hi @Fruchtzwerg94, thank you for this excellent PR! Sorry for the delay in getting to this one (see #2304 for details). These improvements will be in the next release 🚀

PR #2268, opened 2022-11-22
Additional drag events for Span

StendProg:
**Purpose:**

Span is useful for selecting ranges on plot. Often, in addition, the user is given the opportunity to set the range of values ​​manually, for example, through text fields. Keeping these text fields and span values ​​in sync always requires a lot of helper code. Especially when the range is given as From/To and From should always be less than To. The current implementation of Span cannot provide this. The user can easily swap the edges of the Span and Edge1 will become larger than Edge2.

In this PR, the span itself decides which edge has been changed (minimum or maximum) and raise the corresponding event. In addition, the events already contain the changed value in their arguments. This should make it much easier to keep the Span
 boundary values ​​in sync with other UI elements.
Additionally added drag events directly to Edge1 and Edge2. In some cases, they may also come in handy.

```cs
 var vSpan = plt.AddHorizontalSpan(0, 50);
 vSpan.MinDragged += (s, e) => minText.Label = $"Min: {e}";
 vSpan.MaxDragged += (s, e) => maxText.Label = $"Max: {e}";
```


swharden:
Thank you for this new feature @StendProg! It's always great to review your work. Sorry for the delay in merging this (see #2304 for details)

PR #2274, opened 2022-11-27
ScottPlot 5: Refactor layout system to use IPanel, adds ColorBar

bclehmann:
**Purpose:**
This refactors `ILayoutSystem` to act on `IPanel` rather than `IAxis`. It also strips a lot of rendering state from `IAxis` like `Offset`, which is instead stored separately. This means we don't expose (and have to keep in sync) rendering state to the user.

~~This PR is intended to enable colorbar axes, but currently the only non-axis plot is a `BoxPanel` intended for debugging purposes.~~

This enables ColorBars (warning, garish):
image

Source:
```cs
int width = 100;
int height = 100;

double[,] intensities = new double[width, height];

for (int x = 0; x < width; x++)
    for (int y = 0; y < height; y++)
        intensities[x, y] = (Math.Sin(x * .2) + Math.Cos(y * .2)) * 100;

avaPlot.Plot.Plottables.Add(DebugPoint);
var hm = avaPlot.Plot.Add.Heatmap(intensities);

avaPlot.Plot.Add.ColorBar(hm, Edge.Top);
avaPlot.Plot.Add.ColorBar(hm, Edge.Right);
avaPlot.Plot.Add.ColorBar(hm, Edge.Bottom);
avaPlot.Plot.Add.ColorBar(hm, Edge.Left);
```

### DIFF
--- a/src/ScottPlot5/ScottPlot5/Axis/AxisRendering.cs
+++ b/src/ScottPlot5/ScottPlot5/Axis/AxisRendering.cs
@@ -58,8 +58,6 @@ public static class AxisRendering
         surface.Canvas.DrawLine(x1, y1, x2, y2, paint);
     }
 
-    static int i = 0;
-    static int j = 0;
     public static void DrawLabel(SKSurface surface, PixelRect dataRect, Edge edge, Label label, float pixelSize)
     {
         using var paint = label.MakePaint();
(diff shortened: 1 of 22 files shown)


swharden:
Dang, this is a huge amount of work! Thanks for this! I'll get in there and start poking at the details over the next few days (see #2304 for details) and reach out on Discord if any questions pop up

Issue #2278, opened 2022-11-29
If plots are small enough, sometimes there aren't enough ticks


Xerxes004:
## Bug Report

**Issue:** My software sometimes creates very small plots, roughly 340 by 190 pixels. Sometimes, only a single plot tick label shows up. One label is never enough to tell what the scale of the data is, and it's hard to figure out. I tried to solve this by setting Tick Density to 2, but I still have plots with only one label (and sometimes even zero, although this is very rare).

**Reproducing:** Run the "Many Plots" demo, and look around. There's always a few plots in there that only have a single tick label.

![missing tick bug](https://user-images.githubusercontent.com/9276814/204673009-cfaa5bea-9d77-4ff6-bc9a-24a51f71010a.PNG)

See #2280 

### System Details
* **ScottPlot Version:** 4.1.59
* **Operating System:** Windows 10 Pro
* **Application Type:** WPF
* **.NET Version:** .NET 7

Xerxes004:
After digging through the code a bit more, I think I found the bug.

In [`TickCollection.cs`]() there is a check for when `tickPositionsMajor.Length` is less than 2. This is an error for numeric plots; if only a single tick is present, then adding or subtracting 1 tick space from it **by definition** cannot be on the axis. In other words: if you get one tick, it's because the spacing is too large.

This code is moot for numeric labels:
https://github.com/ScottPlot/ScottPlot/blob/c168325f37f1c79bfe9b13072cf8f2806ed9b221/src/ScottPlot4/ScottPlot/Ticks/TickCollection.cs#L310-L316

The corner case needs to be handled in `GetIdealTickSpacing` when the first-choice ideal spacing is greater than half the range. Sometimes, both ticks will fit on the axis. Sometimes they won't. To ensure they always do, the next-smallest spacing needs to be chosen instead.

I will make a PR with a potential solution.

swharden:
resolved by #2280

PR #2280, opened 2022-11-30
Fix bug where only one tick label appears sometimes

Xerxes004:
**Purpose:**

Sometimes, only one tick label would appear on an axis. This was because sometimes the 'ideal' tick spacing wasn't ideal after all. When calculating the ideal tick spacing, I check to see if it's possible the spacing could be too high, then select the next-smaller spacing instead. 

See https://github.com/ScottPlot/ScottPlot/issues/2278


Xerxes004:
I noticed that same Mac test fails in other people's pull requests as well. Not sure what the deal is with that, but not related to this PR. Looks to be some kind of color test.

swharden:
Thanks for this @Xerxes004! Sorry it took me longer to get to this than typical (see #2304 for details)

> I noticed that same Mac test fails in other people's pull requests as well. Not sure what the deal is with that, but not related to this PR. Looks to be some kind of color test.

Something crazy happened to the cloud mac build computers... I'll figure it out later 😅 This PR looks great, thanks again!

PR #2336, opened 2022-12-29
Adding transparency and fixed color feature to Heatmaps

bukkideme:
**Purpose:**
Hello! Following the discussion at #2277 . I tried to add some experimental feature to Heatmaps. Now it is possible to define an alphaMask. If this mask is defined, the transparency will be used from this 2D array. Beside, a fix color option is also added. Maybe this fixed color option is a not nice hack to be included in the Heatmap class, this should be discussed. 

**Motivation:**
We often need to quickly add a transparent mask over another Heatmap. We often need a fixed color for this top layer, when we only need to show the layer over certain areas with fixed color (transparency > 0, < 1, fixed color).

```cs
//Mona Lisa with fixed yellow color mask:

plt.AddHeatmap(imageData, colormap: ScottPlot.Drawing.Colormap.Viridis);

double?[,] transparencyMask = new double?[imageData.GetLength(0), imageData.GetLength(1)];
for (int i = 0; i < transparencyMask.GetLength(0); i++)
{
    for (int j = 0; j < transparencyMask.GetLength(1); j++)
    {
        if (i < transparencyMask.GetLength(0) / 2) transparencyMask[i, j] = 0.5;
    }
}
plt.AddHeatmap(Color.Yellow, transparencyMask);

plt.SaveFig("heatmap_image.png");
```

heatmap_image


swharden:
> Hi @swharden ! I created my very first PR ever, so sorry if I did not do something properly :)

Hi @bukkideme, at first glance this looks great! Congratulations on creating your very first PR 🚀

@bclehmann is the heatmap expert. Ben, do you want to review this PR? I don't want to overwhelm a new contributor with too many issues to resolve, so perhaps we can mark recommended changes as required vs. optional, and if any of them are particularly thorny I'm happy to step in and make the modifications directly to this PR right before merging it. Also if you're too busy for this now that's cool, I can get to it in the next few days.

swharden:
@bukkideme regarding the failing test, run the autoformatter script, commit the changes, and push them up to GitHub. The tests should all pass after that!

Windows:
```bash
src\ScottPlot4\autoformat.bat
```

Linux/MacOS:
```
src/ScottPlot4/autoformat.sh
```

bukkideme:
@swharden , thanks for the hint, it fixed the formatting! By the way, during the modifications, I have found a code line which looks like does not have any purpose, might be a remanent of an older modification:
https://github.com/ScottPlot/ScottPlot/blob/3348154474e8c68f3d970ff458d5c89237fa574f/src/ScottPlot4/ScottPlot/Plottable/Heatmap.cs#L257
Is it allowed to delete such orphaned local variables during dev, or it should be addressed as a separate modification request?

swharden:
> Is it allowed to delete such orphaned local variables during dev, or it should be addressed as a separate modification request?

You're right, this may have been used in the past but it no longer being used. Deleting it is fine by me!

src/ScottPlot4/ScottPlot/Drawing/Colormap.cs
```diff
@@ -151,6 +151,32 @@ public static int[] GetRGBAs(double[] intensities, Colormap colorMap, double min
...
         }
 
+        public static int[] GetRGBAs(double[] intensities, double[] alphaVals, Colormap colorMap)
+        {
+            int[] rgbas = new int[intensities.Length];
+            for (int i = 0; i < intensities.Length; i++)
+            {
+                byte pixelIntensity = (byte)Math.Max(Math.Min(intensities[i] * 255, 255), 0);
+                var (r, g, b) = colorMap.GetRGB(pixelIntensity);
+                byte alpha = (byte)Math.Max(Math.Min(alphaVals[i] * 255, 255), 0);
```
bclehmann, line 161: I think this is equivalent to the simpler `Math.Clamp(alphaVals[i] * 255, 0, 255)`, however, it may be simpler to just take in `alphaVals` as a `byte[]` instead, seeing as we only use it as a byte. This would also use 1/8 as much memory.

I think this applies everywhere but I just wrote it the once.

bclehmann (approved):
Looks good, potentially an option to simplify things by using `byte[]` as an alpha mask rather than `double[]`.

I'm not enthused by all the overloads required but I think that's more a matter of how we've handled data with gaps kinda poorly with heatmaps, requiring lots of nullable and non-nullable overloads.

src/ScottPlot4/ScottPlot/Drawing/Colormap.cs
swharden: Note that `Math.Clamp()` is only available in .NET 6 and newer, so we can't use it since we target .NET Framework 😝

swharden: > Looks good, potentially an option to simplify things by using `byte[]` as an alpha mask rather than `double[]`

This is an interesting choice... I guess it comes down to whether the heatmap API is more like an _image_ or if that concept is abstracted away and we interact with the heatmap more like _data_.

Consider these alternative names for the `alphaVals` argument in `GetRGBAs()`: 

* When regarding heatmap values as image pixel intensities, `byte[,]? alpha` makes sense with a range 0-255

* When regarding heatmap values as data measurements, `double[,]? opacity` makes sense with a range 0-1

I think I'm leaning toward the second option because it's a little more easy to reason about when plotting non-image data as heatmaps. Sure it's converted to images _under the hood_, but I think the heatmap API should abstract that complexity away from the users who don't need to think about it.

bclehmann: I would agree, but the parameter is called `alphaVals`, so we're clearly alluding to an image's alpha channel.

It's also worth noting that if a user is working with colours, they're working with `System.Drawing.Color`, which exposes itself as a struct of four 8-bit colour channels, so dealing with a fractional opacity might even be less intuitive to users.

swharden: I think we're saying the same thing.

* If we keep calling it "alpha", `byte` makes sense with a range 0-255

* If we want to change it to "opacity", `double` makes sense with a range 0-1

My lean is toward the second one, which seems to be to me a little more consistent with `double[] intensities` in being more data-ish and less image-ish

Also I recognize this is splitting hairs 🤷 and I'm happy with either outcome. I'll wait to hear @bukkideme's thoughts and after this conversation is resolved I'm looking forward to merging this in 🚀

bukkideme: Hi! Actually i can agree and reason with both sides. I think it is a bit semantics question too. 

But what about if we rename the names "alpha" to "opacity" everywhere, and keep it double? Even if it is reduced to bytes in this very implementation. We work with data which has more meaning than only the visualizable alpha channel, so keeping it double serves us better. For example, such mask matrix can be used really as double when we use it for data manipulation. 
Another thing, software Gwyddion also defines masks as double 2D arrays, even if using it as a binary 0/1 matrix in present version. Keeping it double makes it more future proof.

bukkideme:
Regarding to @bclehmann :
"I'm not enthused by all the overloads required but I think that's more a matter of how we've handled data with gaps kinda poorly with heatmaps, requiring lots of nullable and non-nullable overloads."

I totally agree with you! I think the fixed color option could be a temporary solution, until we figure out a robust option to handle multiple colorbars. Using multiple colobars would eliminate the need of this overload, since we could clip the second heatmap which is represented as a mask, to a fix color. 
But I am not sure what would be the most compact and best solution in this case. What do you think about this? Nevertheless, the fix color option is a huge help for us, until we handle this better :)

src/ScottPlot4/ScottPlot/Drawing/Colormap.cs
swharden: > I can agree and reason with both sides

We have two good ideas here! I'll step in and pick one and merge this PR, and we can always change it later if a strong reason to arises.

I'm going to resolve this conversation and merge now, but we can continue the discussion or additional modifications in new PRs as needed

> Keeping it double makes it more future proof

I like this plan!

swharden:
I'm adding a cookbook example too 

image

```cs
double?[,] data = DataGen.SampleImageDataNullable();

var hm1 = plt.AddHeatmap(Color.Red, data, lockScales: false);
hm1.OffsetX = 0;
hm1.OffsetY = 0;

var hm2 = plt.AddHeatmap(Color.Green, data, lockScales: false);
hm2.OffsetX = 30;
hm2.OffsetY = 20;

var hm3 = plt.AddHeatmap(Color.Blue, data, lockScales: false);
hm3.OffsetX = 60;
hm3.OffsetY = 40;
```

PR #2340, opened 2022-12-31
ScottPlot 5 Legend: Fixes symbols, reduces remeasuring

bclehmann:
**Purpose:**
- Fixes an issue where children of a legend item could be duplicated
- Fixes an issue where legend symbols did not respect padding
- Eliminates remeasuring of child legend items
- Fixes indentation of child legend items

Outstanding issues:
- You cannot customize what items go in a legend (this might be out of scope for this issue)

Before:
image

After:
image

### DIFF
--- a/src/ScottPlot5/ScottPlot5/Legends/StandardLegend.cs
+++ b/src/ScottPlot5/ScottPlot5/Legends/StandardLegend.cs
@@ -63,10 +63,10 @@ public class StandardLegend : ILegend
         // measure all items to determine dimensions of the legend
         using SKPaint paint = new() { Typeface = Font.ToSKTypeface(), TextSize = Font.Size, IsAntialias = true };
         SizedLegendItem[] sizedItems = GetSizedLegendItems(items, paint);
-        
+
         float maxWidth = sizedItems.Select(x => x.Size.WithChildren.Width).Max();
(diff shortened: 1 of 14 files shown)


src/ScottPlot5/ScottPlot5/Legends/StandardLegend.cs
```diff
@@ -33,65 +60,64 @@ public void Render(SKCanvas canvas, PixelRect dataRect, LegendItem[] items)
         if (!items.Any())
             return;
 
-        // TODO: I don't like the recursion or frequent measurements that happen in the render loop.
```
bclehmann, line 36: I assumed that this TODO is fixed, but you can be the judge of that.

swharden:
Thank you so much for this @bclehmann! This is a great improvement.

I plan to rework many style classes (Fill, Stroke, Font, Marker, etc.) but until that work is done I'd like to prevent other code from using them. For this reason I'll revert your `Stroke` back into distinct `LineWidth` and `LineColor`, but definitely soon in the future when the time is better I will refactor the whole code base to use more intelligent style objects.

I'll make that change and merge within the next several minutes 

> You cannot customize what items go in a legend (this might be out of scope for this issue)

I'll create a new `ManualLegend` object this morning and add a cookbook recipe for it

PR #2350, opened 2023-01-02
ScottPlot 5: Context Menus

bclehmann:
**Purpose:**
Adds context menu support to all controls. The default context menu items are "Save Image" and "Copy to Clipboard"*

*Note that copying images to the clipboard is non-trivial in Avalonia, so AvaPlot only has "Save Image".

### DIFF
--- a/src/ScottPlot5/ScottPlot5/Plot.cs
+++ b/src/ScottPlot5/ScottPlot5/Plot.cs
@@ -350,6 +350,30 @@ public class Plot : IDisposable
         return new(surface.Snapshot());
     }
 
+    public void SaveJpeg(string filePath, int? newWidth = null, int? newHeight = null, int quality = 85)
+    {
+        using Image image = GetImage(newWidth, newHeight);
+        image.SaveJpeg(filePath, quality);
+    }
+
+    public void SavePng(string filePath, int? newWidth = null, int? newHeight = null)
+    {
+        using Image image = GetImage(newWidth, newHeight);
+        image.SavePng(filePath);
+    }
+
+    public void SaveWebp(string filePath, int? newWidth = null, int? newHeight = null, int quality = 85)
+    {
+        using Image image = GetImage(newWidth, newHeight);
+        image.SaveWebp(filePath, quality);
+    }
+
+    public void Save(string filePath, int? newWidth = null, int? newHeight = null, ImageFormat format = ImageFormat.Png, int quality = 85)
+    {
+        using Image image = GetImage(newWidth, newHeight);
+        image.Save(filePath, format, quality);
+    }
+
     #endregion
 
     #region Shortcuts
--- a/src/ScottPlot5/ScottPlot5/Plot.cs
+++ b/src/ScottPlot5/ScottPlot5/Plot.cs
@@ -6,6 +6,7 @@ using ScottPlot.Legends;
 using ScottPlot.Benchmarking;
 using ScottPlot.Control;
 using ScottPlot.Stylers;
+using SkiaSharp;
 
 namespace ScottPlot;
 
@@ -29,7 +30,6 @@ public class Plot : IDisposable
     public IZoomRectangle ZoomRectangle { get; set; }
     internal RenderDetails LastRenderInfo { get; set; } = new();
 
-    public PixelRect FigureRect => LastRenderInfo.FigureRect;
     public PlotStyler Style { get; }
 
     /// <summary>
@@ -330,8 +330,11 @@ public class Plot : IDisposable
         LastRenderInfo = Renderer.Render(surface, this);
     }
 
-    public byte[] GetImageBytes(int width, int height, ImageFormat format = ImageFormat.Png, int quality = 100)
+    public Image GetImage(int? newWidth = null, int? newHeight = null)
     {
+        int width = newWidth ?? (int)LastRenderInfo.FigureRect.Width;
+        int height = newHeight ?? (int)LastRenderInfo.FigureRect.Height;
+
         if (width < 1)
             throw new ArgumentException($"{nameof(width)} must be greater than 0");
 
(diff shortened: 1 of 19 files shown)


src/ScottPlot5/ScottPlot5/Plot.cs
```diff
@@ -376,6 +379,20 @@ public string SaveWebp(string path, int width, int height, int quality = 85)
         return Path.GetFullPath(path);
     }
 
+    public string Save(string path, int width, int height, int quality = 85)
```
swharden, line 382: I actually omitted this function intentionally, but I'm interested in your thoughts. 

My line of thinking is that some file formats accept `int quality` and others may accept `bool transparency`, and I didn't want to force those arguments for file types that don't support them, hence separate SaveBmp, SavePng, SaveJpg, etc.

swharden: ... I also don't love the idea of users evaluating what extensions are supported by trial and error and seeking runtime exceptions 😅 Having dedicated save methods for specific file formats solves that issue.

src/ScottPlot5/ScottPlot5/Plot.cs
```diff
@@ -28,6 +28,8 @@ public class Plot : IDisposable
     public IBenchmark Benchmark { get; set; } = new StandardBenchmark();
     public IZoomRectangle ZoomRectangle { get; set; }
     internal RenderDetails LastRenderInfo { get; set; } = new();
+
+    public PixelRect FigureRect => LastRenderInfo.FigureRect;
```
swharden, line 32: I'd really like to avoid exposing outdated state like this...

Can we change this to something that makes it really intentional what is being returned, maybe like this?

```cs
public RenderDetails GetLastRenderInfo() => LastRenderInfo;
```

Or I guess we could just make `LastRenderInfo` `public` instead of `internal`

In ScottPlot 5 the plot should store no pixel state. The figure size is state that is passed into the render method, and ideally it never leaves that stack.

src/ScottPlot5/ScottPlot5/Plot.cs
bclehmann: I think it's ok to pass in a `quality` that promptly gets ignored, I think that's what `GetImageBytes` and `SKImage.Encode` do anyways.

Perhaps a good middleground is a `public static ImageFormat FromFilePath(string fpath)` which returns an enum with fields like `ImageFormat.Png`, `ImageFormat.Jpeg`, etc etc, just so that we don't have to handle the extension conversions everywhere.

swharden: Good point!

https://github.com/ScottPlot/ScottPlot/blob/282a94a386ed59e252d877f55defa02990548034/src/ScottPlot5/ScottPlot5/ImageFormat.cs#L1-L9

I wish there were a way we could do something like 

```cs
ImageFormat fmt = ImageFormat.FromFilename("test.png");
```

Maybe there should be an `ImageFormats` class that has a `GetImageFormat(string filename)` function? Perhaps the various logic that saves `SKSurface` objects as files could be moved there and out of the `Plot` module

src/ScottPlot5/ScottPlot5/Plot.cs
bclehmann: I do like keeping `LastRenderInfo` internal, most of its members shouldn't be available publicly and it's a struct that we could well want to modify in the future.

That said, I think we need to be able to know how big the image we create is, though that doesn't necessarily need to live in `Plot`. One option is to create a `PlotSnapshot` class, which holds the image dimensions, and something like an `SKImage`. This class can then implement `SavePng`, `SaveJpeg`, etc etc.

This way we're very explicit that we're only sharing the dimensions as they were when the snapshot was created, and we also make it a lot easier for someone to create a class that exports to their own funky formats, especially those that we're unlikely to support (e.g. 4.1 supported TIFF and BMP, which don't currently work in 5.0, plus I could see people wanting PDF and SVG)

src/ScottPlot5/ScottPlot5/Plot.cs
bclehmann: > I wish there were a way we could do something like

Enums support extension methods

src/ScottPlot5/ScottPlot5/Plot.cs
swharden: > we need to be able to know how big the image we create is

`Render()` is `void` but maybe it should return `RenderDetails` 🤔

src/ScottPlot5/ScottPlot5/Plot.cs
swharden: lol, duh, I don't know why I got confused by that 🤪

I was thinking it would have to be a `string` extension method 🤦 

Yes, enum extension method, great option!

bclehmann: I brainfarted, you can't create an enum through its own extension method since extension methods require an instance. I just created an `ImageFormatHelpers` class with static methods `FromFileExtension` and `FromFilePath`

src/ScottPlot5/ScottPlot5/Plot.cs
bclehmann: I think this is resolved (now that we agreed not to return `RenderDetails`)

swharden:
Thanks again for this PR @bclehmann! It's a bit large so it took me quite a while to find a chunk of free time to review it. I'll merge it in today though 🚀 

Note that this throws an exception
```cs
Plot plt = new();
plt.Add.Signal(Generate.Sin(51));
plt.Add.Signal(Generate.Cos(51));
plt.SaveBmp("test_save.bmp");
```

Without supplying width and height it tries to pull it from the last render info, but without a previous render this is not possible. Before merging I'll refactor this toward statelessness, forcing callers to supply a width and height whenever they request a bitmap image to be created.

PR #2363, opened 2023-01-14
Correction for median calculation in population plot

Syntaxrabbit:
This small correction lets the population plot calculate the median in the correct way.
(If the value count is even, the two values in the middle of a sorted array need to be averaged.)


swharden:
Thanks for this fix @Syntaxrabbit! 🚀

PR #2369, opened 2023-01-27
ScottPlot5: DateTime Axes

bclehmann:
**Purpose:**
Adds a DateTime axis

Before this is merged in I'd like to add a fixed interval axis/tick generator. Hopefully with support for gaps, for example to support a plot without weekends or holidays.

```cs
var axis = new ScottPlot.Axis.DateXAxis();
WpfPlot.Plot.XAxes.Clear();
WpfPlot.Plot.XAxes.Add(axis);

DateTime[] dates = Enumerable.Range(1, 31).Select(i => new DateTime(2023, 1, i)).ToArray();
double[] convertedDates = axis.ConvertToCoordinateSpace(dates).ToArray();

double[] sins = Generate.Sin(31);
double[] coses = Generate.Cos(31);

WpfPlot.Plot.Add.Scatter(convertedDates, sins);
WpfPlot.Plot.Add.Scatter(convertedDates, coses);
```
image


swharden:
Hey @bclehmann, I refined some things but left the overall structure the same.

I added a Snap() function to `ITimeUnit` so the first tick can be "snapped" to a whole time unit, then subsequent ticks (spaced by `interval`) are snapped to the same time "grid". This features (and the bug caused by not having it) is best seen when panning plots. 

I added a new cookbook recipe to make investigation easier:

image

## Remaining Tasks

I'm going to set this down for a while so feel free to make more commits if you like, otherwise I'll pick it up again soon. I think the remaining tasks are:

* Use a smarter snap - e.g., the first tick should be preferably placed at March 1 instead of March 2
* Zoom bug - when only one point is on the screen the ticks get kinda weird still

swharden:
Thanks again for this Ben! I got it far enough I'm happy to merge it now and we can refine it with future PRs.

The tick label spacing is a little tight for some of the densities, so maybe some extra logic can be added in the future to make it feel a little more natural. Either way, this landed in a great place!

PR #2374, opened 2023-02-05
WinUI plot control

rafntor:
This PR adds ScottPlot.Uno.WinUI.UnoPlot multi-targeting control for Uno & WinUI(3)

Based on and should be equivalent to the existing ScottPlot.WPF.WpfPlot

Hopefully this does not collide with any other ongoing initiatives ;-) 
Solves https://github.com/ScottPlot/ScottPlot/issues/2039 ?

image

PR #2380, opened 2023-02-09
Sandbox: Interactive ScottPlot in a browser using WebAssembly

rafntor:
Follow-up to https://github.com/ScottPlot/ScottPlot/pull/2374

It is mostly just uno-template-generated stuff, adapted for the sandbox app

You may need to add additional vs-workloads to successfully build and run it

browser


rafntor:
Come to think of it ; If the Cookbook Demo was ported to WinUI you could host the Uno.Wasm variant directly on scottplot.net as a live demoapp - that would be cool ;-)

swharden:
Hey @rafntor, thank you for this PR!

I'm not very experienced with this topic though. I've used Blazor a few times (to build server and client-side web apps), but I'm uncertain about the capabilities of the method demonstrated in this PR. Can it be run entirely on client-side WebAssembly? I'm struggling to get it to build such that I can host it on a static site.

https://scottplot.net/test/wasm/

```
Uncaught (in promise) TypeError: Failed to execute 'compile' on 'WebAssembly': Incorrect response MIME type. Expected 'application/wasm'.
```

Thanks for your input!

swharden:
I think the issue is my host is not serving .wasm files properly, but I think I can correct this with a .htaccess modification

https://scottplot.net/test/wasm/_compressed_br/package_d331d1de8e414e5bc20ec7528b786f19892df023/dotnet.wasm

rafntor:
I'm also not very experienced with webapps (any kind) ; what I have done in this regard is limited to ;
- Launch it directly from vs and run it on localhost with the auto-generated ASP.NET profile in 'launchSettings.json' - this worked out of box for me,  does that work OK on your machine?
- I added [server.py](https://github.com/ScottPlot/ScottPlot/files/10704688/server.py.TXT) to the output-folder ('dist'), uploaded it to a linux-box, then in a shell did "cd /path/to/dist" and  "python -u server.py". Pointing my browser to the ip of that box then serve demo-app fine.

rafntor:
- I found server.py [here](https://github.com/zwcloud/MonoWasm/blob/master/server.py)
- Here are some more [notes](https://github.com/unoplatform/uno/blob/master/doc/articles/how-to-host-a-webassembly-app.md)
Hope this helps ;-)

rafntor:
> Can it be run entirely on client-side WebAssembly?

Yes that is definitely how it works. The webserver only delivers the entire app and after that does nothing more ; the app runs entirely in the browser. On my laptop when I double-click i see ~50 FPS , in Release-build with Firefox and 1920/1080 screen res. Impressive , but it looks like the WinUI app is ~6x faster.

swharden:
> Yes that is definitely how it works

I am really motivated to figure out how to get this working!

* `dotnet publish --configuration Release`
* I copied the output to https://scottplot.net/test/wasm/
* It does not run

image

I added these to a `.htaccess` file in that folder, but it still does not work

```
AddType application/wasm .wasm
AddType application/octet-stream .clr
AddType application/octet-stream .dat
AddType application/octet-stream .pdb
AddType application/font-woff .woff
AddType application/font-woff .woff2
```

swharden:
Actually, I flushed the cache and the error message is different now

* https://scottplot.net/test/wasm/

image

rafntor:
> * I copied `bin/Release/net6.0/publish/` to https://scottplot.net/test/wasm/

I suspect you need to deploy : `bin\Release\net6.0\dist` instead, worth a try

swharden:
I uploaded the `dist` folder and the same error is present

* https://scottplot.net/test/wasm3/

swharden:
Got it! I'll work to refine this then merge it soon 🚀 

* https://scottplot.net/test/wasm4/

swharden:
@rafntor I made some good progress but also got stuck and I want to start working on some other issues now so I'll leave it where it is, but if you can figure out what it takes to build/run this in release mode that would be fantastic. I think it's just the final linking step that's failing. Considering how old some of the dependencies are, maybe starting a new project with all the latest packages would fix the problem?

Thanks again for all your work on this!

rafntor:
- I have regenerated code with newest uno-template-generation and copied & adapted the projects accordingly.
- I have made most nugets floating minor and also stripped out all that are added just in case but are actually unnecessary.
  - Skia for WinUI is held back since the latest 2.88.3 does not build - creates some mismatching version dependency.
  - Notice the comments added to wasm.cspro under release-mode (from template) maybe that can give clues to your release-problems? (I dont see any difference when running debug or release locally)

rafntor:
Now there is a github build-problem, no problem at my machine but I have latest VS with msbuild 17.4.4 while github is 17.4.1.
I get the impression that WinUI & Uno is very fragile when it comes to environment .. (will try some downgrade to try to make it build again)

rafntor:
I tried outside VS like you with  "cd C:\Program Files\IIS Express" and ;
- iisexpress.exe /path:"C:\work\github\ScottPlot\src\ScottPlot5\ScottPlot5 Sandbox\Sandbox.Uno.Wasm\site" 
  - this freezes when I try to open in browser similar to what you have seen 

But , if instead I do cd "C:\work\github\ScottPlot\src\ScottPlot5\ScottPlot5 Sandbox" (sic!)
- then "[dotnet serve](https://github.com/natemcmaster/dotnet-serve)" .. I can successfully navigate to the release dist ;
image
- My takeaway is that the compressed release-output must be OK .. but perhaps not in a way that all web-servers can handle?

swharden:
Thanks for looking into it! I think we took it as far as we can, and I'll make some small changes and merge this in now.

## Working Demo

I modified the csproj so releases don't include debug tools, but they also don't use the linker, and the output seems to run fine on Apache. Here's the URL I'll keep up moving forward:

* https://scottplot.net/test/wasm/
* Loads after 142 requests and 52 MB downloaded

## Slow Build

Just observing that WinUI and Wasm really slow down the solution builds

* rebuild solution with Wasm project loaded: **34 sec**
* rebuild solution with Wasm project unloaded: **12 sec**
* rebuild solution with Wasm and WinUI projects unloaded: **6 sec**

swharden:
Thinking out loud, it would be great if this could be adapted to be a Blazor control! I see that potentially getting a lot of use. It looks like it should be pretty easy since there are already Skia canvas controls for Blazor

* https://nerocui.com/2022/01/08/when-blazor-meets-skiasharp/

PR #2383, opened 2023-02-11
ScatterPlottable which uses raw OpenL for rendering

StendProg:
**Purpose:**
This PR add `ScatterGL` plottable which uses raw OpenGL for rendering.
A section for testing `ScatterGL` has been added to WinformsDeto.

`ScatterGL` is implemented in the `WinForms.Control` project so as not to bring the OpenTK dependency into the main project. And actually intended only for Winforms. WPF has no control with OpenL support.

In fact, this is actually `SkatterConst`. We cannot change the elements of the original array and immediately see the result (as I understand at the moment, ScottPlot 5 still does not allow this even for a regular `Scatter`).

How it works: 
The original dataset is loaded into GPU memory on the first render. When rendering, each time the transformation matrices are dynamically adjusted, which transform our original data array into screen coordinates. Then draw is called, which takes the data already in the GPU memory and applies transformations to them (a typical and very fast operation for the GPU).
The performance of this approach is very high, I was able to get interactive rendering speed for 20 million points on my hardware.

Supposed problems:
1. The GPU can only work with `float`, so visually visible rounding errors are possible, which are significantly higher than for calculations using `double`.
2. There are no checks for the presence of OpenGL. If the control is switched to the software rendering mode, then most likely the application will crash.
3. There are suspicions that this code may not work (work incorrectly) on many hardware configurations. I'm a complete newbie to OpenGL and can't answer any questions that come up.

image

* Consider including a picture if it would help clarify what this pull request does

```cs
            int pointCount = 20_000_000;
            double[] ys = ScottPlot.Generate.NoisySin(new Random(), pointCount);
            double[] xs = ScottPlot.Generate.Consecutive(pointCount);
            var spGL = new ScatterGL(new ScatterSourceXsYs(xs, ys), formsPlot1.GRContext);
            formsPlot1.Plot.Plottables.Add(spGL);
```

### DIFF
--- a/src/ScottPlot5/ScottPlot5 Controls/ScottPlot.WinForms/FormsPlot.cs
+++ b/src/ScottPlot5/ScottPlot5 Controls/ScottPlot.WinForms/FormsPlot.cs
@@ -1,5 +1,6 @@
 ﻿using ScottPlot.Control;
 using ScottPlot.Extensions;
+using SkiaSharp;
 using SkiaSharp.Views.Desktop;
 using System;
 using System.ComponentModel;
@@ -12,6 +13,8 @@ namespace ScottPlot.WinForms;
 public class FormsPlot : UserControl, IPlotControl
 {
     readonly SKGLControl SKElement;
+
+    public GRContext GRContext => SKElement.GRContext;
 
     public Plot Plot { get; private set; }
 
--- a/src/ScottPlot5/ScottPlot5 Controls/ScottPlot.WinForms/GLShader.cs
+++ b/src/ScottPlot5/ScottPlot5 Controls/ScottPlot.WinForms/GLShader.cs
@@ -1,0 +1,107 @@
+﻿using OpenTK.Graphics.OpenGL;
+using System;
+
+namespace ScottPlot.WinForms
+{
+    public class GLShader : IDisposable
+    {
+        private int Handle;
+
+        string _vertexShaderSource =
+        @"# version 330 core
+        layout(location = 0) in vec3 aPosition;
+        uniform mat4 transform;
+
+        void main()
+        {
+            gl_Position = vec4(aPosition, 1.0) * transform;
+        }";
+
+        string _fragmentShaderSource =
+        @"#version 330 core
+        out vec4 FragColor;
+        uniform vec4 pathColor;
+
+        void main()
+        {
+            FragColor = pathColor;
+        }";
+
+        public GLShader()
+        {
+            var VertexShader = GL.CreateShader(ShaderType.VertexShader);
+            GL.ShaderSource(VertexShader, _vertexShaderSource);
+
+            var FragmentShader = GL.CreateShader(ShaderType.FragmentShader);
+            GL.ShaderSource(FragmentShader, _fragmentShaderSource);
+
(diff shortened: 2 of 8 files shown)


swharden:
Thanks for your work on this @StendProg! I'm still learning about SkiaSharp and its capabilities, and I don't have very much experience with GPU programming, but this has been fascinating to learn about.

## GPU Math

> The GPU can only work with float, so visually visible rounding errors are possible, which are significantly higher than for calculations using double

That makes sense! I suspect it is the cause of this effect I see when I zoom at high magnification

signal-zoom

## Unsupported Hardware

> There are suspicions that this code may not work (work incorrectly) on many hardware configurations

I think my system is unsupported 😅  I get very poor performance with 2 million points, and it crashes if I try to go full screen. It would be great if there were some way to determine if a system is unsupported and display an error message.

signal-slow

## This code is useful as documentation

I do not wish to maintain classes which behave so differently on different systems (and which behave poorly on my computer), but this `ScatterGL` class is great to have as a reference. I can link to it on the FAQ page for people interested in making their own GPU-driven plot types. I plan to merge this PR, but I do not see myself making new plot types using this technique commonly in the future.

## Existing plot types are very limited

> We cannot change the elements of the original array ... ScottPlot 5 still does not allow this even for a regular Scatter

I am still mostly focusing on ScottPlot's internal code, so there are very few plot types and they have very few features or restricted features. It is very important to me that high quality plot types are designed well, and I intend to push that effort toward the end of development when all the other important components are in place so I can put a lot of focused effort on their design. That day is getting closer, and I hope to begin making more robust and flexible plot types soon!

---

Thanks again for this PR! It's great to see what GPU code can look like using OpenGL, and I look forward to merging this soon! 🚀

StendProg:
Now the GPU does all calculations on double and there are no more point jumps during interactions.

The minimum supported version of OpenGL is now 4.1 and performance could drop significantly, especially for GPUs with double emulation.

Still acceptable speed at 20 million points on my hardware.
20Mscatter

BambOoxX:
Hey @StendProg, nice work on this one! Did you see any difference in terms of performance between the 20mil example and 20 plots of one mil points? Just out of curiosity.

swharden:
## Performance is fantastic

I was mistaken about my comments earlier regarding poor performance on my machine. I accidentally compared _Signal_ plot vs. ScatterGL and found ScatterGL significantly less performant. Of course it's less performant! I should have compared scatter vs. scatter 🤦 

This works very well! @BambOoxX here are some benchmarks from a 760x412 scatter plot with 1 million poins
* Standard: 170.122 ms
* OpenGL: 0.804 ms

***ScatterGL is over 200 times faster!!!***

## This could be its own package

I modified this PR and added extension methods to this code works:

```cs
formsPlot1.Plot.Add.ScatterGL(xs, ys, formsPlot1.GRContext);
```

This should allow me to create a `ScottPlot.OpenGL` package which, if installed, supplies users with all the extra OpenGL plottables. This way they could be used in any control that supports them.

For now I'm happy to keep these in the WinForms control, but the option exists to move these into their own package in the future if there is a lot of interest in these plot types.

## Thanks!

Thanks @StendProg, this is a fantastic contribution! 🚀

PR #2386, opened 2023-02-13
ScottPlot 5: OHLC

bclehmann:
**Purpose:**
Adds OHLC plots

```cs
int n = 100;
WpfPlot.Plot.Add.OHLC(Generate.Consecutive(n), Generate.RandomWalkOHLC(n).ToArray());
```
image


src/ScottPlot5/ScottPlot5/Generate.cs
```diff
@@ -188,4 +188,27 @@ public static DateTime[] ConsecutiveDateTimes(DateTime start, TimeSpan step, int
...
+
+            double high = gen.RandomNumber(max, max + span * 1.02);
+            double low = gen.RandomNumber(min - span * 0.98, min);
+
+            return new(open, high, low, close);
+        }
+
+        return openingPrices
+            .Take(count)
+            .Select((p, i) => GetOHLC(p, openingPrices[i + 1]));
```
StendProg, line 212: Can be done via `Zip`
```c#
openingPrices
  .Take(count)
  .Zip(openingPrices.Skip(1), (open, close) => GetOHLC(open, close));
```

src/ScottPlot5/ScottPlot5/Plottables/OHLCPlot.cs
```diff
@@ -0,0 +1,64 @@
...
+
+        public AxisLimits GetAxisLimits() => Data.GetLimits();
+
+        public void Render(SKSurface surface)
+        {
+            using SKPaint paint = new();
+            using SKPath path = new();
+
+            foreach ((double x, OHLC y) in Data.GetOHLCPoints())
+            {
```
StendProg, line 39: If we are concerned about performance, then it would be more efficient to have 2 paths: Growing and Falling.
Accumulate them in a loop.
Then, after the loop, call for drawing 1 time each Path.

src/ScottPlot5/ScottPlot5/DataSources/OHLCSourceXsYs.cs
```diff
@@ -0,0 +1,53 @@
...
+        Ys = ys;
+    }
+
+    private Tuple<double, OHLC> GetCoordinatesAt(int index)
+    {
+        return new (Xs[index], Ys[index]);
+    }
+
+    public IReadOnlyList<Tuple<double, OHLC>> GetOHLCPoints()
+    {
```
StendProg, line 26: The use of Zip is more obvious
```c#
  return Xs.Zip(Ys, (x, y) => (x, y)).ToArray();
```

bclehmann: I changed it, but I believe this is copy-pasted from `ScatterSourceXsYs`. I changed it there too but it may have made its way elsewhere.

src/ScottPlot5/ScottPlot5/Plottables/OHLCPlot.cs
bclehmann: I don't think it's necessary (or even noticeable) on performance grounds, but I do think the resulting code is simpler (no pesky `path.Reset()` to remember) so I implemented it anyway. It was something I thought of about 2 seconds after I put this PR down for the day -_-

StendProg (approved):

src/ScottPlot5/ScottPlot5/Plottables/OHLCPlot.cs
swharden: > performance

I agree that performance is probably not an issue for OHLC plots. 

Only fit a few hundred OHLC symbols can fit on a plot before they become too small and overlapping to be interpretable. Maybe in the future some sort of "high density mode" could be devised, but for now I'm satisfied with whatever is simplest and not optimizing financial plots for performance

src/ScottPlot5/ScottPlot5/DataSources/OHLCSourceXsYs.cs
swharden: I always forget about Zip! Thanks for the reminder 😄 

https://learn.microsoft.com/en-us/dotnet/api/system.linq.enumerable.zip

swharden:
@bclehmann this PR is looking really good, thank you so much for implementing this!

Currently OHLC just stores price information ([OHLC.cs](https://github.com/ScottPlot/ScottPlot/blob/3c64944b9d4115708340323c3f2d916df0a7184a/src/ScottPlot5/ScottPlot5/OHLC.cs))

I recognize the `OHLCSourceXsYs` pairs `double[]` times and `OHLC[]` prices together into a single object, but I suspect it would be useful to store position information in the `OHLC` itself. That way it's easier to generate OHLCs and not have to keep track of times and timespans in a separate array.

```cs
// maybe this
public DateTime Start { get; }
public TimeSpan Duration { get; }
public DateTime End => Start + Duration;
```

```cs
// or maybe this
public double StartTime { get; }
public double Duration { get; }
public double EndTime => StartTime + Duration;
```

What do you think? I'm happy to implement this myself before merging if I'm able to get to this before you are.

PS: One could argue that the word "OHLC" only describes price movement so it's fine as it is, but a counter-argument is that OHLCs actually describe price movement _over a specific period of time_ so it should be aware of the time range 🤔

bclehmann:
> a counter-argument is that OHLCs actually describe price movement over a specific period of time so it should be aware of the time range 🤔

In principle I agree, but considering that the user has to convert the datetime to a double through methods which return `double[]`, I think it's a lot more convenient to keep it as a a separate parameter. Otherwise they have to get that `double[]` and do something like `ohlcs.Zip(dates, (ohlc, date) => new OHLC(ohlc.Open, ohlc.High, ohlc.Low, ohlc.Close) { X = date } )` to glue them back together

swharden:
I'm still nooding the idea of not using OADate in this library in the future.

I think there are no references to OADate except in a single extension method.

Maybe we can use DateTime and TimeSpan, then just call that extension method to convert it to coordinate space in the render loop 🤔

bclehmann:
> Maybe we can use DateTime and TimeSpan, then just call that extension method to convert it to coordinate space in the render loop 🤔

I'm ok with this, assuming performance is ok, but I don't want to use the extension method, I think `IDateAxis` has a conversion method? This way people can specify their own mapping, and this is what lets them use a date axis that skips dates without needing to have gaps. Which is especially relevant as we're talking about financial charts right now.

It also might be the case that someone wants a financial chart without a date axis? This was sometimes necessary in SP4 due to limitations in our date axes, but it may be desirable even in SP5, whether it be because some limitations persist or because numeric axes are simply just preferred.

Consider an analyst comparing companies 2 quarters after a major event (IPO, bankruptcy, acquisition, etc). It would be a lot easier to just say "day 0" or "day 7" rather than needing a different x axis for each company so that they can line up the dates how they want to.

swharden:
> We could perhaps add an IOHLCDataSource to do both

I think this would be the best way to go. It makes sense in my head, but I'll try coding it up and seeing if it feels natural to use.

I think this is the trend I'd like to pursue for most plot types in ScottPlot 5. That "data source" approach lets users make their own data sources if they want to do something custom or support some data type that we haven't thought to code into our numerical optimization library or add generic support for

swharden:
This is landing in a great place!

finance

PR #2395, opened 2023-02-17
ScottPlot.WPF based on GL accelerated control

StendProg:
**Purpose:**
SkiaSharp haven't GPU accelerated control for WPF.
But there is PR in SkiaSharp with such control started at Nov 22. https://github.com/mono/SkiaSharp/pull/2317
I tested a bit, and it looks finished.
So i grab the code and place it in ScottPlot.WPF.
If and when the PR is approved we can simply remove the code and change the link.

### DIFF
--- a/src/ScottPlot5/ScottPlot5 Controls/ScottPlot.WPF/SKGLElement.cs
+++ b/src/ScottPlot5/ScottPlot5 Controls/ScottPlot.WPF/SKGLElement.cs
@@ -1,0 +1,274 @@
+﻿using System;
+using System.Collections.Generic;
+using System.ComponentModel;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+using System.Windows;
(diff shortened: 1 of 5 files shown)


swharden:
> I tested a bit, and it looks finished.

This is a great find! Thank you so much for this PR 😄 

> If and when the PR is approved we can simply remove the code and change the link.

I love this plan!

StendProg:
Hi @swharden,

PR with `SKGLElement` has finally been accepted into SkiaSharp. Now we can give up our own copy, but there is a catch.
This PR is in the `Preview` version of `SkiaSharp 3.0` and we will probably have to wait until version 3 goes to release.
I've tested the `SKGLElement` replacement and at a quick glance it works as it should.
I didn't do a PR because you probably don't want to use the `Preview` version of `SkiaSharp` in `ScottPlot.WPF`.

PR #2397, opened 2023-02-17
Scatter gl improvements

StendProg:
**Purpose:**
Some improvements for ScatterGL:
1. Before `Benchmark.Stop()` we call `GL.Finish()`. This block CPU thread until all GPU operations finished. So benchmark now shows correct numbers when using ScatterGL plottables.
2. Optimization for  data to Vertices array copy.
3. `GRContext` => `Func<GRContext>`. This allow us create `ScatterGL` at form constructor. GRContext is null at creation time.
4. `ScatterGL` can be configured to skip render or make `Scatter.Render()` if `GRContext` is null.
5. `Add.ScatterGL` extension method now return added `ScatterGL`.

Extends:
* #2383

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,9 +4,10 @@
 _not yet published on NuGet..._
 * Signal Plot: Support X and Y offset (#2378) _Thanks @minjjKang_
 * WebAssembly: New sandbox demonstrates interactive ScottPlot in a browser (#2380, #2374) _Thanks rafntor_
-* OpenGL: Added experimental support for direct GPU rendering (#2383) _Thanks @StendProg_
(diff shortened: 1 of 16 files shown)


swharden:
Thank you for these improvements @StendProg! I'm going to refine this a little bit and merge it shortly.

The only thing I hesitate about is `Func<GRContext>` because I feel like `func` is a bit awkward, and `GRContext` is not `GRContext?` but sometimes it can be `null`, so I am interested in exploring options for making this part a little simpler.

swharden:
This works now, and I think it's a little simpler for users. The control's `GRContext` is checked for `null` at render time.

```cs
formsPlot1.Plot.Add.ScatterGL(formsPlot1, xs, ys);
```

PR #2398, opened 2023-02-18
ScottPlot5: Fix Plot.Style.Background() bug

Jonathanio123:
**Issue:** If I'm understanding this code correctly, then setting a custom background color currently does nothing. The arguments are simply not used and the colors are hard coded in. I noticed this when following the [Cookbook Style Helper Functions](https://scottplot.net/cookbook/5.0/styling-plots/#style-helper-functions). 

```cs
namespace ScottPlot.Stylers;
public class PlotStyler
{
...
    /// <summary>
    /// Apply background colors to the figure and data areas
    /// </summary>
    public void Background(Color figure, Color data)
    {
        Plot.FigureBackground = Color.FromHex("#07263b");
        Plot.DataBackground = Color.FromHex("#0b3049");
    }
...
```

This pull request simply replaces the hardcoded values with the function arguments.


swharden:
Good catch @Jonathanio123, thanks for identifying this bug and creating this PR to resolve it! 🚀

Issue #2389, opened 2023-02-14
Histogram: class for accumulating binned value counts and outliers


bukkideme:
The recent histogram method requires an already existing double array. 
I started to think about a special method we could call it Rolling Histogram, or like Point by point Histogram as they call such things in LabVIEW (for example [https://www.ni.com/docs/en-US/bundle/labview/page/ptbypt/prob_and_stat_vis_ptbypt.html,](https://www.ni.com/docs/en-US/bundle/labview/page/ptbypt/prob_and_stat_vis_ptbypt.html) ) where we can Enqueue values, and the histogram bin counts are automatically refreshed at each iterations. We do not want to store the data array (would cause a memory leak on long term run), but only the counts for the bins.

What do you think?

bukkideme:
I think now the most simple approach is just to wait for N samples, for example N=100 values, do the binning, then wait again for a new 100 samples, do the binning, etc. After each binning, we just need to add the new "counts" array to the previous one, so we accumulate the values. With a fixed min/max limit, the bin positions will be fixed, so I think this is the most straightforward way to handle "rolling" statistics...?

swharden:
> Rolling Histogram ... where we can Enqueue values ... and the histogram bin counts are automatically updated

> What do you think?

I love this idea! I wouldn't modify the existing classes, but we could certainly create a `LiveHistogram` class that does this using pre-defined bins.

If the bins are predefined and evenly spaced, the bin counts could be a `readonly int[]`, and adding each new datapoint would bump the value in that count array at the correct index by one.

Is this something you'd consider creating and sharing? You could copy/paste the code of this class in this comments area, or submit a pull request. A cookbook example would be great, but I'm happy to create that if it makes your job easier.

As far as usage, I'm open to suggestions, but I'm envisioning something like:

```cs
int firstBin = 0;
binSize = 5;
binCount = 20;
LiveHistogram hist = new(firstBin, binSize, binCount);

for (double y in GetSensorReadings())
{
    hist.Add(y); // automatically updates counts as new items are added
}

for (int i=0; i<hist.BinCount; i++)
{
    double binLow = hist.Bins[i];
    double binHigh = hist.Bins[i] + hist.BinSize;
    int count = hist.Counts[i];
    Console.WriteLine($"There are {count } values between {binLow} and {binHigh}");
}
```

Then it could be added to a plot and would always display live data like this:
```cs
formsPlot1.Plot.AddScatter(hist.Bins, hist.Counts);
[...]

PR #2401, opened 2023-02-18
Statistics.Common: add Histogram functions that report outliers

bukkideme:
Include outliers option and min/max outlers count as return values for the Common.Histogram method.


bukkideme:
Sorry, I totally forgot about the side effects toward DEMO code, etc. 
I have changed the return type from 2 element tuple to 4 element tuple. Maybe we could create 2 method overloads for the two methods only having 2 tuple returns as it was before?

swharden:
Would a solution be to add outliers into the counts for the histogram itself? Outliers could be collected in the edge bins? This way the API doesn't have to change and there's no risk of breaking existing code 🤔 

image

bukkideme:
First I thought, it is better to add the option to the user to decide whether the outliers are included or not. 
Another thing however, we did not think about yet, the probability density option. If we do not include the outlier count in the edge bins, the sum of normalized counts will not add up to 1 (in case of bin size 1). 

I would propose another approach. 

1. we do not use the "includeOutliers" bool parameter, so outliers will be added to the first and last bin
2. However, we keep the "int minOutliers, int maxOutliers" return values optionally, by using method overloads. So we will not break existing code, but if the user wants, then possible to access the outlier counts min/max.

Then it is up to the user, what to do with the outlier count info. These values can be even used to change the first/last bin elements of the returned hist array, or the user only indicates these outlier count values as warning indicators on the UI, etc...

What do you think? Should I add the two extra overload methods? So then we have 2 x 2 Histogram methods, and we do not break the existing code...

bukkideme:
Well, I have realized that method overload only works with differing parameter list, not applicable for different return types :)
I think the method overload could return the two outliers count as out parameters...But I do not really like out parameters. What if we keep the Histogram methods just as before, so they ignore the outliers. No breaking existing code (many people might have use of the existing Histogram and they expect to keep the behavior)! 

I started to feel, we should introduce two new methods, "HistogramWithOutliers" ? I make the required changes into the PR. Please have a look, what do you think?

swharden:
Hi @bukkideme, thanks for starting this PR!

> Please have a look, what do you think?

I'll take a look now, likely make some small changes, and merge this in.

> I would propose another approach.

Maybe these other approaches could be added to the `LiveHistogram` class, so users wanting to get information about outliers can retrieve it from fields of that class. I agree that functions that return a large number of tuples or have `out` parameters are both very confusing, so adding all these features to a new class (instead of sticking it into this method) may be a good long term solution!

> I am quite a beginner in GIT things https://github.com/ScottPlot/ScottPlot/issues/2389#issuecomment-1435944957

I see you're using git inside Visual Studio. A simpler option you may find a lot easier to work with is to use GitHub Desktop to manage code changes. https://desktop.github.com/ It only has a couple buttons, integrates better with GitHub, and in my opinion makes it a lot easier to know how to create new branches and turn them into pull requests. You probably don't need a tutorial to learn how to use it, but even so I bet there are many tutorials and YouTube videos demonstrating how to use it aimed at beginners.

I favor GitHub Desktop over Visual Studio's git tools because I enjoy the simpler interface 😄

swharden:
Following-up, I see there's a `ScottPlot.Statistics.Histogram` class which has been marked obsolete for two years, so I'm super happy repurposing this for something new!

I'll use this PR to clear the content of this class, then you can use it in a future PR to build-out the features of the live histogram. I'll add some fields to the class I think will be useful, but feel free to modify it however you think would be best! We can explore together what the best design will be 🚀

Issue #2299, opened 2022-12-14
Histogram function ignores requested bucket size


Xerxes004:
## Bug Report

**Issue:** Histogram bucket size is ignored when producing bin edges.

`Common.Histogram` has an overload that looks like:

```C#
(double[] hist, double[] binEdges) Histogram(double[] values, double min, double max, double binSize, bool density = false);
```

The `binEdges` returned by this function don't have a spacing of `binSize` as expected.

For example, if I specify binSize of 1.5, and my data goes from 1 to 3, I would expect the edges to be `[1.5, 3.0]`. Instead, they are `[1.0, 2.0]`. These buckets work just fine and the histogram is valid, but the bucket sizes are not the size I requested.

**Reproducing:** Run the following code and print out the results:

```C#
double[] data = new double[] {1.0, 2.0, 3.0};
var (_, leftEdges) = ScottPlot.Statistics.Common.Histogram(data, 1.0, 3.0, binSize: 1.5);
```

### System Details
* **ScottPlot Version:** 4.1.59
* **Operating System:** Windows 10
* **Application Type:** WPF
* **.NET Version:** .NET 7.0

Xerxes004:
I'll code up a fix for this if I find some time soon.

swharden:
Hi @Xerxes004, I'd like to better define the current behavior by adding a test (in #2403) to demonstrate it, but then also add a test to show the new Histogram class performs as expected. For this reason, I'd like to clarify what the expected behavior actually is.

> if I specify binSize of 1.5, and my data goes from 1 to 3, I would expect the edges to be [1.5, 3.0]

If we start at the minimum value (1.0) and new bens start every 1.5 units, wouldn't we expect the bins (lower edges) for this dataset to be 1.0 and 2.5?

swharden:
Following-up, I figured it out! 

I'm happy you brought my attention to this. The same bug that affected the old static method also affected the new Histogram class, so I'm glad we were able to fix it right out of the gate 😄 

I'll share details in #2403

PR #2403, opened 2023-02-19
ScottPlot.Statistics.Histogram

bukkideme:
Hi! This was in my mind as a first implementation. Please have a look.
Thanks!

### DIFF
--- a/src/ScottPlot4/ScottPlot/Statistics/Histogram.cs
+++ b/src/ScottPlot4/ScottPlot/Statistics/Histogram.cs
@@ -64,7 +64,7 @@ public class Histogram
         Max = max;
         AddOutliersToEdgeBins = addOutliersToEdgeBins;
         Counts = new int[binCount];
-        Normalized= new double[binCount];
+        Normalized = new double[binCount];
         BinEdges = new double[binCount];
 
         // create evenly sized bins
@@ -117,7 +117,7 @@ public class Histogram
         {
             double binScale = Counts.Sum() * BinSize;
             for (int i = 0; i < Counts.Length; i++) Normalized[i] = Counts[i] / binScale;
-        }        
+        }
     }
 
     /// <summary>
--- a/src/ScottPlot4/ScottPlot/Statistics/Histogram.cs
+++ b/src/ScottPlot4/ScottPlot/Statistics/Histogram.cs
@@ -43,12 +43,12 @@ public class Histogram
     public bool AddOutliersToEdgeBins { get; }
 
     /// <summary>
-    /// Lower edge of the first bin (inclusive)
+    /// Lower edge of the first bin
     /// </summary>
     public double Min { get; }
 
     /// <summary>
-    /// Upper edge of the last bin (exclusive)
+    /// Upper edge of the last bin
     /// </summary>
     public double Max { get; }
 
--- a/src/ScottPlot4/ScottPlot/Statistics/Histogram.cs
+++ b/src/ScottPlot4/ScottPlot/Statistics/Histogram.cs
@@ -86,7 +86,8 @@ public class Histogram
 
     public static Histogram WithFixedSizeBins(double min, double max, double binSize, bool addOutliersToEdgeBins = false)
     {
-        int binCount = (int)((max - min) / binSize);
+        int binCount = (int)Math.Ceiling((max - min) / binSize);
+        max = binCount * binSize + min;
         return new Histogram(min, max, binCount, addOutliersToEdgeBins);
     }
 
--- a/src/ScottPlot4/ScottPlot/Statistics/Histogram.cs
+++ b/src/ScottPlot4/ScottPlot/Statistics/Histogram.cs
@@ -4,9 +4,6 @@ using System.Linq;
 
 namespace ScottPlot.Statistics;
 
-#pragma warning disable RS0016 // Ignore warnings that this class is not yet documented
-
-[Obsolete("This class is a work in progress. Its API may change as it continues to evolve.")]
 public class Histogram
 {
     /// <summary>
@@ -33,12 +30,12 @@ public class Histogram
     /// <summary>
     /// Lower edge of the first bin
     /// </summary>
-    private readonly double Min;
+    public readonly double Min;
 
     /// <summary>
     /// Upper edge of the last bin
     /// </summary>
-    private readonly double Max;
+    public readonly double Max;
 
     /// <summary>
     /// The calculated bin size.
--- a/src/ScottPlot4/ScottPlot/Statistics/Histogram.cs
+++ b/src/ScottPlot4/ScottPlot/Statistics/Histogram.cs
@@ -77,6 +77,12 @@ public class Histogram
     /// <param name="addFinalBin">if true, one more bin will be added so values equal to <paramref name="max"/> can be counted too</param>
     public Histogram(double min, double max, int binCount, bool addOutliersToEdgeBins = false, bool addFinalBin = true)
     {
+        if (min >= max)
+            throw new ArgumentException($"{nameof(max)} must be greater than {nameof(min)}");
+
+        if (binCount < 1)
+            throw new ArgumentException($"must have at least 1 bin");
+
         BinSize = (max - min) / binCount;
         AddOutliersToEdgeBins = addOutliersToEdgeBins;
 
--- a/src/ScottPlot4/ScottPlot/Statistics/Histogram.cs
+++ b/src/ScottPlot4/ScottPlot/Statistics/Histogram.cs
@@ -12,11 +12,6 @@ public class Histogram
     public readonly int[] Counts;
 
     /// <summary>
-    /// Normalized values calculated for each bin.
-    /// </summary>
-    public readonly double[] Normalized;
-
-    /// <summary>
     /// Lower edge for each bin.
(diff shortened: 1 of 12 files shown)


swharden:
This new histogram class is so much simpler to work with! I'm going to go all the way and update all Cookbook recipes to use this new class. The old code will still work, but I'd like to highlight this newer and simpler way to create and maintain histograms.

Soon I will put this PR down for tonight, but I'll keep working on it tomorrow and hopefully merge it in when all the Cookbook recipes and demo apps are using it 🚀

bukkideme:
Hi! Good to see the progress! :) I think I will look into my old PhD thesis, and check some logarithmic binning in my old C codes. Logarithmic binning is very useful when for example you need to deal with incoming data following power law distribution (shows as a line in a log-log graph). One of the many examples is earthquake magnitude distribution: lots of small earthquakes, but very few large ones. Simple linear binning is not useful in such cases (the tail will look terrible), so in scientific papers it is common to use logarithmically growing X axis bins. I hope I can put together some example.

swharden:
> I hope I can put together some example.

Examples would be great! I have to be careful not to put _too_ much stuff in this library though because it's a lot to maintain, and I want to keep ScottPlot focused as much as possible on plotting (there are already good statistics libraries out there). Perhaps copy/pasting useful code in an Issue or Discussion would be a good place to put it so it's searchable and can be found by those with similar needs, but doesn't have to be deployed and maintained as part of this core library.

> Good to see the progress! :)

I'm very happy with this new histogram class because
* It's way simpler to use, and the cookbook examples are now much better
* It supports live data (I still need to create a demo)

bukkideme:
I totally understand your point. Log binning is not really a usual stuff, so i agree not to put into the main code. But we could show a use case maybe. Anyway, ScottPlot does not really support log scales yet, which is another point why not include this log binning feature :)

Xerxes004:
Is it possible to address #2299 as part of this PR?

bukkideme:
> Is it possible to address #2299 as part of this PR?

That is not an issue anymore. The new Histogram method takes inputs min, max and binCount. Binsize is calculated, and not an input parameter.

src/ScottPlot4/ScottPlot/Statistics/Histogram.cs
```diff
@@ -1,59 +1,197 @@
...
+        BinSize = (Max - Min) / binCount;
+
         for (int i = 0; i < binCount; i++)
         {
-            BinEdges[i] = firstBin + binSize * i;
+            BinEdges[i] = min + BinSize * i;
+        }
+    }
+
+    public static Histogram WithFixedSizeBins(double min, double max, double binSize, bool addOutliersToEdgeBins = false)
```
Xerxes004, line 87: @bukkideme to clear up any confusion, this is why I mentioned #2299

swharden: Hi @Xerxes004, thanks for reminding us about #2299! It would be great to solve that issue with this PR. I'll follow-up with a comment on this PR in a few minutes when I learn more about that issue...

swharden: Fixed by https://github.com/ScottPlot/ScottPlot/pull/2403/commits/43d0664729ba20a3658af97e739c43a20579f74a 🚀

swharden:
I may have been too quick to say that https://github.com/ScottPlot/ScottPlot/pull/2403/commits/43d0664729ba20a3658af97e739c43a20579f74a fully resolved issue #2299

It's pretty close! The `min` is inclusive but the `max` is exclusive (values equal to it roll over into the next bin) so I think the returned data should have one extra bin that currently is not being accounted for... will fix in a follow-up commit

swharden:
Mimicking numpy, I'll make all these methods return one more bin than asked for. This is a weird result of what happens when you need a bin to contain the `max` value, but bins are `min` inclusive but `max` exclusive 🙄 

https://numpy.org/doc/stable/reference/generated/numpy.histogram.html

image

swharden:
I'm now wondering if the default behavior should be changed. Should the bin size be automatically increased by 1 bin so that the `max` can get counted? This is the current behavior... which may be best... but it's not clear to me if this is the behavior most people would expect 🤔 

```cs
var hist1 = ScottPlot.Statistics.Histogram.WithFixedSizeBins(min: 0, max: 10, binSize: 1);

hist1.BinEdges.Should().BeEquivalentTo(new double[] { 0, 1, 2, 3, 4, 5, 6, 7, 8, 9 });
hist1.Counts.Should().BeEquivalentTo(new double[] { 0, 0, 0, 0, 0, 0, 0, 0, 0, 0 });

hist1.Add(10); // since bins are max-exclusive, this counts as an outlier
hist1.Counts.Should().BeEquivalentTo(new double[] { 0, 0, 0, 0, 0, 0, 0, 0, 0, 0 });
```

bukkideme:
I think either options are ok, if the behaviour is clearly explained to the user.
To my personal taste i would leave the thing max exclusive. I think asking for N bins, and getting N+1 is a bit strange, no? :)

swharden:
> I think either options are ok, if the behaviour is clearly explained to the user.

I agree!

> I think asking for N bins, and getting N+1 is a bit strange, no? :)

True, but that's been the existing behavior in this library for a long time and changing it now could break others' code running in production 😅 

I think a good compromise is to not change the existing static method behavior, but to provide more logical (and better documented) behavior for the new Histogram class.

I'm putting together a live demo now and it's feeling very natural to work with! I think this is landing in a great spot 🚀

swharden:
I think this PR may be complete! I'll leave this PR open for another day to give @bukkideme or @Xerxes004 time to look it over in case they have any suggestions. Otherwise I'll merge it tomorrow evening.

Thanks again everyone for your hard work and great input on this!

hist-live

bukkideme:
There is this:
/// <summary>
    /// Upper edge of the last bin (exclusive)
    /// </summary>
    public readonly double Max;

Maybe the help could be changed to:
"Upper edge of the last bin (exclusive if addFinalBin true, otherwise inclusive)" ?

Another thing, why the Add and AddRange methods are same? As i see, the Add method accepts an IEnumerable, and it's summary is the same as of the AddRange methods'.

Xerxes004:
> True, but that's been the existing behavior in this library for a long time and changing it now could break others' code running in production 😅

Ironically, the current static function caused a bug in our production code. I was requesting buckets of size `n`, which I assumed (without checking 🙃) would make the edges exactly `n` apart from each other. We caught the bug when we noticed that the bucket size I displayed on screen was different than our actual exported data's bucket sizes. We were able to rejigger the math to make it work, but it just felt weird that the bucket size I requested was being ignored. We are doing some math so that there are `x` buckets per standard deviation, which gives us very nice-looking histograms regardless of the input data.

For the record, the way that feels the most correct for bucket size is to completely honor the user's `bucket_size` as the stride at which data is collected. This means that the number of buckets would be: 

```C#
int NumBuckets(double min, double max, double bucket_size)
{
    return (int) Math.Ceiling((max - min) / bucket_size);
}
```

Xerxes004 (approved):

swharden:
> Ironically, the current static function caused a bug in our production code

Thanks for sharing your story @Xerxes004! I'm in a bind because on one hand the "surprising" behavior could cause production bugs, but changing behavior of the existing methods to produce the expected output will change behavior of code which may already be running in production, so the fix risks introducing new bugs 😝 

Connecting some dots, this PR updates those static methods to use the ceiling strategy your comment demonstrates (resolving #2299)

I think I'll update this PR to mark those methods obsolete and direct people to the new Histogram class.

https://github.com/ScottPlot/ScottPlot/blob/a16fe2ce2457c510faafdad7fd5a47d5327ad393/src/ScottPlot4/ScottPlot/Statistics/Common.cs#L327-L332

https://github.com/ScottPlot/ScottPlot/blob/a16fe2ce2457c510faafdad7fd5a47d5327ad393/src/ScottPlot4/ScottPlot/Statistics/Common.cs#L342-L347

> Note that my function above doesn't check the sanity of the input params. The real function should.

Good point! I'll add that in before merging.

---

> Add() vs. AddRange()

@bukkideme I'm still deciding which API I prefer... an overload seems simplest, but I recognize `IList` uses `AddRange()` so I'll probably put it back to that

> Maybe the help could be changed to: "Upper edge of the last bin (exclusive if addFinalBin false, otherwise inclusive)" ?

https://github.com/ScottPlot/ScottPlot/blob/a16fe2ce2457c510faafdad7fd5a47d5327ad393/src/ScottPlot4/ScottPlot/Statistics/Histogram.cs#L50-L53

I think this is confusing too, and for simplicity I think I can improve it by removing thee word "exclusive" from the existing XML comment

> maybe the Histogram Demo could have a checkbox "Show normalized" too, it would look even better to have option to see normalized values on Y axis?

I thought about it, but then we might as well add CPH options and outlier options ... then in the effort to demonstrate all the cool things it can do we accidentally make it difficult to learn from 😅 

I think I'll favor ultimate simplicity for these demo apps, and users can dig deeper by poking around the classes they instantiate (and also advanced features demonstrated in the cookbook)

> I think the merge is good to go for me :)

woohoo! I'll merge it and make a new release very soon. Thanks again for getting this PR started! 🚀

Issue #2396, opened 2023-02-17
Make BmpHeatmap field public for Heatmap() class?


bukkideme:
I think making the BmpHeatmap field public (readonly) could be useful in certain cases.
One example: after creating a heatmap, I want to save it as a png. Recently this is not possible, only with the other Plot elements...

```
/// <summary>
/// Pre-rendered heatmap image
/// </summary>
protected Bitmap BmpHeatmap
```

Then we could use it like this:

```
var hm = formsPlot1.Plot.AddHeatmap(Intensities, colormap: color, lockScales: true);            
hm.Update(Intensities, min: ColorScaleMin, max: ColorScaleMax);
var bitMap = hm.BmpHeatmap;
bitMap.Save(saveFileDialog.FileName, ImageFormat.Png);
```

bukkideme:
What about if we create this in Heatmap.cs:
```
public Bitmap GetBitmapDeepCopy()
{
       return BmpHeatmap.Clone(new Rectangle(0, 0, BmpHeatmap.Width, BmpHeatmap.Height), BmpHeatmap.PixelFormat);
}
```

Then we can get only the Heatmap without the other things:

```
var plt = new ScottPlot.Plot(600, 400);
double[,] imageData = DataGen.SampleImageData();
var hm = plt.AddHeatmap(imageData);            
Bitmap bitMap = hm.GetBitmapDeepCopy();
Bitmap resized = new Bitmap(bitMap, new Size(bitMap.Width * 4, bitMap.Height * 4));
resized.Save("testfile.png", ImageFormat.Png);
```

PR #2410, opened 2023-02-22
Heatmap.GetBitmap()

bukkideme:
Following the question in #2396 .
What do you think about this change? It would be really helpful, when I want to save only the heatmap picture.
But then I wonder, if this should be introduced in a more generalized way for other elements also?


swharden:
I like it! My recommendation is to name it `GetBitmap()` to be consistent with some other classes which do similar things. The XML docs can clarify that the returned bitmap is a copy.

Let me know when you're done with your changes and I'll merge it in! 🚀

bukkideme:
Done! If all is ok, please merge it, thanks! :)

Issue #2399, opened 2023-02-18
ScottPlot.OpenGL


swharden:
Create a NuGet package to contain OpenGL plottables and helper methods

* #2397
* #2383

StendProg:
Hi @swharden,

I already thought about it, but there is a problem.
We actually need 2 versions of this package. One version is built on a dependency on OpenTK 2.0 and another on OpenTK 3.0. Versions of the OpenTK API have differences, but at first glance they are not critical. So I think we can get by with one code base and conditional operators in some places. But how to organize it and how it will look, here I am lost, there is little experience with Nuget.

Why is this needed?
1. Winforms control is built on a dependency on OpenTK 2.0.
2. WPF NET462 is built on dependence on OpenTK 2.0.
3 WPF NETCORE is built on a dependency on OpenTK 3.0.

If we try to add package built on OpenTK 2.0 to the WPF NETCORE control, then we get the `cousin dependencies` problem, which, as I understand it, has no solution using Nuget.

swharden:
Hi @StendProg, I think this is easier now that the Windows Forms control uses OpenTK 3

https://www.nuget.org/packages/SkiaSharp.Views.WindowsForms#dependencies-body-tab

![image](https://user-images.githubusercontent.com/4165489/226129466-ac6bd91f-12d3-4b3f-a3fb-7aa516387ea5.png)

I'm moving all the OpenGL code out of the WinForms control and into its own project (#2496). I'll publish it as a separate NuGet package when I make the next ScottPlot 5 release. I think this solves our issues, but if there's a case I did not think of let me know and I can work to improve it or roll the changes back.

StendProg:
Hi @swharden,

My bad. When I wrote 2 I meant 3, when I wrote 3 I meant 4 (the SciaSharp version must have confused me).

Simply put, the problem is in this code:
https://github.com/ScottPlot/ScottPlot/blob/33fca4cbcb884161d61e05615604b008a0e7aac1/src/ScottPlot5/ScottPlot5%20Controls/ScottPlot.WPF/ScottPlot.WPF.csproj#L39-L46

SKGLElement uses OpenTK4 for modern .NET.
All other controls use OpenTK3.
We can of course try to downgrade, but only as long as we have control over this component. When it will be included in SkiaSharp, we won't have this option

swharden:
> SKGLElement uses OpenTK4 for modern .NET.
> All other controls use OpenTK3.

[...]

PR #2460, opened 2023-03-07
ScottPlot5: data source for caching scatter data limits

StendProg:
**Purpose:**

ScatterGL recalculates limits each time AutoAxes called.
This calculation starts to take a noticeable time at millions of points. Since `ScatterGL` uses a constant data set, the calculation of the limits can be cached on the first call.

I left the CacheScatterLimitsDecorator in the main namespace because it might come in handy for other purposes. In fact, all `ScatterDataSources` are constant at the moment.
If its location in the main project is invalid, it can always be moved to the OpenGL folder.


swharden:
Hi @StendProg, this is a very good idea! Thanks for this PR

The ability to cache axis limits (and turn that feature on and off) seems very useful. I may add this functionality into the `IScatterSource` interface.

> If its location in the main project is invalid, it can always be moved to the OpenGL folder.

I'll move this into the new ScottPlot.OpenGL project (#2399, #2496)

Issue #2463, opened 2023-03-07
Histogram function can't handle all-same data


Xerxes004:
## Bug Report

**Issue:** Given an array of data where all the values are the same, the `ScottPlot.Statistics.Common.Histogram` function throws a `System.IndexOutOfRangeException`. This happens because the min and max of this data are the same, which causes the bin count to be 0.

https://github.com/ScottPlot/ScottPlot/blob/3b67060daadd30bb813704ac5bfeb508131b4f04/src/ScottPlot4/ScottPlot/Statistics/Common.cs#L327-L333

Similarly, the new Histogram class has an incorrect assumption about min and max:

https://github.com/ScottPlot/ScottPlot/blob/3b67060daadd30bb813704ac5bfeb508131b4f04/src/Shared/ScottPlot/Statistics/Histogram.cs#L78-L82

Histograms of all-the-same data may be useless, but are still valid. The code should be updated to handle min and max being the same value.

**Reproducing:** 

1. Create an array of all zeros called `data`
1. Attempt to create a Histogram like: `ScottPlot.Statistics.Common.Histogram(data, min: 0, max: 0, binSize: 1)`
1. Crash to desktop

### System Details
* **ScottPlot Version:** 4.1.61
* **Operating System:** Windows 10
* **Application Type:** WPF
* **.NET Version:** .NET 7

PR #2464, opened 2023-03-07
Fix histogram exception

Xerxes004:
**Purpose:**
* Fix the bug mentioned in #2463 
* Add two tests to catch regression
* Add `autoformat.sh` for Shared project


swharden:
Thanks @Xerxes004, this looks great! 🚀

PR #2479, opened 2023-03-10
Cookbook: fix datetime recipe

chrisxfire:
Add newline before `class TicksDateTimeSignal` so that .cs file parses correctly for the website.

Related:
* https://github.com/ScottPlot/ScottPlot.NET/pull/25


swharden:
Hi @chrisxfire, this PR may set the record for the smallest number of characters required to fix a bug 🤣 

The CI pipeline won't let me merge this yet because it wants that file you edited to be autoformatted.

In visual studio you can do that with Edit > Advanced > Format Document, or you could run this console command in the ScottPlot4 folder:

```
dotnet format
```

swharden:
... actually I think you may have used the web editor to make that change, so rather than have you fire up an IDE to fix whitespace I'll just do it on my end right now and merge this. Thanks again for the fix! 🚀

chrisxfire:
Copy!  I’d do it from VS but I’m away from my machine 😊

Chris

Issue #2461, opened 2023-03-07
single double opacity value optionally in Heatmap.Update() method?


bukkideme:
Hi @swharden !

What do you think, would it make sense to make an option to define opacity globally for a whole heatmap? Recently we need to supply this: double?[,] opacity = null .
However it happens often, that we need a global opacity for the whole heatmap, so maybe we could have a method overload to accept single double opacity value?
If you agree, I will make a PR.
Thanks!

PR #2484, opened 2023-03-12
ScottPlot4 Heatmap: Add Opacity property

bukkideme:
Following #2461. Please have a look, what do you think? 
Thanks!


swharden:
Hi @bukkideme, thanks for opening this PR!

I spent way too much time looking into this topic 🤪 but I learned something System.Drawing can do that I didn't previously appreciate

https://learn.microsoft.com/en-us/dotnet/desktop/winforms/advanced/how-to-use-a-color-matrix-to-set-alpha-values-in-images?view=netframeworkdesktop-4.8

```cs
ColorMatrix cm = new() { Matrix33 = (float)Opacity }; // define opacity using a matrix
ImageAttributes att = new();
attr.SetColorMatrix(cm, ColorMatrixFlag.Default, ColorAdjustType.Bitmap);

gfx.DrawImage( // draw the heatmap using the custom matrix
        image: BmpHeatmap,
        destRect: destRect,
        srcX: 0,
        srcY: 0,
        BmpHeatmap.Width,
        BmpHeatmap.Height,
        GraphicsUnit.Pixel,
        attr);
```

This is a shortcut that lets us control global opacity without actually calculating any `Color` values. By adding an `Opacity` property (not passing it into functions) we can keep the original API the same and also not add any logic to check whether multiple types of opacity settings are being used. I'll update this PR to use this simpler method and merge it in shortly 🚀

PR #2502, opened 2023-03-22
Add a public ErrorBarAlignment property in PopulationPlot.cs for boxplot.

benton-anderson:
Box plot in population plots does not have public option for setting error bar alignment. Current setting is align to the right. Aligning to the center is by far the most common error bar alignment in box plots.  

Added public property `ErrorBarAlignment` that sets the `HorizontalAlignment` enum. Default value is now `Center`. 

Also added `errorAlignment: ErrorBarAlignment` parameter to two calls to `Box()` method. 

This is my first ever pull request, so I apologize if I'm not doing it right.


swharden:
Hi @benton-anderson, thank you for opening this issue!

Allowing users to customize this alignment is a big win - thanks for this PR!

This type of plot is what the default behavior is designed for:

https://scottplot.net/cookbook/4.1/category/plottable-population/#multiple-populations

image

While I agree that center is more common, I don't want to change the default behavior as it may break existing apps. I'll certainly make the center alignment the default in the next major version of ScottPlot (version 5) which permits breaking changes like this 👍

PR #2506, opened 2023-03-24
SP4 Cookbook: Customize Bar Label Font

TrapperHell:
**Purpose:**
* Currently there doesn't seem to be a way to customize the font colour used to display the values above the bar graphs (assuming that `.ShowValuesAboveBars` property is enabled).
* By having the `BarPlotBase.cs` accept a Font parameter we can customize the font colour used for values (while also keeping it `readonly`

```cs
double[] values = { 10, 15, 12, 6, 8, 4, 12 };

// font optional-parameter has been added
var bar = plt.AddBar(values, font: new ScottPlot.Drawing.Font { Color = System.Drawing.Color.Red });
bar.ShowValuesAboveBars = true;
```
ScottPlotTests PlotTypes Bar Test_Bar_Series2

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,8 @@
 * Heatmap: Added `FlipVertically` to invert vertical axis of heatmap data (#2444, #2450) _Thanks @Neopentane1_
 * Histogram: Improved support for datasets with low variance (#2464, #2463) _Thanks @Xerxes004_
 * Heatmap: Added `Opacity` property (#2461, #2484) _Thanks @bukkideme_
+* DataOperations: New static class with helper methods for working with 1D and 2D data (#2497) _Thanks @bukkideme and @StendProg_
+* Population: Added option for customizing horizontal errorbar alignment (#2502) _Thanks @benton-anderson_
 
 # Releases
 
--- a/src/ScottPlot4/ScottPlot.Cookbook/Recipes/Plottable/Bar.cs
+++ b/src/ScottPlot4/ScottPlot.Cookbook/Recipes/Plottable/Bar.cs
@@ -146,6 +146,27 @@ namespace ScottPlot.Cookbook.Recipes.Plottable
         }
     }
 
+    public class BarShowValueCustomFont : IRecipe
+    {
+        public ICategory Category => new Categories.PlotTypes.Bar();
+        public string ID => "bar_values_custom_font";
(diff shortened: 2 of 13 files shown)


swharden:
> Currently there doesn't seem to be a way to customize the font colour used to display the values above the bar graphs (assuming that .ShowValuesAboveBars property is enabled).

Hi @TrapperHell, this is possible but poorly documented.

I'll modify this PR to add a cookbook recipe demonstrating this 👍 

```cs

double[] values = { 27.3, 23.1, 21.2, 16.1, 6.4, 19.2, 18.7, 17.3, 20.3, 13.1 };

var bar = plt.AddBar(values);
bar.ShowValuesAboveBars = true;
bar.Font.Color = Color.Green;
bar.Font.Size = 18;
bar.Font.Bold = true;

plt.SetAxisLimits(yMin: 0);

plt.SaveFig("bar_values_custom_font.png");
```

image

TrapperHell:
Hi @swharden,

Apologies for the silence from my end, but I was hoping to take a closer look at the PR during the weekend.

I should thank you though for taking it upon yourself to fix all this and even documenting it in the cookbook.

While checking the code, I was mistakenly operating under the assumption that the `Font` property is of type `System.Drawing.Font` whose properties are get-only. But as you showed in your example, your `Font`implementation is not get-only and could thus be assigned after declaration.

(Also, apologies for having the `System` directives sorted first in the usings. I have that enabled by default and I forgot to check how you're handling it in your project. Perhaps, setting `dotnet_sort_system_directives_first` in the .editorconfig would streamline this in the future.)

Thanks again for your time and keep up the good work!

swharden:
> Apologies for the silence from my end, but I was hoping to take a closer look at the PR during the weekend.

No worries! I'm glad I was able to make some improvement and merge it in, but feel free to open new PRs if you have ideas for further improvements

> mistakenly operating under the assumption that the Font property is of type System.Drawing.Font whose properties are get-only

This is an easy mistake to make! The API isn't super clear about how that works 😅 I'm hoping the new cookbook example will make it a little easier to figure out for new users

> Also, apologies for having the System directives sorted first in the usings

Again, no worries! I only reverted those changes to make it easier to identify what files this PR was meaningfully changing so I could get a better understanding of the PR.

This issue comes up a lot, so to prevent it I should enforce some style rules like you suggest. I created an issue to track progress on this front:

* #2511

PR #2547, opened 2023-04-02
SP5: Throw in designer mode of .NET framework projects

swharden:
This error message is more useful than the one thrown when the SkiaSharp dll fails to load


swharden:
Unfortunately this doesn't solve the issue... the designer still crashes and I can't seem to customize the exception message that is thrown 😩

PR #2565, opened 2023-04-05
ScottPlot5: WpfPlot as CustomControl

KroMignon:
**Purpose:**

Use Control as base class for WpfPlot instead of UserControl to enable customization. (cf https://github.com/ScottPlot/ScottPlot/issues/2509).

**Changes:**
- update `ScottPlot5-wpf` Solution Filter to add missed `ScottPlot.OpenGL` project
- `WpfPlot` use now a template and is based on `System.Windows.Controls.Control`, very close to ScottPlot4 version (#2526).

I am not totally happy with this control. The method `WpfPlot::Replace()` has a _bad taste_ to me. It looks weird as the control hold by the new `Interaction` instance my be another one as the control to which it will be set. 

This looks very error prone to me 💣


swharden:
@KroMignon this is fantastic, thanks so much! I'll make a few small changes to improve mirroring between the WPF and WinForms controls and merge this into main in a few minutes 🚀

swharden:
> The method `WpfPlot.Replace()` has a bad taste to me

I agree! This strategy was originally created so users could customize the interactivity (swapping what mouse buttons do, changing zoom and pan behavior, etc) but I think the API can benefit from improvement in the future

Issue #1301, opened 2021-09-30
Violin Plot (suggestion for new plot type)


bclehmann:
## Features Suggestion

**Feature description:** Add a new Violin plottable.
![image](https://user-images.githubusercontent.com/8635304/135386339-ea11a669-e03d-4815-9979-dd4c7447cb20.png)

A violin plottable is like a box plot surrounded by a [kernel density estimation](https://en.wikipedia.org/wiki/Kernel_density_estimation) (KDE), which is analogous to a histogram that has been smoothed. There are variants that stack multiple KDEs to represent data in different confidence intervals.

There are also variants that are asymmetrical, showing KDEs for two different categories. These can include the box plot in the middle if the box plot is for the whole set, but the box is often not shown.
![image](https://user-images.githubusercontent.com/8635304/135387006-febda299-91aa-436c-91d9-a046000b4b8d.png)

I think this feature makes the most sense as an addition to population plots.

**Code example:** If applicable, use this space to demonstrate how you wish the new feature could be used in code.

```cs
var popPlot = plt.AddPopulations(multiSeries);
popPlot.Violin = true;
```

swharden:
I'll close this for now since it's not being actively worked on, but it's added to the triage list #1028 👍

bclehmann:
Yeah I avoided starting this because the math looks non-trivial, and then I forgot about it.

PR #2589, opened 2023-04-14
ScottPlot 5: Box Plots

bclehmann:
**Purpose:**
Adds box plots to ScottPlot 5

```cs
Random rand = new(0);

ScottPlot.Plottables.Box CreateBox()
{
    int N = 50;
    double mean = rand.NextDouble() * 3;
    double stdDev = rand.NextDouble() * 3;

    double[] values = Generate.RandomNormal(N, mean, stdDev);
    Array.Sort(values);
    double min = values[0];
    double q1 = values[N / 4];
    double median = values[N / 2];
    double q3 = values[3 * N / 4];
    double max = values[N - 1];

    return new ScottPlot.Plottables.Box
    {
        WhiskerMin = min,
        BoxMin = q1,
        BoxMiddle = median,
        BoxMax = q3,
        WhiskerMax = max,
    };
}

int numBoxesPerSeries = 3;
int numSeries = 2;
ScottPlot.Plottables.BoxSeries[] series = new ScottPlot.Plottables.BoxSeries[numSeries];
var colorPalette = new Category10();
for (int i = 0; i < series.Length; i++)
{
    series[i] = new ScottPlot.Plottables.BoxSeries
    {
        Fill = new FillStyle { Color = colorPalette.GetColor(i) },
        Boxes = new ScottPlot.Plottables.Box[numBoxesPerSeries],
    };

    for (int j = 0; j < series[i].Boxes.Count; j++)
    {
        var box = CreateBox();
        box.Position = j;
        series[i].Boxes[j] = box;
    }
}

var boxPlot = myPlot.Add.Box(series);
```
image

I hope you appreciate the restraint it took for me to not reduce that code snippet to one big nested `Enumerable.Range(...).Select()` statement


bclehmann:
I elected to make box plots their own plottable rather than the population plot from ScottPlot 4.1. I think the previous approach was quite confusing and asked users to accept a lot of extra stuff just to put a box on the screen.

It is worth noting that the whisker style differs, in 4.1 it's to the right of the box and bends inwards. My high school teacher called these "flamingo box and whisker plots": 
image

I went with a more traditional whisker, though I can change them or implement both if you think it's necessary:
image

swharden:
Awesome Ben, this is looking great!

> I think the previous approach was quite confusing and asked users to accept a lot of extra stuff just to put a box on the screen

Well stated!

> flamingo-style

I agree the center-aligned default style is indeed the way to go!

> I can change them or implement both if you think it's necessary

Not necessary, but I'll create (and close) an issue tagged "good first issue" in case someone wants to add this in the future

swharden:
I spent a little more time working with this today and there are some decisions which are worth considering a little deeper so I'll revisit it tomorrow. The big questions I'm stuck with presently are:

* The `Box` class has things like `WhiskerMin` and `BoxMiddle`, but I wonder if boxes should be able to individually hold their styling options (stroke thicknesses, colors, etc.)

* It may be easier to maintain this plot type if it only has relatively dumb `List<Box>` (with a smart helper function to generate the styled collection of boxes) rather than put high level multi-series logic in the plot type itself.

* While I'm in here, I'll probably use this PR to refine the old bar plot API. It's pretty rough 😅

bclehmann:
I don't know if you're working on it or not so I won't push anything, but I just remembered that I neglected to add any logic for handling horizontally oriented box plots after I refactored them, so it would surprise me greatly if they weren't broken.

I can work on that if you'd like, though it'll probably have to wait until morning.

swharden:
I don't want to add too much complexity to this PR, but I will point out that there is some extra functionality that may be expected of box plots in the future (e.g., min/max markers, and a mean marker in addition to a median line)

https://www.originlab.com/doc/origin-help/create-box-chart

This type of functionality may be better implemented in another plot type, and there is some overlap with violin plots (#1301)

swharden:
I left things mostly how they were, but I'd love to work toward having a `DataSource` for bar plots, box plots, and grouped versions of both.

I'll merge this now and explore these higher level organization options in #2596

Thanks for your work on this @bclehmann! 🚀

PR #2597, opened 2023-04-17
WPF: fix XAML component naming

KroMignon:
Always use `x:Name` for naming components in XAML code.

This is to most "common" way to name components in XAML/WPF as the XAML compiler will ensure this name will be unique and it cannot be changed later in code behind via `SetValue()`. 

This way, any naming conflicts would be avoid.


swharden:
Thanks @KroMignon! I'll update the WPF Quickstart page too https://scottplot.net/quickstart/wpf/

Additional information:
* https://learn.microsoft.com/en-us/dotnet/desktop/xaml-services/xname-directive
* https://stackoverflow.com/a/593151
* https://stackoverflow.com/a/23936061

PR #2599, opened 2023-04-17
ScottPlot 4: DataLogger

### LINKED ISSUE #2377: DataLogger and DataStreamer: new plot types
This is a recurring problem and it would be nice if there were a simpler way to solve it.

Consider a situation where you have incoming sensor data you wish to plot in real time:

* Multi-channel readings
* Hundreds or thousands of readings per second
* Readings stream in faster than we need to update the plot
* The plotted data grows infinitely (and data structure should expand as needed)
* Autoscaling should be smart (don't traverse all the data to rescale on every redraw)

This is hard because (1) signal plots cannot easily grow and (2) automatic axis scaling is not currently optimized for grow-only data.

## Demo

I created a fake data generator and added a demo WinForms app that tries to accomplish this goal using signal plots, but since signal plots can't change the length of their data arrays the plot has to be cleared and signals re-added on every frame redraw. This is a poor design, but I hope that a new plot type can make this better.

https://github.com/ScottPlot/ScottPlot/blob/98744cf82ce915b8aca8bd4a4d244d38553dddcc/src/ScottPlot4/ScottPlot.Demo/ScottPlot.Demo.WinForms/WinFormsDemos/DataLogger.cs

data-logger

## Related
* https://scottplot.net/faq/live-data/
* https://github.com/swharden/FtdiSharp
* https://github.com/swharden/ADC-10-F103C
* https://github.com/swharden/SeriPlot

swharden:
This PR will track progress toward creating a new `DataLogger` plot type optimized for recording and displaying streaming data (#2377)


swharden:
I made a lot of progress here so I'm going to merge this in, but I want to test this new plot type on real data in an IoT application so I'll wait a few days until I begin working on that sensor project:

* https://github.com/swharden/FtdiSharp

New data logging plot types are in in that they:
* Manage data privately, only exposing `Add()` and `Clear()`
* Track axis limits as data comes in (not when `GetAxisLimits()` is called)
* Know the `Plot` that contains them, allowing them to read/set axis limits
* Set axis limits within the `Render()` function

The code is actually very simple, so different plot types should be managed by creating different plot types (e.g., `ScatterDataLogger`, `SignalDataLogger`, etc.)

Issue #2609, opened 2023-04-26
SP5: Two plot control and keyboard input will stackoverflow


taya92413:
## Bug Report

**Issue:** Two plot controls and then keyboard input will System.StackOverflowException: 'Exception_WasThrown' in WindowsBase.dll or CoreLib.dll.
I tested 5.0.4 5.0.3 5.0.2 have the same problem(.NET 6.0),5.0.1 5.0.0 can be used normally,I also tested .NET Framework 4.6.2, 5.0.4 can be used normally.

**Reproducing:** add two plot control and focus on the window and type something with the keyboard

![螢幕擷取畫面 2023-04-26 161931](https://user-images.githubusercontent.com/35124424/234515138-5f1f757b-1129-4f22-b0b4-b0b35a0d2272.png)

```xaml
// a code sample 
<Grid>
   <Scottplot:WpfPlot x:Name="wpfplot" HorizontalAlignment="Left" VerticalAlignment="Top" Height="249" Width="352" Margin="10,10,0,0" />
   <Scottplot:WpfPlot x:Name="wpfplot_Copy" HorizontalAlignment="Left" VerticalAlignment="Top" Height="249"Width="352" Margin="400,10,0,0" />       
</Grid>
```

### System Details
* **ScottPlot Version:** 5.0.4 beta
* **Operating System:** Windows11
* **Application Type:** WPF
* **.NET Version:** .NET6.0

PR #2616, opened 2023-05-06
SP5 OpenGL Control: Fix keyboard input bug #2609

StendProg:
**Purpose:**
Two plot control and keyboard input will stackoverflow #2609

This bug refers to `GLWPFControl`. https://github.com/opentk/GLWpfControl/issues/82
They made it possible to bypass incorrect keypress handling in version 4.2.3.
This PR updates the version of `GLWPFControl`, and disables incorrect event handling by `GLWPFControl` itself. ScottPlot.WPF's keyboard handlers continue to work correctly if the control gets focus.


swharden:
Thanks @StendProg! I was sick for a few weeks so it took me longer than typical to get to this, but I'm merging it in now 👍

PR #2632, opened 2023-05-17
SP4 Ticks: Improve layout when empty ticks are used

### LINKED ISSUE #2606: SP4: Poor automatic layout behavior for empty tick labels
**ScottPlot Version:** 4.1.63

**Operating System:** Windows 11

**Application Type:** WPF

**Question:**
As the title says.
My current approach is to use `TickLabelFormat`, making it returns an empty string when it's a decimal:

```xaml
<WpfPlot x:Name="plotControl" />
```

```cs
plotControl.Plot.BottomAxis.MinimumTickSpacing(0.5);
plotControl.Plot.BottomAxis.TickLabelFormat((d) => d % 1 == 0 ? d.ToString() : string.Empty);
```

However this approach causes inconsistency on layout when zooming with mouse wheel:

swharden:
* resolves #2606


swharden:
This PR is an error, and the commit does not adequately solve the issue. Closing.

PR #2639, opened 2023-05-18
SP4 BinnedHistogram

### LINKED ISSUE #2453: 2D Histogram
We recently added a new Histogram class (#2389, #2403) but it would be great if we could achieve similar functionality in two dimensions. 

Histograms would be built by accumulating X/Y pairs into binned ranges. Histogram output could be plotted as heatmaps, with color representing the count in each 2D bin.

An example use case would be for visualizing [flow cytometry](https://en.wikipedia.org/wiki/Flow_cytometry) data

image

swharden:
Work toward a new plot type that consumes X/Y coordinates, bins them to create a 2D histogram, and displays the result as a heatmap. Useful for plotting data such as [flow cytometry](https://en.wikipedia.org/wiki/Flow_cytometry)

binned-histogram

* Resolves #2453


swharden:
This is done now, and here's how to use it:

```cs
// create a binned histogram
var hist2d = plt.AddBinnedHistogram(100, 100);

// data is a collection of X/Y points
Coordinate[] flowData = DataGen.FlowCytometry();

// add X/Y points to the histogram
hist2d.AddRange(flowData);
```

binned-histogram

bigsteiny:
Very handy, thankyou - however I seem to be having quite some trouble if my Coordinate has any negative values. This seems unsupported at the moment?

PR #2646, opened 2023-05-20
Fixed  LogDistributed hard code about logScale mintick

hellfo:
Fixed LogDistributed hard code about logScale mintick


swharden:
Thanks for this PR @hellfo!

Note that this PR changes the default minor tick count in log axis mode from 5 to 10, and I think this is an improvement

5 ticks | 10 ticks
---|---
image|image

PR #2675, opened 2023-06-03
Fix for Test_FormatCanBeReversed() crashing than decimal separator differ from "."

DmitryKotenev:
**Purpose:**
Avoid crashing Test_FormatCanBeReversed() when decimal separator differ from ".".

On my computer ScottPlot.Tests.Axis.TickLabel.Test_FormatCanBeReversed() crashes due to hard coded string for expected label, that does not count different decimal separators in different Cultures may be. This pull request suggests fix this test crash.

Crashing of the test can be reproduced by setting decimal separator different from "." after plt initialisation:

```cs

//...
var plt = new ScottPlot.Plot(800, 600);
plt.SetCulture(null, ","); // Set decimal separator to ","
//...

```

I've added commit to reproduce test crash:
https://github.com/ScottPlot/ScottPlot/commit/6c01229734398b1d312442bd293c5180d88180aa


swharden:
Hi @DmitryKotenev, thanks for identifying this issue and fixing it! I'm disappointed it took me so long to get to this open PR, but I really appreciate your effort on this! 🚀

PR #2679, opened 2023-06-08
Fixed `ToString` in HSpan.cs

RachamimYaakobov:
Fixed a small error in the function `ToString`


swharden:
Thanks for identifying this issue and fixing it @RachamimYaakobov!

PR #2708, opened 2023-06-30
population plot: add line color options

johndoh:
**Purpose:**
make the box border and error/std dev bar color configurable in population plot, currently hardcoded as black and that is kept as the default

```cs
PopulationPlot popPlot = plt.AddPopulations(populations);
popPlot.BoxBorderColor = Color.Red;
popPlot.ErrorStDevBarColor = Color.Blue;
```


swharden:
Thanks @johndoh, this looks great!

Issue #2664, opened 2023-05-26
Support for Avalonia 11.0.0-preview8 and later


Developer-Alexander:
## Feature Suggestion

**Feature description:** 
Please support Avalonia 11.0.0-preview8 and upcoming versions in the future.

swharden:
To keep things simple, the ScottPlot targets the latest _release_ version of Avalonia (not the latest _preview_ version)

Related issues:
* #2507
* #2321
* #2184
* #2183

Issue #2507, opened 2023-03-24
Support for Avalonia 11


PremekTill:
## Feature Suggestion

**Feature description:**

Hello :)

The Avalonia framework is moving closer to releasing v11 from preview, with many improvements over the previous version. Currently, there is a feature freeze in place to iron out some of the remaining bugs and hopefully move to a RC and full release as soon as possible ([https://github.com/AvaloniaUI/Avalonia/discussions/10599](https://github.com/AvaloniaUI/Avalonia/discussions/10599)).

Unfortunately, there are quite a few breaking changes between the versions, meaning that the current AvaPlot cannot be used in v11 as-is ([https://github.com/AvaloniaUI/Avalonia/wiki/Breaking-Changes](https://github.com/AvaloniaUI/Avalonia/wiki/Breaking-Changes)). Fortunately, for ScottPlot 5, the relevant changes are mostly some namespace shuffling and a switch from SaveFileDialog to IStorageProvider, which provides better support for mobile and browser APIs.

I have tried my hand on getting ScottPlot 5's AvaPlot to work with Avalonia 11 in [my fork](https://github.com/PremekTill/ScottPlot/tree/avalonia_11) (currently only tested on Win10 with very simple stationary plots). If there is interest, I can submit a PR to prepare the changes for the eventual release of v11 and test them a bit more.

However, there are some questions which should probably be discussed first: should the changes be made to both ScottPlot 4 and 5? Should we support both v0.10 and v11, or only one of them? There seems to be more code which might need changing in ScottPlot 4, but it should still be fairly doable. From my point of view, an interesting solution could be too keep ScottPlot 4 at 0.10.18 and drop it from ScottPlot 5, which would only support the latest version - this way, ScottPlot 4 would be there to support legacy apps, whereas ScottPlot 5 would stay at the cutting edge. For most apps with more than basic UI, the porting process from Avalonia 0.10 to 11 will likely be a lot more involved than from ScottPlot 4 to 5, so I wouldn't expect that many to decide to upgrade Avalonia but find upgrading ScottPlot to be a dealbreaker. This also depends on the planned release schedule for ScottPlot 5, which I am not familiar with.

swharden:
Hi @PremekTill, thanks for this thoughtful message! As you point out it will require effort for users moving from Avalonia 10 to 11, so getting the timing right is important. My gut reaction is to keep things simple and move everything (ScottPlot 4 and 5) to Avalonia 11 once it is out of preview.

[...]

Issue #2321, opened 2022-12-23
Avalonia 11


petryaevve:
## Feature Suggestion
Upgrade to Avalonia 11 version
**Feature description:** Describe your idea here
Update Avalonia version for 11 preview.

swharden:
Hi @petryaevve, 

It looks like the current release of Avalonia is 0.10.18
https://www.nuget.org/packages/Avalonia/

I think the plan is to only target the latest _released_ version of Avalonia, but I'm happy to collect notes of people who have adapted ScottPlot to work with preview versions too.

Related
* #2183

PR #2720, opened 2023-07-05
ScottPlot4: Migrate from Avalonia 10 to Avalonia 11

Fruchtzwerg94:
**Purpose:**
* Update ScottPlot to the latest Avalonia 11 release

### DIFF
--- a/src/ScottPlot4/ScottPlot.Avalonia/AvaPlot.axaml.cs
+++ b/src/ScottPlot4/ScottPlot.Avalonia/AvaPlot.axaml.cs
@@ -52,17 +52,26 @@ namespace ScottPlot.Avalonia
         /// </summary>
         public event EventHandler PlottableDragged;
 
+        [Obsolete("use 'PlottableDragged' instead", error: true)]
+        public event EventHandler MouseDragPlottable;
+
         /// <summary>
         /// This event is invoked right after a draggable plottable was dropped.
         /// The object passed is the plottable that was just dropped.
         /// </summary>
         public event EventHandler PlottableDropped;
 
+        [Obsolete("use 'PlottableDropped' instead", error: true)]
+        public event EventHandler MouseDropPlottable;
+
         private readonly Control.ControlBackEnd Backend;
         private readonly Dictionary<ScottPlot.Cursor, Ava.Input.Cursor> Cursors;
         private readonly Ava.Controls.Image PlotImage = new Ava.Controls.Image();
         private float ScaledWidth => (float)(Bounds.Width * Configuration.DpiStretchRatio);
         private float ScaledHeight => (float)(Bounds.Height * Configuration.DpiStretchRatio);
+
+        [Obsolete("Reference Plot instead of plt")]
+        public ScottPlot.Plot plt => Plot;
 
         static AvaPlot()
         {
--- a/src/ScottPlot4/ScottPlot.Avalonia/AvaPlot.axaml.cs
+++ b/src/ScottPlot4/ScottPlot.Avalonia/AvaPlot.axaml.cs
@@ -13,6 +13,8 @@ using Avalonia.VisualTree;
 
 using Ava = Avalonia;
 
+#pragma warning disable CS0067 // don't warn about obsolete events that aren't used
+
 namespace ScottPlot.Avalonia
 {
     /// <summary>
--- a/src/ScottPlot4/ScottPlot.Avalonia/AvaPlot.axaml.cs
+++ b/src/ScottPlot4/ScottPlot.Avalonia/AvaPlot.axaml.cs
@@ -7,21 +7,17 @@ using Avalonia;
 using Avalonia.Controls;
 using Avalonia.Input;
 using Avalonia.Interactivity;
+using Avalonia.Platform.Storage;
 using Avalonia.Threading;
 using Avalonia.VisualTree;
 
 using Ava = Avalonia;
-
-#pragma warning disable IDE1006 // lowercase public properties
-#pragma warning disable CS0067 // unused events
-#pragma warning disable CS0618 // TODO: Storage API
 
 namespace ScottPlot.Avalonia
 {
     /// <summary>
     /// Interaction logic for AvaPlot.axaml
     /// </summary>
-
     [System.ComponentModel.ToolboxItem(true)]
     [System.ComponentModel.DesignTimeVisible(true)]
     public partial class AvaPlot : UserControl, ScottPlot.Control.IPlotControl
@@ -56,26 +52,17 @@ namespace ScottPlot.Avalonia
         /// </summary>
         public event EventHandler PlottableDragged;
 
-        [Obsolete("use 'PlottableDragged' instead", error: true)]
-        public event EventHandler MouseDragPlottable;
-
         /// <summary>
         /// This event is invoked right after a draggable plottable was dropped.
         /// The object passed is the plottable that was just dropped.
         /// </summary>
         public event EventHandler PlottableDropped;
 
-        [Obsolete("use 'PlottableDropped' instead", error: true)]
-        public event EventHandler MouseDropPlottable;
-
         private readonly Control.ControlBackEnd Backend;
         private readonly Dictionary<ScottPlot.Cursor, Ava.Input.Cursor> Cursors;
         private readonly Ava.Controls.Image PlotImage = new Ava.Controls.Image();
         private float ScaledWidth => (float)(Bounds.Width * Configuration.DpiStretchRatio);
         private float ScaledHeight => (float)(Bounds.Height * Configuration.DpiStretchRatio);
-
-        [Obsolete("Reference Plot instead of plt")]
-        public ScottPlot.Plot plt => Plot;
 
         static AvaPlot()
         {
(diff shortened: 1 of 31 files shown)


swharden:
Hi @Fruchtzwerg94, thanks so much for this PR! I'm excited Avalonia 11 is finally here 🚀 

At first glance this PR looks great to me, but I want to give two others the opportunity to look it over and weigh in if they wish: @bclehmann who made the first Avalonia ScottPlot control (#503) and @Xerxes004 who recently made a ScottPlot control targeting a preview version of Avalonia 11 (#2184). Since this PR is large I recommend providing non-blocking feedback where possible so we can merge soon, then we can pursue refinements in future issues/PRs.

There's been a lot of discussion about Avalonia 11 and I want to recognize the others who participated along the way:
* #2720 
* #2184 
* #2664 
* #2507 
* #2321 
* #2184 
* #2183

bclehmann:
My only note is that the cookbook looks wrong, Avalonia 11 automatically sets dark mode according to system preferences and that causes an issue here:
image

bclehmann (approved):
Looks good, just the thingy about dark mode, which you can fix if you want or you can just force light mode like it was before.

Fruchtzwerg94:
Hi,

I've checked some of the remaining points and updated my PR including the following changes:
- Replacement of OpenFileDialog by Storage API like already proposed by @Xerxes004 
- Replacement of deprecated CheckedChanged events
- Used light mode by default
- Minor code cleanup

So all in all, it should be fine to merge from my side.

BR

src/ScottPlot4/ScottPlot.Avalonia/README.md
```diff
@@ -14,7 +14,6 @@ Add the following to your start-up sequence:
 double[] dataX = new double[] { 1, 2, 3, 4, 5 };
 double[] dataY = new double[] { 1, 4, 9, 16, 25 };
 
-AvaPlot avaPlot1 = this.Find<AvaPlot>("AvaPlot1");
 avaPlot1.Plot.AddScatter(dataX, dataY);
```
Xerxes004, line 17: Should be `AvaPlot1` since the variable is gone now.

src/ScottPlot4/ScottPlot.Demo/ScottPlot.Demo.Avalonia/AvaloniaDemos/PlotViewer.axaml.cs
Xerxes004, line 1: Line 58 still uses the `Find<>` function

src/ScottPlot4/ScottPlot.Demo/ScottPlot.Demo.Avalonia/CookbookControl.axaml
```diff
@@ -13,6 +13,7 @@
 		<Grid.RowDefinitions>
 			<RowDefinition Height="auto"/>
 			<RowDefinition Height="*"/>
+            <RowDefinition Height="auto"/>
```
Xerxes004, line 16: Formatting

src/ScottPlot4/ScottPlot.Demo/ScottPlot.Demo.Avalonia/CookbookControl.axaml
```diff
@@ -30,7 +31,9 @@
 			</DockPanel>
 		</Border>
 
-		<Border Grid.Column="0" Grid.Row="2" Margin="5" BorderThickness="2" BorderBrush="#2a3a56" Background="#2a3a56" >
+        <TextBlock Name="BenchmarkLabel" Grid.Row="2" Margin="5,0,0,0" Text="Benchmark"/>
```
Xerxes004, line 34: Formatting

Xerxes004 (commented):
I had just a couple of notes in there, mostly on formatting stuff. Looks good!

Xerxes004 (approved):

src/ScottPlot4/ScottPlot.Avalonia/README.md
Fruchtzwerg94: Good catch, will be fixed.

src/ScottPlot4/ScottPlot.Demo/ScottPlot.Demo.Avalonia/AvaloniaDemos/PlotViewer.axaml.cs
Fruchtzwerg94: This is because `avaPlot1` is a private field, which means it can't be accessed withoiut the find function.
If this is a problem, a small designchange of the `AvaPlotViewer` would be necessary.

src/ScottPlot4/ScottPlot.Demo/ScottPlot.Demo.Avalonia/CookbookControl.axaml
Fruchtzwerg94: Uuuh, some ugly tabs are used all around here ... Will be fixed :D

src/ScottPlot4/ScottPlot.Demo/ScottPlot.Demo.Avalonia/CookbookWindow.axaml.cs
```diff
@@ -14,14 +14,14 @@ public CookbookWindow()
...
-            TreeView DemoTreeview = (TreeView)sender; ;
+            TreeView DemoTreeview = (TreeView)sender;
 
             Cookbook.TreeNode selectedDemoItem = null;
             if (DemoTreeview.SelectedItems.Count > 0)
             {
                 selectedDemoItem = (Cookbook.TreeNode)DemoTreeview.SelectedItems[0];
             }
-            if (selectedDemoItem != null && selectedDemoItem.ID != null)
+            if (selectedDemoItem?.ID != null)
```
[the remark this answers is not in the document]
Fruchtzwerg94: Not sure what you mean here, but the two statements should work exactle the same way. Could you please give me some more details about your thoughts here?

src/ScottPlot4/ScottPlot.Demo/ScottPlot.Demo.Avalonia/CookbookControl.axaml
Fruchtzwerg94: Fixed

src/ScottPlot4/ScottPlot.Demo/ScottPlot.Demo.Avalonia/CookbookWindow.axaml.cs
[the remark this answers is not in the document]
bclehmann: In general for a null check you're supposed to prefer `is null` or `is not null` rather than `== null` or `!= null`. We've been inconsistent about this throughout the codebase but this is what Microsoft and most other style guides recommend.

`==` is affected by operator overloading which is rarely the behaviour that one wants for a null check. An operator overload that changes comparison with null should be rare, but it also allows more helpful warnings and errors (for example it can prevent null checks on non-nullable value types).

And there are potentially performance differences, `is` can always be compiled to MSIL's `ceq` instruction, where `==` is only compiled that way if `==` isn't overloaded (and you're using a sufficiently recent version of Roslyn).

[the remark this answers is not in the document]
Fruchtzwerg94: Got it, I was not fully aware of it. Thanks for the explanation. I've fixed it.

src/ScottPlot4/ScottPlot.Demo/ScottPlot.Demo.Avalonia/AvaloniaDemos/PlotViewer.axaml.cs
Xerxes004: Ok sounds good

Xerxes004: @Fruchtzwerg94 you can resolve this comment whenever

src/ScottPlot4/ScottPlot.Avalonia/AvaPlot.axaml.cs
```diff
@@ -56,27 +52,18 @@ public partial class AvaPlot : UserControl, ScottPlot.Control.IPlotControl
         /// </summary>
         public event EventHandler PlottableDragged;
 
-        [Obsolete("use 'PlottableDragged' instead", error: true)]
```
bclehmann, line 57: Sorry I didn't catch these earlier, but I think we need to preserve these events.

src/ScottPlot4/ScottPlot.Demo/ScottPlot.Demo.Avalonia/AvaloniaDemos/PlotViewer.axaml.cs
Fruchtzwerg94: Fixed.

src/ScottPlot4/ScottPlot.Avalonia/AvaPlot.axaml.cs
Fruchtzwerg94: Is there a reason why the events should remain?
Since they are never called, this could cause a lot of confusion for users of the library. Thats was my intention to remove them completely.
But if necessary, I could add them like they where again ...

bclehmann: They're part of the public API, we don't use them but our users might.

Fruchtzwerg94: With "using", you mean users are subscribing to it? Since the event is never called / fired it still wouldn't make sense to keep it.

swharden: Those obsolete decorators are there to help guide people who are upgrading from very old versions of ScottPlot. Without the obsolete message, these users' code would break with no indication of how to fix it. Messages like this provide a path for them to get their code working again.

I'll stick the messages back in now and merge this PR in the next few minutes! 👍

swharden: Oh, `<TreatWarningsAsErrors>true</TreatWarningsAsErrors>` made that harder than it needed to be lol

I'll try to fix that, then merge now

PR #2732, opened 2023-07-09
adjust the height to show the link

heartacker:
highter to show the link

image


swharden:
Hi @heartacker, thanks for this fix!

Issues like this are difficult to identify because the layout looks good with some system fonts but bad with others. I'm glad you were able to correct this issue so the demo will look better for international users! 🚀

(this is how it looks with my system font)

image

Issue #2733, opened 2023-07-10
Datalogger do not work as expected


KroMignon:
## Bug Report

**Issue:** 

I have some issues with the DataLogger:
* when DataLogger trace has been rendered, the renderer will crash after resetting DataLogger and adding 1 point. When DataLogger has at least 2 points it won't crash! My workaround, for this issue is to set Visibility to false when trace has less than 2 points.
* I would like to change `Plot.AddDataLogger()` to be able to define trace color, line width, name, etc. like `Plot.AddScatterList()` method.
* Maybe `Plot.AddDataStreamer()` should also be changed to allow color, width, name, etc.

**Reproducing:** 

Change DataLogger.cs is WinFormsDemo to add a checkbox, called chkBug and add following code to the CheckedChanged event:
```cs
private void chkBug_CheckedChanged(object sender, EventArgs e)
{
    Logger.Clear();
    formsPlot1.Refresh();
}
```

and change AddRandomWalkData() as follow
```cs
private void AddRandomWalkData(int count)
{
    if (chkBug.Checked == false)
    {
        for (int i = 0; i < count; i++)
        {
            LastPointValue = LastPointValue + Rand.NextDouble() - .5;
            Logger.Add(Logger.Count, LastPointValue);
        }
    }
    else if (Logger.Count == Logger.CountOnLastRender)
    {
        LastPointValue = LastPointValue + Rand.NextDouble() - .5;
        Logger.Add(Logger.Count, LastPointValue);
    }
} 
```

When chkBug is checked, the application will crash!

### System Details
* **ScottPlot Version:** 4.1.65
* **Operating System:** Windows 10
* **Application Type:** WPF
* **.NET Version:** .NET 6.0

PR #2736, opened 2023-07-10
SP4: Add optional arguments to AddDataLogger() and DataStreamer()

KroMignon:
Use color palette to define DataLogger and DataStream trace color as described in #2733


swharden:
Hi @KroMignon, thanks for your work here and in #2735!

After working on the first bullet of #2733 I forgot there were more recommendations there. Thanks for following-up with this PR!

PR #2749, opened 2023-07-13
SP5 Avalonia: Sandbox should directly depend on Avalonia.Desktop

bclehmann:
**Purpose:**
Previous to #2748 `Sandbox.Avalonia` was depending on Avalonia.Desktop through `ScottPlot.Avalonia`. Now `ScottPlot.Avalonia` doesn't include `Avalonia.Desktop`, so we need to depend on it directly.

Without this the sandbox fails to compile as `AppBuilder.UsePlatformDetect` does not exist (see [Program.cs:20](https://github.com/ScottPlot/ScottPlot/blob/8b26a908a5d978a8ce4d0e5df83b690f41eb8d6d/src/ScottPlot5/ScottPlot5%20Sandbox/Sandbox.Avalonia/Program.cs#L20)). This method is an extension method from `Avalonia.Desktop`.


bclehmann:
That CI error looks like it comes from https://github.com/ScottPlot/ScottPlot/commit/8b26a908a5d978a8ce4d0e5df83b690f41eb8d6d

PR #2753, opened 2023-07-14
SP5: `Fonts.Default` and `Fonts.Detect()`

### LINKED ISSUE #2746: SP5: Improve support for Chinese characters
wpf app
ScottPlot Version="5.0.6-beta"
image

swharden:
adds support for international character sets

* resolves #2746


src/ScottPlot5/ScottPlot5/FontService.cs
```diff
@@ -1,26 +1,27 @@
...
+        return new(SKFontManager.Default.FontFamilies, StringComparer.InvariantCultureIgnoreCase);
     }
 
-    private static SKTypeface GetDefaultSansFont()
+    public static SKTypeface GetDefaultSansFont()
     {
-        HashSet<string> installedFonts = new(SKFontManager.Default.FontFamilies);
+        var installedFonts = GetInstalledFonts();
 
         string[] preferredFonts = { "Open Sans", "Segoe UI", "Lato", "DejaVu Sans", "Helvetica" };
```
bclehmann, line 26: I think users should probably be able to just edit this, one might want to modify the default _and_ set a fallback font, and that's inconvenient to do with only `SetDefaultFont`

src/ScottPlot5/ScottPlot5/FontService.cs
```diff
@@ -65,4 +66,44 @@ private static SKTypeface GetDefaultSerifFont()
...
+        // https://github.com/ScottPlot/ScottPlot/issues/2746
+        DefaultFontName = DetectFont(c);
+    }
+
+    /// <summary>
+    /// Set the default font to the best one for displaying the given character.
+    /// This method helps display Chinese, Japanese, and Korean (CJK) characters.
+    /// </summary>
+    public static void SetDefaultFont(string fontName, bool throwIfNotFound = false)
+    {
```
bclehmann, line 95: I think this should at least return `bool success` or something like that. I can understand not wanting to throw by default, but this current signature makes it pretty hard to tell if the method has failed, and this is a method that will likely fail often.

Consider that if a user wants to detect a failure to set the font (i.e. to attempt a fallback font) they need to do this:
```cs
try
{
    SetDefaultFont("Comic Sans MS", true);
}
catch (InvalidOperationException e) 
{
    SetDefaultFont("Fallback Font"); // Potentially multiple fallback fonts
}
```

This should probably look more like this:
```cs
if (!SetDefaultFont("Comic Sans MS"))
    SetDefaultFont("Fallback Font");
```
Or my personal preference:
```cs
SetDefaultFont("Comic Sans MS", fallbackList); // Perhaps this should even just be SetDefaultFont(fontList);
```

Also `InvalidOperationException` is probably the wrong exception, maybe `FileNotFoundException` or `NotSupportedException`. Or a custom exception class.

src/ScottPlot5/ScottPlot5/FontService.cs
```diff
@@ -34,9 +35,9 @@ private static SKTypeface GetDefaultSansFont()
         return SKTypeface.Default;
     }
 
-    private static SKTypeface GetDefaultMonospaceFont()
+    public static SKTypeface GetDefaultMonospaceFont()
     {
-        HashSet<string> installedFonts = new(SKFontManager.Default.FontFamilies);
+        var installedFonts = GetInstalledFonts();
 
         string[] preferredFonts = { "Roboto Mono", "Consolas", "DejaVu Sans Mono", "Courier" };
```
heartacker, line 42: why not use system by default？

src/ScottPlot5/ScottPlot5/FontService.cs
swharden: Thanks for pointing this out! I'll try to simplify this system by making all these methods private (solving the issue) and only exposing a few strings which the user can set how they want if they don't like the defaults:

https://github.com/ScottPlot/ScottPlot/blob/418665bb0e6ed053a29ee53fbc4b0fd22218d2d8/src/ScottPlot5/ScottPlot5/FontService.cs#L11-L15

I don't think we even need a fallback. If the font doesn't exist, SkiaSharp renders it using whatever the system default is

src/ScottPlot5/ScottPlot5/FontService.cs
swharden: I'm adding a `bool Exists(string fontName)` method to make this easier, then the user can choose how to implement their own fallback logic

src/ScottPlot5/ScottPlot5/FontService.cs
swharden: Great question! Most Windows systems use Segoe UI by default. I find the anti-alising looks bad. Look at the "V" in the word "Vertical"

image

Compare to Open Sans which looks fantastic, and I prefer as a default

image

swharden:
After reviewing the feedback I got and thinking about this problem some more, I decided to simplify this class. The user can handle fancy things like font lookups and fallback logic. I'll merge it in like it is now, but I welcome continued feedback and am happy to change it with a new PR/issue if improvements can be made.

## Default
```cs
ScottPlot.Plot plt = new();
plt.Add.Signal(Generate.Sin());
plt.Add.Signal(Generate.Cos());
plt.TopAxis.Label.Text = Fonts.Default;
plt.BottomAxis.Label.Text = "Horizontal Axis";
plt.LeftAxis.Label.Text = "Vertical Axis";
```
image

## Changing the Default
```cs
Fonts.Default = "Comic Sans MS";

ScottPlot.Plot plt = new();
plt.Add.Signal(Generate.Sin());
plt.Add.Signal(Generate.Cos());
plt.TopAxis.Label.Text = Fonts.Default;
plt.BottomAxis.Label.Text = "Horizontal Axis";
plt.LeftAxis.Label.Text = "Vertical Axis";
```
image

## Best Font Identification

```cs
Fonts.Default = Fonts.Detect('频') ?? Fonts.System;

ScottPlot.Plot plt = new();
plt.Add.Signal(Generate.Sin());
plt.Add.Signal(Generate.Cos());
plt.TopAxis.Label.Text = "频率信号削减1/32";
plt.BottomAxis.Label.Text = "Horizontal Axis";
plt.LeftAxis.Label.Text = "Vertical Axis";
```

image

swharden:
> I still believe that we should render correctly for all of user, not be beautiful firstly

Thanks for being persistent about this point! I improved it by always favoring the system default font (and only using Open Sans if the default font is Segoe UI)

https://github.com/ScottPlot/ScottPlot/commit/e83eeed439d3ef43a6a705fd518347b6c3224d06

PR #2782, opened 2023-07-21
Remove "experimental" from Scatter List

prime167:
The Scatter List as been added for over a year. I think it's time to graduate it from an experimental feature and make it an official fully launched feature.


swharden:
I certainly agree ScatterPlotList is no longer "experimental" - thanks for this PR!

PR #2787, opened 2023-07-21
SP5: RandomDataGenerator.Random()

Silent0Wings:
RandomDataGenerator


swharden:
Thanks @Silent0Wings! I used this PR as a jumping-off point to add XML docs and a few more methods.

Issue #2790, opened 2023-07-21
SP5: `CoordinateRect` needs a `Contains()` method


swharden:
https://github.com/ScottPlot/ScottPlot/blob/main/src/ScottPlot5/ScottPlot5/CoordinateRect.cs should be extended to add methods like:

```cs
public bool Contains(Coordinates point)
{
    /* ... */
}
```

and maybe an overload

```cs
public bool Contains(double x, double y)
{
    /* ... */
}
```

with logic like

```cs
return X >= XMin && X <= XMax && Y >= YMin && Y <= YMax;
```

Issue #2789, opened 2023-07-21
SP5: `CoordinateRect` needs a `Coordinates Center` property


swharden:
Add a line below this

https://github.com/ScottPlot/ScottPlot/blob/1a4568c469b6e705846399ae6f13e8930c7a8d86/src/ScottPlot5/ScottPlot5/CoordinateRect.cs#L15-L16

something like

```cs
public Coordinates Center => new(XCenter, YCenter);
```

tijin-abe-thomas:
Hey, Could I work on this issue?

swharden:
Hi @tijin-abe-thomas, indeed! If you create a PR I will be happy to review and merge it quickly 🚀

PR #2809, opened 2023-07-24
SP5:  Added a Center property and Contains method in CoordinateRect

tijin-abe-thomas:
**Purpose:**
* This PR aims to solve issue #2789 
* Added a new Center property
* This PR aims to also solve issue #2790


KroMignon:
@tijin-abe-thomas to avoid code duplication, I would suggest to change code as follow:
```C#
    public bool Contains(Coordinates point) => Contains(point.X, point.Y);
```

tijin-abe-thomas:
@KroMignon Yes, that seems like a better implementation. Have updated the code in the new commit. Thanks :))

tijin-abe-thomas:
Switched the PR to #2813

PR #2817, opened 2023-07-26
SP4 Heatmap: `CenterOfRotation` property

bukkideme:
New property to set the center of axis when we use Rotate.


swharden:
Thanks, I love it!

image

```cs

for (int i=0; i<5; i++)
{
    var hm = plt.AddHeatmap(imageData, lockScales: true);
    hm.XMin = 0;
    hm.XMax = 1;
    hm.YMin = 0;
    hm.YMax = 1;
    hm.Rotation = i*10;
}

for (int i=0; i<5; i++)
{
    var hm = plt.AddHeatmap(imageData, lockScales: true);
    hm.XMin = 2;
    hm.XMax = 3;
    hm.YMin = 0;
    hm.YMax = 1;

    hm.CenterOfRotation = Alignment.MiddleCenter;

    hm.Rotation = i * 10;
}

plt.SetAxisLimits(-1, 4, -1, 2);
```

PR #2820, opened 2023-07-28
SP4 WPF: Fix PlottableDragged add event accessor

tadmccorkle:
Corrects the `PlottableDragged` event accessor adding a handler to the dropped event.


swharden:
Thanks for this fix @tadmccorkle!

PR #2822, opened 2023-07-28
SP4 & SP5 Update Avalonia to 11.0.1

oktrue:
This fix Avalonia package version formatting in SP5 and bump version to 11.0.1


swharden:
Thanks @oktrue!

Issue #2833, opened 2023-08-04
SP5: frequent typeface lookups result in poor performance on Linux


taya92413:
## Bug Report

**Issue:** 
Very poor performance, only 10-20FPS
Referring to #2419 could solve the performance issue, but I am not sure about the problems encountered on .NET Framework, so it was not ultimately resolved.

#2419 method:
Modify source code 
Path: `ScottPlot-main\src\ScottPlot5\ScottPlot5\Extensions\SkiaSharpExtensions.cs`
Add a static SKTypeface to avoid searching for SKTypeface every time rendering occurs.
Finally build the .dll

```cs
    //Add
    private static SKTypeface? _typeface = null;

    public static void ApplyToPaint(this FontStyle fontStyle, SKPaint paint)
    {
        SKFontStyleWeight weight = fontStyle.Bold ? SKFontStyleWeight.Bold : SKFontStyleWeight.Normal;
        SKFontStyleSlant slant = fontStyle.Italic ? SKFontStyleSlant.Italic : SKFontStyleSlant.Upright;
        SKFontStyleWidth width = SKFontStyleWidth.Normal;
        SKFontStyle skfs = new(weight, width, slant);

        //SKTypeface typeface = SKTypeface.FromFamilyName(fontStyle.Name, skfs);

        //Add choose one of the two, they are the same
        SKTypeface typeface = _typeface ?? (_typeface = SKTypeface.FromFamilyName(fontStyle.Name, skfs));
        //SKTypeface typeface = (_typeface == null) ? _typeface = SKTypeface.FromFamilyName(fontStyle.Name, skfs) : _typeface;


        paint.Typeface = typeface;
        paint.TextSize = fontStyle.Size;
        paint.Color = fontStyle.Color.ToSKColor();
        paint.IsAntialias = fontStyle.AntiAlias;
    }
```

Low performance picture:
<img width="600" alt="螢幕擷取畫面 2023-08-04 095502" src="https://github.com/ScottPlot/ScottPlot/assets/35124424/e94bb86b-6858-49b8-840b-4ac9e5f18e32">

**Reproducing:** 
a code sample
Xaml:
```Xaml
<ScottPlot:AvaPlot Name="AvaPlot1" />

```
C#:
```cs
   double[] dataX = { 1, 2, 3, 4, 5 };
   double[] dataY = { 1, 4, 9, 16, 25 };

   AvaPlot avaPlot1 = this.Find<AvaPlot>("AvaPlot1");
[...]

PR #2848, opened 2023-08-07
SP5: Cache typefaces

KroMignon:
Update FontStyle to add a typeface cache to avoid to call too often the font face lookup table.

This should solve issue #2833

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 * https://scottplot.net/contributors/ shows all of ScottPlot's contributors
 
 ## ScottPlot 5.0.8-beta (in development)
+* Rendering: Improved performance by caching typefaces (#2833, #2848) _Thanks @KroMignon and @taya92413_
 
 ## ScottPlot 4.1.67 (in development)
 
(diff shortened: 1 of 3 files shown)


swharden:
Thank you for this fantastic solution @KroMignon! I really appreciate your work improving ScottPlot 5 performance with this pull request 🚀

A few thoughts about this work:

* I hope to to benchmark performance before/after this PR using docker: #2851

* Your original `SetField<T>` function was very elegant! I'm copying the code below so it can be appreciated. However, to simplify the code I chose to put all the logic in the getters for font `Name`, `Bold`, and `Italic`. Your solution worked well and would have been preferred if there were more properties. However, moving this logic to the getters allowed explicit typing (preventing the need for casting) and also avoided multiple levels of `if` statements. In the end I favored the simplicity at the cost of a small amount of code duplication.

```cs
// strategy in the original PR
protected bool SetField<T>(ref T fieldValue, T value)
{
    if (value is string)
    {
        if (string.Compare(fieldValue as string, value as string) == 0)
            return false;
    }
    else if (Equals(fieldValue, value))
        return false;
    fieldValue = value;
    _typeface = null;
    return true;
}
```

Thank you again for this excellent work! 🙏

PR #2862, opened 2023-08-10
SP4: Add markers support to DataLogger

KroMignon:
**Purpose:**

This PR is to enable `Markers` support for `DataLogger`.
Per default, markers are disabled to not change current behavior.

I know that SP4 is in "frozen" state, but it would be nice if this PR could be included 😄

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@
 ## ScottPlot 4.1.67 (in development)
 * DataLogger: Improved appearance of legend items (#2829, #2850) _Thanks @KroMignon and @p4pravin_
 * Radial Gauge Plot: Improved layout for plots with a large number of gauges (#2722) _Thanks @tinuskotze_
+* DataLogger: Added support for markers (#2862) _Thanks @KroMignon_
 
 ## ScottPlot 5.0.7-beta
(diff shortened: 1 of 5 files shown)


swharden:
Hi @KroMignon, thanks for this PR! I'll extend this PR by adding a cookbook recipe (which I just noticed doesn't exist for the data logger plot type yet) then merge in a few minutes and it will be in the next release 🚀

swharden:
To keep the API consistent with `DataStreamer` (and prevent breaking code of people who are already using the `label` argument) I'll modify this PR to remove the arguments customizing markers. Instead, users can opt-in to markers like this:

```cs
var logger = plt.AddDataLogger();
logger.MarkerShape = MarkerShape.openCircle;
logger.MarkerSize = 5;
```

AzureGulf:
Is this available in ScottPlot 5 (am using 5.0.34, but don't see it) and, if so, how is a marker added to a position/coordinate in the plot (not just it's shape etc)?
At the moment just add directly to the Plot as defined in the xaml, WinUI 3, and seems to work OK, as per code below:
```
 var mkrCoords = new Coordinates(DateTime.Now.ToOADate(), lastPointValue);
 xaml_winUIPlot.Plot.Add.Marker(mkrCoords, MarkerShape.FilledCircle, 12, Colors.Red);
```

Issue #2871, opened 2023-08-14
SP5: `CoordinateRect` needs methods for expanding to include `Coordinates`


swharden:
needs methods to expand to include a point or rectangle, something like

```cs
CoordinateRect rect1 = new(-10, 10, -10, 10);

Coordinates coord = new(13. 42);

CoordinateRect rect2 = rect1.Expanded(coord);
```

notably the values are mutable, so perhaps this could run in place

https://github.com/ScottPlot/ScottPlot/blob/42987b60b6fb90e94d0fb67a60c7dfafc73d8d9a/src/ScottPlot5/ScottPlot5/Primitives/CoordinateRect.cs#L8-L13

aespitia:
Is the idea here that rect1 grows to a size to include the coordinate, and returns a new rectangle?

swharden:
Correct! That's my thinking of what would be most useful

aespitia:
I've never contribued before, so apparently I don't know how to make a pull request for this repo.  Is this something like what you had in mind?

public CoordinateRect Expanded(Coordinates point) {
        double exLeft = Left;
        double exRight = Right;
        double exBottom = Bottom;
        double exTop = Top;

        if (!Contains(point))
        {
            if (point.X < Left) exLeft = point.X;
            if (point.X > Right) exRight = point.X;
            if (point.Y < Bottom) exBottom = point.Y;
            if (point.Y > Top) exTop = point.Y;
        }

        return new CoordinateRect(exLeft, exRight, exBottom, exTop);
    }

swharden:
> Is this something like what you had in mind?

Indeed! 

> I don't know how to make a pull request for this repo

If you feel like figuring it out you can probably find a YouTube video that walks through the steps, and I'm happy to support you along the way by answering any questions that may pop up! If it ends up being too much work or you don't want to create your own pull request, let me know and I can create the pull request myself 👍

swharden:
> Is this something like what you had in mind?

Maybe it could be even simpler with something like:

```cs
return new CoordinateRect(
    left: Math.Min(Left, point.Left),
    right: Math.Max(Right, point.Right),
    bottom: Math.Min(Bottom, point.Bottom),
    top: Math.Max(Top, point.Top),
);
```

swharden:
> YouTube video

Here's one that describes how to use GitHub Desktop (a GUI for git) 

https://www.youtube.com/watch?v=8x6V5IOuXog

[...]

PR #2890, opened 2023-08-17
feat: adding Expandable Rectangle issue 2871

aespitia:
This feature adds an Expandable method to the rectangle that takes in a Coordinate.  The method checks if the coordinate already exists in the rectangle, if it does, then it returns a copy of the existing rectangle, if it doesn't, the method will return an expanded version of the rectangle that does contain the new coordinate.  

#2871

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,12 @@
 * https://scottplot.net/changelog/ is a formatted version of this document
 * https://scottplot.net/contributors/ shows all of ScottPlot's contributors
 
-## ScottPlot 5.0.8-beta (in development)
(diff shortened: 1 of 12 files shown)


swharden:
Thanks @aespitia, this looks great! I got backed-up on some other projects and this PR took a lot longer than typical to get to, but I really appreciate it! 🚀

aespitia:
> Thanks @aespitia, this looks great! I got backed-up on some other projects and this PR took a lot longer than typical to get to, but I really appreciate it! 🚀

was a great first check-in experience.  your feedback was super helpful and timely, looking forward to using SP5!

Issue #2886, opened 2023-08-15
SP5: Improve legend support for the `FillY` plot type


msroest:
## Feature Suggestion

**Feature description:**
I'm attempting to create a FillY plot with 2 series 1 for power usage and 1 for Solar power generated I'm currently able to get it to look the way I'm wanting by using the FillY and setting the second series to all 0's so it just fills from my data point to the x-axis.  However I would like to still have a Legend and it appears that the FillY doesn't have a Label to set and is hardcoded to always return no LegendItems
https://github.com/ScottPlot/ScottPlot/blob/main/src/ScottPlot5/ScottPlot5/Plottables/FillY.cs#L8
vs Scatter
https://github.com/ScottPlot/ScottPlot/blob/main/src/ScottPlot5/ScottPlot5/Plottables/Scatter.cs#L28

I can't imagine I'm the only one who wants to do something like this;
What I'm currently getting:
![generation](https://github.com/ScottPlot/ScottPlot/assets/91028/2cd4fd18-b978-4039-a326-1b8440a4818b)

Scatter with the same plot but with legend:
![generation-1](https://github.com/ScottPlot/ScottPlot/assets/91028/d50a8b62-d24d-4bb9-90f1-2bd5ec22fca4)

swharden:
Hi @msroest, thanks for this suggestion! I agree it's a good idea to add legend support. If someone is available to make a pull request, they can probably help make this happen faster than I can!

You correctly identified the line of code requiring modification:
https://github.com/ScottPlot/ScottPlot/blob/42987b60b6fb90e94d0fb67a60c7dfafc73d8d9a/src/ScottPlot5/ScottPlot5/Plottables/FillY.cs#L8

And by adding logic like scatter plot, FillY plots will appear in the legend too:
https://github.com/ScottPlot/ScottPlot/blob/42987b60b6fb90e94d0fb67a60c7dfafc73d8d9a/src/ScottPlot5/ScottPlot5/Plottables/Scatter.cs#L28-L34

I'll leave this issue open until this is resolved. If no one fixes it after a week or two, I'll fix it myself before the next release 👍

msroest:
Hey @swharden,
  Thanks for the note that you'd be up for the support, I'll take a shot at trying to submit a PR here over the next few days.

PR #2896, opened 2023-08-22
Add legend support to FillY Plottable

msroest:
Closes https://github.com/ScottPlot/ScottPlot/issues/2886

Seems to work as expected:
image

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@
 ## ScottPlot 5.0.9-beta (in development)
 * PixelPadding: `TotalHorizontal` and `TotalVertical` renamed to `Horizontal` and `Vertical` (#2874, #2878) _Thanks @viktoriussuwandi_
 * CoordinateRect: Added `Expanded()` method for creating a copy of the rectangle expanded to include a given point (#2871, #2890) _Thanks @aespitia_
+* FillY: Added legend support (#2886, #2896) _Thanks @msroest_
 
 ## ScottPlot 4.1.68 (in development)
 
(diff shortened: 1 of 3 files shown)


msroest (commented):
Moved to Draft as FillY with no linestyle defined the legend doesn't show properly (it doesn't show the color just the label).  Need to work on it more

msroest:
> Moved to Draft as FillY with no linestyle defined the legend doesn't show properly (it doesn't show the color just the label). Need to work on it more

I've changed the code to ignore the line style and set the legend to the fill color.  I don't love it so I'm open to alternate ideas.

swharden:
Looks like the error is
```
Predefined type 'System.ValueTuple`3' is not defined or imported
```

ScottPlot 5 also targets .NET Framework 4.6.2 so I'm guessing there's a language feature you're using that isn't supported by that framework

https://github.com/ScottPlot/ScottPlot/blob/42987b60b6fb90e94d0fb67a60c7dfafc73d8d9a/src/ScottPlot5/ScottPlot5/ScottPlot.csproj#L4

I can take a closer look when I review this PR later today or tomorrow if it keeps failing unexpectedly build in the cloud. 

Thanks for this contribution, I look forward to merging it in and making it available in the next release! 🚀

swharden:
> Not really sure why it's now trying to compile net462 on linux/macos when it wasn't before (or wasn't failing before)

Actually, don't worry about it!

I just realized the errors are in the color code (unrelated to this PR), so let's keep this PR clean and focused just on FillY, and I'll swoop in and fix the color issues causing the build failure right before I merge 👍 

Thanks again for this PR!

PR #2897, opened 2023-08-22
SP4 Heatmap: bugfix and parallel option

bukkideme:
Playing with a 2048x2048 array, i have got about 3X faster rendering with the Update + Refresh method calls in case of parallel=true option.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@
 * LinearRegression: Added `Statistics.Regression` (see cookbook) for fitting lines to collections of X/Y data points (#2901) _Thanks @anewton_
 
 ## ScottPlot 4.1.68 (in development)
+* Heatmap: Added a `UseParallel` option which can improve `Update()` performance for large datasets () _Thanks @bukkideme_
 
 ## ScottPlot 5.0.8-beta
 _Published on [NuGet](https://www.nuget.org/profiles/ScottPlot) on 2023-08-13_
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,7 +12,7 @@
 * LinearRegression: Added `Statistics.Regression` (see cookbook) for fitting lines to collections of X/Y data points (#2901) _Thanks @anewton_
 
 ## ScottPlot 4.1.68 (in development)
-* Heatmap: Added a `UseParallel` option which can improve `Update()` performance for large datasets () _Thanks @bukkideme_
+* Heatmap: Added a `UseParallel` option which can improve `Update()` performance for large datasets (#2897) _Thanks @bukkideme_
 
 ## ScottPlot 5.0.8-beta
 _Published on [NuGet](https://www.nuget.org/profiles/ScottPlot) on 2023-08-13_
--- a/src/ScottPlot4/ScottPlot.Cookbook/Recipes/Plottable/Heatmap.cs
+++ b/src/ScottPlot4/ScottPlot.Cookbook/Recipes/Plottable/Heatmap.cs
@@ -546,4 +546,22 @@ namespace ScottPlot.Cookbook.Recipes.Plottable
             hist2d.AddRange(flowData);
         }
     }
+
+    public class HeatmapParallel : IRecipe
+    {
+        public ICategory Category => new Categories.PlotTypes.Heatmap();
+        public string ID => "heatmap_parallel";
+        public string Title => "Parallel Processing";
+        public string Description =>
+            "Heatmaps have opt-in parallel processing which may improve performance when calling Update() for large datasets.";
+
+        public void ExecuteRecipe(Plot plt)
+        {
+            double[,] data = Generate.Sin2D(width: 1_000, height: 1_000);
+            var hm = plt.AddHeatmap(data, lockScales: false);
+
+            // opt into parallel processing
+            hm.UseParallel = true;
+        }
+    }
 }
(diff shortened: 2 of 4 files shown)


StendProg:
It is very surprising that parallelism gives such a gain. In this PR, parallel loops contain only memory copying without complex calculations.

bukkideme:
> It is very surprising that parallelism gives such a gain. In this PR, parallel loops contain only memory copying without complex calculations.

still, i see a 3X speed gain, so in my opinion it is worth to use it, no?

StendProg:
Well if it gives a gain then why not use it, especially since you made it customizable and can always be turned off.
It's just very surprising, from my experience such parallelization usually does not yield anything. You need to load the loop body properly to cover the overhead of parallelism.
There is either something overlooked here, or such a specific task, or modern multi-core systems change the rules of the game.
I will test it on my machine when I have time.

bukkideme:
> Well if it gives a gain then why not use it, especially since you made it customizable and can always be turned off. It's just very surprising, from my experience such parallelization usually does not yield anything. You need to load the loop body properly to cover the overhead of parallelism. There is either something overlooked here, or such a specific task, or modern multi-core systems change the rules of the game. I will test it on my machine when I have time.

Parallel.For or Parallel.Foreach can be extremely powerful. In one of my application, I need to load and parse data from ~10 thousands of csv files, process data, etc. So it includes HDD operations (IO) and also CPU intensive operations. Using a Parallel.For with a ConcurrentBag yields about 100X speed up! And I am talking about conventional magnetic HDD...
Actually there was also a trap: not configuring the MaxDegreeOfParallelism to a sane number (like 8 in my intel i7 CPU PC), could lead to a kind of "thread exhaustion": the whole process slowed down, then finally hanged forever. But this might be to do something with my programming skills, not necessarily the issue with the Parallel.Foreach :D

StendProg:
I found the problem.

First of all the initial tests showed 3x the gain as in your case.
The problem lies in this Casts:
https://github.com/ScottPlot/ScottPlot/blob/42987b60b6fb90e94d0fb67a60c7dfafc73d8d9a/src/ScottPlot4/ScottPlot/Plottable/Heatmap.cs#L241
https://github.com/ScottPlot/ScottPlot/blob/42987b60b6fb90e94d0fb67a60c7dfafc73d8d9a/src/ScottPlot4/ScottPlot/Plottable/Heatmap.cs#L243


If you rewrite it to a normal double loop without any parallelism, it will execute about 10 times faster.

As a result, after the changes I have:
~ 400 ms to update Heatmap 3000x3000 with parallel=true
~500-550 ms to update Heatmap 3000x3000 with parallel=false

Which is very similar to the expected result. About 30% gain due to frying 24 threads.

bukkideme:
Hmm, i do not get it. You wrote you changed the Cast to a normal double FOR loop, and got it 10X faster. Then between the normal FOR loop and the parallel FOR, you got about 30% speed increase.

But then how it was possible to get 3X speed up changing the Cast to Parallel.For at the first test comparison?

bukkideme:
ok, i tested also myself:
normal double FOR loop: ~600 msec
parallel FOR: ~450 msec
original Cast version: ~1250 msec

So i see a 2X speed increase between the original Cast version and the normal double For loop version.

StendProg:
I apologize for being misleading. 10x referred to the time specifically for one line of code and its replacement. All other numbers refer to the high-level test and execution time of the whole Update() method.

bukkideme:
> I apologize for being misleading. 10x referred to the time specifically for one line of code and its replacement. All other numbers refer to the high-level test and execution time of the whole Update() method.

No need to apologize, and thanks for the tests! :) Also it was important to point out, that the Cast is not optimal. So i change the Casts everywhere in the class for normal double for loops, and keep the parallel.For loops for the parallel=true cases.

swharden:
Hi @bukkideme and @StendProg, thanks for this PR and useful discussion!

This one took me longer than typical to get to because I've been busy with some other projects and also some teaching obligations for my university job 😅 I'm back now though and happy to review this PR

I refactored the `Update()` methods to simplify them a bit and isolate the parallel logic in their own methods. I also tried to keep the public API simple (and unchanged) by not passing `parallel` arguments, but instead having a public `UseParallel` property (similar to `SignalPlot`) that users can opt into using. I added a cookbook recipe demonstrating this functionality, but since it only optimizes `Update()` and not `Render()` it's not easy to evaluate the difference from the cookbook example alone.

@bukkideme when you have time can you look over my changes and be sure they're consistent with your original goals and also meet your needs? If I don't hear anything for a while I'll merge this and we can refine it later, but I know you use heatmaps a lot so I'm really interested in your feedback.

Thanks!

bukkideme:
Hi @swharden ! Glad to hear from you! :) I look through your changes, and give a prompt reply! Thanks!

bukkideme:
Ok, everything looks good, i like how to brought the parallel option into the game!
However, I do not find my another extra change, which is an overload of the AddHeatmap method. Maybe my commit did not get updated?
If we want to pre- specify the min/max scale for the color heatmap calculation, we would avoid an extra Update method call in this way. Makes a huge difference when we have to deal with large matrices...

public Heatmap AddHeatmap(double?[,] intensities, Drawing.Colormap colormap = null, bool? lockScales = true,
                          double? min = null, double? max = null, double?[,] opacity = null, bool parallel = false)

(I am not sure if we need the parallel option here, if we do not use it, then we do not have optimization in the very first step, right?)

swharden:
Hi @bukkideme, thanks for your feedback! It took me longer than typical to get to this one (reasons discussed in the Discord #1966)

> I do not find my another extra change, which is an overload of the AddHeatmap method

> If we do not use it, then we do not have optimization in the very first step, right?

You correctly point out that `Plot.AddHeatmap()` calculates the scale using the old method.

I decided not to add additional overloads in this PR to keep the API simple. Special overloads that customize behavior with prespecified scales and options for parallelism require a lot of arguments and I'm hesitant to put that in the Plot API (and keep adding new overloads as new features are conceived).

Instead I recommend users that desire these advanced features `new` their own heatmap, setup those options as desired, then `Add()` it to the plot. Cookbook examples can demonstrate how to do this with advanced features. 

There's not one solution that suits everyone, but I think this strategy makes the library easiest to maintain and extend with minimal disruption. Let me know if you disagree! Otherwise I'm happy to merge this in and I can make a release today 👍

PR #2920, opened 2023-09-19
Fix AvaPlot rendering issues

nightfog-git:
For AvaPlot to be displayed correctly alongside other controls, ClipToBounds should be true.
Correct rendering size is Bounds.Size and correct rendering position is always origin (0,0).

### DIFF
--- a/src/ScottPlot5/ScottPlot5 Controls/ScottPlot.Avalonia/AvaPlot.cs
+++ b/src/ScottPlot5/ScottPlot5 Controls/ScottPlot.Avalonia/AvaPlot.cs
@@ -37,13 +37,9 @@ public class AvaPlot : Controls.Control, IPlotControl
         new("All Files") { Patterns = new List<string> { "*" } },
     };
 
-    static AvaPlot()
-    {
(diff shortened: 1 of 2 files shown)


src/ScottPlot5/ScottPlot5 Controls/ScottPlot.Avalonia/AvaPlot.cs
```diff
@@ -123,13 +128,15 @@ public void Render(ImmediateDrawingContext context)
             var surface = lease.SkSurface;
             if (surface is null) return;
```
oktrue, line 129: Unnecessary variable declaration and null checking.

oktrue (approved):

swharden:
Thanks for this improvement @nightfog-git! Also thanks @oktrue for your feedback along the way. I'm merging this now and will publish a release in the next few minutes 🚀

swharden:
@nightfog-git and @oktrue you're more familiar with this topic than me so I seek your recommendation: should this change also be made to ScottPlot 4's Avalonia control? #2926

Issue #2933, opened 2023-10-07
SP5: signal plots with repeating values


swharden:
when a signal plot flips into high density mode, a line is not rendered if its min and max values are equal

![density](https://github.com/ScottPlot/ScottPlot/assets/4165489/e560d463-809b-4579-b57d-8f4460b78c9d)

swharden:
resolved in #2935

PR #2935, opened 2023-10-10
SP5 Signal: fix horizontal line rendering issue

StendProg:
**Purpose:**
Fix for #2933.

The problem was this line:
https://github.com/ScottPlot/ScottPlot/blob/5d377025c0fa6ce63eeca4114b17aeeb8bded629/src/ScottPlot5/ScottPlot5/Plottables/Signal.cs#L142

In the case of identical `y`'s in the entire pixel column, we draw a line from (x,y) to the same (x, y). The graphics library probably tries to optimize this and draws absolutely nothing, thus we lose the line thickness.
In this PR the drawing of unconnected vertical lines is changed to drawing a connected path along all vertical columns, which eliminates the effect described in #2933.

In the process of working on PR it turned out that this bug is always present and was even useful for drawing areas outside the signal. In fact, the same single point lines are always drawn there that were not displayed. Switching to path drawing started to display those invisible lines as well.
As a result, it had to be significantly redesigned, in particular to filter out pixel columns that lie outside the signal.


swharden:
Hi @StendProg, thanks for this fix!

I identified a rendering issue that caused downward sloping lines to appear too thick. I figured out it was because `LineTo()` was drawing extra pixels as it jumped from the top of one vertical bar to the bottom of the next. I think I fixed it in a reasonable way, but I always welcome your input! 

I'll merge this in now and we can create a new PR to improve or refine this code if needed 👍 

### Before
image

### After
image

StendProg:
Hi @swharden,

Yes, it does look better with your extra fix. But we can find a counterexample for it too. For example, when neighboring pixel columns do not overlap by Y, then they will be unconnected.

If we continue these improvements, we can do what SignalXY draws for a single column:

1. Line (Column[i-1].Right, Column[i].Left)
2. Line (Column[i].Top, Column[i].Bottom)
3. Line (Column[i].Right, Column[i+1].Left)
SignalConnectionExample

The blue color indicates the original signal.
Rectangles mark the pixel column.
Green color indicates the lines to be drawn.

swharden:
Thanks for the feedback @StendProg!

> when neighboring pixel columns do not overlap by Y, then they will be unconnected

I'm having trouble demonstrating this with code... do you have a similar code example I can use to check that the issue is present and confirm when it has been fixed? 

Here's my attempt so far:

```cs
[Test]
public void Test_Signal_VerticalGap()
{
    double[] data = Generate.Sin(10_000, oscillations: 100);
    for (int i = (int)(data.Length * .3); i < (int)(data.Length * .7); i++)
    {
        data[i] = data[i] + 10;
    }

    ScottPlot.Plot plt = new();
    var sig = plt.Add.Signal(data);
    sig.LineStyle.AntiAlias = false;
    plt.Grids.Clear();
    plt.SaveTestImage();
}
```

image

swharden:
In theory the artifact you describe with the green squares makes sense, but in practice it's pretty hard to demonstrate

```cs
using ScottPlot;
using ScottPlot.WinForms;

double[] data = Generate.Sin(10_000, oscillations: 100);
for (int i = (int)(data.Length * .3); i < (int)(data.Length * .7); i++)
    data[i] = data[i] + 10;

Plot plot = new();
plot.Add.Signal(data);

Form form = new()
{
    Width = 600,
    Height = 400,
    StartPosition = FormStartPosition.CenterScreen
};

FormsPlot formsPlot1 = new()
{
    Dock = DockStyle.Fill
};

formsPlot1.Reset(plot);
formsPlot1.Refresh();
form.Controls.Add(formsPlot1);
form.ShowDialog();
```

step

StendProg:
I was able to get this on the latest version:
SignalExample

I think the code will need a lot of reworking, but you're right that it should be a separate task.

You failed to reproduce the effect I described earlier because the X ranges for the pixel columns overlap (which should not be the case) and the data points can be included in both columns.
The overlap is most likely due to rounding in these rows:
https://github.com/ScottPlot/ScottPlot/blob/c61bb771bd79c36a43cc056bf62f9a22781307a9/src/ScottPlot5/ScottPlot5/Plottables/Signal.cs#L43-L45
When iterating, we rounded the width, but when calculating the density of points per pixel, we didn't:

PR #3005, opened 2023-11-24
SP4: Support LineStyle in DataLogger and DataStreamer

Gray-lab:
**Purpose:**
* Adds test for DataLogger LineStyle
* Fixes bug 2972 - SP4: DataLogger should support `LineStyle`

**Notes:**
* Used `IsDifferentThan` in the test assert to detect change in `LineStyle`, but it doesn't feel like the most precise way of testing this - open to suggestions here


swharden:
Thanks @Gray-lab, this looks great! 🚀

swharden:
Oh yeah, bonus points for the test that demonstrate this fix too! 🚀 ✨

PR #3006, opened 2023-11-25
SP5: Improve multi-control layout and axis matching

### LINKED ISSUE #2941: SP5: `IPanel` should have option for fixed size
Currently axes are `IPanel` objects which `Measure()` themselves so the layout manager can pack the figure tightly before each frame is rendered

For multi-control layout matching it is required that panels can be assigned a fixed dimension (e.g., to match a similar panel from another control)

This issue tracks implementing a strategy for manually assigning a size to panels. It could be achieved by adding:
```cs
public float? FixedSize { get; set; }
```

near:
https://github.com/ScottPlot/ScottPlot/blob/5d377025c0fa6ce63eeca4114b17aeeb8bded629/src/ScottPlot5/ScottPlot5/Interfaces/IPanel.cs#L18

Then updating the layout manager to respect the override:

https://github.com/ScottPlot/ScottPlot/blob/5d377025c0fa6ce63eeca4114b17aeeb8bded629/src/ScottPlot5/ScottPlot5/LayoutEngines/Automatic.cs#L19-L22

Also make sure the matched layout engine respects it too
https://github.com/ScottPlot/ScottPlot/blob/5d377025c0fa6ce63eeca4114b17aeeb8bded629/src/ScottPlot5/ScottPlot5/LayoutEngines/Matched.cs#L25-L28

swharden:
This PR works to make it easy to match layouts and axes across controls.

Originally this is difficult because of the lack of suitable events to update one control's layout from another.

resolves #2941


swharden:
The core issue of this PR has been resolved, but OpenGL rendering issues prevent the demo from working will. These will be resolved in a future PR and are currently tracked by #3007

Issue #3007, opened 2023-11-25
SP5: OpenGL rendering fail when multiple controls are in the same window


swharden:
Extreme artifacts and hard crashes appear if both plots render really close to each other in time. I think it's related to the frame buffer switching. I found https://github.com/opentk/GLWpfControl/issues/58 which indicates the problem may be multiple OpenGL context objects, but a solution could be to force multiple controls to use the same context.

This issue will track adding a way to enable/disable OpenGL rendering functionality. At first I'll probably create two controls for this, but in the future something smarter may be a solution. I'm experimenting in WinForms but presumably all controls have this issue when OpenGL is employed.

## Standard
![nongl](https://github.com/ScottPlot/ScottPlot/assets/4165489/0208ed5b-2059-41b4-9016-1b22d511a1a0)

## OpenGL
![gl](https://github.com/ScottPlot/ScottPlot/assets/4165489/abdc2223-9fac-468f-8fd0-11e0129dfb9d)

swharden:
To keep ScottPlot 5 development unblocked I'm resolving this by giving users the option to opt-in to the OpenGL control. I'll go full speed ahead developing using the non-accelerated controls (#3008) and this issue can be fixed after ScottPlot 5 is released.

Issue #2950, opened 2023-10-17
SP5 WpfPlot: Support Windows ARM64


swharden:
**Does anyone have experience running ScottPlot 5 or using SkiaSharp in Windows on an ARM64 platform?** 

I received an email indicating that a WPF app with a ScottPlot 5 control crashes on opening. I'll share additional debug information here if it becomes available. I see SkiaSharp builds a "win-arm64" runtime `libSkiaSharp.dll` so it seems like it should be supported out of the box.

It's possible that the issue is related to the custom `SKGLElement` we created to support OpenGL in the WPF control:
https://github.com/ScottPlot/ScottPlot/blob/4d772dae2644d575cd3a69a13a7a5b341667a9fe/src/ScottPlot5/ScottPlot5%20Controls/ScottPlot.WPF/SKGLElement.cs#L1-L266

Crown0815:
I was able to pull this stack trace from my ScottPlot demo application running on Windows 11 ARM (M2) VM through Parallels.

```
Application: ChartingDemo.exe
CoreCLR Version: 7.0.923.32018
.NET Version: 7.0.9
Description: The process was terminated due to an unhandled exception.
Exception Info: System.AccessViolationException: Attempted to read or write protected memory. This is often an indication that other memory is corrupt.
Stack:
   at OpenTK.Graphics.Wgl.Wgl.DXOpenDeviceNV(IntPtr)
   at OpenTK.Wpf.DxGlContext..ctor(OpenTK.Wpf.GLWpfControlSettings)
   at OpenTK.Wpf.GLWpfControlRenderer..ctor(OpenTK.Wpf.GLWpfControlSettings)
   at OpenTK.Wpf.GLWpfControl.Start(OpenTK.Wpf.GLWpfControlSettings)
   at SkiaSharp.Views.WPF.SKGLElement.Initialize()
   at SkiaSharp.Views.WPF.SKGLElement..ctor()
   at System.RuntimeType.CreateInstanceDefaultCtor(Boolean, Boolean)
   at System.RuntimeType.CreateInstanceImpl(System.Reflection.BindingFlags, System.Reflection.Binder, System.Object[], System.Globalization.CultureInfo)
   at System.Activator.CreateInstance(System.Type, System.Object[])
   at System.Xaml.Schema.SafeReflectionInvoker.CreateInstanceCritical(System.Type, System.Object[])
[...]

PR #3008, opened 2023-11-25
SP5: Create OpenGL and Non-OpenGL controls

swharden:
OpenGL rendering can cause issues when displaying multiple controls (#3007) and problems for rendering on newer hardware platforms (#2950)

This PR creates separate `FormsPlot`, `FormsPlotGL`, `WpfPlot`, and `WpfPlotGL` controls

One advantage is that the `WpfPlot` can use the _official_ `SKElement` and not the unofficial `SKGLElement` we've been carrying in ScottPlot's code base (#2395)  because it doesn't seem to be in the official package yet https://github.com/mono/SkiaSharp/pull/2317

I don't think this is an ideal long term solution. A better solution would be an easier way for users to opt-in to OpenGL acceleration using the same control. This was feasible for the WinForms control, but I had trouble getting this to work in the WPF control because of issues related to templating (#2565)


swharden:
Following-up, wow, the OpenGL control doesn't do much to improve performance on my system

image

PR #3013, opened 2023-11-26
SP4 Population: add option to define box transparency

Gray-lab:
**Purpose:**
* Resolve Issue 2967 - Add option to set custom alpha value for scatter markers and boxes in population plots
* Add xml docs for new public fields `MarkerAlpha` and `BoxAlphaOverride` to explain behavior
* Add tests for these two fields


Gray-lab:
@swharden Thank you for cleaning up the XML docs and making the legend respect the box alpha - I completely missed that when I was looking at test plots.

PR #3022, opened 2023-11-28
SP4: Improves Generate class.

### LINKED ISSUE #2893: SP4: Improve `Generate` seeding
Spillover from #2861, calling `RandomWalk()` multiple times in rapid succession does not properly re-seed. Fix this internally using a private static instance of the random data generator.

```cs
ScottPlot.Plot plt = new();
for (int k = 0; k < 50; k++)
{
    double[] data = ScottPlot.Generate.RandomWalk(10000);
    plt.AddSignalConst(data);
}
```

KroMignon:
Use a global Random number generator when `seed` is null to enforce values randomness. Should fix #2893

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@
 * Controls: Added `Interaction.Disable()` and `Interaction.Enable()` methods for easy control of mouse interactivity
 * Render: Improve axis frame and tick mark rendering for SVG export (#2944) _Thanks @Crown0815_
 * Controls: Created OpenGL controls `FormsPlotGL` and `WpfPlotGL` distinct from `FormsPlot` and `WpfPlot` (#3008, #3007, #2950, #2395, #2565)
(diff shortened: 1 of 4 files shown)


swharden:
Thanks for this improvement @KroMignon! This topic made me double-check what's going on in ScottPlot5's `RandomDataGenerator` class. It's using a `System.Security.Cryptography.RandomNumberGenerator` to generate the seed for `Random()`. I suspect this is thread safe and truly random, but if you have more familiar with this topic I'd appreciate any feedback you may have on this class

KroMignon:
Thanks for merging my PR 😄 

AFAIK, the only thread-safe random number generator in .NET is [`System.Random.Shared`](https://learn.microsoft.com/en-us/dotnet/api/system.random.shared?view=net-6.0), which is available since .NET 6.
For all other random number generators, you should use `ThreadLocal` if you are using a static instance of them.

Issue #2969, opened 2023-10-26
SP4 ScatterPlotDraggable: Bug after `Update()`


SasKayDE:
## Bug Report

**Issue:**
After updating the X and Y values of a ScatterPlotDraggable object programmatically by calling the Update(Xnew, Ynew) and Refresh(), the plot is visualized correctly, considering the updated Xnew and Ynew values, but the draggable point coordinates are not updated and remain at the previous positions Xold, Yold, indicated by the DragCursor still appearing at the positions Xold, Yold.

**Reproducing:**

Correct/expected behavior after initialization:

```cs
// Initialization with initial dataX, dataY
myScatterPlotDraggable = new ScatterPlotDraggable( dataX, dataY ) { DragEnabled = true, DragCursor = ScottPlot.Cursor.Hand };
myFormsPlot.Plot.Add( myScatterPlotDraggable );
myFormsPlot.Refresh();
```

Change of the X and Y data programmatically results in the described/unexpected behavior:

```cs
// change of X and Y values (keeping the array sizes)
myScatterPlotDraggable.Update( Xnew, Ynew );
myFormsPlot.Refresh();
```

While debugging it has been observed that the changed values (in this example Ys[1]) are not updated consistently in the attributes "Ys (ScottPlot.Plottable.ScatterPlot)" and "Ys" of the myScatterPlotDraggable object after calling myScatterPlotDraggable.Update( Xnew, Ynew );

![image](https://github.com/ScottPlot/ScottPlot/assets/16592829/e268cb59-c211-4723-b2ed-2c993f174424)

The result is a plotted point at 1800 (Ynew[1]) and a draggable point at 1024 (Yold[1]).

### System Details
* **ScottPlot Version:** 4.1.68
* **Operating System:** Windows 10
* **Application Type:** WinForms
* **.NET Version:** .NET 6.0

SasKayDE:
Same root cause, different observation:
When increasing the array size of Xnew and Ynew and calling myScatterPlotDraggable.Update() an exception is thrown when the ScatterPlotDraggable tries to refresh/render the plot because only Xs and Ys of the base class is updated, but not Xs andYs of the ScatterPlotDraggable class. The following snapshot shows the myScatterPlotDraggable object after calling myScatterPlotDraggable.Update() with the Xnew and Ynew array increased from 5 to 6 elements:

[...]

Issue #2870, opened 2023-08-13
SP4: ScatterPlotDraggable `IsUnderMouse()` crashes after calling `ScatterPlot.Update()`


onur-akaydin:
## Bug Report

**Issue:** ScatterPlotDraggable.IsUnderMouse method crashes after calling ScatterPlotDraggable.Update method

**Reproducing:** 

Calling ScatterPlotDraggable.Update method with a new pair of Xs and Ys arrays, "larger" than the existing ones, then waiting for any event that is triggering ScatterPlotDraggable.IsUnderMouse method (i.e. a mouse down event) will eventually crash. Because in ScatterPlotDraggable, Xs and Ys are defined as NEW double arrays, IsUnderMouse references to them, however ScatterPlot.Update() method updates the other Xs and Ys arrays in the base class.

```cs
myScatterPlotDraggable.Update(newXsWithLargerSize, newYsWithLargerSize);
// make a mouse interaction
```

I found a solution by removing Xs and Ys properties from ScatterPlotDraggable, and changing the relevant set accessors in ScatterPlot class to "protected set"

In ScatterPlotDraggable.cs:
```cs
//public new double[] Xs { get; private set; }
//public new double[] Ys { get; private set; }
```

In ScatterPlot.cs:
```cs
public double[] Xs { get; protected set; }
public double[] Ys { get; protected set; }
```

But I'm not sure if there is a side effect to doing it like this.

### System Details
* **ScottPlot Version:** 4.1.65.0
* **Operating System:** Windows 11
* **Application Type:** WPF
* **.NET Version:** .NET 7.0

swharden:
Hi @onur-akaydin, thank you for reporting this!

Considering how long this issue has been open and how I'm focusing the next several weeks on ScottPlot 5 (#2984) it's unlikely I'm going to be able to look into this soon, so I'll close this issue but I welcome contributions if others have a recommendation.

Note that the method that crashes is here:
https://github.com/ScottPlot/ScottPlot/blob/c7ce83274728f35977dc7d4c55aadc56ed8ee997/src/ScottPlot4/ScottPlot/Plottable/ScatterPlotDraggable.cs#L101-L114

...so an an alternative fix that is less likely to produce side effects would be to wrap that in some logic that catches index exceptions and returns `false` if they occur. If someone opens a PR with this fix I'm happy to review it!

swharden:
Following-up, I see this issue may also be causing #2969

[...]

PR #3025, opened 2023-11-28
SP4: Fix ScatterPlotDraggable

KroMignon:
**Purpose:**

This should resolve issues with `ScatteerPlotDraggable` (#2870 and #2969):
- remove duplicated members `Xs` and `Ys`, they are already managed by base class `ScatterPlot`
- ensure a valid point is selected before trying to update its position
- avoid possible exception with `PointCount` when no data is set


swharden:
Fantastic fix @KroMignon! I _really_ appreciate all your help on these recent issues 😄🙏🚀

PR #3052, opened 2023-12-04
SP5: Avoid code duplication for WPF controls

KroMignon:
**Purpose:**

Add base virtual custom control class to hold common code between ``WpfPlot` and `WpfPlotGL` to avoid code duplication.

Also update `SandBox` application to show `WpfPlot` and `WpfPlotGL`.


KroMignon:
Happy I could help you 😃

PR #3062, opened 2023-12-05
SP5: render Legend into Image

### LINKED ISSUE #2934: SP5: GetLegendBitmap()
Can Legend support being placed outside the chart

KroMignon:
This is a first step to allow rendering legend alone as `Image` , which can be converted into many type of picture formats (BMP, JPEG, etc.).

This should fix  #2934.

- fix initialisation bug in `PixelRects`
- add `GetLegendImage()` into `Plot`

@swharden tell me what you are thinking about this.


swharden:
Hi @KroMignon, thank you for working on #2934! At first glance this looks fantastic 

I see this is still "WIP" so let me know when you're ready to merge and I'll review it a little closer and merge it in good time

My only suggestion is to add XML docs to `GetImage()` to clarify what it does. Perhaps "Return the content of the legend as a standalone image" or something like that

KroMignon:
Hello @swharden, sorry for the long delay since last update, I was sick with COVID 😷.

I made some changes to fix shadow rendering, but I am not sure if shadow is required in this case?
I also add optional parameters to limit height and/or width of the legend.

I would also propose to change the `Image` class to allow access to `SKImage` member to show it on screen.
Or perhaps create a shadow copy of it is required, I am not really aware about the good practices with Skia.

KroMignon:
Hello @swharden, what do you think about this implementation?
Converting Legend into `SKImage` has for me the advantage that the user could convert it in any kind of picture format with help of the `ScottPlot.Image` class or display it on screen quickly with help of package [`SkiaImageView`](https://github.com/kekyo/SkiaImageView)

bclehmann:
I don't think it's ideal that `Render` and `GetImage` are near duplicates of each other.

Instead, I think both `GetImage` and `Render` should be implemented in terms of a private function which just draws the legend onto a given canvas at a given point. Then `Render` simply calls this on the canvas it's given once it has calculated where it should be placed based on its alignment, padding, etc. And `GetImage` simply creates a canvas, calls this private function, and returns the resulting image. Both will require sizing calculations, which probably also should be refactored into a private function.

As for `SKImage` vs `ScottPlot.Image`, I'm a little on the fence. If the user wants to modify the image or composite it onto their plottables, I don't think `ScottPlot.Image` is fit for purpose. But if they want to save the image to disk, I don't believe `SKImage` is fit for purpose. It _is_ tempting to simply give them an `SKImage` and document that `ScottPlot.Image` provides facilities to save that to disk, but I'm unsure if we want to tie our API to Skia so strongly.

KroMignon:
@bclehmann many thanks for your feedback.

> Instead, I think both `GetImage` and `Render` should be implemented in terms of a private function which just draws the legend onto a given canvas at a given point. Then `Render` simply calls this on the canvas it's given once it has calculated where it should be placed based on its alignment, padding, etc. And `GetImage` simply creates a canvas, calls this private function, and returns the resulting image. Both will require sizing calculations, which probably also should be refactored into a private function.

I generally try to avoid code duplication, but sometimes it is more effective. I will check your suggestion more deeply, perhaps I find a way which is not "overengineered" / overcomplicated.
 
>It _is_ tempting to simply give them an `SKImage` and document that `ScottPlot.Image` provides facilities to save that to disk, but I'm unsure if we want to tie our API to Skia so strongly.

I do not understand the problem here, AFAIK `ScottPlot 5` backend is Skia and not multi-backend targeted, or did I misunderstood something?

KroMignon:
@swharden this PR comes with the following changes!
- legend can be rendered into `SKImage`
- legend image size can be constrained
- fix legend item selection to not add hidden plotables
- fix initialization bug in `PixelRects`

src/ScottPlot5/ScottPlot5/Legends/Legend.cs
```diff
@@ -25,48 +28,74 @@ public void Render(RenderPack rp)
...
+    }
+    public void AsSvg(ScottPlot.Plot plot, Stream svgStream, int maxWidth = 0, int maxHeight = 0)
+    {
+        if (svgStream is null)
+            throw new NullReferenceException($"invalid Stream");
 
-        // render all items inside the legend
-        float yOffset = legendRect.Top + Padding.Top;
-        for (int i = 0; i < sizedItems.Length; i++)
+        var canvas = RenderToObject(plot, svgStream: svgStream, maxWidth: maxWidth, maxHeight: maxHeight) as SKCanvas;
```
bclehmann, line 54: I'm really not a fan of `RenderToObject` being capable of returning two very different objects. I think you were on the right track with `RenderLegend`, which can take any canvas (whether an `SKSvgCanvas` or not).

Currently `RenderToObject` gives you either an `SKSurface` or an `SKCanvas` because it can do two different things. The code which is actually shared is the sizing code (measuring how big the legend should be) and the call to `RenderLegend`. The rest is creating either a surface or an SVG canvas, which isn't actually shared between the two usecases. So if we stop trying to share it (instead making the callers deal with that mess) we get a simpler implementation which is easier to reason about and to test.

What I would do is create a function called something like `MeasureLegend`, which returns `legendRect`, `maxWidth`, `maxHeight`, etc. Then the consuming functions do this:

- `Render` calls `MeasureLegend` and then calls `RenderLegend` with `rp.Canvas`
- `AsSvg` calls `MeasureLegend` and uses that to create an `SKSvgCanvas`. It passes that to `RenderLegend`
- `GetImage` calls `MeasureLegend` and uses that to create an `SKSurface`. It passes that to `RenderLegend`

bclehmann: All of this is assuming we need SVG support for the legend, which I hadn't thought was necessary. If we don't then you can probably move the surviving code from `RenderToObject` back into `GetImage` and there'd be no problem.

Although I would still like the measurement code to be factored out. Otherwise it tends to get copy-pasted around and it can introduce lots of subtle bugs if one isn't very disciplined. If it's just a function that everyone calls rather than copy-pastes then that discipline is unnecessary.

bclehmann (changes requested):

src/ScottPlot5/ScottPlot5/Legends/Legend.cs
swharden: Thanks for this feedback @bclehmann! I'm going to mark this as resolved so I can merge now, but there's a lot to be desired about performance and design of the Legend class. Luckily the public API is simple and raster and vector exporting is supported now, and we can loop back around and refine the internals as time allows. I'll share some details in a comment at the bottom of this PR

swharden:
Hi @KroMignon, thanks for your extensive work on this! Thanks @bclehmann too for your input along the way.

I reworked the code a bit and now this is possible:

```cs
Plot plt = new();
var sig1 = plt.Add.Signal(Generate.Sin());
sig1.Label = "Sine";
plt.Legend.IsVisible = true;

Image img = plt.GetLegendImage();
img.Save("demo.png");

SvgImage svg = plt.GetLegendSvg();
svg.Save("demo.svg");
```

I created a `SvgImage` class that creates and exposes a SVG-backed `Canvas` that can get passed into the `Legend` for rendering, then it can get saved by the user at any time. Although the API original to this PR had users pass a filename into `SaveLegendAsSvg()`, I think this new design is a little more consistent internally. It also prevents exposing SkiaSharp types to the user.

The basic functionality works now so I'm going to merge this, but there is definitely room for improvement. SkImage does not implement `IDisposable` right now. Also code throughout the `Legend` class could be refactored to reduce duplicate processes like item measurement, paint creation, etc. It's a pretty difficult class to work in because it's so complicated, so there's a lot of room for improvement. 

I don't want refining the legend to block development of the rest of ScottPlot 5 so I think this is fine to merge for now, but I welcome refinements along the way and will probably perform a deep refactor of the Legend class at some point.

Thanks again, I'm glad to have this in ScottPlot 5! 🚀

KroMignon:
@swharden @bclehmann 

I was not 100% happy with actual code, but had no time to reorganize/rewrite it.
@bclehmann gives interesting ideas, perhaps I can find time during the next days.

Thanks for merging it into code base.

Issue #3055, opened 2023-12-05
SP5: Improve descriptive statistics class


swharden:
First, we can use `Average()` here:
https://github.com/ScottPlot/ScottPlot/blob/443a60668e415b5b4ca975c5fb023cbe9e5b3ff6/src/ScottPlot5/ScottPlot5/Statistics/Descriptive.cs#L12

Next, all those `double[]` arguments should be changed to `IEnumerable<double>` if possible
https://github.com/ScottPlot/ScottPlot/blob/443a60668e415b5b4ca975c5fb023cbe9e5b3ff6/src/ScottPlot5/ScottPlot5/Statistics/Descriptive.cs#L10
https://github.com/ScottPlot/ScottPlot/blob/443a60668e415b5b4ca975c5fb023cbe9e5b3ff6/src/ScottPlot5/ScottPlot5/Statistics/Descriptive.cs#L15C24-L15C24
https://github.com/ScottPlot/ScottPlot/blob/443a60668e415b5b4ca975c5fb023cbe9e5b3ff6/src/ScottPlot5/ScottPlot5/Statistics/Descriptive.cs#L20

Third, there should be a `StdErr()` that returns standard deviation divided by the square root of the sample count https://en.wikipedia.org/wiki/Standard_error

PR #3071, opened 2023-12-07
Improve descriptive statistics class

arthurits:
**Purpose:**
* Implements #3055.
* Input parameters are defined as `IEnumerable<T>` instead of `double`.
* Uses `IEnumerable<T>.Average()` to compute the data average.
* Functions `StDev` are renamed to `StdDev`.
* Code for `StdDev` is modified:
  * Booleand parameter `asSample` is added to compute both the population and the sample standard deviations.
  * Code checks for and avoids division by 0 runtime exceptions.
  * Function is overloaded with respect to parameter `mean`.
* Function `StdErr` (standard error of the mean) is added.


swharden:
Hi @arthurits, thanks so much for this PR! This is fantastic, and I look forward to merging it in a few minutes!

> Functions StDev are renamed to StdDev.

Note that I'll keep it `StDev` to be consistent with naming I typically find in Python data analysis packages. I try to mimic the API of those packages (especially matplotlib) where possible
* https://docs.python.org/3/library/statistics.html#statistics.stdev
* https://numpy.org/doc/stable/reference/generated/numpy.std.html

swharden:
Thinking out loud about performance...

I'm tempted to make this an `IReadOnlyList` which will allow it to accept both `List<double>` and `double[]` and I think offer improved performance because `.Count` can be used instead of Linq's `.Count()` ... Probably the decision should be made with proper benchmarks to prove it's meaningfully faster, but I'm not really motivated to micro-optimize these methods at this time.

Similarly, I'll probably make the overloads private that accept the `mean` "for performance". Maybe it saves a few cycles by not calculating mean twice, but calculating mean isn't very costly considering that function is loaded with `Sqrt()` calls which are a lot more significant

swharden:
One more thought, rather than passing a flag to switch between population or sample standard deviation, I'll probably make it two functions to mimic python
* https://docs.python.org/3/library/statistics.html#statistics.stdev
* https://docs.python.org/3/library/statistics.html#statistics.pstdev

swharden:
[sorry about the message bomb here lol] ... thinking _more_ out loud,

In researching Microsoft's naming of this stuff, it looks like they favor `StandardDeviation()` and `StandardDeviationP()` which is a few more keystrokes but may be more readable be a better reference than Python APIs

* https://learn.microsoft.com/en-us/dotnet/api/system.data.entity.core.objects.entityfunctions.standarddeviation?view=entity-framework-6.2.0
* https://learn.microsoft.com/en-us/dotnet/api/system.data.entity.core.objects.entityfunctions.standarddeviationp?view=entity-framework-6.2.0

swharden:
> I look forward to merging it in a few minutes!

A few minutes turned into a few hours - what a rabbit hole! I think this ended in a fantastic place though and I'm merging now. Thanks again for your help getting this started @arthurits! 

There's room for benchmark-driven micro-optimization in the future, but I'm happy where this is.

arthurits:
Hi @swharden. Although it looked quite straightforward, it turns out there were many points to address.
The final code looks great!

PR #3078, opened 2023-12-11
Render only Candlesticks and Bars in viewable area of chart

AndreyPalyutin:
I have added a little logic to FinancePlot and BarPlot to render only visible candles/OHLCs/bars.
In finance plots, vertical bars are frequently used together with candles/OHLCs, to demonstarte volumes. That's why i would appreciate if you will accept this change for BarPlot too.

Solution for issue #3053


swharden:
This looks fantastic @AndreyPalyutin, thanks so much! 🚀

Issue #3030, opened 2023-11-29
SP5 Axis: label is misaligned for right axes


swharden:
```cs
double[] data1 = ScottPlot.RandomDataGenerator.Generate.RandomWalk(count: 51, mult: 1);
var sig1 = formsPlot1.Plot.Add.Signal(data1);
sig1.Axes.YAxis = YAxis1;
sig1.Axes.YAxis.Label.Text = "YAxis1";
sig1.Axes.YAxis.Label.Font.Color = sig1.LineStyle.Color;

double[] data2 = ScottPlot.RandomDataGenerator.Generate.RandomWalk(count: 51, mult: 1000);
var sig2 = formsPlot1.Plot.Add.Signal(data2);
sig2.Axes.YAxis = YAxis2;
sig2.Axes.YAxis.Label.Text = "YAxis2";
sig2.Axes.YAxis.Label.Font.Color = sig2.LineStyle.Color;
```

![image](https://github.com/ScottPlot/ScottPlot/assets/4165489/8016913c-7809-4cb1-acf5-5450fc08d796)

PR #3096, opened 2023-12-15
SP5: Fix right aligned axis label position

albyoo:
**Purpose:**
* Fixes: #3030 

image


swharden:
Hi @albyoo, thanks so much for this fix! Merging now! 🚀

Issue #3089, opened 2023-12-14
Setting IsVisible=false on one plot, hides subsequent plots


raburton:
## Bug Report

**Issue:** Setting IsVisible=false on one plot, hides subsequent plots

**Reproducing:** Add a plot, set IsVisible=False on it, add more plots. Subsequent plots are also invisible (or not there at all?) and even manually setting IsVisible=true on them doesn't help.

```cs
private void Form1_Load(object sender, EventArgs e) {

    double[] data1 = new double[10000];
    double[] data2 = new double[10000];

    for (int i = 0; i < 10000; i++) data1[i] = 100;

    Signal s1 = formsPlot1.Plot.Add.Signal(data1);
    s1.LineStyle.Color = Colors.Red;
    s1.LineStyle.Width = 5;
    s1.IsVisible = false;

    Signal s2 = formsPlot1.Plot.Add.Signal(data2);
    s2.LineStyle.Color = Colors.Blue;
    s2.LineStyle.Width = 5;
    s2.IsVisible = true;

}
```

![image](https://github.com/ScottPlot/ScottPlot/assets/433543/efee8d79-0e5b-4bc7-8d4f-16137758aab6)

### System Details
* **ScottPlot Version:** 5.0.10-beta
* **Operating System:** Windows 10
* **Application Type:** WinForms
* **.NET Version:** .NET 8.0

swharden:
fixed in #3097

PR #3097, opened 2023-12-15
SP5: fix hidden potables rendering

KroMignon:
Fix plotables rendering when one plotable is hidden.

This should fix issue #3089.


swharden:
Great catch @KroMignon, thanks!

Issue #3063, opened 2023-12-05
SVG: legend shadow is not semitransparent


swharden:
extending #1293, #2704, #3004 it looks like legend shadows aren't perfect

## WPF
![image](https://github.com/ScottPlot/ScottPlot/assets/4165489/cf656e44-31a0-4d5f-a2d5-603877243574)

## SVG
![image](https://github.com/ScottPlot/ScottPlot/assets/4165489/2dc8eb51-5300-4cff-a243-3750215efa88)

swharden:
resolved in #3098

PR #3098, opened 2023-12-15
SVG: legend shadow is not semitransparent

KroMignon:
Dirty workaround to enable transparency on SVG with black color.

This should fix issue #3063.


swharden:
Great find @KroMignon! I suspect that would have taken me a really long time to figure out. I moved that logic from the constructor to the `WithAlpha()` method so it's not engaged quite as often.

I'm curious if this is an issue with SVG spec or with SkiaSharp. I haven't looked at the content of the SVGs, but if it's a SkiaSharp bug it may be worth reporting https://github.com/mono/SkiaSharp/issues

KroMignon:
Hello @swharden,

sorry for the late reply, I was quite busy in the last days!

I will try to report this the SkiaSharp with a short code example. I didn't find anything about this in the bug reports.
It is a very nasty bug, I already had such an issue years ago with Qt/QML Framework (opacity and dark color)

PR #3101, opened 2023-12-18
Add missing horizontal line marker shape to enum

SerhiiMahera:
**Purpose: Add missing horizontal line marker shape to enum**
* During development I find out that HorizontalBar.cs is available in MarkerShapes folder. But MarkerShape.horizontalBar was missing. I tried to solve my issue with MarkerShape.cross and it looks not so nice. So, would be great to have it here.
image

```cs
 plt.AddMarker(1, 210, MarkerShape.horizontalBar, color: Color.Black, size: 5);
```


swharden:
Good catch @SerhiiMahera, thanks for identifying the issue and creating this PR! I'm merging now and these changes will be in the next release 🚀

Issue #3110, opened 2023-12-22
SP5: Make Plot Interactions more extensible


albyoo:
## Feature Suggestion

Currently to override the plot interactions, the Interaction class must be overriden to handle custom functionality. A lot of the Interaction's members are private, which makes it more difficult to extend the base functionality.

**Feature description:** Add an interaction interface, similar to the IPlotExtensions experimented with here https://github.com/ScottPlot/ScottPlot/commit/a4ded384c82078f08bc4b251d7d3bde416e3fc31

Additionally, it would be nice to make all the private members on `Interaction` protected.

**Code example:**

```cs
public interface IPlotInteraction
{
    ContextMenuItem[] ContextMenuItems { get; }

    string DefaultSaveImageFilename { get; }

    /// <summary>
    /// Disable all mouse interactivity
    /// </summary>
    void Disable();

    /// <summary>
    /// Enable mouse interactivity using the default mouse actions
    /// </summary>
    void Enable();

    /// <summary>
    /// Enable mouse interactivity using custom mouse actions
    /// </summary>
    void Enable(PlotActions customActions);

    void OnMouseMove(Pixel newPosition);

    void KeyUp(Key key);

    void KeyDown(Key key);

    void MouseDown(Pixel position, MouseButton button);

    void MouseUp(Pixel position, MouseButton button);

    void DoubleClick();

    void MouseWheelVertical(Pixel pixel, float delta);
}
```

swharden:
resolved in #3111

PR #3111, opened 2023-12-22
SP5: Add IPlotInteraction

albyoo:
**Purpose:**
Addresses #3110
* Adds the IPlotInteraction Interface
* Updates IPlotControl to take an IPlotInteraction
* Update Interaction to make private members protected

Originally I was going to wait back to hear what others thought about this change, but while I waited I thought I'd give it a crack :)


albyoo:
I'm not quite sure about are having these methods in the interface:
```cs
ContextMenuItem[] ContextMenuItems { get; }
string DefaultSaveImageFilename { get; }
void Enable(PlotActions customActions);
```

swharden:
Thanks @albyoo, this is a great start! We can refine it over the coming weeks

> I'm not quite sure about are having these methods in the interface

I agree! As I start to work a little more with the Blazor control I can refine this to make it feel more natural.

Thanks again for this PR! It will get merged now and be in the next release 🚀

Issue #3113, opened 2023-12-23
SP5 Statistics: add Mean, Stdev, and StdErr methods for 2D arrays


swharden:
* Ideally could get results vertically or horizontally
* Also there should be paired methods like `NanStdErr()` which ignore NaN values
* It may be worth adding a `Rotate90()` while we're in there

PR #3121, opened 2023-12-24
Statistics for 2D arrays

arthurits:
**Purpose:**
* Proposes initial solution for #3113 
* Adds new overloaded methods: `NaNMean<T>`, `NaNVariance<T>`, `NaNVarianceP<T>`, `NaNStandardDeviation<T>`, `NaNStandardDeviationP<T>`, and `NaNStandardError<T>`, for both `IReadOnlyList` and 2D (not jagged) arrays.
* Auxiliary functions: `RemoveNaN<T>`, `RemoveNaN`, `ArrayTraspose`, and `ArrayToVector`, are defined as `public` but they are only used internally. Consider making them `private`.
* Auxiliary functions: `RemoveNaN<T>`, `RemoveNaN`, `ArrayTraspose`, and `ArrayToVector` do not really belong to `Descriptive.cs`. Should they be relocated to `NumericConversion.cs`?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -33,6 +33,7 @@
 * Box: Refactored box plot API to favor simplicity and user customization (#3072)
 * Rendering: Added `RenderManager.RenderStarting` event to allow modification of plottable properties (#3077) _Thanks GooBad_
 * Rendering: Added `RenderManager.PreRenderLock` event so developers of multi-threaded applications can ensure plottables are stable at render time (#3095) _Thanks @bclehmann_
+* Statistics: Added descriptive statistics methods and improved support for 2D arrays (#3113, #3121) _Thanks @arthurits_
 
 ## ScottPlot 4.1.70 (in development)
 * Population Plot: Improved performance for populations with curves that run off the screen (#3054) _Thanks @Em3a-c and @cornford_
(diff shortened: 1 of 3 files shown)


swharden:
Thanks for kicking this off @arthurits! I'm extending this PR a bit and merging so we have these tools around, but I expect them to continue to evolve as this library matures. A few thoughts:

* What should NanMean() return if all values in the array are NaN? NaN? 🤣 
* Performance can be improved by using `Span<double>` to avoid creating new arrays (some day in the future)
* I agree the namespace is a bit weird. I'd like `ScottPlot.Math.Mean()` but `Math` may conflict with `System.Math`
* I thought about having something like `ScottPlot.ArrayOperations.Mean()` and keeping `Statistics.Descriptive` for working with 1D data
* There may be room for a data type specifically for working with 1D and 2D data where statistics are of interest. For example, when working with 2D arrays I often want to know the vertical non-nan Mean, StandardDeviation, and N, but calculating these 3 separate times seems like a lot of work. Maybe something like a `GetVerticalStatistics(double[,] values)` could return a `PopulationStatistics` struct with `Mean`, `StandardDeviation`, `NanMean`, etc.

There's good progress here so I'll merge in what we have so far and we can continue to work on this in the future. Thanks again for your help here!

Issue #3148, opened 2024-01-01
SP5: need `ShortDash` pattern to compliment `Dash`


swharden:
also the `GetPathEffect()` in SkiaSharpExtensions couild be moved into the `LinePattern` .cs file

swharden:
resolved in #3161

PR #3161, opened 2024-01-04
SP5: LinePattern.DenselyDashed

MCF:
Addresses the requirement for a ShortDash LinePattern in https://github.com/ScottPlot/ScottPlot/issues/3148

I have chosen a dash pattern of 6 on 6 off as being shorter than the Dash LinePattern and longer than the Dot LinePattern.  I inspected V4 line patterns and would note:

1. V5 Dash, Dot (and now ShortDash if this PR is accepted) do not seem to match the patterns laid out in V4.
2. V5 is also missing a few other line types (happy to add those as well if needed).

Are we trying to recreate the same line patterns in V5 that are currently in v4?


swharden:
This looks great @MCF, thanks so much! I'm going to make some changes then merge this in, and we can add additional patterns later.

I'm not concerned about having different patterns/names from ScottPlot 4.

Maybe we can look to mimic patterns in matplotlib? https://matplotlib.org/stable/gallery/lines_bars_and_markers/linestyles.html

swharden:
I'm also adding a cookbook recipe that shows all the patterns available, and it'll update itself as new patterns are added 🚀 

image

PR #3169, opened 2024-01-07
SP5 Controls: Refactor `Interaction` and `Menu`

swharden:
This PR significantly modifies how users customize mouse actions and context menu items, making it far easier than previously. See the WinForms demo for details. 

## Custom Right-Click Menu
```cs
// add a custom action to the right-click menu
formsPlot1.Menu.Add("Add Scatter", (formsplot1) =>
{
    formsplot1.Plot.Add.Scatter(Generate.RandomCoordinates(5));
    formsplot1.Plot.Axes.AutoScale();
    formsplot1.Refresh();
});
```

## Custom Mouse Interactions
```cs
ScottPlot.Control.InputBindings customInputBindings = new()
{
    DragPanButton = ScottPlot.Control.MouseButton.Middle,
    DragZoomRectangleButton = ScottPlot.Control.MouseButton.Right,
    DragZoomButton = ScottPlot.Control.MouseButton.Right,
    ZoomInWheelDirection = ScottPlot.Control.MouseWheelDirection.Up,
    ZoomOutWheelDirection = ScottPlot.Control.MouseWheelDirection.Down,
    ClickAutoAxisButton = ScottPlot.Control.MouseButton.Right,
    ClickContextMenuButton = ScottPlot.Control.MouseButton.Left,
};

ScottPlot.Control.Interaction interaction = new(formsPlot1)
{
    Inputs = customInputBindings,
};

formsPlot1.Interaction = interaction;
```

I still question whether `IPlotInteraction` is a necessary interface to keep, but that structure can be refined later

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -28,7 +28,7 @@
 * Scatter Plot: Added support for DateTime types (#3154)
 * Style: Added support for more line patterns (#3161) _Thanks @MCF_
 * Controls: Assigning `Interaction` can be used instead of `Replace()` for customizing mouse actions (#3150)
-* Controls: Added `IMenu` for simpler context menu customization (#3150)
+* Controls: Added `Menu` with `Add()` and `Clear()` methods to simplify context menu customization (#3150)
 
 ## ScottPlot 5.0.11-beta
 _Published on [NuGet](https://www.nuget.org/profiles/ScottPlot) on 2023-12-29_
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -27,6 +27,8 @@
 * Scatter Plot: Added support for generic types including Xs and Ys of different types (#3154)
 * Scatter Plot: Added support for DateTime types (#3154)
 * Style: Added support for more line patterns (#3161) _Thanks @MCF_
+* Controls: Assigning `Interaction` can be used instead of `Replace()` for customizing mouse actions (#3150)
+* Controls: Added `IMenu` for simpler context menu customization (#3150)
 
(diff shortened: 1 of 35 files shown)


albyoo:
> I still question whether `IPlotInteraction` is a necessary interface to keep, but that structure can be refined later

The main use case I had for `IPlotInteraction` was when I needed a single interaction to control several plots. In this case I had too many scatter plots (14) to display on a single chart. To make them still easily comparable with many plots vertically stacked, I need all interactions to update all the other plots. To me it felt wrong to expand the existing `Interaction` because fundamentally it expects a 1:1 relationship with a plot, whereas I needed a different kind of relationship. If you think there's a better way to write this kind of interaction I'm interested to hear your thoughts, but I will always be an advocate for using an interface if possible :)

swharden:
Hi @albyoo, thank you for commenting about your use case! The ScottPlot 5 WinForms demo shows how to connect multiple plot controls together. I suspect this is a simpler way to get the effect you're looking for, but if you're unable to achieve this you can open up an issue and we can explore the topic and come up with a better strategy 👍 

https://github.com/ScottPlot/ScottPlot/blob/f3d3af7936f09f4e944fcd6cb68d7c4cbf91cb28/src/ScottPlot5/ScottPlot5%20Demos/ScottPlot5%20WinForms%20Demo/Demos/SharedAxes.cs#L1-L58

swharden:
... basically the recommended strategy is to tie into each plot's `RenderManager.AxisLimitsChanged` event and when it's fired use that opportunity to update the axis limits for all other plots

PR #3233, opened 2024-01-18
SP5: Remove duplicated render manager initialization

VoteForPedro:
Tiny PR aimed to remove duplicated render manager initialization.


swharden:
Thanks for the PR @VoteForPedro! lol I love the user name

Do you think the next line can be deleted too? 🤔

swharden:
> Do you think the next line can be deleted too?

I misread `Legend` as `Layout` -- my bad! This looks great as it is and I'll merge shortly 👍

Issue #3213, opened 2024-01-16
SP5 RenderManager: `EnableRendering` flag


swharden:
~For consistency it would be nice to add the `RenderLock()` and `RenderLock()` methods for consistency with previous behavior to compliment the `PreRenderLock` event handler added in #3095~

Update: I'm going to add a `EnableRendering` flag to allow this:

```cs
void MyDangerousOperation(){

    // disable new renders
    formsPlot1.Plot.RenderManager.EnableRendering = false;

    // wait for the current render to finish
    while (formsPlot1.Plot.RenderManager.IsRendering) {}

    // perform your dangerous operation
    myScatterPlot.DataSource.Clear();

    // permit new renders
    formsPlot1.Plot.RenderManager.EnableRendering = true;

    // force the plot to redraw
    formsPlot1.Refresh();
}
```

Note that `EnableRendering` _skips_ renders while `PreRenderLock` _hangs_ renders

swharden:
https://scottplot.net/faq/async/

kagerouttepaso:
Thank you for the update. However, I am unable to use `formsPlot1.Plot.Rendermanager.EnableRenderring = true;` as you have implemented the `EnableRenderring` flag as private in the modified code.

PR #3264, opened 2024-01-22
SP5: Change EnableRendering to Public property #3213

kagerouttepaso:
I have created a pull request addressing an issue related to the multi-threading support described at [https://scottplot.net/faq/async/](https://scottplot.net/faq/async/). 
According to this documentation, it is necessary to set `EnableRendering` to `false` for multi-threading purposes. However, this property is currently private.
 In this pull request, I have made the `EnableRendering` property public to resolve this issue.


swharden:
Good catch @kagerouttepaso, thanks so much!

PR #3265, opened 2024-01-22
SP5: Arrow

hockerschwan:
Added `ArrowCoordinated`

- Property name changed: `ArrowheadWidth` -> `ArrowheadWidthPixels` for clarity
- Added: `SetColor(Color)` for convenience
- Changed: `Plot.AddArrow(double xTip, double yTip, double xBase, double yBase, float lineWidth = 5, Color? color = null)` -> `Plot.Add.Arrow(double xBase, double yBase, double xTip, double yTip)` Is this OK?

arrow_quickstart


swharden:
Hi @hockerschwan, I just finished reviewing this PR. Fantastic job! I made a few small changes, but overall I'm really happy this thing landed in my inbox ready to merge and looking great 😄

* I added `Color` and `LineWidth` properties to simplify setting those and avoid `SetColor()`
* I removed the marker functionality. If users want to add a marker they can `Plot.Add.Marker()`
* I renamed `WidthPixels` to `Width` and use the `float` type to indicate it's in pixels

I probably won't do a deep dive into this right now, but there may be a <1px gap between the arrowhead and the line that's only visible due to anti-aliasing. A solution could be to allow 1px of overlap between the two shapes

image

swharden:
following-up, the overlap issue was fixed by this change in a confusing area 😅

image

Issue #3223, opened 2024-01-17
SP5: String measurement ignores leading and trailing whitespace


lindpatr:
**Issue:** (Text display is wrong when string to display contains spaces at start or at end.)
 
**ScottPlot Version:** (5.0.17)
 
**Code Sample:** Based on https://scottplot.net/cookbook/5.0/Text/Formatting/ sample code.

```
// Picture left side
var text = myPlot.Add.Text("Hello, World", 42, 69);

// Picture center
// 3 leading space characters
var text = myPlot.Add.Text("   Hello, World", 42, 69); 

// Picture  right side
// 3 trailing space characters
var text = myPlot.Add.Text("Hello, World   ", 42, 69); 
```

![image](https://github.com/ScottPlot/ScottPlot/assets/95692383/ff716ec6-16a0-47cb-9647-5187522770b7)

lindpatr:
According to this :  https://learn.microsoft.com/en-us/dotnet/api/skiasharp.skpaint.measuretext?view=skiasharp-2.88#skiasharp-skpaint-measuretext(system-byte()-skiasharp-skrect@)

**paint.MeasureText(...)** function returns the width of the text, taking into account all characters including leading and trailing whitespace.

```
public class Label
{
...
    public PixelSize Measure(SKPaint paint)
    {
        ApplyTextPaint(paint);
        SKRect textBounds = new();
        float textWidth = paint.MeasureText(Text, ref textBounds);
        
        // textWidth is not always greater than textBounds.Width (depends of the font name).
        return new PixelSize(Math.Max(textBounds.Width, textWidth), textBounds.Height);
    }
...
```

It seems to work...

devbotas:
![image](https://github.com/ScottPlot/ScottPlot/assets/71982803/1c459c71-e6b6-4a7d-81a6-f55eab86a8b8)

KroMignon:
@devbotas @lindpatr you are close to the solution ;)

According to this [post on StackOverflow](https://stackoverflow.com/a/77764594/13933646), the `SKPaint.MeasureText(string str, ref SKRect rect)` works as follow:
* the returned value is the length of the text with leading and trailing white space
* `rect.Left` contains the width of leading white space
* `rect.width` contains the length of the text **without** leading and trailing white space

So to resume, this should work:

```C#
public PixelSize Measure(SKPaint paint)
{
        ApplyTextPaint(paint);
        SKRect textBounds = new();
[...]

PR #3268, opened 2024-01-22
SP5: Fix text width measurement to include leading and trailing spaces

KroMignon:
According to this [post on StackOverflow](https://stackoverflow.com/a/77764594/13933646), the `SKPaint.MeasureText(string str, ref SKRect rect)` works as follow:

*  the returned value is the length of the text with leading and trailing white spaces
*  `rect.Left` contains the width of leading white spaces
*  `rect.width` contains the length of the text __without__ leading or trailing white spaces

This should fix issue #3223


swharden:
Great find @KroMignon, thanks so much!

PR #3296, opened 2024-01-24
SP5: `Plot.Remove()` overload that accepts `Type`

### LINKED ISSUE #3284: Plot.Remove() overload that accepts a type
it would be great if we could do this:

```cs
formsPlot1.Plot.Remove(typeof(ScottPlot.Plottables.Marker));
```

and under the hood it does something like this:

```cs
var oldMarkers = formsPlot1.Plot.PlottableList.OfType<ScottPlot.Plottables.Marker>().ToArray();
foreach (var oldMarker in oldMarkers)
{
    formsPlot1.Plot.Remove(oldMarker);
}
```

DerekGooding:
Remove by type.
Fix #3284

Returns false if no items matched that type for debugging purposes.


swharden:
Thanks so much @DerekGooding! This will be a great addition to the next release 🚀

swharden:
Wow! Note that this PR was further extended by another contributor in #3297

PR #3297, opened 2024-01-24
SP5: Plot.Remove() generic overloads

KroMignon:
This PR adds overload to `Plot.Remove()` to be able to remove `IPlottable` in generic way.

This should fix issue #3284.

Usage exampes:
```cs

myPlot.Add.Signal(Generate.Sin()).Label = "Toto";

// to remove all Signals
myPlot.Remove<ScottPlot.Plottables.Signal>();

....

// to remove all Signals without Label
myPlot.Remove<ScottPlot.Plottables.Signal>((x) => (x.Label?.Length ?? 0) <= 0);

...

```


swharden:
This is fantastic @KroMignon, thanks so much! I really like the predicate overload 🚀

I modified all the `Remove()` signatures to be `void` because it seems they should either all return `int` or none of them should. I figure it's easy enough for the user to measure the `Count` before/after the action if it's important to them. If there a consideration or use case you think makes it worth adding that functionality back in, let me know 👍

Issue #3157, opened 2024-01-04
SP5 Add support for Polygon gradient fill


faguetan:
## Feature Suggestion

**Feature description:** Would be really nice to have not only solid color fill for polygons, but also linear gradients.

swharden:
Thanks for this suggestion @faguetan! I just merged #3298 and added a cookbook recipe showing how to use it. There's room for improvement still, but I'll mark this issue as resolved and we can loop back around to refine it in the future

faguetan:
Awesome! Thank you, @KroMignon and @swharden !

PR #3298, opened 2024-01-24
SP5 FillStyle: add gradient fill support

KroMignon:
Hi @swharden,

this is a very preliminary version of gradient fill support (issue #3157)
Is this way to do okay for you?

Code example:
```cs

myPlot.Add.Signal(Generate.Sin()).Label = "Toto";

var rect = myPlot.Add.Rectangle(10, 33, 22, 45);
rect.FillStyle.Color = Colors.Blue;
rect.FillStyle.HatchColor = Colors.Red;
rect.FillStyle.Hatch = new Gradient();

...


swharden:
Hi @KroMignon, this is great! Thanks so much! I recognize that it's a WIP but it's already in a good spot so I'm going to merge it now and we can refine it in the future.

In plot types like polygons, maybe we could determine the rectangular boundary of the shape and pass that into the gradient class right before each render so the colors "track" the shape as it moves, rather than being "stuck" to the frame.

gradient-drag

KroMignon:
>I recognize that it's a WIP but it's already in a good spot so I'm going to merge it now and we can refine it in the future.

It is really in early stage, there are many tests I did not do. I have only do basic test with rectangles... I hope there are not too many side effects.

PR #3306, opened 2024-01-25
SP5 Plot: allow optional axes as arguments to GetPixel

MCF:
Now matches the inverse GetCoordinates functions

This is helpful when working with mouse events on multi-axes plots and determining nearest data point under the cursor.


swharden:
Good catch @MCF, thanks so much!

Issue #2419, opened 2023-02-25
SP5: Cache typefaces to improve Linux performance


nightfog-git:
Hello, 
I'm experiencing some problems with ScottPlot version 5 on linux, the rendering time is too big. If you enable the performance meter with double click, it shows 70-100 ms time and 10-15 fps on an empty plot. To run on linux I use an Avalonia gui, with custom-made control (this is not important, the root of the bug is plot.Render(SKSurface).
You may found a reproduction sample application on [this repo](https://github.com/nightfog-git/ScottPlot-AvaloniaTest).

I've looked at the code and conclude, that `Extensions/SkiaSharpExtensions -> ApplyToPaint(this FontStyle fontStyle, SKPaint paint)` is the root of the problem, and the exact call is `SKTypeface.FromFamilyName(fontStyle.Name, skfs)`.
For some unknown reason this call takes about 4ms, and that function is called by axes measure and panel rendering 17 times each rendering time!
To test my assumptions, I've replaced the var assignment with 
`SKTypeface typeface = (_typeface == null) ? SKTypeface.FromFamilyName(fontStyle.Name, skfs) : _typeface;`
and declared a static _typeface nearby at class level (`private static SKTypeface? _typeface = null;`). That is, I just cached the call's result. And, it began to work as expected! The frame rate became the same as I saw on Windows, 200+.
So, it seems some caching layer is needed to negate the slow call on Linux. I don't know the reasons, but I've checked that on two different distributions with the same outcome.

### System Details
* **ScottPlot Version:** 5.0.1-beta
* **Operating System:** Ubuntu Linux 22.04.1 LTS, ArchLinux current
* **Application Type:** Avalonia (actually didn't matter)
* **.NET Version:** .NET 6.0, 7.0

StendProg:
HI @nightfog-git,
The requested font may not be available on your system. The search takes a long time and eventually returns the default font.
Check that the requested and returned font match. Try changing the requested font to one that is exactly on your system.

nightfog-git:
I've checked that, the requested one is Numbus Sans, and `SKTypeface.FromFamilyName` returns the object with the same `FamilyName`. The font file is physically present in /usr/share/fonts/gsfonts.

swharden:
Thanks for reporting this @nightfog-git! 

For added context, the `SKTypeface.FromFamilyName()` call here is what you say is slow on Linux:

[...]

PR #3334, opened 2024-02-02
SP5: Fix typeface caching

Milkitic:
Fixes: https://github.com/ScottPlot/ScottPlot/issues/3335

Checked https://github.com/ScottPlot/ScottPlot/pull/2848 and https://github.com/ScottPlot/ScottPlot/issues/2419, but it seems that the issue is not actually fixed in the latest branch.
It causes critical performance issues in Linux systems.
In KUbuntu 23.10 with Avalonia 11.0.7 , the problem still exists as below:

Screenshot_20240202_105303

After fixing:

Screenshot_20240202_140558

Just simply modifed the code to reach the caching goals:

```cs
// public SKTypeface Typeface => CachedTypeface ?? CreateTypeface(Name, Bold, Italic); // broken
public SKTypeface Typeface => CachedTypeface ??= CreateTypeface(Name, Bold, Italic); // works
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@
 _Not yet on NuGet..._
 * Rendering: Added additional options for gradient fills (#3324) _Thanks @KroMignon_
 * Plot: Improve `GetPixel()` behavior when a custom `ScaleFactor` is in use (#3327) _Thanks @MCF_
+* Fonts: Improve behavior of cached typefaces (#3334, #3335) _Thanks @Milkitic_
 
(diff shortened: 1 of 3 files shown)


Milkitic:
This one will fix https://github.com/ScottPlot/ScottPlot/issues/3335

swharden:
Great catch @Milkitic, thanks so much for my fix! I was sick for a few weeks and got behind on issues and PRs, but I'm happy to be merging this in now. Thanks again!

PR #3352, opened 2024-02-09
Fix race condition in ScottPlot.DataViews.Wipe

drolevar:
Streamer.Data.NextIndex might get updated after it's assigned to newestCount, but before it's used when populating oldest[i], which sometimes leads to array index getting out of bounds.


swharden:
Great catch @drolevar! Thanks for identifying the issue and creating a PR to fix it as well. It took me longer to get to this one than typical because I was out sick the last few weeks, but I'm doing better now and happy to be merging this in! Thanks again 🚀

swharden:
Note that I didn't have permissions to edit this branch before merging, but I credited you in the changelog in #3358 😄 

Race conditions can be really hard to test for and troubleshoot, and again, thanks! I really appreciate this PR.

Issue #3369, opened 2024-02-16
SP5: RandomDataGenerator.RandomNormalSample throws System.IndexOutOfRangeException


DominicBeer:
https://github.com/ScottPlot/ScottPlot/blob/4b0c2def9c34c4d575b49eff0a3a40464bd5e0f4/src/ScottPlot5/ScottPlot5/RandomDataGenerator.cs#L143C26-L143C29

Should be < not <= 

I will probably make a fix and PR for this as it is a very simple fix. Will also likely add a test of some sort.

swharden:
resolved in #3370

PR #3370, opened 2024-02-16
#3369 Fix RandomNormalSample from always throwing

DominicBeer:
Single character change to stop the IndexOutOfRangeException


swharden:
Great catch @DominicBeer, thanks so much for identifying this issue and submitting this PR to fix it!

PR #3375, opened 2024-02-16
SP5: Updated LegendItems in BarPlot

fhannan-ti:
As of 5.0.21, `BarPlot` does not show a legend even after calling `Plot.ShowLegend()` or setting `Plot.Legend.IsVisible = true`. This change is to add that functionality, using the `Scatter` implementation as a reference.


swharden:
Thanks @fhannan-ti, this looks great! I made a few refinements, added a cookbook example, and am excited to merge this in now. It will be a great feature to have in the next release! Thanks again 🚀

PR #3376, opened 2024-02-16
SP5: Fixed legend in cookbook example for Grouped Bar Plot

fhannan-ti:
The current example for Grouped Bar Plot is not showing the legend. This change fixes it by using `Plot.ShowLegend(items, location)`, which is working as of 5.0.21.


swharden:
Hi @fhannan-ti, great catch!

Coincidentally I noticed this same bug and fixed it in #3375 before I saw this PR 😅 

I'll close this PR not because it's incorrect, but because it's already been fixed lol

Thanks again for this PR and the previous one too! I welcome any additional suggestions you have for improvements in the future! 🚀

PR #3373, opened 2024-02-16
SP5 Avalonia: Call InvalidateVisual from Dispatcher.UIThread.

oktrue:
Fixed #3359

PR #3381, opened 2024-02-19
WPF: Redraw plots using a non-blocking background thread

drolevar:
This is almost exactly the same as #3373 but for WPF. Checks whether Refresh is called on a UI thread, and if not, calls asynchronous Dispatcher.BeginInvoke(Refresh).


swharden:
This looks great @drolevar, thanks so much! 🚀

Issue #3397, opened 2024-02-22
SP5 Inverted Y axis loses labelling


BrianAtZetica:
**Issue:** When I invert the Y axis of a chart the Y-axis labels and horizontal grid lines are ommitted.

**ScottPlot Version:** ScottPlot.WPF v5.0.21

**Code Sample:** 
```cs
using ScottPlot;
using System.Drawing;

double[] dataX = { 1, 2, 3, 4, 5 };
double[] dataY = { 1, 4, 9, 16, 25 };

ScottPlot.Plot myPlot = new();
var trace = myPlot.Add.SignalXY(dataX, dataY);
myPlot.SavePng("quickstart1.png", 400, 300);

myPlot.Axes.SetLimits(left: 0, right: 6, bottom: 26, top: 0);
myPlot.SavePng("quickstart2.png", 400, 300);

```

quickstart1.png:
![quickstart1](https://github.com/ScottPlot/ScottPlot/assets/119825052/ca69a075-dffa-4779-95c4-3657e861a800)

quickstart2.png:
![quickstart2](https://github.com/ScottPlot/ScottPlot/assets/119825052/c146c15c-d3f7-4470-9b52-4556f16f4fba)

swharden:
Fixed in #3401

PR #3400, opened 2024-02-23
#3396 SignalXY with Inverted YAxis

BrianAtZetica:
This resolves issue #3396, allowing reversed X axes on SignalXY plots, and reversed Y axes on rotated SignalXY plots.

Code below was used in WPF sample project to verify various scenarios.
Note that inverted axis is currently missing axis labels according to #3397 

```cs
using System;
using System.Linq;
using System.Windows;
using ScottPlot;

namespace Sandbox.WPF;

public partial class MainWindow : Window
{
    public MainWindow()
    {
        InitializeComponent();
        double[] Xs = Enumerable.Range(1, 5000).Select(i => (double)i).ToArray();
        var signal = WpfPlot1.Plot.Add.SignalXY(Xs, Generate.Sin(count: 5000, oscillations: 4));
        //Test with normal X axis
        //WpfPlot1.Plot.Axes.SetLimitsX(left: 100.1, right: 5100);
        
        //Test with reversed X axis
        //WpfPlot1.Plot.Axes.SetLimitsX(left: 5100, right: 100.1);

        //Test with rotated SignalXY plot on normal Y axis
        //signal.Data.Rotated = true;
        //WpfPlot1.Plot.Axes.SetLimitsY(bottom: 100.1, top: 5100);

        //Test with rotated SignalXY plot on reversed Y axis
        signal.Data.Rotated = true;
        WpfPlot1.Plot.Axes.SetLimitsY(bottom: 5100, top: 100.1);

        //test with generic data type (not double) on normal X axis
        //int[] Xs = Enumerable.Range(1, 5000).Select(i => (int)i).ToArray();
        //int[] Ys = Enumerable.Range(1, 5000).Select(i => (int)i).ToArray();
        //var signal = WpfPlot1.Plot.Add.SignalXY(Xs,Ys);
        //WpfPlot1.Plot.Axes.SetLimitsX(left: 100.1, right: 5100);

        //test with generic data type (not double) on reversed X axis
        //int[] Xs = Enumerable.Range(1, 5000).Select(i => (int)i).ToArray();
        //int[] Ys = Enumerable.Range(1, 5000).Select(i => (int)i).ToArray();
        //var signal = WpfPlot1.Plot.Add.SignalXY(Xs,Ys);
        //WpfPlot1.Plot.Axes.SetLimitsX(left: 5100, right: 100.1);

    }
}

```

PR #3401, opened 2024-02-24
#3397 Resovles broken Axes labels and ticks on inverted axes

BrianAtZetica:
This pull request resolves issue #3397 
The issue stems from the use of terms Min and Max, which are not numerically minimum and maximum values when axes are inverted. 

The fix was to find where the assumptions were in correct and locally calculate a "trueMin" or "trueMax" value.

I believe the proper fix would be to have IAxes retain Min and Max properties, but make them a true numeric minimum and maximum. This woul dbe augmented by an "Inverted" property.
Meanwhile, IXAxis would inherit Min, Max and Inverted, and could add Left and Right properties for convenience.
Similarly, IYAxis would inherit Min, Max and Inverted properties, but could add Top and Bottom peroperties for convenience.

Unfortunately that change would require extensive changes and careful planning and woud likely introduce breaking code changes for some users. 

I believe my changes are safe for existing code.

Note that this pull request is dependant on pull request #3400 (I needed the inverted axes to be functioning to test all four scenarios (X/Y axes inverted with rotation)

Test code below.

```cs
using System;
using System.Linq;
using System.Windows;
using ScottPlot;

namespace Sandbox.WPF;

{
    public MainWindow()
    {
        InitializeComponent();
        double[] Xs = Enumerable.Range(1, 5000).Select(i => (double)i).ToArray();
        var signal = WpfPlot1.Plot.Add.SignalXY(Xs, Generate.Sin(count: 5000, oscillations: 4));
        //Test with normal X axis
        //WpfPlot1.Plot.Axes.SetLimitsX(left: 100.1, right: 5100);

        //Test with reversed X axis
        WpfPlot1.Plot.Axes.SetLimitsX(left: 5100, right: 100.1);

        //Test with rotated SignalXY plot on normal Y axis
        //signal.Data.Rotated = true;
        //WpfPlot1.Plot.Axes.SetLimitsY(bottom: 100.1, top: 5100);

        //Test with rotated SignalXY plot on reversed Y axis
        //signal.Data.Rotated = true;
        //WpfPlot1.Plot.Axes.SetLimitsY(bottom: 5100, top: 100.1);

        //test with generic data type (not double) on normal X axis
        //int[] Xs = Enumerable.Range(1, 5000).Select(i => (int)i).ToArray();
        //int[] Ys = Enumerable.Range(1, 5000).Select(i => (int)i).ToArray();
        //var signal = WpfPlot1.Plot.Add.SignalXY(Xs,Ys);
        //WpfPlot1.Plot.Axes.SetLimitsX(left: 100.1, right: 5100);

        //test with generic data type (not double) on reversed X axis
        //int[] Xs = Enumerable.Range(1, 5000).Select(i => (int)i).ToArray();
        //int[] Ys = Enumerable.Range(1, 5000).Select(i => (int)i).ToArray();
        //var signal = WpfPlot1.Plot.Add.SignalXY(Xs,Ys);
        //WpfPlot1.Plot.Axes.SetLimitsX(left: 5100, right: 100.1);

    }
}

```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,7 @@ _Not yet on NuGet..._
 * Color: Added `Lighten()` and `Darken()` properties (#3387, #3390) @KroMignon
 * Color: Modified `ToHSL()` to return improved Hue, Saturation and Luminosity values (#3390) @KroMignon
 * SignalXY: Improve support for displaying data on inverted axes (#3396, #3400) @BrianAtZetica
+* Axes: Improved support for ticks and labels on inverted axes (#3401, #3397) @BrianAtZetica
 
 ## ScottPlot 5.0.21
 _Published on [NuGet](https://www.nuget.org/profiles/ScottPlot) on 2024-01-28_
--- a/src/ScottPlot5/ScottPlot5/AxisManager.cs
+++ b/src/ScottPlot5/ScottPlot5/AxisManager.cs
@@ -215,7 +215,7 @@ public class AxisManager
     {
         yAxis.Min = bottom;
         yAxis.Max = top;
-        AutoScaler.InvertedY = bottom > top? true : false;
+        AutoScaler.InvertedY = bottom > top ? true : false;
     }
 
     public void SetLimitsX(double left, double right)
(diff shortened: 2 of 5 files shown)


swharden:
> Unfortunately that change would require extensive changes and careful planning and woud likely introduce breaking code changes for some users.

You made some great observations, and I'm really happy with the implementation you suggested in this PR - thanks so much!

To be honest I didn't even realize that the cookbook example on https://scottplot.net/cookbook/5.0/AdvancedAxis/InvertedAxis/ was currently broken 🤦 Thanks for identifying the issue and fixing it!

Current recipe | Recipe after this PR
---|---
image|image

PR #3407, opened 2024-02-25
FractionalAutoScaler: do not adjust scale for invisible plottables

levipara:
The AutoScaler currently expands axes ranges to include non-visible plottables. Assumed behavior, unless otherwise documented, would be that hidden plottables are NOT considered. This change simply ignores a plottable if it is hidden when scaling axes ranges. Please evaluate for merge


swharden:
Thanks @levipara! This is a great catch and an excellent fix. It will be in the next release! 🚀

PR #3410, opened 2024-02-26
Controls: pan/zoom in a single dimension if mouse is pressed over an axis panel

drolevar:
Enable separate zooming and panning of axes.


swharden:
Hi @drolevar, thanks for this PR! 

~~Unfortunately I'm really struggling to understand exactly what this does. Is there a particular use case that can help me understand the problem this solves?~~

Update: This PR enables restricts panning and zooming to a single dimension if the mouse is pressed over the axis label (outside the data area).

I look forward to merging this in now and it will be in the next release! 🚀 

pan-zoom-1d

swharden:
I updated this PR by moving some logic around and adding support for axis-specific middle-click autoscale behavior and middle-click-drag zoom rectangle behavior. This is a great improvement, thanks again!

PR #3411, opened 2024-02-26
SP5: Improve DataLogger axis management

drolevar:
Make sure that when a DataLogger updates axis limtis, it only updates the axes it's using.


swharden:
Thanks @drolevar, this looks great!

PR #3413, opened 2024-02-26
Improve automatic tick placement for centiseconds and deciseconds

drolevar:
Fix snapping for centisecond and decisecond.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,6 +22,7 @@ _Not yet on NuGet..._
 * Axes: Updated the auto-scaler to ignore plottables with visibility disabled (#3407) @levipara
 * Axes: Restrict pan, zoom, and autoscale to a single dimension if the cursor is over an axis panel (#3410) @drolevar
 * Controls: Improved behavior of middle-click-drag zoom rectangle actions when CTRL or SHIFT is pressed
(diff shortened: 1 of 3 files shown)


swharden:
Here's a sandbox app I use for testing

```cs
public Form1()
{
    InitializeComponent();

    TimeSpan delta = TimeSpan.FromMilliseconds(100);
    DateTime start = new(2021, 01, 01);
    for (double i = 0; i < Math.PI * 2; i += .1)
    {
        double x = (start + delta * i).ToOADate();
        double y = Math.Sin(i);
        formsPlot1.Plot.Add.Marker(x, y);
    }

    formsPlot1.Plot.Axes.DateTimeTicksBottom();
}
```

image

drolevar:
@swharden If I understand the logic correctly, the Snap method for deciseconds is supposed to round the DateTime to a whole amount of deciseconds. Am I right? The issue is that without this patch I was seeing strange issues. I'll come with an example a bit later.

swharden:
> The issue is that without this patch I was seeing strange issues

It took me over an hour to arrive at the 3-line fix 😅

It was doubly confusing because sometimes centisecond snapping issues were caused by decisecond snapping issues 🤦

Issue #3429, opened 2024-02-29
DataStreamer: Index out of range exception


swharden:
This one was emailed to me* but I don't have much additional context

*The email address name is `境者` which translates to `Those who are in a state of affairs` 🤷 

<img width="988" alt="5F01D3A8@459B452A 6AF4D765" src="https://github.com/ScottPlot/ScottPlot/assets/4165489/d56ba041-e794-4756-964a-fb82f9924734">

KroMignon:
Hi @swharden , hope you are going well.

I think I understand the problem in `DataViews.Scroll` class.
The code bellow expects that the Data are not updated during rendering:
```csharp
public void Render(RenderPack rp)
{
    Pixel[] points = new Pixel[Streamer.Data.Length];

    int oldPointCount = Streamer.Data.Length - Streamer.Data.NextIndex;

    for (int i = 0; i < Streamer.Data.Length; i++)
    {
        bool isNewPoint = i < oldPointCount;
        int sourceIndex = isNewPoint ? Streamer.Data.NextIndex + i : i - oldPointCount;
        int targetIndex = NewOnRight ? i : Streamer.Data.Data.Length - 1 - i;
        points[targetIndex] = new(
            x: Streamer.Axes.GetPixelX(targetIndex * Streamer.Data.SamplePeriod + Streamer.Data.OffsetX),
            y: Streamer.Axes.GetPixelY(Streamer.Data.Data[sourceIndex] + Streamer.Data.OffsetY));
    }

    using SKPaint paint = new();
    Drawing.DrawLines(rp.Canvas, paint, points, Streamer.LineStyle);
}
```

To solve this, the "volatile" value should be cached or the access to the `Data` should be "locked" to avoid changes during rendering.
I have create a PR to solve this by caching the _volatile_ values.

swharden:
> To solve this, the "volatile" value should be cached or the access to the Data should be "locked" to avoid changes during rendering.
> I have create a PR to solve this by caching the volatile values.

Thanks for this clarification! I'm happy to close this issue for now, but in the future we could consider adding some type of locking system so that data modifications don't occur mid-render... for the short term though, I'm happy to leave this where it is.

Thanks again for your work on #3430!

PR #3430, opened 2024-03-01
DataLogger and DataStreamer: fix possible out of range exception when using Scroll view

KroMignon:
Fix out of range exception in `DataViews.Scroll` when `Streamer` is updated will `Render()` is running

This should fix issue #3429


UrbaLaba7:
Very Good, I approve

Issue #3438, opened 2024-03-05
How to create a doughnut chart in Scott Plot 5


Prototipo-Erick-Santander:
**Question:** I Would like to know how can I create a doughnut chart in Scott Plot 5, I know that I can make a pie chart and on V4.1 there used to be a .DonutSize property, which is no longer available.

**ScottPlot Version:** 5

**Code Sample:** 
```cs
public void Create_ScottPlot_Doughnut(string path)
    {
        double[] values = { 44.2, 51.9, .09, 1.1, .8, 1.1 };
        string[] labels = { "External", "Sold", "Personel", "Others", "Depreciation", "Tax" };
        var palette = new ScottPlot.Palettes.Category20();
        
        // Define the slices for the doughnut chart
        var slices = new List<PieSlice>();
        for (var i = 0; i < values.Length; i++)
        {
            var slice = new PieSlice
            {
                Value = values[i], 
                Label = values[i].ToString("N2"),
                LabelStyle = new Label { IsVisible = true },
                FillColor = palette.Colors[i],
            };
            
            slices.Add(slice);
        }

        // Add the doughnut chart to the plot
        Plot myPlot = new();
        var pie = myPlot.Add.Pie(slices);
        
        //pie.ExplodeFraction = 0.05; // Set the donut size to create a hollow center
        pie.ShowSliceLabels = true;
        pie.SliceLabelDistance = 1.3;
        pie.LineStyle.Width = 0;

        myPlot.HideGrid();
        myPlot.Layout.Frameless();
        myPlot.Legend.IsVisible = true;
        myPlot.Legend.Location = Alignment.LowerCenter;
        
        myPlot.SavePng(path, 1000, 650);
        Console.ReadKey();
    }
```

This is something like what I would like to do:
![image](https://github.com/ScottPlot/ScottPlot/assets/62098744/fb1b0027-9d06-4b6c-bcb9-9292de6b61b3)
Is it also possible to make the labels never be on top of one another?

aespitia:
i was actually just looking into the same thing today.  one idea i thought about was just adding a while filled circle in the middle of the chart.

i had also looked into updating the code to add the DonutSize like 4.1 had as well.  doesn't seem like it would be too hard to add for someone that knows what they're doing.  if i have time i might look into this as well and submit a PR.

aespitia:
I've created a PR for this:
https://github.com/ScottPlot/ScottPlot/pull/3447

swharden:
[...]

PR #3447, opened 2024-03-06
Pie chart: add support for "donut" display style

aespitia:
#3438

this is to add the donut functionality that was in SP4 and port to SP5.


aespitia:
would love feedback on this @swharden, my first time messing with the actual rendering code, so I wouldn't be surprised if there's another/better way to do this.

swharden:
Thanks so much for this @aespitia! 

I'm going to modify this PR slightly to draw the outline around the inner donut, then merge 🚀 

image

swharden:
> draw the outline around the inner donut

Oof, this got complicated fast 😱 It means we need to use a path and "ArcTo" instead of the clever clipping method 🤦 

I'm going to merge this now and create a new issue to track that refinement. If you want to have a go at it, awesome! If not, that's okay too - someone from the community may get to it. If not, I'll implement it eventually 😎

swharden:
Got it! That was actually super confusing 🤪 

image

aespitia:
awesome, thanks so much!

PR #3455, opened 2024-03-10
Change `Plot.ScaleFactor` from `float` to `double`

### LINKED ISSUE #3454: Plot: ScaleFactor should be a double
Adding the `f` is a bit annoying

```cs
ScottPlot.Plot plt = new();
plt.ScaleFactor = 1.5f;
```

MCF:
As per the title.  Will fix #3454


swharden:
haha, that change touched a lot more lines of code than I expected it would 😅

I'll see if I can simplify this somehow, and merge shortly. Thanks so much! 🚀

Issue #3453, opened 2024-03-07
Markers: OpenCircle issue


swharden:
I don't think the `OpenCircle` shape is being displayed as expected

```cs
var mk = plt.Add.Markers(xs2, ys2);
mk.MarkerStyle.Shape = MarkerShape.OpenCircle;
```

![image](https://github.com/ScottPlot/ScottPlot/assets/4165489/4cf56e3e-061b-4867-bf9f-2d7882375482)

swharden:
resolved in #3456

Issue #3141, opened 2023-12-31
SP5: refactor marker system


swharden:
Currently markers are structs, but I'm unsure if this is the best strategy. It results in marker designators being decoupled from marker rendering code and makes it really hard for users to inject custom markers. 

The advantage of keeping `MarkerShape` a struct is that it can have clear default values in argument lists, but SP5 tries to move away from methods with long argument lists, so it turns out that this isn't a strong advantage after all.

markers are confusing because some have separate outline and fill colors

PR #3456, opened 2024-03-11
Marker: fix rendering issue affecting filled vs. outlined shapes

KroMignon:
Fix `MarkerStyle` rendering when `Shape` is changed from ___filled___ to ___outlined___ and vice versa.

This should fix issue #3453

**NOTES:** when changing from __filled__ to __outlined__ shape, the width of the outline must be set because it is 0 per default in this case.

```cs

double[] xs = Generate.Consecutive(51);
double[] sin = Generate.Sin(51);
double[] cos = Generate.Cos(51);

myPlot.Add.Markers(xs, sin, MarkerShape.OpenCircle, 15, Colors.Green);
var mk1 = formsPlot1.Plot.Add.Markers(xs, cos);
mk1.MarkerStyle.Shape = MarkerShape.OpenSquare;
// Default shape is MarkerShape.FilledCircle, so we have to set `MarkerStyle.Outline.Width` to be greater than 0!
mk1.MarkerStyle.Outline.Width = 2;
mk1.Color = Colors.DarkOrange;

myPlot.SavePng("MarkerShapes.png", 320, 240);
```
MarkerShapes


swharden:
Thanks so much @KroMignon! I've been meaning to refactor the marker system for a while (#3141), but this fix helps me put it off a little longer 😅 

I'll make a quick change now (having each marker set its own default line width) so line-only markers are visible without requiring users to set that, then I'll merge this in in a few minutes.

Thanks again! 🚀

swharden:
```cs
MarkerShape[] markerShapes = Enum.GetValues<MarkerShape>().ToArray();
ScottPlot.Palettes.Category20 palette = new();

for (int i = 0; i < markerShapes.Length; i++)
{
    var mp = myPlot.Add.Marker(x: i, y: 0);
    mp.MarkerStyle.Shape = markerShapes[i];
    mp.MarkerStyle.Size = 10;
    mp.MarkerStyle.Outline.Width = 1.5f;
    mp.MarkerStyle.Outline.Color = palette.GetColor(i);
    mp.MarkerStyle.Fill.Color = palette.GetColor(i).WithAlpha(.5);
}
```

image

PR #3462, opened 2024-03-13
Scatter: `Plot.Add.ScatterLine()` convenience methods

### LINKED ISSUE #3452: Plot.Add.ScatterLine()
It would be nice to have a helper method which under the hood does this:

```cs
var sp = plt.Add.Scatter(xs, ys);
sp.MarkerSize = 0;
```

It could be used like:
```cs
Plot.Add.ScatterLine(xs, ys);
```

A similar method that adds a scatter point with markers but no line would be nice to have too

MCF:
These new methods match the Plot.Add.Scatter methods in argument and functionality, except the plot markers are not shown.  

The old way of adding a scatter plot without markers:

```csharp
var sp = plt.Add.Scatter(xs, ys);
sp.MarkerSize = 0;
```

The new way:

```csharp
plt.Add.ScatterLine(xs, ys);
```

Fixes #3452


swharden:
Thanks @MCF, this looks great! I simplified the cookbook recipe and added a complimentary `AddScatterPoints()` to match the API from ScottPlot 4.1

PR #3466, opened 2024-03-13
DateTime ticks: simplify snapping logic

drolevar:
@swharden I have screwed in my previous PR. It was originally correct, but then I for some strange reason decided it was not (tired I guess). 
So, to put things short, check if this one works for you. If you think this logic is worse, just drop it.


swharden:
>  It was originally correct, but then I for some strange reason decided it was not

haha, no worries! This modulus strategy is more consistent with the other tick scale classes and I'm happy to merge it in 🚀

PR #3468, opened 2024-03-13
Optimize GenerateTicks logic

drolevar:
This optimizes GenerateTicks and avoids several unnecessary Drawing.MeasureString(...) calls.


swharden:
Thanks @drolevar, this looks great!

PR #3498, opened 2024-03-17
Controls: Improve ZoomRectangle behavior and axis locking

BrianAtZetica:
When using the middle button to draw a zoom rectangle, it was easy to accidentally roll the wheel and cause the canvas to zoom. This resulted in the zoom rectangle no longer representing the intended extents. 

This simple fix just ignores mousewheel zooming if a zoom window is currently being drawn.


swharden:
Thanks @BrianAtZetica, this is great! 🚀 

Typos always hide in the sneakiest places 😜

Issue #3503, opened 2024-03-18
Demo: Add example for Scatter.GetNearest() using multiple scatter plots


RubensMigliore:
I successfully implemented suggestion for handling the MouseMove event in charts with only one dataset with ScottPlot 5 in VB .NET. However, I’m facing another situation with multiple curves being drawn, but MyCrosshair is only triggered for the last dataset. I suspect this problem is related to the following method that uses ‘PltPerfElev.Plot.LastRender’:

Dim nearest As DataPoint = splt.Data.GetNearest(mouseLocation, PltPerfElev.Plot.LastRender)

Is it possible to review the following code to present CrossHair for the entire dataset?

Here is my VB. NET implementation.
![Capturar-2](https://github.com/ScottPlot/ScottPlot/assets/104206448/80780cc0-6e87-48c4-b13c-79de50f05322)

```vv
 Private Sub MouseMoveHandler(
      s As Object, e As MouseEventArgs)

  Dim mousePixel As New Pixel(
      e.Location.X, e.Location.Y)
  Dim mouseLocation As Coordinates =
      PltPerfElev.Plot.GetCoordinates(mousePixel)
  Dim nearest As DataPoint = splt.Data.
      GetNearest(mouseLocation, PltPerfElev.Plot.LastRender)

  If nearest.IsReal Then
      MyCrosshair.IsVisible = True
      MyCrosshair.Position = nearest.Coordinates
      PltPerfElev.Refresh()
      lbPerfElev.Text = $"  X = {nearest.X:#,##0.0} m " &
          $"   -    Y = {nearest.Y:#,##0.0} cm"
  ElseIf MyCrosshair.IsVisible Then
      MyCrosshair.IsVisible = False
      PltPerfElev.Refresh()
      lbPerfElev.Text = $""
  End If
End Sub
```

swharden:
Here's the C# version of that VB code (created using https://converter.telerik.com/). Providing the C# may help this question get answered quicker 👍 

```cs
private void MouseMoveHandler(object s, MouseEventArgs e)
{
    Pixel mousePixel = new Pixel(e.Location.X, e.Location.Y);
    Coordinates mouseLocation = PltPerfElev.Plot.GetCoordinates(mousePixel);
    DataPoint nearest = splt.Data.GetNearest(mouseLocation, PltPerfElev.Plot.LastRender);

    if (nearest.IsReal)
    {
        MyCrosshair.IsVisible = true;
[...]

PR #3506, opened 2024-03-18
Text: add Offset support

jamaa:
When rendering a `Text` plottable, add its `PixelOffset` to the rendering location.

```cs
ScottPlot.Plot myPlot = new();
ScottPlot.Plottables.Text myText = myPlot.Add.Text("test", 0, 0);
myText.Label.Alignment = Alignment.LowerLeft;
myText.PixelOffset = new PixelSize(5, -5); // offset text 5 pixels right and up
myPlot.SavePng("Text with offset.png", 600, 400);
```

grafik

PR #3507, opened 2024-03-18
Demo: show value near cursor with multiple scatter plots

jamaa:
As suggested in #3503, this adds a a demo for hovering over the nearest point of multiple scatter plots.

Title is "Show Value Under Mouse, Multiple Series", please advise whether to name it differently to differentiate it more from the existing demo "Show Value Under Mouse".

Also, maybe I went a little overboard by also adding a marker and a text to the hovered point?

grafik


jamaa:
btw, the text placement being slightly offset to the point relies on #3506

Issue #3524, opened 2024-03-23
Render within an AxisChaged event handler causes incorrect zoom


BrianAtZetica:
**Issue:** if a user creates an AxisChanged event handler and calls Render() within that handler, the axis limits change unexpectedly.

For example if I start with the basic two sinusoids:
![image](https://github.com/ScottPlot/ScottPlot/assets/119825052/39f58f16-f8c9-46c1-945c-2f37de953027)

Then position my cursor at x = 31, y = -0.7 (where the sinusoids cross each other) then mousewheel zoom out two clicks I get this:
![image](https://github.com/ScottPlot/ScottPlot/assets/119825052/a6e18a0b-af87-47ca-bc4d-2cb58bca836c)

(note change to rendering size seems to be part of the problem?)

**ScottPlot Version:** Current github Main 

**Code Sample:** WPF example code. The example is a little contrived, but the point is a user may very easily decide to force a render within the axischanged handler event.

```cs
        InitializeComponent();
        WpfPlot1.Plot.RenderManager.AxisLimitsChanged += AxisChanged;
        WpfPlot1.Plot.Add.Signal(Generate.Sin(51));
        WpfPlot1.Plot.Add.Signal(Generate.Cos(51));

        void AxisChanged(object sender, RenderDetails e)
        {
            WpfPlot1.Plot.Render();
        }
```

PR #3510, opened 2024-03-19
Proposed implementation of an Annotation Plottable

dlampa:
Reference to issue  #3356 
Outstanding issues:

- Currently missing the `MarginX/MarginY` implementation
- Needs firming up on the default formatting
- Needs firming up the default (minimum) offset from respective axes to ensure uniform and predictable spacing
- Additional tests?
- Example code/contrib. to the Cookbook

Issue #3508, opened 2024-03-19
AxisLimitsChanged event: the emitted `RenderDetails` contains old information


BrianAtZetica:
**Question:** Why are the RenderDetails one render behind when implementing an AxisChanged event handler? Is this a WPF only issue?

**ScottPlot Version:** current github main branch

**Code Sample:** 
In the following code sample (modified from the WPF sandbox from the repo) I have subscribed to the AxisLimitsCHanged event with my AxisChanged handler. In my handler I set the title of the plot to show the "AxisLimits" from the RenderDetails object.

```cs
using System.Linq;
using System.Windows;
using ScottPlot;

namespace Sandbox.WPF;

public partial class MainWindow : Window
{
    public MainWindow()
    {
        InitializeComponent();
        WpfPlot1.Plot.RenderManager.AxisLimitsChanged += AxisChanged;
        double[] Xs = Enumerable.Range(1, 5000).Select(i => (double)i / 1000).ToArray();
        var signal = WpfPlot1.Plot.Add.SignalXY(Xs, Generate.Sin(count: 5000, oscillations: 4));

        void AxisChanged(object sender, RenderDetails e)
        {
            var changedPlot = (ScottPlot.Plot)sender;
            var newLimits = e.AxisLimits;
            WpfPlot1.Plot.Title($"New Y limits: {newLimits.Bottom:F1}, {newLimits.Top:F1} New X limits: {newLimits.Left:F1}, {newLimits.Right:F1}");
        }
    }
}

```

When first rendered, the AxisChanged event must have been called during initialisation and plotting of the data, and the title has the correct limits:
<img width="372" alt="image" src="https://github.com/ScottPlot/ScottPlot/assets/119825052/4fd09c3b-4771-461d-bafb-745b7799ae79">

But if I zoom out once with the mousewheel the axis limits in the title are old (note: I had to screengrab this, because a copy to clip board from the right click context menu would force another render and the saved image title is then correct!):
<img width="379" alt="image" src="https://github.com/ScottPlot/ScottPlot/assets/119825052/0c907657-df41-4346-9031-469ae6ec715b">

If I then reset zoom with mouse wheel click
the graph resets correctly, but the axis title shows the previous zoom extent:
<img width="374" alt="image" src="https://github.com/ScottPlot/ScottPlot/assets/119825052/2204584d-dae3-4134-8e3a-8a59b7920b48">

[...]

PR #3525, opened 2024-03-23
Rendering: Improve event handling early during a render

BrianAtZetica:
This PR addresses #3524 

If a render is still in progress any call to render will be returned. 
This prevents unpredictable axis changes when a user calls for a render within an AxisChangedEvent, but it does mean it is not possible to force a render within any SizeChanged or AxisLimitChanged events.

Is this OK?

Code for testing:

```cs
        InitializeComponent();
        WpfPlot1.Plot.RenderManager.AxisLimitsChanged += AxisChanged;
        WpfPlot1.Plot.Add.Signal(Generate.Sin(51));
        WpfPlot1.Plot.Add.Signal(Generate.Cos(51));

        void AxisChanged(object sender, RenderDetails e)
        {
            WpfPlot1.Plot.Render();
        }

```


BrianAtZetica:
I solved both #3524 and #3508 with this PR. 
I have added a new event during the render that fires BEFORE render actions take place.  
The event is called AxisLimitsChanging.

This allows things like labels to be set in response to the new axis limits, and yet be rendered in the same render.

It also solves my original requirement, which was to give me knowledge of the direction and scale of the zoom change, so I can snap to a tick interval.

Example code using the new event:

```cs
        InitializeComponent();
        WpfPlot1.Plot.RenderManager.AxisLimitsChanging += AxisChanging;
        WpfPlot1.Plot.Add.Signal(Generate.Sin(51));
        WpfPlot1.Plot.Add.Signal(Generate.Cos(51));

        void AxisChanging(object sender, AxesChangingArgs e)
        {
            var changedPlot = (ScottPlot.Plot)sender;
            var newLimits = e.NewAxisLimits;
            var oldLimits = e.OldAxisLimits;

            if ( Math.Abs(newLimits.Right - newLimits.Left) < Math.Abs(oldLimits.Right - oldLimits.Left))
            {
                WpfPlot1.Plot.Title($"You zoomed In on X");
            }
            else if(Math.Abs(newLimits.Right - newLimits.Left) > Math.Abs(oldLimits.Right - oldLimits.Left))
            {
                WpfPlot1.Plot.Title($"You zoomed Out on X");
            }
            else
            {
                WpfPlot1.Plot.Title($"You didn't Zoom X");
            }
            WpfPlot1.Plot.YLabel($"New Y limits: {newLimits.Bottom:F1}, {newLimits.Top:F1}");
            WpfPlot1.Plot.XLabel($"New X limits: {newLimits.Left:F1}, {newLimits.Right:F1}");
        }
```

The EventArgs being sent is from a new class called AxisChangingEventArgs (It didn't seem apprpriate to send a RenderDetails object as the data woudl be form the previous render not in in progress one).
I wasn't sure where to put this AxisChangingEventArgs, as I couldn't find any other similar EventArgs classes, so take a look and see if I got it right. 

image

swharden:
Thanks @BrianAtZetica! I'm really liking the direction this is going. I have a bit of feedback that may result in code changes, and you can make them or I can based on what you think.

## Events
I see what you're doing with the new `AxisLimitsChanging` event (and the `AxesChangingArgs` it emits). We may get more general-purpose functionality if we change it to something like `RenderStarting` to compliment the existing `RenderFinished` event. It could emit the `RenderPack` which has all sorts of information (including the plot, which can access previous axis limits etc).

## Skipping Renders
Regarding exiting `Render()` early if the plot is already rendering, this is probably okay, but that's a really interesting method that is called only a few very special times. I think it's called to force a render only in memory (notice that nothing is returned) to setup the axis, layout, etc. Leaving that method early makes me a bit concerned. What would you think about replacing the "if IsRendering, return" with a "while IsRendering, delay 1 ms" or something? That way that manually-requested render is guaranteed to happen, and a render will be completed before that function returns. I've only looked at this PR in the browser so far, but I can dig a little deeper and see when/why that odd function exists and what it's needed for

swharden:
... sorry about the accidental commits. I'm going to revert them in this branch and leave this like I found it 😅 https://github.com/ScottPlot/ScottPlot/pull/3527#issuecomment-2016648180

BrianAtZetica:
Hi @swharden, 

### Skipping Renders
I took a look at the Skipping Renders problem further. I explored using a Semaphore instead of a "while IsRendering, delay 1 ms" (thinking this would be less harmful on CPU usage). Unfortunately, both methods have the same issue. 
The situation I was mitigating was the call of plot.Render() from within the AxisLimitsChanged event. Unfortunately, the AxisLimitsChanged event is happening on the same thread, so whether we use a Semaphore or a simple Delay loop, the second render will delay indefinitely waiting for (and holding up) the previous render that invoked the event to finish:

```cs
        if (EnableEvents)
        {
            RenderFinished.Invoke(Plot, LastRender);

            if (LastRender.SizeChanged)
            {
                SizeChanged.Invoke(Plot, LastRender);
            }
            if (LastRender.AxisLimitsChanged)
            {
                AxisLimitsChanged.Invoke(Plot, LastRender);
            }
        }

        // TODO: event for when layout changes
        RenderCount += 1;
        IsRendering = false;
```
We could simply move the RenderCount and IsRendering = false above the firing of events. This seems appropriate given the rendering has by all accounts finished (as per the first event invoked!) It means nothing stops a new render (even my original code) because the previous render has actually finished before the events fire. 

That is a simple change (basically revoking my mess and moving two lines of code above the events). 

I'll actually commit my version that implemented the Semaphore (mainly for your interest/thoughts). It might help if there are other scenarios that could trigger simultaneous calls to render that aren't locked on the same thread? But I dont know if that's even likely, so I will follow that up with a commit that revokes it. 

### Pre-Rendering Events
I'm happy with the solution of using the RenderStarting event. I was able to achieve the same real-time updates of title and axis labels with the following as the event handler:
```cs
        void renderstarted(object sender, RenderPack e)
        {
            var newLimits = e.Plot.Axes.GetLimits();
            var oldLimits = e.Plot.LastRender.AxisLimits;

            if (Math.Abs(newLimits.Right - newLimits.Left) < Math.Abs(oldLimits.Right - oldLimits.Left))
            {
                WpfPlot1.Plot.Title($"You zoomed In on X");
            }
            else if (Math.Abs(newLimits.Right - newLimits.Left) > Math.Abs(oldLimits.Right - oldLimits.Left))
            {
                WpfPlot1.Plot.Title($"You zoomed Out on X");
            }
            else
            {
                WpfPlot1.Plot.Title($"You didn't Zoom X");
            }
            WpfPlot1.Plot.YLabel($"New Y limits: {newLimits.Bottom:F1}, {newLimits.Top:F1}");
            WpfPlot1.Plot.XLabel($"New X limits: {newLimits.Left:F1}, {newLimits.Right:F1}");
        }
```
Interestingly, this was already being fired by a RenderAction, while the RenderFinished event is simply fired within the Render method. I've reverted my event in the latest commit (which will be pushed when I finish this epic comment!)

### Wrapping Up
I think this PR and associated issues turned into a bit of a storm in a teacup. Ultimately, very little has changed, except that I discovered how to access the changed axis extents before rendering is complete. I guess my specific named event would have been easier to find for someone wanting to do something similar.

It seems to me that three minor issues remain:

1. It should be posisble to update titles and legends in the AxisLimitsChanged events - the render has, by all accounts, actually finished by the time this event fires, so the user should be able to make a change to the plot and have it re-render the new change (I guess there is some risk of a rendering loop if the user puts bad code into the event handler).
2. Calling of Plot.Render() was probably misguided on my part, but it is accessible to a user, and the doc string implies that it would force any new changes to be rendered. Maybe the method needs some form of protection to make it unavailable outside of ScottPlot? or at least some more detailed guidance in the docstring?
3. If I understand better now (not gauranteed) the better way for a user to force a redrawing of the plot is to call Refresh() on the control. This also does not seem to update the axis labels or title in my AxisLimitsChanged event handler. I don't need this now that I am using the RenderStarting event, but I am keen to learn what's happening.

Thanks!

swharden:
> We could simply move the RenderCount and IsRendering = false above the firing of events. 

Simple, but elegant! I like it!

> I think this PR and associated issues turned into a bit of a storm in a teacup

I never heard that phrase before, I love it 😆 

> It should be possible to update titles and legends in the AxisLimitsChanged events - the render has, by all accounts, actually finished by the time this event fires, so the user should be able to make a change to the plot and have it re-render the new change (I guess there is some risk of a rendering loop if the user puts bad code into the event handler).

Actually I don't think this is the case. Consider this sequence of events that happens in the render cycle:
* The title string is measured to determine how much space it needs
* The data area is adjusted to make room for the title
* Plottables are rendered in the data area

It's a bit of a chicken-and-egg problem, but I think the title has to be known before the data can be rendered.

An alternative solution is to create a custom plot type that renders text outside the data area. This way it's always rendered near the end of the render cycle, so it can be aware of the pixel/axis dimensions. This may be a really useful plot type to have, and may have shared functionality with new Annotation plot type that was recently merged #3510 but isn't on NuGet yet. Adding a flag to let annotations render outside the data area would be great! Maybe it already works if you give it negative padding? This is worth looking into...

> Calling of Plot.Render() was probably misguided on my part, but it is accessible to a user, and the doc string implies that it would force any new changes to be rendered. Maybe the method needs some form of protection to make it unavailable outside of ScottPlot? or at least some more detailed guidance in the docstring?

I'll look into this further after posting this comment.

> If I understand better now (not guaranteed) the better way for a user to force a redrawing of the plot is to call Refresh() on the control. This also does not seem to update the axis labels or title in my AxisLimitsChanged event handler. I don't need this now that I am using the RenderStarting event, but I am keen to learn what's happening.

I think leaning on the Annotation plot type may be the solution here! Use the RenderStarting event to change the Annotation's text.

swharden:
> Calling of Plot.Render() was probably misguided on my part, but it is accessible to a user

It turns out that method was only used by the test system to make sure nothing crashes during a render. It wasn't intended to be called by end users. I'm marking the method obsolete now 😈

swharden:
Okay I'm going to merge this PR because it's in a good place... and I'll open a new issue to track progress toward adding the feature where a title or annotation may change its text inside the render loop.

FYI this is the code I'm using to demonstrate the issue, and the rendered scale is always 1 frame behind

```cs
using ScottPlot;
using ScottPlot.WinForms;
using System.Windows.Forms;

namespace Sandbox.WinFormsFramework;

public partial class Form1 : Form
{
    public Form1()
    {
        InitializeComponent();
        FormsPlot formsPlot1 = new() { Dock = DockStyle.Fill };
        Controls.Add(formsPlot1);

        formsPlot1.Plot.Add.Signal(Generate.Sin());
        formsPlot1.Plot.Add.Signal(Generate.Cos());

        var an = formsPlot1.Plot.Add.Annotation("test", Alignment.UpperCenter);
        formsPlot1.Plot.RenderManager.AxisLimitsChanged += (s, e) =>
        {
            an.Label.Text = e.AxisLimits.ToString(2);
        };
    }
}
```

swharden:
Hey, I think this works!

```cs
using ScottPlot;
using ScottPlot.WinForms;
using System.Windows.Forms;

namespace Sandbox.WinFormsFramework;

public partial class Form1 : Form
{
    public Form1()
    {
        InitializeComponent();
        FormsPlot formsPlot1 = new() { Dock = DockStyle.Fill };
        Controls.Add(formsPlot1);

        formsPlot1.Plot.Add.Signal(Generate.Sin());
        formsPlot1.Plot.Add.Signal(Generate.Cos());

        var an = formsPlot1.Plot.Add.Annotation("test", Alignment.UpperCenter);

        formsPlot1.Plot.RenderManager.RenderStarting += (object sender, RenderPack rp) =>
        {
            AxisLimits thisRenderLimits = rp.Plot.Axes.GetLimits();
            AxisLimits lastRenderLimits = rp.Plot.LastRender.AxisLimits;

            if (thisRenderLimits == lastRenderLimits)
            {
                // test this by resizing the window
                an.Label.Text = "limits unchanged";
            }
            else
            {
                an.Label.Text = thisRenderLimits.ToString(2);
            }
        };
    }
}
```

Issue #3523, opened 2024-03-23
Prevent too many calls to SkiaSharp's Canvas.Restore()


BrianAtZetica:
**Suggestion:** 

In SP5, the RenderManager calls SkiaSharp's Canvas.Save() before every RenderAction and Canvas.Restore() after every RenderAction.
However, it also calls Canvas.Restore a second time within the DisableClipping() method

```cs
        foreach (IRenderAction action in RenderActions)
        {
            sw.Restart();
            rp.Canvas.Save();
            action.Render(rp);
            rp.Canvas.Restore();
            rp.DisableClipping();
            actionTimes.Add((action.ToString() ?? string.Empty, sw.Elapsed));
        }
```

According to SkiaSharp method docstring, Restore should not be called more often than save. 

Suggestion is to use Canvas.SaveCount to prevent excessive callas to Canvas.Restore.

swharden:
Related? https://github.com/ScottPlot/ScottPlot/pull/3517

BrianAtZetica:
Note to Self:

Canvas.SaveCount never seems to drop below one, even if Restore is called too many times. I assume something under-the-hood is preventing this.

SkiaSharp.SKCanvas has this remark:

```CS
    // Remarks:
    //     This will equal the number of SkiaSharp.SKCanvas.Save calls minus SkiaSharp.SKCanvas.Restore
    //     calls + 1. The save count on a new canvas is 1.
   public int SaveCount => SkiaApi.sk_canvas_get_save_count(Handle);
```

Note also the following RenderActions that may need checking: 

- [ ] RenderDataBackground performs some clipping and calls DisableClipping without a canvas.save (so is unsafe). Should replace DisableClipping with an explicit save and restore call.
- [ ] RenderPlottables calls a save, and finishes with a call to DisableClipping (so is safe). But it doesn't always do a clip. DisableClipping should be replaced with an explicit call to canvas.Restore.
- [ ] HorizontalLine.RenderLast() callas DisableClipping without a canvas.save (so is unsafe).
- [ ] VerticalLine.RenderLast() callas DisableClipping without a canvas.save (so is unsafe).

StandardZoomRectangle.Render() is an example that looks most correct (save, clip, restore)

swharden:
resolved in #3527

PR #3527, opened 2024-03-23
Render Manger: wrap clipping functionality in an abstraction

BrianAtZetica:
This may be a risky PR to adopt - I never observed any issues, and couldn't identify any changes resulting form my code fixes. 
Nevertheless, the changes here address #3523, makign sure any call to cavas.Save() is ballanced by a call to canvas.Restore() and vice versa.

cavas.SaveCount seems to be limited to a minimum value of 1, making it difficult to recognise when there is an imbalance. 
I was able to verify that the code before was unballanced. To do this I inserted an additional canvas.Save() before the render actions and an additional canvas.Restore() after the render actions. 

```cs
        canvas.Save()
        Stopwatch sw = new();
        foreach (IRenderAction action in RenderActions)
        {
            sw.Restart();
            rp.Canvas.Save();
            action.Render(rp);
            rp.Canvas.Restore();
            actionTimes.Add((action.ToString() ?? string.Empty, sw.Elapsed));
        }
        canvas.Restore();
```
(note this was a temporary change not committed to code).

This way I could track the cavas.SaveCount after each render action, which should have started and ended on a value of 2.
Before my code changes, the SaveCount dropped to one during the RenderDataBackground action.
After the changes, it remains at 2 for all actions. 

I ran all tests, and didn't spot anythign odd on any of the png outputs. 
See what you think. I'll understnad if changes seem too risky/uncertain.


swharden:
Thanks for looking into this @BrianAtZetica! I think I'll adapt this PR and implement a more heavy-handed approach. I recently added source code parsing capabilities to the test system, so I'll scan the code base and fail the build if anyone calls `canvas.Save()` directly 😈

Instead, I'll add a `Clipping` class to the `RenderPack` which manages save/restore with a counter to make sure we never overdo it.

I'll modify this PR now to add that functionality and merge it in later today if it works as expected 👍

swharden:
oof... I accidentally "fixed" this code but committed to the wrong branch 😩 I'm sorry! I'll reverse these changes in #3525 and make the commits on this branch instead.

swharden:
I think things are good now! I'm going to merge this PR, but I'd appreciate it if you could glance over it and make sure you like what you see! We can refine it as needed in future PRs 👍

BrianAtZetica:
Thanks @swharden (and @chjrom for #3528).
I was punching way above my weight to even be poking around in this area of the code. 
Satisfied that my effort, if only raised awareness of a posisble issue. Don't really need any attribution for the change log.

PR #3528, opened 2024-03-23
Prevent too many calls to Canvas.Restore()

### LINKED ISSUE #3523: Prevent too many calls to SkiaSharp's Canvas.Restore()

chjrom:
Changes due to comments in https://github.com/ScottPlot/ScottPlot/issues/3523#issuecomment-2016237841.

Added the  `RenderPack.PushClip` method that returns an` IDisposable` so that the caller is responsible for disable clipping.

The `RenderPack.ClipToDataArea` and `RenderPack.DisableClipping` methods are marked as obsolete.

Resolve #3523.
Same as #3527.

### DIFF
--- a/src/ScottPlot5/ScottPlot5/Control/StandardZoomRectangle.cs
+++ b/src/ScottPlot5/ScottPlot5/Control/StandardZoomRectangle.cs
@@ -72,8 +72,7 @@ public class StandardZoomRectangle : IZoomRectangle
 
         SKRect rect = new(MouseDown.X, MouseDown.Y, MouseUp.X, MouseUp.Y);
 
-        canvas.Save();
-        canvas.ClipRect(dataRect.ToSKRect());
+        using RenderPack.RestoreState _ = rp.PushClipToDataArea();
 
         if (HorizontalSpan)
         {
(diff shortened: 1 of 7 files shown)


swharden:
Hi @chjrom, thanks so much for this PR! It looks like this issue was resolved by #3527. I took some of the ideas in this PR and implemented them over there. I added your name and this PR in the changelog because you both fixed the problem even though only one of those branches was merged 🙃 

I'd love it if you could look over my implementation and let me know if you have ideas for improvements! We can always refine this system in future PRs, and I'd be happy to review and merge them in good time 🚀 

Thanks again!

Issue #3229, opened 2024-01-18
SP5: Signal is slow in low density mode


swharden:
This affects signal plots after they zoom in enough to allow visualizing individual points

I suspect all points are being rendered rather than just those in the view

Thanks Marven from the Discord (#1966) for reporting this

EDIT: it's reported to happen when working with `float[]` but not `double[]` ... need to benchmark #2677 this

minjjKang:
Hello.

I'm using in ScottPlot5 5.0.14-beta, and i have same problem.

I proceeded Benchmark for all SignalSources, I got the following result.

  | FPS |  
-- | -- | --
  | Low Density | High Density
SignalSourceDouble (double[]) | 700 | 65
SignalSourceGenericList (float[]) | 26 | 39
SignalSourceGenericArray (List<double>) | 25 | 39

For other sources except SignalSourceDouble, it is basically slow & much slow in LowDensity.

I think this is caused by GenericToDouble and GetYs.

RenderHighDensity -> GetPixelColumn, only a specific index of Ys was converted to GenericToDouble,
so there was only a small speed decrease,

RenderLowDensity -> GetYs : it converts all indexes.

To solve this problem,

using Enumerator for not convert Ys values that are not used in RenderLowDensity.
(In this case, the slowdown in High Density remains the same )

or

When creating SignalSource, all values must be converted to double[].

Thanks.

swharden:
Hi @minjjKang, thanks for adding context! 

The library has undergone a lot of evolution since this issue was first opened so consider benchmarking it now. ScottPlot 5 is still only a few months into release and I'm focused on major features and bugfixes over performance optimization, but if you or someone else from the community is able to narrow-down the issue and open a PR, I would be happy to review it!

PR #3530, opened 2024-03-23
improve performance of Signal.RenderLowDensity

minjjKang:
Hi Scott,

This is PR to solve the issue of #3229 

# Issue
at the Signal.LowRenderDensity,
Very slow on **SignalSourceDouble and SignalSourceGenericArray** with large size data.

Because that all of SignalSource are uses GetYs() from RenderLowDensity, and it copy all the data,
and that two sources convert all data by **NumericConversion.GenericToDoubleArray**, it reduced performance.

# Solution
I refer to the TODO of ISignalSource that you written,
I added IEnumerable<double> GetYs function, have arguments of index range.
(I used IEnumerable instead of IReadOnlyList,
You just decide what you think is better.)

And, SignalSource used it in the following way:
```cs
public IEnumerable<double> GetYs(int i1, int i2)
{
    for (int i = i1; i <= i2; i++)
    {
        yield return NumericConversion.GenericToDouble(ref Ys[i]);
    }
}
```

Signal.RenderLowDensity changed it to have Pixel:
```cs
IEnumerable<double> Ys = Data.GetYs(i1, i2);
int i = i1;
foreach (double yValue in Ys)
{
    float x = Axes.GetPixelX(Data.GetX(i));
    float y = Axes.GetPixelY(yValue + Data.YOffset);
    Pixel px = new(x, y);
    points.Add(px);
    i++;
}
```

I checked that the performance imporved from LowDensity with like this:

  | RenderLowDensity avg FPS| |
-- | -- | --
Source  | Before | After
SignalSource | 242 | 243
SignalSourceDouble | 3 | 248
SignalSourceGenericArray<double> | 3 | 246
(50M 0-500 random data, display only 50 data)

 Thanks!

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,7 @@ _Not yet on NuGet..._
 * Ticks: Added `MinimumTickSpacing`, `TickDensity`, and `TargetTickCount` properties to the automatic tick generator (see Cookbook)
 * Avalonia: Fixed transparent background issue introduced in the previous version (#3502, #3516) @chjrom @MrOldOwl @kebox7
 * Rendering: Improved canvas state management to prevent duplicate restoration calls (#3527, #3523, #3528) @BrianAtZetica @chjrom
+* Signal: Improved performance of large signal plots when zoomed in (#3530, #3229) @minjjKang 
 
 ## ScottPlot 4.1.73
 _Not yet on NuGet..._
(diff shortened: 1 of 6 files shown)


swharden:
Thank you for this fantastic improvement @minjjKang!

Your `IEnumerable` solution was elegant, but I extended the `ISignalSource` to add a `GetY()` which allows me to iterate across a range using a `for` loop which I think is a little more predictable. I didn't benchmark this solution though, so it may be a good idea to run your existing benchmark test one more time to ensure the performance is still high as expected.

Let me know if you have any hesitations about my solution! We can revert to using the range if it's better, and track progress in a new PR 👍

Issue #3520, opened 2024-03-21
SP5 the behavior of Bar.Label is weird when BarPlot.Horizontal is true


yui1227:
**Issue:** the behavior of Bar.Label is weird when BarPlot.Horizontal is true. the position of value is wrong.

**ScottPlot Version:** 5.0.22

**Code Sample:** 

```cs
using ScottPlot;

var plot = new Plot();

double[] values = { 1, 2, 3, 4, 5 };
var barPlot = plot.Add.Bars(values);

foreach (var bar in barPlot.Bars)
{
    bar.Label = $"{bar.Value}";
}

plot.Save("vert.png", 500, 500);
barPlot.Horizontal = true;
plot.Save("hori.png", 500, 500);
```
![vert](https://github.com/ScottPlot/ScottPlot/assets/77109827/0fd9292b-5a74-40dd-a2cd-b165b53d49f3)
![hori](https://github.com/ScottPlot/ScottPlot/assets/77109827/19232274-cf58-4aae-b8ae-d5bb49b8d810)

swharden:
This is the code that needs to be modified:

https://github.com/ScottPlot/ScottPlot/blob/df89ed6e9286c4ccb4ef751ced541a4002a0f67f/src/ScottPlot5/ScottPlot5/Primitives/Bar.cs#L107-L108

An `if` condition should be add to place the label on the correct side given the value of `this.Orientation`

... also maybe logic should be added to support "negative" vs. "positive" bars too

swharden:
Thanks for reporting this @yui1227! I'll drop this information here in case someone wants to try adding this feature before I'm able to get to it. If someone creates a PR, I'm happy to review it in good time! 🚀 

This is the code that needs to be modified:

https://github.com/ScottPlot/ScottPlot/blob/df89ed6e9286c4ccb4ef751ced541a4002a0f67f/src/ScottPlot5/ScottPlot5/Primitives/Bar.cs#L107-L108

An `if` condition should be add to place the label on the correct side given the value of `this.Orientation`

... also maybe logic should be added to support "negative" vs. "positive" bars too

aespitia:
PR submitted for this

https://github.com/ScottPlot/ScottPlot/pull/3546

PR #3546, opened 2024-03-25
feat: bar label position for horizontal bar charts

aespitia:
This PR is regarding issue #3520, the label placement for horizontal bar charts was always on top, this moves labels for horizontal bar charts to either left or right, depending on negative or positive value of the bar.

Included a recipe called Bar with Value Labels (horizontal), looks like this attached image.

Some work may need to be done somewhere around adjusting the size of the container graph if the labels go outside of the drawing area?

image


swharden:
Thanks @aespitia, this looks great!

> Some work may need to be done somewhere around adjusting the size of the container graph if the labels go outside of the drawing area?

```cs
myPlot.Axes.Margins(bottom: 0, top: .2); // before
myPlot.Axes.Margins(left: .2, right: .2); // after
```

image

Issue #3374, opened 2024-02-16
LockedVertical right axis or secondary axis doesn't work


Boris-Lebeau:
Hi Scott, thanks a lot for your great and useful work.

I have noticed a bug ever reported on Locked axis (horizontal or vertical) on SP5. I think there is another consequence of this bug : a problem when locking the right vertical axis (or locking a secondary axis) :

```cs
var sig1 = myPlot.Add.Signal(Generate.Sin(mult: 1));
var sig2 = myPlot.Add.Signal(Generate.Cos(mult: 100));

sig1.Axes.YAxis = myPlot.Axes.Left;
sig2.Axes.YAxis = myPlot.Axes.Right;

// No problem if sideAxis = PlotA.Plot.Axes.Left;
var sideAxis = myPlot.Axes.Right;
var rule = new ScottPlot.AxisRules.LockedVertical(sideAxis);
myPlot.Axes.Rules.Add(rule);
```

The use of Render() not seems to solve the problem in this case. 

Thanks for you help.
Boris

BendRocks:
Yes, I see the same on this in Avalonia. ScottPlot 5.0.21 Demo Axis Rules section works as expected under WinForms, but if I move that same code into an Avalonia app, the axis doesn't lock. @Boris-Lebeau, are you using Winforms or other?

PS. It looks like 5.0.22 which came out last week didn't publish (it's not on Nuget) and the changelog indicates _Warning: NuGet package date is missing._

BendRocks:
I don't know if fixed in 5.0.22 or not.

For anyone looking to repro, you can take the Ava sandbox example and do it there. Every time the LockedVertical.Apply() is called, the axis limits are updated, but at some point after the YAxis range getting set in Apply(), they get changed by the mouse drag. And then for the next time Apply() is called, the wrong values are applied. Seems like the LockedVerticalAxis.Apply() needs to be called earlier in the pipeline.

```
public partial class MainView : UserControl
{
    public MainView()
    {
        InitializeComponent();

        AvaPlot.Plot.Add.Signal(Generate.Sin());
        AvaPlot.Plot.Add.Signal(Generate.Cos());
        AvaPlot.Refresh();

        AvaPlot.Plot.Axes.Rules.Clear();

        ScottPlot.AxisRules.LockedVertical leftAxisRule = new ScottPlot.AxisRules.LockedVertical(AvaPlot.Plot.Axes.Left);
[...]

PR #3547, opened 2024-03-26
New Axis rules

BrianAtZetica:
I required a graph to snap to integer number of tick intervals. The reason for this was so that a user of my graph could make multiple images from differnet data using reproducible axis limits (but without having to type in axis limits). 

For this I created two new Axis rules, SnapToTicksX and SnapToTicksY.
These work as expected with two minor side-effects:

- every now and then, the tick interval updates as a result of the snapping, which has the effect of the axis no longer being snapped to a major interval. An easy fix for the user is to rock the zoom in and out once (or out and in). 
- Panning no longer looks smooth. 

In the process I added some other new rules.
Locked Left/Right/Top Bottom coordinates (useful for example if you want to always include the origin in a plot, but still allow free zooming).
LockedCenterX / Y (useful if you want a graph to be always centred on the 0,0, orgin, but can also be used with a custom origin).

The interaction of the rules was interesting. I've set the new rules to check what other rules are applied to the same axis. LockedHorizontal / LockedVertical rules always take priority. If any two coordinates are locked (left and centrte for example) it is treated as a locked axis and wont pan or zoom. 

I have tested that it works on invertedd Y axes, and on dual Y axes (in the process I have addressed #3374 )

To demonstrate teh axis rules (including inverted axes) I created a WPF app. The code is too long to paste here, so I have committed to the Sandbox WPF app in this PR. 

image


BrianAtZetica:
Spotted an issue with the snapping to ticks rule during an autoscale. When tick interval was laerger than the data extents, the rescale would snap to the same value and the axis would disappear. 
The fix is in three parts:
- No longer applying the rule if the new limits woudl be the same value
- regenerating the ticks before calculating the snap location.
- preferring a snap direction that increases the extents except where this would override the user intent.

swharden:
Thanks @BrianAtZetica, this looks great! 

> To demonstrate the axis rules (including inverted axes) I created a WPF app. The code is too long to paste here, so I have committed to the Sandbox WPF app in this PR.

I gave the WPF app a spin and it's cool to see some of these axis rule ideas that I never thought of before.

I'm merging in all the rules, but I'll revert to the original WPF sandbox app to keep things simple. We could translate that app you made to a WinForms app and bundle it with the demo app if it would be useful to demonstrate these for new users.

Issue #3095, opened 2023-12-15
SP5: Add lock and unlock render actions to improve support for multi-threading environments


swharden:
Following discussion with @bclehmann in the Discord (#1966), let's add some empty actions that get invoked at the start and end of each render which advanced users could add their own code into to wait for an optical lock state to be achieved before rendering and stay locked until after rendering.

This functionality should go in the RenderManager class https://github.com/ScottPlot/ScottPlot/blob/main/src/ScottPlot5/ScottPlot5/Rendering/RenderManager.cs

bclehmann:
As a basic summary of our discussion:

For users on multithreaded platforms the plot may be mutated and be rendered on separate threads, which can cause unexpected issues. A pre-render action could be a mechanism to ensure plot mutation happens right before rendering on the same thread, eliminating the possibility for interleaving by essentially eliminating multithreading. Note that if someone puts an expensive pre-render action it may delay rendering by a lot, perhaps we should advise them to trigger a render once they've registered an action to avoid paying this cost at an inopportune time.

For users who specifically want multithreading, this gives them an opportunity to get their ducks in a row before rendering starts. For example, if they use locks to synchronize their threads they can wait on their locks in this action. This ensures that once the later actions start running there is no possibility for interleaving.

swharden:
Done! Here's an example use:

```cs
formsPlot1.Plot.RenderManager.PreRenderLock += (s, e) =>
{
    // code here must complete before the frame renders
    System.Threading.Thread.Sleep(100);
};
```

PR #3559, opened 2024-03-29
Control: use locks to prevent concurrent processing of UI events

### LINKED ISSUE #3557: Avalonia: graphical artifacts during mouse interactions
**Issue:**
I'm using a simple static plot with 2 vertical axes and 6 signals. Everything is working as expected... expect when the mouse is panning, we can see some flashing artifacts (as if the series are drawn multiple times). 
I've tried to simplify the plot as much as possible (keeping one axis, one signal) but it's always there.

https://github.com/ScottPlot/ScottPlot/assets/150119089/8ec9c78d-7135-44c4-bf28-6bb793bec850

Am I doing something wrong? Or... could it be an Avalonia issue? 
The WinForms demo are working well on the same system.

**ScottPlot Version:**
ScottPlot.Avalonia 5.0.23
Avalonia 11.0.10

chjrom:
The user control that hosts the plot can have a separate thread for rendering. For example, in AvaloniaUI.
Event processing and plot rendering are not synchronized. This results in artifacts while rendering.

This PR added a sync object to `Plot` to prevent multithreaded access to the plot state. 

Fix #3557.


swharden:
Hi @chjrom, thanks for this fix! It will be in the next release 🚀 

I also like exposing the `Sync` object for external locking as an alternative to using the `PreRenderLock` event handler described in #3095

Issue #3543, opened 2024-03-25
Controls: `InvalidCastException` when starting mouse interaction in the Title area


pkstrsk:
**Issue:** System.InvalidCastException: 'Unable to cast object of type 'ScottPlot.Panels.TitlePanel' to type 'ScottPlot.IAxis'.'

**ScottPlot Version:** 5.0.23 (Avalonia)

**Code Sample:** Default plot settings, empty app with any plot

Plot.Interactions set to default. Setting them to only zoom or pan also introduces the issue.

I have tested on 5.0.21 and the issue is not there.

pkstrsk:
Just tested on the Demo 5.0.22 and I can get it to crash as well, just by click and drag from the plot title.
![image](https://github.com/ScottPlot/ScottPlot/assets/14115952/05d345ab-e944-49da-a611-49724fc4e5e1)

Full exception
```
See the end of this message for details on invoking 
just-in-time (JIT) debugging instead of this dialog box.

************** Exception Text **************
System.InvalidCastException: Unable to cast object of type 'ScottPlot.Panels.TitlePanel' to type 'ScottPlot.IAxis'.
   at ScottPlot.Plot.GetAxis(Pixel pixel) in /_/src/ScottPlot5/ScottPlot5/Plot.cs:line 171
   at ScottPlot.Control.StandardActions.DragPan(IPlotControl control, MouseDrag drag, LockedAxes locked) in /_/src/ScottPlot5/ScottPlot5/Control/StandardActions.cs:line 84
   at ScottPlot.Control.Interaction.MouseDrag(Pixel from, Pixel to, MouseButton button, IEnumerable`1 keys, MultiAxisLimitManager start) in /_/src/ScottPlot5/ScottPlot5/Control/Interaction.cs:line 106
   at ScottPlot.Control.Interaction.OnMouseMove(Pixel newPosition) in /_/src/ScottPlot5/ScottPlot5/Control/Interaction.cs:line 78
   at ScottPlot.WinForms.FormsPlotBase.SKElement_MouseMove(Object sender, MouseEventArgs e) in /_/src/ScottPlot5/ScottPlot5 Controls/ScottPlot.WinForms/FormsPlotBase.cs:line 83
   at System.Windows.Forms.Control.WndProc(Message& m)
   at System.Windows.Forms.Control.ControlNativeWindow.WndProc(Message& m)
[...]

PR #3571, opened 2024-04-01
Controls: fix `InvalidCastException` when starting mouse interaction in the Title area

bwedding:
Fixes [#3543](https://github.com/ScottPlot/ScottPlot/issues/3543)
The exception was caused because the logic in the ternary operator was reversed. When axesOnly was true, it was returning all panels and vice versa:

```cs
        // Reverse here so the "highest" axis is returned in the case some overlap.
        var panels = axesOnly
            ? Axes.GetPanels().Reverse().OfType<IAxis>()
            : Axes.GetPanels().Reverse();
 ```
Added check to ensure type is IAxis before returning it rather than casting the wrong type, which was what threw the exception.

```cs
    public IAxis? GetAxis(Pixel pixel)
    {
        IPanel? panel = GetPanel(pixel, axesOnly: true);
        return panel is IAxis axis ? axis : null;
    }

```


swharden:
Great catch @bwedding, thanks so much! 😄 🚀

Issue #3538, opened 2024-03-24
Label: draws a border even if border width is 0


swharden:
I think the solution is to draw the rectangle using one of ScottPlot's Drawing methods instead of a SkiaSharp method directly

PR #3572, opened 2024-04-01
Label: Don't draw border when line width is 0

bwedding:
For issue #3538 
In SkiaSharp, setting the StrokeWidth to 0 does not guarantee that no border will be drawn. A StrokeWidth of 0 is interpreted as the thinnest line that can be drawn on the device, which is not the same as not drawing a line at all.

So I just wrapped the code to draw the border with a conditional to only paint it if borderwidth is greater than 0. This works fine on the recipe code when setting BorderWidth = 0;

image

[Issue 3538](https://github.com/ScottPlot/ScottPlot/issues/3538)

```cs
[ScottPlot.Plot myPlot = new();
if (BorderWidth > 0)
{
    ApplyBorderPaint(paint);
    canvas.DrawRect(backgroundRect.ToSKRect(), paint);
}
```


swharden:
Another great fix @bwedding, I'm happy to merge this in and it will be in the next release! 🚀

PR #3577, opened 2024-04-03
Scatter: Allow `NaN` to draw noncontinuous lines

drolevar:
Enable plotting noncontinuous lines. For example, when a DataLogger has several NaN values, those are not plotted and just skipped.


drolevar:
Also, using ICollection in Drawing.DrawLines enables to use the Count property instead of the Count() LINQ extension, which prevents a double enumeration.

swharden:
Great new feature and optimizations @drolevar, thanks so much!

I made some revisions to ensure that the automatic axis limit detection respects the gaps in the data. I think it's all working as expected, but I welcome feedback! We can refine this if needed in another PR.

I'm excited that this will be in the next release! 🚀

PR #3590, opened 2024-04-04
SP5: Add Sequential property for OHLCPlot

oktrue:
Add Sequential property in the same way as in the CandlestickPlot


swharden:
Thanks @oktrue, this looks great!

PR #3592, opened 2024-04-04
SquareZoomOut Axis Rule: add support for inverted axes

VisMotrix:
…om: 1.5, top: -1.5);
This PR fixes the axis range setting if you use the SquareZoomOut  and AutoScaler.InvertedX / Y

This code did not "keep" the SetLimit / AutoScaler.InvertedY setting cause SquareZoomOut rule just sets YAxis.Range.Set(yMin, yMax); we need to distinguish if the axis got inverted.
```cs
myPlot.Add.Signal(Generate.Sin());
myPlot.Axes.SetLimitsY(1.5, -1.5);
myPlot.Axes.SquareUnits();
```


swharden:
Hi @VisMotrix, thanks for this fix! I extended this PR to make some additional improvements to the axis system. I'm merging this in now so the square zoom rules work support inverted axes, but other axis rules still don't support them so I opened #3597 to track progress toward that fix in the future.

Issue #3570, opened 2024-03-31
Scatter plot: rendering fails on inverted horizontal axis


swharden:
Recreating:
```cs
formsPlot1.Plot.Add.Signal(Generate.Sin());
formsPlot1.Plot.Axes.SetLimitsX(0, -100);
```

This line throws:
https://github.com/ScottPlot/ScottPlot/blob/b12cb0c1ae5ddd76fe4e7a686fac79cc57519fbb/src/ScottPlot5/ScottPlot5/Plottables/Signal.cs#L111

```
System.ArgumentOutOfRangeException: 'Index was out of range. 
Must be non-negative and less than the size of the collection. (Parameter 'index')'
```

bwedding:
I took a quick look at this. The reason it throws is obviously because there are no points to move to. If you just handle that and return from RenderLowDensity(), it won't throw, but it won't plot anything either.

The reason there are no points to render is much deeper. Signal, by virtue of how it handles the X axis with a period and not actual point values, doesn't easily invert its x-axis. **Perhaps it shouldn't either. It is a signal. Inverting it means plotting from the future to the past, which doesn't make sense.**

If you do want to handle it, it's going to take a major refactor, from what I can see.

All this code here in SignalBase.cs will need to change because it doesn't expect negative values. 

```
public int MinimumIndex { get; set; } = 0;
public int MaximumIndex { get; set; } = int.MaxValue;

public int MinRenderIndex => Math.Max(0, MinimumIndex);
public int MaxRenderIndex => Math.Min(Length - 1, MaximumIndex);

public double XOffset { get; set; }
public double YOffset { get; set; }

public int GetIndex(double x, bool visibleDataOnly)
{
    int i = (int)((x - XOffset) / Period);

    if (visibleDataOnly)
    {
        i = Math.Max(i, MinRenderIndex);
        i = Math.Min(i, MaxRenderIndex);
    }

    return i;
}

```

PR #3594, opened 2024-04-04
Signal: support inverted axes

Excustic:
Fixed issue  #3570 - enabling the option to invert axes of Signal plots.

- Added direction constants for inverting the axes in Signal plots. 
- Added GetVisibleYRange private method in Signal.
- Refactored variable names inside RenderLowDensity method of Signal.

image

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,7 @@
 * Axes: Added optional arguments to `Plot.Axes.AutoScale()` to add support for nonstandard axes (#3592)
 * Axis Rules: Improved `Plot.Axes.SquareUnits()` to support inverted axes (#3592) @VisMotrix
 * WinForms: Improve `FormsPlot` disposal so the control displays properly when re-launched (#3593, #3589) @bwedding @Kruno313
+* Signal: Added support for inverted horizontal axes (#3594) @Excustic
 
 ## ScottPlot 5.0.23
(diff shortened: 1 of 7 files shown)


swharden:
Hi @Excustic, thanks for working on this! I'm about out of time right now so I can't take a closer look, but I can tell something's not exactly right because I'd expect the sine wave to be displayed in the X range between 0 and +50.

```cs
// this should display data from X = 0 to 50
WpfPlot1.Plot.Add.Signal(Generate.Sin());

// this inverts the X axis (so it goes from 50 down to 0)
WpfPlot1.Plot.Axes.SetLimitsX(50, 0);
```

image

If you're able to figure it out before I get to it again, feel free to update this PR! Otherwise I'll pick this up where I left off next time I'm free. Thanks again for helping out on this one! It will be great to have inverted axis support for signal plots 🚀

Excustic:
I see, seems like it's a bit more complicated than I thought. It certainly requires refinement and further testing, I'll try to get on it next week and help you with it 👍

Excustic:
I see, seems like it's a bit more complicated than I thought.

swharden:
Figured it out! The code change required seems simple, but wow it took a lot to figure out. Thanks for getting this PR started @Excustic! I'm happy this new feature will be in the next release 🚀 

Note that this doesn't actually resolve #3570 because this PR addresses _signal_ plots but that issue is about _scatter_ plots

Issue #3595, opened 2024-04-05
FunctionPlot: RangeX is not respected (likely because min/max are flipped)


Matthew-Chidlow:
If I pass an `IFunctionSource` into `plot.Add.Function()` which has specified `RangeX` to limit the function to the range, then this is ignored and the function will plot over the whole visible range.

I'm pretty sure its because the Min/Max are backwards here:
https://github.com/ScottPlot/ScottPlot/blob/5c02d82432cebe5018d22ebd8be41738e80f0ebc/src/ScottPlot5/ScottPlot5/Plottables/FunctionPlot.cs#L23-L24
I believe `MinX` should be the `Max(...)` of the two mins, and `MaxX` should be the `Min(...)` of the two maxs, from what I understand.

```cs
ScottPlot.Plot myPlot = new();
myPlot.Axes.SetLimits(-10, 10, -10, 10);
myPlot.Add.Function(new FunctionSource(x => x)
{
    RangeX = new CoordinateRange(-1.0, 1.0)
});
myPlot.SavePng("bug.png", 300, 300);
```

swharden:
Hi @Matthew-Chidlow, thanks for reporting this issue! ... and drilling down to the lines of code that underlies it. I'll take a closer look within the next few days (but am happy to review a PR if someone gets to it before I do).

For added context, the present behavior is that the line is shown over the whole visible X range, but really it should only be drawn in the X range [-1, 1]

```cs
    [Test]
    public void Test_Function_Bounds()
    {
        FunctionSource fs = new(x => x)
        {
            RangeX = new CoordinateRange(-1.0, 1.0)
        };

        Plot myPlot = new();
        myPlot.Axes.SetLimits(-10, 10, -10, 10);
        myPlot.Add.Function(fs);

        myPlot.SaveTestImage();
    }
```

![image](https://github.com/ScottPlot/ScottPlot/assets/4165489/cee19629-55be-4361-9b52-5d3a5224a82b)

swharden:
resolved in #3603

PR #3603, opened 2024-04-05
fix:MinX and MaxX functions swap #3595

Dibyanshuaman:


swharden:
Thanks @Dibyanshuaman, this looks great! 🚀

PR #3609, opened 2024-04-07
SignalXY: add support for markers

Giviruk:
If I understood correctly, then the solution you propose looks like this. But I'm not sure what's in Drawing.DrawMarkers should be passed pixels, but I haven't quite figured out how I can get the points I need.


swharden:
Hi @Giviruk, thanks so much for opening this pull request! I have the code open right now so now that you got the ball rolling, I think I can refine this and merge it in shortly. Thanks again!

swharden:
Looks fantastic! This will be in the next release 🚀 

```cs
var sig = myPlot.Add.SignalXY(xs, ys);
sig.MarkerStyle.Shape = MarkerShape.FilledCircle;
```

image

PR #3612, opened 2024-04-07
AxisLines: Allow them to be excluded from the legend

MCF:
If an AxisLine (horizontal or vertical) has its Text property set it would always be included in any legend - both the automatic or manually built legends. The Text property is also used to display the text along the appropriate axis for the line so it is a valid use case to want the Text set but not add the line to the legend.

This change adds an ExcludeFromLegend property to the AxisLine base class which defaults to false (no change to current default behavior). Setting this property to true keeps the line out of the legend regardless of the Text property being set or not.

A cookbook recipe was added demonstrating how the AxisLines and legend work together.  The recipe code and result is shown below.

```cs
myPlot.Add.Signal(Generate.Sin());

var axLine1 = myPlot.Add.VerticalLine(24);
axLine1.Text = "Line 1";

var axLine2 = myPlot.Add.HorizontalLine(0.75);

var axLine3 = myPlot.Add.VerticalLine(37);
axLine3.Text = "Line 3";
axLine3.ExcludeFromLegend = true;

var axLine4 = myPlot.Add.HorizontalLine(0.25);
axLine4.Text = "Line 4";

var axLine5 = myPlot.Add.HorizontalLine(-.75);
axLine5.Text = "Line 5";
axLine5.ExcludeFromLegend = true;

myPlot.ShowLegend();
```

AxisLineInLegend

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 _Not yet on NuGet..._
 * Function: Improved autoscaling behavior and respect for user-defined horizontal ranges (#3618) @Matthew-Chidlow  
 * Function: Exposed `MinX` and `MaxX` to allow users to restrict display to a horizontal range (#3595, #3603) @Matthew-Chidlow @Dibyanshuaman
+* Axis Lines: Added `ExcludeFromLegend` so text can be added to axis line labels without appearing in the legend (#3612) @MCF
 
 ## ScottPlot 5.0.25
 _Published on [NuGet](https://www.nuget.org/profiles/ScottPlot) on 2024-04-08_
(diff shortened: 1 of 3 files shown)


MCF:
I've realized there is a corollary use case that is not fixed by this change: if the Text/Label is only wanted in the legend and not at the end of the drawn line on the axis.  That is currently impossible to do and still impossible with this change.

A better, general solution to all of this may be:

1. Separate the Label from the Text property.  They can both be set independently.
2. The Text property controls the visibility of what is displayed on the axis.
3. The Label property controls the visibility of any legend item.

This feels more consistent with (all?) other plottables that are added to the legend.  AFAIK it is the Label property that controls legend visibility for all the rest.  I think it would be clearer if the Text property is separate and only controls the along axis annotation.  If you want both, you just set Text and Label to the same thing.  I think it is very unlikely you would want both, but there is not harm in supporting all uses cases.

If you feel this proposed solution is better I'd be happy to close this PR and put a new one together that implements what I've described above.

swharden:
Hi @MCF, this looks great, thanks!

Regarding alternative approaches, I'll add a flag in the `Label` class to allow the legend to exclude it even if text is present.

Issue #3585, opened 2024-04-04
WpfPlot: Need sample app to demonstrate display scaling


MagicFawkes:
**Issue:** 
Mouse position is shiftet if windows 11 screen scale is set to 125% (e.g. I use the application on a notebook), on 100% it is okay and works fine

**ScottPlot Version:** 
ScottPlot.WPF 5.0.23

**Code Sample:** 

```cs
    var position = System.Windows.Input.Mouse.GetPosition(WpfPlot1);
    var coordinates = WpfPlot1.Plot.GetCoordinates(new Pixel(position.X, position.Y));

    double x = coordinates.X;
    double y = coordinates.Y;
```

![AxisLineQuickstart](https://github.com/ScottPlot/ScottPlot/assets/107568517/0d44f29e-9ebf-4e58-bf19-a52bc7b071be)
![windows scale](https://github.com/ScottPlot/ScottPlot/assets/107568517/5392609a-db36-4aee-8c82-9dd1e5dbb9cd)

swharden:
Hi @MagicFawkes, perhaps this will help! Adapted from #3550

```cs
double factor = PresentationSource.FromVisual(this).CompositionTarget.TransformToDevice.M11;
var position = System.Windows.Input.Mouse.GetPosition(WpfPlot1);
double x = position.X * factor;
double y = position.Y * factor;
var coordinates = WpfPlot1.Plot.GetCoordinates(x, y);
```

This seems to relate to #3465 which got closed by accident. A minimal WPF demo should be created to demonstrate WPF-specific functionality (like this)

Issue #3469, opened 2024-03-13
Plot: not all methods or behaviors properly support `ScaleFactor`


swharden:
This issue will list the methods or behaviors which need to be updated to better support custom scale factors (useful for DPI scaling). Listing here will make it easier to fix them later. This topic came-up as part of #3465

If anyone identifies a method which is broken, they can mention it below as a comment and it can be fixed later

Issue #3465, opened 2024-03-13
SP5: Minimal WPF Demo


swharden:
I plan to use the WinForms demo to demonstrate virtually all GUI-specific functionality (mouse interactions, etc), but there are a few WPF-specific topics that may be nice to demonstrate and reference in a small standalone WPF demo: display scaling and mouse tracking.

* @MCF started to add this information to the WPF quickstart page (https://github.com/ScottPlot/ScottPlot.NET/pull/34) but a working code example may be a more useful way to document these behaviors

MCF:
FWIW, there are some bugs in some of the Plot functions that must take into account ScaleFactor.  I have been slowly picking away at a PR to fix them but it is relatively complicated and I've been busy.  It may be ready in the next few days.  But that is why I couldn't fix up the documentation PR - I couldn't describe how to use these functions because they don't work properly.  These bugs will have to be fixed before any demo app can make use of the functions as well.

Regarding the use of ScaleFactor and DisplayScale to deal with high DPI displays: some of the same issues need to dealt with in the WinForms control as well.  For example the WinForms demo app shows all of the plots as being much too small on my laptop (which is set at 200% display scaling by Windows as it is a high DPI display).

swharden:
> there are some bugs in some of the Plot functions that must take into account ScaleFactor

Thanks for reporting this! Maybe having a WPF demo for scale factor will help shine a light on some of those bugs and help us knock them out one by one.

> But that is why I couldn't fix up the documentation PR

Makes sense!

> Regarding the use of ScaleFactor and DisplayScale to deal with high DPI displays: some of the same issues need to dealt with in the WinForms control as well. For example the WinForms demo app shows all of the plots as being much too small on my laptop (which is set at 200% display scaling by Windows as it is a high DPI display).

Good point! Solving these issues will be a win for all the controls (not just the WinForms control)

Also, oof, integration testing using different system DPI scaling settings is so complicated and annoying 🤪

Even if we just make a list of all the broken functions that don't support scaling well, that's a good start! I created #3469 for this purpose

Thanks for all your help on this @MCF! I'd love to get these all fixed up in the next release or two 🚀

Rusty:
I replicated the issue here.
FWIW, my main monitor (win11) has the Scale set to 200%. My other monitors (x2) are at 100%.  
The example is just trying yo create a plot with a cross hair that tracks the mouse.
The app misbehaves on all monitors.

WPF code here->

XAML---
[...]

Issue #3186, opened 2024-01-11
Controls: refactor input binding system


MohammadSadeghSalimian:
## Feature Suggestion

**Feature Description:** To change the mouse and keyboard keys for panning, zooming, etc., for Actions in Plot Viewer in WPF, I had to create a class from the IPLotInteraction interface and change it based on my needs. However, I think in some cases, users need to assign a key combination to these plot actions. For example, I prefer to pan the axis while the space key is being held and dragged with the mouse's middle button. 

**Code example:** If applicable, use this space to demonstrate how you wish the new feature could be used in code.

```cs
public InputBindings Inputs = new InputBindings()
{
    DragPanButton = MouseButton.Middle+KeyBoardKeys.Space.Holding,
    DragZoomRectangleButton = MouseButton.Right,
    DragZoomButton = MouseButton.Middle+KeyBoardKeys.Alt.Holding,
    ClickAutoAxisButton = MouseButton.Middle,
    ClickContextMenuButton = MouseButton.Unknown,
    DoubleClickButton = MouseButton.Left,
    ZoomInWheelDirection = MouseWheelDirection.Up,
    ZoomOutWheelDirection = MouseWheelDirection.Down,
    LockHorizontalAxisKey = Key.Shift,
    LockVerticalAxisKey = Key.Ctrl,
    PanZoomRectangleKey = Key.Alt,
};
```

swharden:
> I prefer to pan the axis while the space key is being held and dragged with the mouse's middle button

This is a great example of a behavior that is a challenge with the current design! I'll leave this issue open and continue to think about how I can streamline control behavior customizations.

I'm thinking that I may have gotten too complex by all those interfaces, binding classes, etc. A simple `ControlBehavior` class may be the way to go, every method could be `virtual` so the user can inherit-and-override to customize specific behaviors without having to re-implement everything.

I'll keep thinking about this topic and prioritize refactoring this system in a future release 👍 

PS: https://scottplot.net/demo/ has updated information about customizing some elements of user inputs

swharden:
Spillover from #3622, when this refactoring is done ensure there is consistency across controls for how to get the mouse position. This is platform-specific when it comes to display scaling and styling, especially for WPF controls.

bclehmann:
I'm not sure if I love the idea of just throwing strings around, stringly-typed fields are typically quite difficult to work with. I think if you want that sort of generality you should make it an interface, where users are required to override all the relevant classes to handle that interface.

[...]

PR #3622, opened 2024-04-08
SP5 / WpfPlot: simplify mouse coordinate with display scaling

KroMignon:
Mouse interaction with `ScottPlot` 5 WPF control is not very user friendly.

There are some internals to know:
- the WPF control is a **Custom Control** to allow customization via a user define style
- the style must contains an item called `PART_SKElement`. For "standard" SkiaSharp it must be a `SKElement`, and for the OpenGL version it must be a `SKGLElement`.
- the plot rendered output will be draw on `PART_SKElement`. So if the style do not contain it, there will be nothing visible for the user.
- as the style can be customized, the origin of `PART_SKElement` may not be the upper left corner of the WPF control

Because this maybe very confusing for the control user, I have updated the SP5 WPF control to:
- use `PART_SKElement` as reference to calculate mouse pixel position on plot rendering
- add helper methods to get mouse position on rendered plot which also take active scaling factor in account

This should partially fix issues #3585, #3622, #3469 and #3465

This methods can than be used in C# code for example to show the crosshair like this in the WPF Sandbox App:

```cs
{
    readonly ScottPlot.Plottables.Crosshair Crosshair;
    public MainWindow()
    {
        InitializeComponent();
        WpfPlot1.Plot.Add.Signal(Generate.Sin());
        Crosshair = WpfPlot1.Plot.Add.Crosshair(0, 0);
        Crosshair.IsVisible = false;
        WpfPlot1.MouseEnter += (o, e) =>
        {
            if (Crosshair?.IsVisible == false)
            {
                Crosshair.IsVisible = true;
                WpfPlot1.Refresh();
            }
        };
        WpfPlot1.MouseLeave += (o, e) =>
        {
            if (Crosshair?.IsVisible == true)
            {
                Crosshair.IsVisible = false;
                WpfPlot1.Refresh();
            }
        };
        WpfPlot1.MouseMove += (o, e) =>
        {
            if (Crosshair is not null)
            {
                var p = WpfPlot1.GetPlotPixelPosition(e);
                var mouseCord = WpfPlot1.Plot.GetCoordinates(p, Crosshair.Axes.XAxis, Crosshair.Axes.YAxis);

                if (mouseCord.AreReal)
                {
                    Crosshair.IsVisible = true;
                    Crosshair.Position = mouseCord;
                    WpfPlot1.Refresh();
                }
                else if (Crosshair.IsVisible)
                {
                    Crosshair.IsVisible = false;
                    WpfPlot1.Refresh();
                }
            }
        };
    }
}
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@ _Not yet on NuGet..._
 * Function: Improved autoscaling behavior and respect for user-defined horizontal ranges (#3618) @Matthew-Chidlow  
 * Function: Exposed `MinX` and `MaxX` to allow users to restrict display to a horizontal range (#3595, #3603) @Matthew-Chidlow @Dibyanshuaman
 * Axis Lines: Added `ExcludeFromLegend` so text can be added to axis line labels without appearing in the legend (#3612) @MCF
+* WPF: Added `GetPlotPixelPosition()` for getting mouse position relative to the figure (#3622) @KroMignon
 
 ## ScottPlot 5.0.25
 _Published on [NuGet](https://www.nuget.org/profiles/ScottPlot) on 2024-04-08_
--- a/src/ScottPlot5/ScottPlot5 Controls/ScottPlot.WPF/WpfPlot.cs
+++ b/src/ScottPlot5/ScottPlot5 Controls/ScottPlot.WPF/WpfPlot.cs
@@ -11,6 +11,7 @@ namespace ScottPlot.WPF
         private const string PART_SKElement = "PART_SKElement";
 
         private SkiaSharp.Views.WPF.SKElement? SKElement;
+        protected override FrameworkElement PlotFrameworkElement => SKElement!;
 
         public override GRContext GRContext => null!;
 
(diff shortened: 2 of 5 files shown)


swharden:
Hi @KroMignon, thanks for this PR! We are very fortunate to have a WPF expert contributing to this project 😄

I'll merge this in now and work on the documentation when I address #3465

There's one issue I'm not very happy about, but I'm not sure if it's a problem that can be solved easily: `GetPlotPixelPosition()` is a special function that only the WPF control has, so the docs on how to determine where the cursor is change depending on what control is being used. One solution could be to add a similar method to all controls... I'll take a quick look if I can do this using generics, and if it gets complicated I'll just merge this in now and pick it up when I start work on #3465.

Thanks again for all your work on this project! 🚀

swharden:
> I'll take a quick look if I can do this using generics, and if it gets complicated I'll just merge this in now

it got complicated 😅 

I still think a generic solution is good, and I'll implement that when I work on #3186

KroMignon:
> it got complicated 😅
> 
> I still think a generic solution is good, and I'll implement that when I work on #3186

To be honest, I am also not really 100% satisfy with the implementation.

I first wanted to update the `IPlotInteraction` to add a method to get last known cursor position but this would not work when interaction are disabled.
Could not estimate if this would be a problem, so I leave it.

I also think about make it available for each backend, but I am not really comfortable with all of them 😞

PR #3625, opened 2024-04-09
Vector Field

bclehmann:
This example is lifted from the 4.1 cookbook:
```cs
var vectors = new List<RootedCoordinateVector>();
double[] xs = Generate.Consecutive(20, 0.5, -5);
double[] ys = Generate.Consecutive(20, 0.5, -5);
float r = 0.5f;

for (int i = 0; i < xs.Length; i++)
{
    for (int j = 0; j < ys.Length; j++)
    {
        vectors.Add(new RootedCoordinateVector(new(xs[i], ys[j]), new((float)ys[j], -9.81f / r * (float)Math.Sin(xs[i]))));
    }
}

var field = new VectorField(new VectorFieldDataSourceCoordinatesList(vectors));
WpfPlot1.Plot.PlottableList.Add(field);
```
image

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@ _Not yet on NuGet..._
 * Axis Lines: Added `ExcludeFromLegend` so text can be added to axis line labels without appearing in the legend (#3612) @MCF
 * WPF: Added `GetPlotPixelPosition()` for getting mouse position relative to the figure (#3622) @KroMignon
 * Scatter: Upgraded the default smooth behavior to use cubic spline interpolation (#3623, #3606, #3274, #3566) @drolevar
+* Vector Field: Added a new plot type to display a collection of rooted vectors (#3625, #3626) @bclehmann
 
 ## ScottPlot 5.0.25
 _Published on [NuGet](https://www.nuget.org/profiles/ScottPlot) on 2024-04-08_
--- a/src/ScottPlot5/ScottPlot5 Cookbook/Recipes/PlotTypes/VectorField.cs
+++ b/src/ScottPlot5/ScottPlot5 Cookbook/Recipes/PlotTypes/VectorField.cs
@@ -1,0 +1,43 @@
+﻿namespace ScottPlotCookbook.Recipes.PlotTypes;
+
+public class VectorField : ICategory
+{
+    public string Chapter => "Plot Types";
+    public string CategoryName => "Vector Field";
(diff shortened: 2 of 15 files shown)


swharden:
Thanks @bclehmann, this is a great start! Pulling some convo from Discord (#1966)

> I would like at a minimum to have less ugly arrows, a method in the plottable factory, and options to anchor the arrows on their midpoint, rather than the vector tail. Colormap support would be cool too

These are all good extensions for this plot type! I'll review and merge this PR in now, and refinements can be made in future PRs 👍

swharden:
Also (let me know if I'm messing up here) I'm going to rename `RootedCoordinateVector.Distance` to `RootedCoordinateVector.Magnitude` because I think that word better matches the concept of a vector, and it may be representing things which are not spatial

bclehmann:
> Also (let me know if I'm messing up here) I'm going to rename `RootedCoordinateVector.Distance` to `RootedCoordinateVector.Magnitude` because I think that word better matches the concept of a vector, and it may be representing things which are not spatial

`Magnitude` is actually what I named it originally but I realized some of our other classes used `Distance` so I tried to be consistent. And then I learned `Vector2` uses `Length` so I guess consistency is going to be hard.

bclehmann:
I think the main thing that's missing is in the plottable factory, I don't think it automatically assigns a colour to the vector field (so it will be black by default). I don't think we should release a new version without that, but fortunately that's a 2 minute fix.

I also wonder if the legend should have a little arrow (right now it's just a line, which isn't great). As for all the colourmap stuff, that would be cool and probably pretty easy, but maybe not essential.

swharden:
Good points! Thanks for your feedback.

> Magnitude is actually what I named it originally but I realized some of our other classes used Distance so I tried to be consistent. And then I learned Vector2 uses Length so I guess consistency is going to be hard.

Now that you mention it, I can see multiple sides to this. Since the `RootedCoordinateVector` and `RootedPixelVector` both have a unit baked in (coordinates and pixels), perhaps distance or length makes sense over magnitude.

> I think the main thing that's missing is in the plottable factory, I don't think it automatically assigns a colour to the vector field (so it will be black by default). I don't think we should release a new version without that, but fortunately that's a 2 minute fix.

#3630

> I also wonder if the legend should have a little arrow (right now it's just a line, which isn't great). As for all the colourmap stuff, that would be cool and probably pretty easy, but maybe not essential.

#3631

Issue #3442, opened 2024-03-05
CI: treat warnings as errors


swharden:
It's annoying to enable this during development, but as part of the CI checks performed when PRs are opened it would be great to fail the build if warnings are present to prevent them from accidentally ending-up in the main branch, then others who fork that have to work around them

KroMignon:
You could `TreatWarningsAsErrors`, `WarningsAsErrors` and `WarningsNotAsErrors` in the project files to force error for given warnings. 

==> [C# Compiler Options to report errors and warnings](https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/compiler-options/errors-warnings)

swharden:
I found it's really annoying to add it at the csproj level because (1) I have lots of warnings while I'm developing and I don't want it to fail the build and (2) we have lots of csproj files...

However, I'm thinking we can `dotnet build /p:TreatWarningsAsErrors="true"` in our CI system 🤔

PR #3629, opened 2024-04-09
Scatter: improve cubic spline interpolation and expose SmoothTension property

drolevar:
Fix detection of consecutive segments. Previously segments were detected incorrectly, all NaN  points were just ignored.
Make tension consistent with a common definition.


drolevar:
@swharden I wonder if introducing the new way of indexing arrays is OK with you.

swharden:
Great fixes @drolevar, thanks so much!

Regarding the array code, I'm happy to include files like this to add modern language support to the older platforms we target as long as we keep the special features `internal`. 

Can you clarify where you obtained Index.cs? The file in this PR is a bit different than the one I found in the dotnet runtime repo https://github.com/dotnet/runtime/blob/main/src/libraries/System.Private.CoreLib/src/System/Index.cs

drolevar:
@swharden I got it from here: https://github.com/bgrainger/IndexRange
It's available as a Nuget package, but that would make it an extra dependency. The code gets anyway optimized out in release builds.

swharden:
> It's available as a Nuget package, but that would make it an extra dependency. The code gets anyway optimized out in release builds.

Thanks for the clarification! I found a newer version of the file from an "official" source so I feel a little better about using that one. Either way, these little compiler niceties are great to have in the code base -- thanks!

drolevar:
@swharden That new file doesn't have the necessary ##if guards for the newer versions, which leads to a lot of warnings.

swharden:
Thanks for pointing that out @drolevar! If I'd implemented #3442 by now, I'd have caught that 😅

I'll mark #3442 high priory and both issues later today 👍

PR #3632, opened 2024-04-09
VectorField Improvements

bclehmann:
- Default colour drawn from colour palette (not defaulted to black).


bclehmann:
I might return to do colourmaps, but I think this is good as-is.

PR #3638, opened 2024-04-10
Used two lines in Crosshair to reuse their functionalities

Fruchtzwerg94:
Used two AxisLines as Crosshair. This allows more flexible reusage of their features like the Text.

MouseTracker Demo:
Croshair


heartacker:
can MouseTracker be an api for general purpose，
user can enable and get/set h/v color, width,styler and so on

swharden:
Thanks for your work on this @Fruchtzwerg94, it looks great! 🚀 

I refactored this a bit and added @heartacker's suggestions to make it easy to assign line styling options for both lines at the same time, while still allowing users to reach in and directly customize styling of the lines individually in case they want vertical and horizontal lines styled differently.

jvsanchezo:
That was great! Thanks a lot!

Issue #3637, opened 2024-04-10
DataLogger ViewSlide() does not work as expected when using right axes


jpgarza93:
**Issue:** 
When I add a right Axis to my plot the DataLogger.ViewSlide() stops working as expected.

**ScottPlot Version:**
5.0.25

**Code Sample:** 
To reproduce, I added the last 2 lines in your demo datalogger file
```cs
Logger = formsPlot1.Plot.Add.DataLogger();

var axis = formsPlot1.Plot.Axes.AddRightAxis();
Logger.Axes.YAxis = axis;
```

swharden:
Hi @jpgarza93, thanks for reporting this! I think it has been fixed in #3641 and will be working better in the next release (5.0.56). If not, let me know and we can re-open this issue and take a closer look 👍

jpgarza93:
It appears to me that using the function `formsPlot1.Plot.Axes.AddRightAxis();` still causes the issue. I have tested in the demo project with the latest version 5.0.26

swharden:
Hi @jpgarza93, thanks for following-up! Can you copy/paste the code you used to recreate that issue? Thanks!

jpgarza93:
Hello @swharden, It is the same as in the first comment, does that work? Else, I can provide anything needed.

swharden:
Thanks for the tip! I'll look into it now 👍 
![3637](https://github.com/ScottPlot/ScottPlot/assets/4165489/7adde1b8-b244-4e17-9fc4-5839dbb195df)

swharden:
Actually, do you need to ever call `var axis = formsPlot1.Plot.Axes.AddRightAxis();`?

The demo plots data on the _built-in_ right axis (`formsPlot1.Plot.Axes.Right`) without having to create a new (additional) right axis

You can tell your plotted objects to use the built-in `formsPlot1.Plot.Axes.Right` and you may be set!

jpgarza93:
Okay, you tell me. Maybe I can accomplish it differently. But currently I use `Axes.AddRightAxis()` because I allow the users to trace 1-N dataloggers and display it as below image:
![image](https://github.com/ScottPlot/ScottPlot/assets/79015116/c42fcdfa-6b95-4ac5-abe2-227d0f3f070b)

swharden:
Ah, gotcha. It sounds like you're using multiple dataloggers. Thanks for the clarification! I'll modify the demo app to test this and follow-up shortly 👍

[...]

PR #3641, opened 2024-04-11
SP5: fix DataLogger scrolling when not using default axis

KroMignon:
There was a little bug in `FractionalAutoScaler`, data were always taken from default axes left and bottom and not from the axes given as parameter.

I have changed the Demo application to add a checkbox to toggle between left and right axis for the `DataLogger`.

This should fix issue #3637.


swharden:
Thanks @KroMignon! That was a complicated issue to track down, and I'm surprised to see the issue was all the way down in the autoscaler 😅 

I really appreciate you looking into this issue and implementing a fix! This is a great addition for the next release 😄 🚀

Issue #3589, opened 2024-04-04
FormsPlot: does not redraw after Dialog window opens second time


Kruno313:
I have a DialogWindow in WinForms (NET8) with FormsPlot on it. When I show the modal dialog first time, the plot draws itself.
But when I open the same dialog instance again, the dialog window shows but the FormsPlot does not draw it self on it.

This behaviour is reletad to V5.0.23.  Same code works with V4.1.73.
I have made a test solution (see attachment).
[ScottPlotTest.zip](https://github.com/ScottPlot/ScottPlot/files/14865031/ScottPlotTest.zip)

bwedding:
I just submitted a PR for this if you want to look. Easy fix.

Kruno313:
Is it meant to be in V5.0.24?
The 5.0.24 does not fix it (I updated the test sample and tried today).

Kruno313:
Confirm this working with Version V5.0.26.

PR #3646, opened 2024-04-11
FormsPlot: Fix issue affecting plots on windows that have been closed and reopened

CodeBehemoth:
My proposal to fix this issue is to instantiate the SKControl each time the form is shown.

In this PR:
1. in Sandbox.WinForms: Launch-button added to Winforms Sandbox to reproduce issue #3589
2. in ScottPlot.WinForms: Fix in FormsPlot


CodeBehemoth:
I tested this solution in 2 different applications with different chart types: it works without problems.

swharden:
Hi @CodeBehemoth, thanks for this fix! I refactored your implementation a bit and moved your example out of the sandbox and into the demo app. I also added a related demo and a few Color niceties to this PR. I'm merging now and these features will be in the next release! 🚀

PR #3666, opened 2024-04-15
Add style interfaces to plottables

### LINKED ISSUE #3658: LabelStyle: add interface and helper base class
Plottables that have `LabelStyle` often expose shortcuts like `FontSize`, `FontColor`, `BackgroundColor`, etc. It would be nice if there were something like an `IHasLabel` which requires those shortcut properties and ensures they all have consistent naming, and they could all be added with one swoop by having a base class like `FontStyleShortcuts` that could be inherited.

swharden:
This PR tracks creation of abstract "style properties" classes which can be inherited by plottable classes which have objects like `LabelStyle`, `LineStyle`, `MarkerStyle`, etc. to provide top-level shortcut properties to things like `LabelRotation`, `LineWidth`, and `MarkerSize`.

This strategy minimizes code duplication inside plottables and standardizes how these configurations are set (e.g., `TextSize` vs. `FontSize`)

Resolves #3658

- [x] `LabelStyle`
- [x] `LineStyle`
- [x] `FillStyle`
- [x] `MarkerStyle`
- [x] `ArrowStyle`


swharden:
Oof... can't do multiple inheritance in C# 💀 

I can use interfaces, but that's a pain because the plottable developer will have to implement tons of little helper methods. I may have to rethink this strategy.

Issue #3664, opened 2024-04-14
Blazor Cookbook: Add features and publish it to the website


swharden:
- [x] Give each recipe a distinct URL so it can be linked to directly
- [x] Improve the layout to support mobile views
- [x] Clicking a recipe fails for the first one you click
- [x] Menu horizontal slider needs to be hidden
- [x] The plot size is weird and doesn't adjust well to all screen sizes
- [x] Allow more pages than just cookbook recipes to be displayed in the menu
- [x] Recipe JSON is loaded after every click, but it should be loaded once when the app starts and passed around
- [x] A built-in search box would be great too
- [ ] Support multi touch on mobile
- [ ] More informative quickstart page
- [ ] Disable right-click context menu that launches when zooming
- [ ] Make the hamburger float instead of taking up a column
- [ ] Add icons to change the plot size
- [ ] Add icon to loading screen
- [ ] Add OpenGraph meta tags

swharden:
### New live demo: https://swharden.com/tmp/sp/blazor/

![image](https://github.com/ScottPlot/ScottPlot/assets/4165489/00014c88-e370-41ec-a309-b6dad3a87b79)

KroMignon:
Hello @swharden, looks nice 👍 

I see some annoying bugs:
- on mobile version, the dropdown menu is always open
- on mobile version, when selecting an item in dropdown menu, the menu is not hidden
- links for "performance tests" and "quick start" are absolute and not relative (which not work when app has not an absolute URL)
- there seems to be a mistake between 2 functions for the navigation bar:
    - on small displays (eg mobile devices): the menu is a top bar which can be dropdown with the burger icon (`collapseNavMenu`)
    - on large displays: the menu is a sidebar (on the left) which can be collapsed to save space on screen (`IconMenuActive`)
    - those functions seem to be merged is a not very good way

I will try to create a PR

heartacker:
Would it be possible to add a web-based sandbox/playground environment, similar to a Notebook? This would allow users to directly execute some demos in the website

swharden:
> Would it be possible to add a web-based sandbox/playground environment, similar to a Notebook? This would allow users to directly execute some demos in the website

I love this idea, but I suspect this is not possible because it would require a C# _compiler_ to also be available in the browser and I'm not sure if that's currently possible.

heartacker:
https://github.com/ashmind/SharpLab

swharden:
[...]

PR #3667, opened 2024-04-15
Blazor Cookbook fixes

KroMignon:
- fix first recipe load
- add filter
- load JSON file once

It is only a preview... fix some issues from #3664 

Listed issues:
- [x] Give each recipe a distinct URL so it can be linked to directly
- [x] Improve the layout to support mobile views. Bootstrap is already there so we can use their hamburger menu.
- [x]  Clicking a recipe fails for the first one you click
- [x] Menu horizontal slider needs to be hidden
- [x] The plot size is weird and doesn't adjust well to all screen sizes
- [x] We should allow more things to be added to the menu than just cookbook recipes (e.g., comparing the OpenGL control vs. standard one)
- [x] Recipe JSON is loaded after every click, but it should be loaded once when the app starts and passed around
- [x] A built-in search box would be great too
- [ ]  support multi touch on mobile


KroMignon:
Hi @swharden 

I am on holiday for the next 2 weeks and would stop update this PR.
I think I have done all I can do.
As far as I can see, the application is functional and all recipes can be called for standard or OpenGL control.
I have notice bug on "Signal" page when switching between Standard and OpenGL control, the grid and axes are not drawn for Standard version. But when changing the signal size, the grid is coming back!

Perhaps someone else could add some updates to improve the UI/UX.

swharden:
Hi @KroMignon, I just look a look at this and **WOW!** You did fantastic work here! Thank you so much 🚀 😄 

I'll refine the UI a bit, merge this in, and publish it to the website. It's looking fantastic!

Enjoy your holiday!

swharden:
### Live demo: https://swharden.com/tmp/sp/blazor/

image

PR #3694, opened 2024-04-18
Helper function to retrieve the SvgXml

aespitia:
Similar to how there was a helper function for GetLegendSvgXml, I didn't see an equivalent to get the SvgXml for the whole graph, I hope this accomplishes that.  

Please let me know if there was already an alternate way to do this.  I was trying to save having to SaveSvg to a file, and reading the file back.

```cs
       Plot plt = new();
        plt.Add.Signal(Generate.Sin(51));
        plt.Add.Signal(Generate.Cos(51));
        var svgXml = plt.GetSvgXml(400, 300);
```


swharden:
Great solution, thanks! 😄

Issue #3674, opened 2024-04-16
AxisRules: implement a double first-render


swharden:
Axis rules (like "lock vertical") can't be applied before the first render (because the rule relies on information from the previous render).

Add logic to determine if axis rules are in place before the first render, and if so the first render could be a double render (once without the rules, then apply the rules, then re-render).

Topic came up in Discord #1966 by Boris.

swharden:
resolved in #3701

PR #3701, opened 2024-04-21
AxisRules: refactor to cut dependence on previous renders

swharden:
Rewrites logic of axis rules to force the user to specify the desired limits rather than taking from from the previous render. This ensures all axis rules can be applied and used immediately upon the first render.

* #3674 
* #3547

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,7 @@ _Not yet on NuGet..._
 * Legend: Created a `LegendPanel` to allow legends to be displayed outside the data area (#3672, #3635) @Graat @mikeKuester
 * Axis: Prevent left axis from appearing if no plottables use it (#3637) @jpgarza93
 * Label: Added `BorderRadius` to support backgrounds and outlines with rounded edges (#3659)
+* Axis Rules: Changed behavior of axis rules to reduce reliance on previous renders (#3674, #1966, #3547)
 
 ## ScottPlot 5.0.26
 _Published on [NuGet](https://www.nuget.org/profiles/ScottPlot) on 2024-04-14_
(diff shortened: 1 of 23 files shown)


swharden:
@BrianAtZetica I made some changes to code you added in #3547 so if you are interested and have a few minutes to check it out, I welcome feedback on this 👍

BrianAtZetica:
> @BrianAtZetica I made some changes to code you added in #3547 so if you are interested and have a few minutes to check it out, I welcome feedback on this 👍

@swharden thanks for the tag. I haven't had time to run the changed code (am preparing for a work trip so it may be a few weeks yet). But looking at the changes, I have just one concern. It looks like the SnapToTicks rules now only expand the graph extents. They never zoom in. There is a change that a user will be trying to zoom in, but because their change doesn't reach the next tick, the rule snaps back out to the previous extent, effectively locking the axes.

I had tried to catch that in my original (far more verbose) implementation by comparing the user's new limit (before the rule is applied) with the old limit (to detemine intent) and with the "proposed" outward snapped limit (to establish if the intent was achieved). If the user intended to zoom out then it is easy - we snap outwards. Otherwise if the user intended to zoom in, but the snapping results in a zoom out, we should override the default and snap inwards. I implemented this seperately for left and right limits, shown here for the right limit:

```
            var oldRight = rp.Plot.LastRender.AxisLimitsByAxis[XAxis].Max;
            var proposedNewRight = Math.Ceiling(newLimits.Max / tickInterval) * tickInterval;

            if (newLimits.Max >= oldRight || proposedNewRight < oldRight)
            { // we'll snap outwards if we can 
                newRight = proposedNewRight;
            }
            else
            { // but if the snapped limit expands the range when a user requested to reduce the range we'll snap inwards
                newRight = Math.Floor(newLimits.Max / tickInterval) * tickInterval;
            }
```
 
I'll aim to test eventually (maybe I've misunderstood some of the changes) and if I find the problem I foresee, I'll raise a new issue to resolve it.

PR #3717, opened 2024-04-24
Improve spline shape

drolevar:
This fix gives better shape around endpoint in an open spline.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 ## ScottPlot 5.0.28
 _Not yet on NuGet..._
 * Marker: Refactored marker system to support improved styling and the ability to use custom markers (#3716, #3141)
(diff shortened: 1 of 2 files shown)


swharden:
Thanks for this PR @drolevar! The difference looks pretty small. Do you have a shape you recommend trying to best showcase the change?

before | after
---|---
image|image


before | after
---|---
image|image

drolevar:
@swharden It is indeed very small, but you can look for example at the top end of the orange curve (tension 0.3).

swharden:
Wow, it's so subtle! How did you realize it was incorrect? 😅 Thanks for the fix either way!

## Before
spline-before

## After
spline-after

drolevar:
@swharden Because I knew that both options are possible, and later I decided that the second is better :)

PR #3718, opened 2024-04-24
Signal: experimental SignalSource that caches binned range extrema

Cardroid:
exp

This PR can improve the performance of Signal Plot by about three times by applying `MinMaxCache`. (I named it 😅)

Currently, it is only applied to Double Array Sources. (=`FastSignalSourceDouble`)
The cache implementation code is separate, so I think it can be applied elsewhere.

We are providing a demonstration by slightly fixing the code in the `SignalPerformance` part.

I'm using the good library well.
I hope this PR helps 👍


bclehmann:
This looks quite interesting, but I think caching will break our current contract, since the user can change the data of a signal plot at any time. You do create a separate data source, so we could document this and add a method for invalidating the cache, but then it seems to get pretty close to the role of `SignalConst`.

SignalConst already uses a segmented tree for these sorts of queries (which would have been my next suggestion, though I haven't done the testing to check if a segmented tree actually outperforms this implementation). Strangely, it seems that SignalConst doesn't really allow a user to pass in their own data source class, instead it takes a plain array and builds a `SignalConstSourceDoubleArray` over it. Ideally, it would be refactored to take something like an `ISignalConstDataSource`, and then it would be easy to test whether your implementation is faster than the tree-based approach. Also `SignalConst` should probably have a way to update the data (and invalidate the cache) like it did in v4.1, since as long as these updates are rare it should still outperform `Signal`.

Ideally, in my view, there should be only `Signal`, and `SignalConst`-like behaviour is achieved by using `Signal` with a caching datasource, provided that `SignalConst` does not have any additional optimizations (I believe it does not). If Scott is receptive to this kind of change, then I think this PR is a good idea, though we'll need to document the caching stuff. In the meantime, you might want to switch to using `SignalConst` if you need the performance boost.

swharden:
Hi @Cardroid, thanks for your original PR and @bclehmann for your throughout feedback!

I think there may be room for this plot type to be a "sweet spot" between signal plots and signal const plots. SignalConst offers fantastic performance improvements for extremely large datasets, but it takes a very large amount of memory (4x what signal plots require, if I remember right). The strategy presented here may offer many of SignalConst's speed benefits, but by using binning to achieve a lower resolution of cached ranges, the memory requirement and pre-processing time may be significantly reduced. I look forward to reviewing this more carefully soon!

I'll ping @StendProg (the original author of the SignalConst plot type) in case they're interested in commenting about the strategy implemented here.

---

Thoroughly reviewing this PR carefully requires a lot of performance testing, which is the main reason I've been delaying my review as I focus on resolving the currently open high priority issues. I think I'll proceed by modifying this PR a bit and merging it in mostly as it is now, then we can refine it as needed in future PRs. I won't "advertise" it in the cookbook until it's a little more thoroughly evaluated and tested. 

Thanks again for your hard work on this! 👍🚀

swharden:
I added a new demo window to facilitate interactive evaluation.

I'm noticing that the SiganlConst performance isn't as good as I would expect, and I think it may be because I never properly split the `double` optimized signal source from the generic one. To keep this PR clean I'll merge this in now, investigate/improve SignalConst in another PR, then report back here.

image

swharden:
> I'm noticing that the SignalConst performance isn't as good as I would expect, and I think it may be because I never properly split the double optimized signal source from the generic one. To keep this PR clean I'll merge this in now, investigate/improve SignalConst in another PR, then report back here.

Following-up, I experimented with a non-generic signal const (#3771) and confirmed the performance is the same, so that isn't the issue. In ScottPlot 4 I recall that SignalConst was much faster than Signal, so it's possible that something in the migration to ScottPlot 5 impaired performance. A closer inspection of ScottPlot 4 vs. ScottPlot 5 may help reveal what the issue is.

StendProg:
Hi all, 
thanks @swharden for pinging me in this tread.

The above implementation `MinMaxCache` reduces the computation time, but technically it is still `0(N)`. For interval min/max search we get `0(1)` in best case, `O(CachePeriod+ N/CachePeriod)` in worst case, and average is still `O(CachePeriod +N/CachePeriod)` = `O(N)`.  A significant advantage of this method is the small memory consumption `N/CachePeriod`, but technically it's still `0(N)`.

The segment tree gives guaranteed `O(log(N))` for any interval. The tree takes `2*N` memory (but N must be augmented to pow2, so from `2*N` to `4*N` depending on the original N).

Both methods take `O(N)` time to build the cache.

I don't mean to say that the `MinMaxCache` is bad. It does its job and it is better than nothing (this is roughly the solution where attempts to speed up signals started ;-) ).
 
But we already have a better algorithmic solution and if it suddenly breaks, we should try to fix it.

StendProg:
It looks like `SignalConstSourceDoubleArray` is not fully finalized. 
https://github.com/ScottPlot/ScottPlot/blob/7e4669bb86da09f6beb8e8b6fdbd6293096d95f8/src/ScottPlot5/ScottPlot5/DataSources/SignalConstSourceDoubleArray.cs#L49-L62
This method uses linear search, while `SegmentedTree.MinMaxRangeQuery(firstIndex, lastIndex, out min, out max);` should be called

Issue #3723, opened 2024-04-24
FillStyle: disable border by default


Fruchtzwerg94:
After the change 695cb78 fills are constanty showing a black border around them. Even if the LineStyle is changed, nothing changes.
It seems like the LineStyle should be forwareded to Fills internal Polygon instead of tjust keeping it and not using it.
What do you think? Currently I see no way to change the border around the fill since the [corresponding polygon is private](https://github.com/ScottPlot/ScottPlot/blob/main/src/ScottPlot5/ScottPlot5/Plottables/FillY.cs#L13).

![grafik](https://github.com/ScottPlot/ScottPlot/assets/29866610/888bfe53-094a-4601-8367-811e5e60397c)

swharden:
Thanks for reporting this! It seems related to #3720 and will be fixed shortly 👍

swharden:
resolved in #3726

Fruchtzwerg94:
Thank you @swharden , love to see that. Works like a charm.

PR #3726, opened 2024-04-24
FillY bug fixes: Update poly coordinates instead of creating a new po…

### LINKED ISSUE #3720: Marker: closed marker shapes should disable outlines by default
Otherwise things like scatter plots have markers with black outlines

asis_gridAdvanced

SebastianDirks:
…lygon to prevent loss of all set properties (marker, line and fill styles) when setting (updating) data source. Redirect LineStyle property to Poly to actually apply it.

Use this space to describe what this pull request accomplishes. 
FillY bug fixes: Update poly coordinates instead of creating a new polygon to prevent loss of all set properties (marker, line and fill styles) when setting (updating) data source. Redirect LineStyle property to Poly to actually apply it.
Refer to open issues by their number to help link them with this pull request.
should fix #3720 and #3723


swharden:
Thanks for the fix @SebastianDirks! I improved a few additional edge cases, and these fixes will be in the next release 👍 🚀

PR #3767, opened 2024-05-03
Avoid excessive memory allocation in Markers

drolevar:
Calling ToArray() causes extra memory allocation, as specified in TODO:

```
Large Object Heap: code that allocates a lot of memory in LOH
Allocated object type: Coordinates[]
Last observation: 2024-04-24 11:59 MyApp.exe
  Allocated size: 4 278,4 MB

at Enumerable+ListPartition<Coordinates>.ToList()
at Markers.Render(RenderPack) in /_/src/ScottPlot5/ScottPlot5/Plottables/Markers.cs:line 38 column 13
at RenderPlottables.Render(RenderPack) in /_/src/ScottPlot5/ScottPlot5/Rendering/RenderActions/RenderPlottables.cs:line 22 column 17
```

Use lazy Linq query instead.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
-## ScottPlot 5.0.32
+## ScottPlot 5.0.33
 _Not yet on NuGet..._
+* Markers: Reduced memory allocations and improved performance during rendering (#3767) @drolevar
+
+## ScottPlot 5.0.32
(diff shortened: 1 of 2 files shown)


bclehmann (approved):
I like this, this was the main reason I pushed for `IEnumerable` on most of the drawing APIs, we had a lot of array-copying in previous versions of ScottPlot because we so often need to convert large arrays from coordinate to pixel space.

PR #3768, opened 2024-05-03
Prevent exception when no ticks have been generated

drolevar:
If the zoom level is too big, no ticks are generated an Measure methods throw an exception.

### DIFF
--- a/src/ScottPlot5/ScottPlot5/AxisPanels/XAxisBase.cs
+++ b/src/ScottPlot5/ScottPlot5/AxisPanels/XAxisBase.cs
@@ -14,7 +14,7 @@ public abstract class XAxisBase : AxisBase, IXAxis
         if (!IsVisible)
             return 0;
 
-        if (!Range.HasBeenSet)
+        if (!Range.HasBeenSet || TickGenerator.Ticks.Length == 0)
             return SizeWhenNoData;
 
         using SKPaint paint = new();
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 ## ScottPlot 5.0.33
 _Not yet on NuGet..._
 * Markers: Reduced memory allocations and improved performance during rendering (#3767) @drolevar
+* Axes: Prevent exceptions for conditions where tick generation produces no ticks (#3768) @drolevar
 
 ## ScottPlot 5.0.32
 _Published on [NuGet](https://www.nuget.org/profiles/ScottPlot) on 2024-05-01_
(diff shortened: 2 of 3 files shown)


src/ScottPlot5/ScottPlot5/AxisPanels/XAxisBase.cs
```diff
@@ -14,7 +14,7 @@ public float Measure()
         if (!IsVisible)
             return 0;
 
-        if (!Range.HasBeenSet)
+        if (!Range.HasBeenSet || TickGenerator.Ticks.Length == 0)
             return SizeWhenNoData;
```
bclehmann, line 18: I like this, but I think instead of `SizeWhenNoData` (which is seemingly just hardcoded to 15) this should return `axisLabelHeight`. This is probably also necessary even when the range hasn't been set, so this is a problem that preexists your change.

I guess I'd also like a screenshot of how the plot looks when there's no ticks, to make sure it still looks right (or as right as it can look), even though a slightly broken axis is preferable to throwing.

bclehmann (changes requested):

src/ScottPlot5/ScottPlot5/AxisPanels/XAxisBase.cs
swharden: I may be misunderstanding the discussion, but the intent of returning 15 is to cover the case where there are no ticks and no axis labels. It's not aesthetic for the edge of the data area to extend right to the edge of the image, mostly because tick labels on other axes tend to get cut off if this is one.

Also FWIW, the hard-coded 15 pixels of padding for empty axes is user settable, so maybe we can call it soft coded 💁

bclehmann: My point was that there could be no ticks, but still an axis label. So we need to measure the axis label if its present and return its height.

swharden: Ah, now I understand - thanks for the clarification! I'll make that fix in main and tag this PR in it 👍

swharden: This should do the trick! 🚀 
https://github.com/ScottPlot/ScottPlot/commit/8a12c5debde4b54ee0431f0f52d32539c0826ad9

drolevar: @swharden I think from the user's perspective it would be beneficial to disallow zooming to the levels when no ticks can be generated. Not sure if this functionality exists already.

swharden: hi @drolevar, I suspect you can achieve that with "axis rules" https://scottplot.net/demo/#axis-rules

PR #3776, opened 2024-05-04
AxisManager: plottable-specific `AutoScale()` methods

levipara:
Allow a plot consumer to define what elements (IPlottable's) of the plot the AutoScaler should size axis limits to for an individual call. This can be used to ignore plottables that the user does not want to focus on if a chart contains multiple data sets that do not overlap but share the same axes OR IPlottables used only to enhance visual clarity such as spans or vertical/horizontal lines, text etc.  UI/UX specific requirements may not allow for switching certain plottable's visibility on/off to artificially adjust autoscale limit results, this circumvents that. Also adds to overall user interactivity possibilities


levipara:
AutoScaleLimited Gif

swharden:
This looks great @levipara, thanks! I'm renaming the methods a bit and modifying to accept `IEnumerable<IPlottable>` so they can accept lists and arrays, and this feature will be in the next release! 🚀

PR #3779, opened 2024-05-05
Ticks, Tests, Cookbook: Improve DateTimeFixedInterval

### LINKED ISSUE #3408: DateTimeFixedInterval: Add option to snap to the first of each month
**Issue:**
Excuse me if I sound vague because it's pretty difficult to explain and I can't get my recording to work. If ticks are set to DateTimeFixedInterval (for TimeUnits.Day() for instance), then the first tick is always all the way on the left of the plot, with following ticks at the correct interval apart. This means that if you then pan over to either side, the ticks and gridlines do not move along with the graph. Instead, they stay stationary and snap to the next value as soon as the TimeUnit has been passed.
I considered that this may be intented behaviour, but I could not imagine in what case this would be useful. I would like the tick to be positioned at the start of the day and for this tick to move along with the points at this time when panning. Now it is not clear when a day starts or ends.
Using any other TimeUnit has the same issue.

**ScottPlot Version:**
ScottPlot.WinForms 5.0.21

**Code Sample:**

```cs
//This code is taken from the ScottPlot 5 Cookbook:

// plot data using DateTime units
DateTime[] dates = Generate.DateTime.Days(100);
double[] ys = Generate.RandomWalk(100);
myPlot.Add.Scatter(dates, ys);

// tell the plot to display dates on the bottom axis
myPlot.Axes.DateTimeTicksBottom();

//Only the following line has been added to the example:
myPlot.Axes.Bottom.TickGenerator = new DateTimeFixedInterval(new ScottPlot.TickGenerators.TimeUnits.Day(), 30);

myPlot.SavePng("demo.png", 400, 300);
```
This results in the following plot, note how the ticks start all the way to the left. I can't record video so I can't show how panning this in an interactable FormsPlot affects it
demo

EricEzaM:
Fixes #3408
No breaking API changes, only optional params/properties added.

https://github.com/ScottPlot/ScottPlot/assets/41730826/2fbd2061-994a-4b98-a122-f27ff00b0fa2

It would be nice if fixed-width tick labels also had logic to hide when overlap started... but not in this PR.

### DIFF
--- a/src/ScottPlot5/ScottPlot5 Cookbook/Recipes/Axis/DateTimeAxes.cs
+++ b/src/ScottPlot5/ScottPlot5 Cookbook/Recipes/Axis/DateTimeAxes.cs
@@ -1,4 +1,7 @@
-﻿namespace ScottPlotCookbook.Recipes.Axis;
+﻿using ScottPlot.TickGenerators;
+using ScottPlot.TickGenerators.TimeUnits;
+
+namespace ScottPlotCookbook.Recipes.Axis;
 
 public class DateTimeAxes : ICategory
 {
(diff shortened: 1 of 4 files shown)


src/ScottPlot5/ScottPlot5 Cookbook/Recipes/Axis/DateTimeAxes.cs
```diff
@@ -91,4 +94,56 @@ public override void Execute()
...
+            // Customise gridlines to make the ticks easier to see
+            myPlot.Grid.XAxisStyle.MajorLineStyle.Color = Colors.Black.WithOpacity();
+            myPlot.Grid.XAxisStyle.MajorLineStyle.Width = 2;
+
+            myPlot.Grid.XAxisStyle.MinorLineStyle.Color = Colors.Gray.WithOpacity(0.25);
+            myPlot.Grid.XAxisStyle.MinorLineStyle.Width = 1;
+            myPlot.Grid.XAxisStyle.MinorLineStyle.Pattern = LinePattern.DenselyDashed;
+            
+            // Remove labels on minor ticks, otherwise there is a lot of tick label overlap
+            myPlot.RenderManager.RenderStarting += (s, e) =>
```
EricEzaM, line 134: Not sure if this is really the best way to handle the minor tick labels, but I saw it used in other samples. The other option is passing another delegate to the constructor for what the label format should be, but this method in RenderStarting introduces less complexity into the `DateTimeFixedInterval` class.

swharden: I like it! Those tick generator classes are super confusing to work with, and I love that this solution fits in a cookbook recipe and can be easily modified by the end user. Thanks so much for this, it looks great!

PR #3783, opened 2024-05-05
DateTimeAutomatic: Add `LabelFormatter` property

loyvsc:
Hi! With this request, I want to add a label formatter for the datetimeautomatic tick generator. At the moment, we cannot set a custom label formatter - we are content with "presets" (timeunits), but with labelformatter we can customize the display of date and time.
For example:
```cs
WpfPlot1.Plot.Axes.DateTimeTicksBottom();
((DateTimeAutomatic)WpfPlot1.Plot.Axes.Bottom.TickGenerator).LabelFormatter = LabelFormatter;

string LabelFormatter(DateTime arg)
{
    return arg is { Hour: 0, Minute: 0, Second: 0 } ? arg.ToShortDateString() : arg.ToLongTimeString();
}
```
I hope I've brought a useful feature.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 * Legend: Added `Plot.ShowLegend()` overload that accepts an `Edge` for quickly adding a legend outside the data area (#3672, #3635)
 * Radar Plot: New plot type (also called a spider charts or star charts) for representing multi-axis data as a 2D shape on a circular axis system (#3457, #3780) @bclehmann
 * Coxcomb Plot: New plot type like a pie graph where the angle of slices is constant but the radii are not (#3457, #3780) @bclehmann
+* Axes: Added `LabelFormatter` property to `DateTimeAutomatic` for custom formatting of DateTime tick labels (#3783) @loyvsc
 
 ## ScottPlot 5.0.34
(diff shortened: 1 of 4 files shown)


swharden:
Hi @loyvsc, thanks so much for this PR! I like the idea and will be happy to merge it in soon. 

It would be helpful if you can also add a cookbook recipe demonstrating how this feature can be used. You can duplicate the existing custom tick formatter recipe and modify it by having it work on DateTime axes using the new feature you recently added. 

Note that the cookbook recipe https://scottplot.net/cookbook/5.0/CustomizingTicks/CustomTickFormatter/ is achieved from this block of code, and you can demonstrate your feature by adding a similar class just beneath this one: 
https://github.com/ScottPlot/ScottPlot/blob/093994e1b020d2d1f349e4e0a06c9d67e2577b44/src/ScottPlot5/ScottPlot5%20Cookbook/Recipes/Axis/CustomizingTicks.cs#L11-L45

loyvsc:
Thanks, I added the recipe

swharden:
Thanks so much @loyvsc, this is fantastic! I'm going to refine a bit and merge.

Note that I ran into an interesting edge case, but I think this behavior is okay... midnight is date but noon is time

image

PR #3818, opened 2024-05-09
SP5: Apply ScaleFactor when rendering the figure background

### LINKED ISSUE #3786: Rendering: Saved image size is incorrect when `ScaleFactor` is less than `1`
**Issue:** When I change the ScaleFactor of the ScottPlot.Plot the GetImageBytes returns a image with a blackout sections

**ScottPlot Version:** 5.0.34

**Code Sample:** 

```cs

ScottPlot.Plot plt = new();
plt.ScaleFactor = 0.5;

// use code to express information related to the the discussion
plt.Add.Signal(ScottPlot.Generate.Sin());
plt.Add.Signal(ScottPlot.Generate.Cos());

plt.SaveBmp("demo.png", 200, 200);

```

demo

MCF:
The figure background was not applying the scale factor when determining its bounds.  Added a property called ScaledFigureRect to the RenderPack and changed the figure background render function to use it when determining bounds.

The issue was present in all renders, including to the screen.  It just didn't show up easily in most application since the default background is white for both the plot's background and the window background for most applications.  For image saves the default image background is black, making the incorrect background standout.  When ScaleFactor was 1 the background bounds was correct, when ScaleFactor was > 1 the background was drawn to large, but then clipped back to the Figure bounds by other mechanisms.

Fixes #3786 

Sample application and before and after results shown below.

```cs

myPlot.ScaleFactor = 0.75;

myPlot.FigureBackground.Color = Colors.Yellow;

var sp1 = myPlot.Add.Signal(Generate.Sin());
sp1.LegendText = "Sine Wave";
sp1.LineWidth = 4;

myPlot.Legend.Orientation = Orientation.Horizontal;
myPlot.Legend.Alignment = Alignment.LowerRight;
myPlot.Legend.TightHorizontalWrapping = true;

myPlot.SavePng("test.png", 300, 200);
```

Before:

test-before

After:

test-after


swharden:
Thanks @MCF! Adding that custom `ScaleFactor` solved a great problem, but lots of little issues like this popped up as a result and are hard to track down #3469, #3786

Thanks for identifying this issue and fixing it with this PR! 😄 🚀

PR #3823, opened 2024-05-13
Add ability to create custom axes and add them to plots.

EricEzaM:
Just adds:
* Measure and Render are virtual on `YAxisBase` and `XAxisBase`
* `Add<T>` methods on `AxisManager` to add any type of `IXAxis` and `IYAxis`

And a cookbook recipe :)

Note that due to the way Cookbook recipes are generated, the custom axis class needed to be moved to a different folder, outside of `Recipes`. If I put it in any file inside `Recipes`, it would try and read it as a recipe and fail for one reason or another. Let me know if there is a better place it put it, or if there is some other way to resolve the issue.

image


swharden:
Thanks so much for this @EricEzaM! I'm going to adapt it to be part of the demo app rather than a cookbook recipe to make the source code a little more accessible, then will merge it in tonight 🚀 

image

EricEzaM:
Ok, thanks @swharden.

My only concern there is that features which do not have a cookbook recipe are less discoverable in my opinion. When searching for how to do something, or available features, I typically just use the cookbook online as it essentially acts as the available documentation for the library.

Since the demo app already references `ScottPlot Cookbook.csproj`, could it exist in both places?

EricEzaM:
Ah after your pushes I see that is what you have done, all good! 👍

swharden:
> I typically just use the cookbook online as it essentially acts as the available documentation for the library and is the most accessible.

> Ah after your pushes I see that is what you have done, all good! 👍

image

Having a `ScottPlot.AxisPanels.Experimental` namespace will be a nice way to collect useful custom axis panels! I'll demonstrate their usage in the cookbook, and users who want to create their own can look deeper at the code of the demo app. This is great timing because https://github.com/ScottPlot/ScottPlot/issues/3827 was just requested, and if work gets put into reaching a solution that class will probably end up here too.

Now that you're well versed in creating custom axes, would you be interested in working on creating a custom left (and bottom) axis that supports that "corner notation" for multiplication and/or offset? Funny it was in ScottPlot 4 and is demonstrated in the animated screenshot on https://scottplot.net but that feature doesn't currently exist in ScottPlot 5 😅 If you don't have the time or aren't interested, no worries, but I figured I'd ask! If you want to give it a go I'll mark #3827 high priority and tag you in it

Thanks again for this PR! I'm excited to merge it now and it will be in the next release 🚀 🚀 🚀

PR #3833, opened 2024-05-15
Prevent legend overflow

drolevar:
Currently the legend overflows the DataRect if the DataRect size is too small:

image

With this change it will fit the DataRect and the contents will be clipped to the LegendRect:

image


swharden:
Great fix @drolevar, thanks so much! 🚀

PR #3838, opened 2024-05-16
Add override support for axis lock keys in InputBindings

JinjieZhao:
Allow users to customize the lock logic for X and Y axes. 

In my use case, the mouse wheel only zooms X. 
If the right mouse button is pressed, the logic remains consistent with previous behavior.

This is due to the fact that under typical circumstances, I am examining signal data and don't need to zoom the Y-axis.

Here is my code
```
public override bool ShouldLockY(IEnumerable<Key> keys, MouseButton? button = null)
    {
        if (button == DragZoomButton)
        {
            return base.ShouldLockY(keys);
        }

        return !base.ShouldLockY(keys);
    }
```


swharden:
This looks great @JinjieZhao, thanks for the PR! This feature will be in the next release 🚀 

Note that there is an upcoming breaking change where the whole input binding system will be refactored toward a simpler system which is easier to customize. While I think this change will be good in the long wrong, the classes you make that override these methods may need to be updated in the future to accommodate the change.  #3186

Issue #3722, opened 2024-04-24
Demo: show how to use a custom font


kebox7:
How can I use my own fonts in ScottPlot, which are located in application resources? For example, there is the [PDFsharp ](https://github.com/empira/PDFsharp) library. It has a font substitution mechanism that allows you to get a font in any available way. It looks something like this (very simplified):
```cs
using Avalonia.Platform;
using PdfSharp.Fonts;

namespace TestApp.Main;

internal sealed partial class PdfDocument {
    private sealed class CustomFontResolver : IFontResolver {
        public byte[]? GetFont(string faceName) {
            using (var memory = new MemoryStream()) {
                using (var asset = AssetLoader.Open(new Uri(faceName))) {
                    asset.CopyTo(memory);
                    memory.Position = 0;
                    return memory.ToArray();
                }
            }
        }

        public FontResolverInfo? ResolveTypeface(string familyName, bool isBold, bool isItalic) =>
            new($"avares://TestApp.Fonts/Assets/NotoSans-{(isBold ? "Bold" : "Regular")}.ttf");
    }
}
```
Is there something similar for ScottPlot?

**ScottPlot Version:** 5.0.25

kebox7:
The issue was resolved, although not in the best way (using reflection).
```cs
public static class PlotHelper {
    private static readonly FieldInfo? LabelCachedTypeface = typeof(Label).GetField("CachedTypeface", BindingFlags.NonPublic | BindingFlags.Instance);
    private static readonly SKTypeface? CachedTypeface = CreateTypeface();

    public static void SetPlotTypeface(AvaPlot control) {
        if (LabelCachedTypeface is null || CachedTypeface is null) {
            return;
        }

        foreach (var axis in control.Plot.Axes.GetAxes()) {
            LabelCachedTypeface.SetValue(axis.TickLabelStyle, CachedTypeface);
        }
    }

    private static SKTypeface CreateTypeface() {
        using (var asset = AssetLoader.Open(new Uri("avares://TestApp.Fonts/Assets/NotoSans-Regular.ttf"))) {
            return SKTypeface.FromStream(asset);
        }
    }
}
```
[...]

PR #3841, opened 2024-05-17
Fonts: improve support for fonts loaded from true-type files

kebox7:
Discussed in https://github.com/ScottPlot/ScottPlot/issues/3722

- [x] Create custom font resolver
- [x] Improve management of cached typeface
- [x] Demo


src/ScottPlot5/ScottPlot5/Fonts.cs
```diff
@@ -5,6 +5,37 @@
 /// </summary>
 public static class Fonts
 {
+    /* Typefaces are cached to improve performance.
+     * https://github.com/ScottPlot/ScottPlot/issues/2833
+     * https://github.com/ScottPlot/ScottPlot/pull/2848
+     */
+    private static readonly Dictionary<int, SKTypeface> TypefaceCache = new();
```
bclehmann, line 12: I'm not sure why you're going to the trouble of making this a `Dictionary<int, SKTypeface>` and all the associated hash combining, I'm pretty sure you can just use a tuple like `(string fontname, bool isBold, bool isItalic)` as a dictionary key. This saves you work, and it will be robust to hash collisions.

src/ScottPlot5/ScottPlot5/Fonts.cs
```diff
@@ -5,6 +5,37 @@
...
+
+    /// <summary>
+    /// Gets or sets the global font resolver
+    /// </summary>
+    private static IFontResolver? _fontResolver;
+    public static IFontResolver? FontResolver
+    {
+        get => _fontResolver;
+        set
```
bclehmann, line 21: If the user changes the font resolver that should probably invalidate the typeface cache as well.

src/ScottPlot5/ScottPlot5/Primitives/FontStyle.cs
```diff
@@ -73,28 +22,6 @@ public override string ToString()
         return $"{Name}, Size {Size}, {Color}";
     }
 
-    private void ClearCachedTypeface()
-    {
-        CachedTypeface = null;
-    }
-
-    public static SKTypeface CreateTypefaceFromName(string font, bool bold, bool italic)
```
bclehmann, line 81: Unfortunately we need to keep `CreateTypefaceFromName` and `CreateTypefraceFromFile` here for backwards compatibility. `CreateTypefaceFromName` should just forward to `Fonts.GetTypeface`, and `CreateTypefacfeFromFile` is similar, although you'll have to wire it up such that it gets cached properly.

src/ScottPlot5/ScottPlot5/Primitives/Label.cs
```diff
@@ -30,57 +30,11 @@ public class Label
...
-        set { _FontSize = value; ClearCachedTypeface(); }
-    }
-
-    private bool _Bold = false;
-    public bool Bold
-    {
-        get => _Bold;
-        set { _Bold = value; ClearCachedTypeface(); }
-    }
+    public bool AntiAliasText { get; set; } = true;
```
bclehmann, line 33: Very nice, this cleans it up well

src/ScottPlot5/ScottPlot5/Primitives/Label.cs
```diff
@@ -30,57 +30,11 @@ public class Label
...
-
-    private string _FontName = Fonts.Default;
-    public string FontName
-    {
-        get => _FontName;
-        set { _FontName = value; ClearCachedTypeface(); }
-    }
-
-    private string? _FontFile = null;
-    public string? FontFile
```
bclehmann, line 65: We need to preserve this for backwards compatibility.

src/ScottPlot5/ScottPlot5/Fonts.cs
```diff
@@ -35,7 +66,45 @@ public static class Fonts
...
+        {
+            typeface = FontResolver.CreateTypeface(fontName, bold, italic);
+            if (typeface is { })
+            {
+                TypefaceCache.Add(hashCode, typeface);
+                return typeface;
+            }
+        }
+
+        SKFontStyleWeight weight = bold ? SKFontStyleWeight.Bold : SKFontStyleWeight.Normal;
```
bclehmann, line 96: I think all of this stuff should just live in a default `IFontResolver`, rather than being the fallback if nothing can be found by font resolver.

bclehmann (commented):
Very nice, it cleans up some parts quite well. Just a few suggestions about keeping backwards compatibility and some other tweaks.

src/ScottPlot5/ScottPlot5/Fonts.cs
kebox7: To be honest, I don’t know what to say here. I thought about it. On the one hand, you are right, but on the other... I even had an idea to prohibit repeated changes the font resolver.

src/ScottPlot5/ScottPlot5/Fonts.cs
kebox7: I agree with your suggestion to simplify the code.

src/ScottPlot5/ScottPlot5/Primitives/Label.cs
kebox7: Please explain your comment, I don’t quite understand what it means.

src/ScottPlot5/ScottPlot5/Fonts.cs
kebox7: This has a slightly different meaning. `IFontResolver `should only handle custom fonts. If a standard font is specified (for example: `ScottPlot.Fonts.Sans`, `ScottPlot.Fonts.Serif`, `ScottPlot.Fonts.Monospace`), then we simply pass them to `SKTypeface.FromFamilyName`. `IFontResolver `should not create them.

src/ScottPlot5/ScottPlot5/Primitives/FontStyle.cs
kebox7: That's something I probably disagree with. The `CreateTypefraceFromFile` method as well as the `FontFile` property were added in version **5.0.33 (2 weeks ago)**. It seems to me that while not much time has passed since that moment, we can painlessly break the API. Otherwise, this will lead to duplication of functionality.

There is also one more point to consider: `IFontResolver `is a more universal solution than the existing one. It allows you to set fonts from any source, for example from application resources:
```cs
public class ResFontResolver : IFontResolver
{
    public bool Exists(string fontName)
    {
        return ...
    }

    public SKTypeface? CreateTypeface(string fontName, bool bold, bool italic)
    {
        ...
        using (var asset = AssetLoader.Open(new Uri($"avares://TestApp/Assets/{fontName}.ttf"))) {
            return SKTypeface.FromStream(asset);
        }
    }
}
```

src/ScottPlot5/ScottPlot5/Primitives/Label.cs
kebox7: Already answered a little higher.

src/ScottPlot5/ScottPlot5/Fonts.cs
bclehmann: I think a setter should just be a setter, if there's a good reason to add such behaviour then I think it would ideally be in a `UpdateFontResolver` method such that we can properly document its behaviour.

I also wonder what purpose this behaviour would have (beyond just a normal setter). I can't see a reason why it should be restricted in such a way, and if the user wants this kind of "set-if-not-exists" behaviour then they can do it themselves.

src/ScottPlot5/ScottPlot5/Primitives/FontStyle.cs
bclehmann: I agree that it's unfortunate that we have to preserve a 2 week old API (and one that I don't think really adds much that SkiaSharp doesn't do already). But I think we're stuck with it, the method has shipped and so it should be preserved at least until there's a new major version. The effort of preserving this pales in comparison to the effort of removing it once you consider the support side of things. However, I would recommend annotating it with `[Obsolete]`.

src/ScottPlot5/ScottPlot5/Fonts.cs
bclehmann: This is for custom fonts is it not? A user can request a font by name (e.g. Arial), and provided that it is installed on their machine, they will get Arial. Even though Arial is (to my knowledge) not a standard font. This isn't for retrieving the Sans Serif font that ScottPlot would use by default, I don't see a reason why that code shouldn't be left as it is.

swharden:
Hey @kebox7, thanks for this awesome PR and @bclehmann for your insightful feedback! I'm about to pick this up now and I'll make a few commits of my own to this PR, resolve the open comments, and merge hopefully in the next hour or so.

I will give a little feedback on this comment though, because it's an important topic that comes up from time to time...

> the method has shipped and so it should be preserved at least until there's a new major version

If ScottPlot were a commercial product that a lot of customers pay for, I would tend to agree more with this sentiment. If paid to develop and support this thing, I would happy support all those existing users 🤑 Maybe one day we'll be there, but it's not quite where I am today.

In practice I tend to be all right with breaking changes as long as (1) they make the API simpler in the long run, (2) properties are marked obsolete where possible, (3) they're deep in the code base in areas already only touched by advanced users.

I recognize the advantages of strictly adhering to semantic versioning and not breaking things, but I have to balance that sentiment with one where I have to support this thing on my nights and weekends 🤪 so whatever makes this thing simplest to support in the upcoming years I'm generally in favor of 🚀

src/ScottPlot5/ScottPlot5/Fonts.cs
swharden: > This has a slightly different meaning

> This is for custom fonts is it not?

These are some good points! ... and an example of where "naming is hard". One could say that a default system font made bold has been "customized", so the language is a bit squishy. I may rename this class to `FontFileResolver` and supply it with ScottPlot, but not use it as the default because the default behavior is to use system fonts.

src/ScottPlot5/ScottPlot5/Fonts.cs
swharden: I'll probably side with the simplest option (for now at least) and make this a basic getter/setter without a backing field. It's already nullable, so that signals to the user that it may be null (and also that newing classes may require disposal considerations on behalf of the one doing the newing)

src/ScottPlot5/ScottPlot5/Primitives/FontStyle.cs
swharden: See https://github.com/ScottPlot/ScottPlot/pull/3841#issuecomment-2123580075

swharden:
I'm going to merge this because I think it's mostly in an all right state, but feel free to refine my changes with new pull requests! I worked on this for a while and it's a bit late here so it could be that I overlooked something.

I changed the nullable `FontResolver` to a readonly `List<IFontResolver>`  to add support for using multiple TTF files simultaneously. I also took @bclehmann's advice and moved the system font lookup logic into its own resolver that implements `IFontResolver`.

image

Note that this strategy adds support for font styles beyond bold and italic (e.g., light)

src/ScottPlot5/ScottPlot5/Fonts.cs
```diff
@@ -1,105 +1,103 @@
...
 
-    #region PRIVATE
-
-    private static HashSet<string> GetInstalledFonts()
+    [Obsolete("To determine if a font exists, call GetTypeface() and check for null", true)]
+    public static bool Exists(string fontName, bool bold, bool italic)
     {
-        return new(SKFontManager.Default.FontFamilies, StringComparer.InvariantCultureIgnoreCase);
+        throw new NotFiniteNumberException();
     }
```
kebox7, line 66: Maybe `NotImplementedException` is better?

PR #3846, opened 2024-05-18
Add option for clear type text

bforlgreen:
Clear type text is a process to make text appear crisper at the same resolution. Whilst basic antialiasing makes text look less pixelated, it can also then make text look more blurry. Clear type is an extension to antialiasing that uses colours to "trick" us into perceiving the text is crisper.  

On the left, no AA, so looks pixelated.
In the middle, AA has been applied (but in certain situations, like with a grey colour, it can look blurry)
On the right, AA with clear type

image

To enable ClearType in Skia, you have to set LcdRenderText on the Paint class to true, which will then enables "SubpixelAntialias". I am not sure why they use the name "LcdRenderText" instead of 'SubpixelAntialias'.

Unfortunately, this is not the only change you have to make to enable ClearType. You also have to set the SkSurfacesProperties within the SKElement class. To achieve this, you kind of have to copy the the SKElement class from the SkiaSharp code base.

Results ( you will need to zoom in)
image

### DIFF
--- a/src/ScottPlot5/ScottPlot5/Primitives/Label.cs
+++ b/src/ScottPlot5/ScottPlot5/Primitives/Label.cs
@@ -32,6 +32,12 @@ public class Label
     public bool AntiAliasBackground { get; set; } = true;
     public bool AntiAliasText { get; set; } = true;
 
+    /// <summary>
+    /// To make use of ClearType text, you must also set the  SKSurfaceProperties 
+    /// of your SKSurface to be something other than SKPixelGeometry.Unknown
+    /// </summary>
+    public bool SubPixelAntiAliasText { get; set; }
+
     // TODO: use a class for cached typeface management
 
     private SKTypeface? CachedTypeface = null;
@@ -170,6 +176,7 @@ public class Label
         paint.TextSize = FontSize;
         paint.Color = ForeColor.ToSKColor();
         paint.IsAntialias = AntiAliasText;
+        paint.LcdRenderText = SubPixelAntiAliasText;
         paint.Shader = null;
     }
 


src/ScottPlot5/ScottPlot5/Primitives/Label.cs
```diff
@@ -32,6 +32,12 @@ public class Label
     public bool AntiAliasBackground { get; set; } = true;
     public bool AntiAliasText { get; set; } = true;
 
+    /// <summary>
+    /// To make use of ClearType text, you must also set the  SKSurfaceProperties 
+    /// of your SKSurface to be something other than SKPixelGeometry.Unknown
+    /// </summary>
+    public bool SubPixelAntiAliasText { get; set; }
```
bclehmann, line 39: Wonder if we should make this an enum, since setting `LcdRenderText` to true has no effect unless `paint.IsAntialias` is also true.

I'm thinking something along the lines of 
```cs
enum TextAntialiasType {
  None,
  Antialias,
  SubpixelAntialias,
}
```

Then we can set the booleans but allow a simpler API. But we'll need to keep `AntiAliasText` to keep backwards compatibility. I think we should use getters and setters to enforce the current behaviour (if `AntiAliasText` is set to true we get the current AA behaviour, if it's false we get no AA). Not sure if setting it to true while subpixel AA is on should have no effect or if it should be set to non-subpixel AA.

src/ScottPlot5/ScottPlot5/Primitives/Label.cs
```diff
@@ -32,6 +32,12 @@ public class Label
     public bool AntiAliasBackground { get; set; } = true;
     public bool AntiAliasText { get; set; } = true;
 
+    /// <summary>
+    /// To make use of ClearType text, you must also set the  SKSurfaceProperties
```
bclehmann, line 36: I believe this is something that has to be enabled on the control, and I don't like the idea of having AA flags being placed in two very different places. I think the solution is probably making some `TextAntialiasType` a global field which is read in both places (maybe on the `Plot` class?). And if necessary we can preserve some way for the user to mix unantialiased text with AA text. But I'm not convinced that the user would want something like that.

bforlgreen: Yes, the problem is that you have to be able to set SKSurfaceProperties when creating the SKSurface, which you can only do by implementing your own SKElement class (SKElement is implemented in SkiaSharp)

bclehmann: We already use `SKElement` on (most of) the controls. For example, see https://github.com/ScottPlot/ScottPlot/blob/main/src/ScottPlot5/ScottPlot5%20Controls/ScottPlot.WPF/Themes/Generic.xaml

Worst case scenario, we can subclass `SKElement` to provide our own `OnRender` function (relevant code for WPF is here: https://github.com/mono/SkiaSharp/blob/main/source/SkiaSharp.Views/SkiaSharp.Views.WPF/SKElement.cs). But I'm admittedly not that pleased at the idea of duplicating most of the function just to allow a slightly different `SKSurface.Create` call. Granted it doesn't look like these views are changed much, but it doesn't fill me with joy. I do also see some private methods and fields which we'd need to copy-paste for this to work.

This could also be achieved by copying , but that's not ideal, one of the big benefits of ScottPlot 5 is we don't normally need to copy bitmaps between the plot and the controls.

PR #3873, opened 2024-05-24
Improve SharedAxes.

StendProg:
The previous version of SharedAxes didn't work very well. They were synchronized by timer to avoid recursive calls. As a result, synchronization occurred with a large visible delay.

To solve the problem of recursive call of axis change when linking axes, a special method SetSilent() in Range was added. Range now keeps track of any of its changes by setting a special 'WasChanges' flag. The only method that allows you to change the Range without setting the flag is 'SetSilent()'.

So to synchronize the second linked axis we need to set its limits with 'SetSilent', and then its change will not trigger the axis change event of the second Plot. This way we break the recursion, and we can synchronize several Plot very quickly (without timer).

I have not managed to overcome recursive call (leading to an exception) at initial Plot initiation. That's why event subscription is delayed with the help of a timer.

https://github.com/ScottPlot/ScottPlot/assets/53831487/3848dd81-a64a-4d46-9cc9-b1ba3240dfe4


swharden:
Hi @StendProg, thanks for working on this! This is a much better solution. Shared axes are a common question in issues, so I'm going to put some extra time into this PR to try to adapt the demo window to be as educational and simple as possible.

I really like this solution, but I'm going to keep working to try and find a way to simplify this line:
```cs
formsPlot1.Plot.Axes.Bottom.Range.SetSilent(formsPlot2.Plot.Axes.Bottom.Range.Min, formsPlot2.Plot.Axes.Bottom.Range.Max);
```

> I have not managed to overcome recursive call (leading to an exception) at initial Plot initiation. That's why event subscription is delayed with the help of a timer.

I find that this works! It's simpler than setting-up a one-shot timer.

```cs
if (formsPlot1.Plot.RenderManager.RenderCount == 0)
    return;
```

swharden:
This strategy isn't necessarily better, it's just different. It lets users use more traditional methods for setting axis limits, and also avoids the infinite loop at initialization.

```cs
private void CopyLimits(IPlotControl source, IPlotControl target)
{
    AxisLimits sourceLimits = source.Plot.Axes.GetLimits();

    if (checkShareX.Checked)
        target.Plot.Axes.SetLimitsX(sourceLimits.Left, sourceLimits.Right);

    if (checkShareY.Checked)
        target.Plot.Axes.SetLimitsY(sourceLimits.Bottom, sourceLimits.Top);

    target.Plot.RenderManager.DisableAxisLimitsChangedEventOnNextRender = true;

    target.Refresh();
}
```

swharden:
...after all that, I ended-up removing the "CoordinateRangeMutableSilent" class. Although the solution worked well, the primary goal of this PR was achieved by removing the shared axis timer (improving performance) and leaning on the new `DisableAxisLimitsChangedEventOnNextRender` flag (preventing infinite loop at startup) and I think this is a simpler solution.

Also I may want to refactor `MutableAxisLimits` in the future (possibly storing axis limits in readonly structs), so I am hesitant to encourage inheritance by making its properties `virtual`.

Thanks again for your help with this PR! I'm happy that the shared axis demo is now significantly better 😎

StendProg:
You've come at it from a completely different angle, but it works just as well 🚀

PR #3878, opened 2024-05-27
Fix avalonia performance on Linux

### LINKED ISSUE #3439: ScottPlot5 Avalonia control is very slow on Linux
**Question:** ScottPlot5 skiasharp is terribly slow on Linux (Debian 12 (linux-x64), i3-10105, Intel UHD Graphics 630).

**ScottPlot.Avalonia Version:** 5.0.21
**SkiaSharp.NativeAssets.Linux.NoDependencies Version:** 2.88.7
**Avalonia Version:** 11.0.6

**Code Sample:**

```xaml
<ScottPlot:AvaPlot Name="avaPlot1"/>
```

```cs
        avaPlot1.Plot.Title("UDP data graf");

        avaPlot1.Plot.XLabel("Time [s]");
        avaPlot1.Plot.ShowLegend();

        // create a timer to update the GUI
        _renderTimer = new DispatcherTimer();
        _renderTimer.Interval = TimeSpan.FromMilliseconds(20);
        _renderTimer.Tick += Render;
        _renderTimer.Start();
```
```cs
void Render(object sender, EventArgs e) {
avaPlot1.Refresh();
}
```

When I use ScottPlot4 or ScottPlot5 on Windows then the UI page with the Scottplot plot performance is smooth and fast. The Scottplot5 is even faster than Scottplot4 on Windows. When I use the same on Linux (same machine) then the application is terribly slow, regardless the refreshing render time period. Scottplot4 performance is satisfactory on Linux.

The rendering time (FPS) get by dobble-click in graph is more than 100 times worse on Linux.

- I am using **dotnet** 6.0.
Build command: `dotnet publish -r linux-x64 -c Debug -p:publishsinglefile=true --self-contained true` 

The following packages **libfreetype6**, **libfontconfig1** and **fontconfig** have been installed.

I am curious whether the Scottplot5 is using the GPU acceleration or what I am doing wrong.

Thank you for support.

pkstrsk:
Thanks to @zii-dmg for finding the solution for the issue with unset CachedTypeface causing very slow performance under Avalonia in Linux.

Tested on RaspberryPi 5 under DRM, performance improvement from 1-3FPS to 40-80FPS rendering speed on large data set.

Fixes #3439

### DIFF
--- a/src/ScottPlot5/ScottPlot5/Primitives/Label.cs
+++ b/src/ScottPlot5/ScottPlot5/Primitives/Label.cs
@@ -77,6 +77,19 @@ public class Label
         FontName = Fonts.Detect(Text);
     }
 
+    private SKTypeface? CachedTypeface = null;
+
+    private SKTypeface Typeface
+    {
+        get
+        {
+            if (CachedTypeface is null)
+            {
+                CachedTypeface = Fonts.GetTypeface(FontName, Bold, Italic);
+            }
+            return CachedTypeface;
+        }
+    }
 
     private void ApplyPointPaint(SKPaint paint)
     {


src/ScottPlot5/ScottPlot5/Primitives/Label.cs
```diff
@@ -77,6 +77,19 @@ public void SetBestFont()
...
+    {
+        get
+        {
+            if (CachedTypeface is null)
+            {
+                CachedTypeface = Fonts.GetTypeface(FontName, Bold, Italic);
+            }
+            return CachedTypeface;
+        }
+    }
```
bclehmann, line 92: Forgive me if I'm missing something, but this isn't used anywhere is it? I believe #3841 changed `Label` such that this no longer does anything.

I also believe #3841 is also supposed to have addressed the performance concerns? I believe I tested it when it came out, but I'd appreciate if you could confirm that it helped? Unfortunately I tested it under WSL, which isn't ideal for making performance conclusions.

swharden:
Following-up, it looks like this was implemented in #3841

https://github.com/ScottPlot/ScottPlot/issues/3439#issuecomment-2133926779

Thanks @pkstrsk for your input along the way! It helped make that large PR a success and I'm looking forward to releasing these new features in a new version shortly 😎

PR #3879, opened 2024-05-27
Disable() double call fix.

StendProg:
A double call to `Interactions.Disable()` resulted in the fact that interactions could not be restored with `Enable()` afterwards.

In this PR a special flag `Disabled` was introduced. It is now possible to call `Disable()` many times and then restore interactions with `Enable()` afterwards.

### DIFF
--- a/src/ScottPlot5/ScottPlot5/Control/Interaction.cs
+++ b/src/ScottPlot5/ScottPlot5/Control/Interaction.cs
@@ -11,6 +11,11 @@ public class Interaction(IPlotControl control) : IPlotInteraction
     public IPlotControl PlotControl { get; private set; } = control;
 
     /// <summary>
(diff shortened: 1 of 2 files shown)


src/ScottPlot5/ScottPlot5/Control/Interaction.cs
```diff
@@ -51,6 +60,7 @@ public void Disable()
     public void Enable()
     {
         Actions = ActionsWhenDisabled;
```
KroMignon, line 62: I would suggest to change `Disable()` to only use `ActionsWhenDisabled` if it was disabled:
```csharp

public void Enable()
{
    if (Disabled)
        Actions = ActionsWhenDisabled;
    Disabled = false;
}
```

PR #3919, opened 2024-06-08
Snap to ticks axis rule fixed

BrianAtZetica:
I previously implemented the axis rules SnapToTicksX and SnapToTicksY in #3547 . 
These had a few limitations, most notably:
- panning was very jumpy
- the tick interval sometimes changed as a result of the snap, which meant is wasn't snapping to a tick every time
- it was untested with inverted axes
- It had had only limited testing in combination with other rules

Subsequently this rule was simplified, which removed most of the special case handlign, and resulted in it functioning only for zooming out (zooming in was effectively prevented). This was discussed in #3701.

These changes:
- reinstate the original functionality
- handle panning so that it is intuitive and smooth
- check for and correct the occasional situation where tick interval changes during the snapping
- ensure the rules work for inverted axes (although problems might occur if the axes are inverted while the rule is already set)
- ensure the rules work with an end limit set.

I confess, this was a challenge, and the code does not read well. I've tried to document intention with comments as much as I can, but if it is still unclear, I welcoime a discussion an dsome help in writing simpler logic (without loosing functionality). 

The final commit in this branch is once again an updated sandbox project to easily test the rules in combination. I understand that won't be merged.


swharden:
Thanks @BrianAtZetica! I'm not too worried about the complexity since it's all bundled neatly inside those classes and it can always be refactored later if/as needed. I'll merge this in now and it will be in the next release 👍

Issue #3812, opened 2024-05-08
SignalXY: last point sometimes does not get rendered


swharden:
Perhaps because a binning issue?
![sigxylast](https://github.com/ScottPlot/ScottPlot/assets/4165489/fe13f7f8-ec59-4a7a-b727-6e657cb92166)

swharden:
This is reproducible and only affects "high density mode"

```cs
using ScottPlot;

namespace Sandbox.WinForms;

public partial class Form1 : Form
{
    public Form1()
    {
        InitializeComponent();

        int count = 1000;
        double[] xs = Generate.Consecutive(count);
        double[] ys = Generate.RandomWalk(count);
        ys[^1] = -100;

        formsPlot1.Plot.Add.SignalXY(xs, ys);
    }
}
```

swharden:
Interestingly, it's only the very last point, and only about half the frames. It "flickers" as you zoom horizontally.

![image](https://github.com/ScottPlot/ScottPlot/assets/4165489/2c3a9ad7-45c3-4c7d-af65-fa89d23b3edd)

```cs
using ScottPlot;

namespace Sandbox.WinForms;

public partial class Form1 : Form
{
    public Form1()
    {
        InitializeComponent();

        int count = 1000;
        double[] xs = Generate.Consecutive(count);
        double[] ys = Generate.RandomWalk(count);
        ys[0] = -100;
        ys[499] = -100;
        ys[999] = -100;

        var sig = formsPlot1.Plot.Add.SignalXY(xs, ys);
        sig.MarkerShape = MarkerShape.OpenCircle;
        sig.MarkerColor = Colors.Black;
        sig.MarkerSize = 10;
    }
}
```

swharden:
I'm pretty sure the issue is in here

https://github.com/ScottPlot/ScottPlot/blob/6468dbe93b7cb5dd686183cc0329d1f93fed35b9/src/ScottPlot5/ScottPlot5/DataSources/SignalXYSourceDoubleArray.cs#L156-L187

swharden:
I'm thinking the issue is here: https://github.com/ScottPlot/ScottPlot/blob/6468dbe93b7cb5dd686183cc0329d1f93fed35b9/src/ScottPlot5/ScottPlot5/DataSources/SignalXYSourceDoubleArray.cs#L170

For the last point, it's possible where the last pixel column contains a single value, so start and stop index positions will be the same

BrianAtZetica:
[...]

PR #3921, opened 2024-06-09
BugFixes to Rotated SignalXY plots

BrianAtZetica:
Reolution for #3812 had introduced an incorrect final pixel location for rotated SignalXY plots.
This was resolved by a swapping of Xs and Ys coordinates when defining the additional "lastPoint" value. 
_Note, we could consider renaming Xs and Ys to something like independentValues and dependentValues respectively. Not sure if that will be intuitove to all developers though._

Using the sample code provided in comment on #3812  the fix provided no longer shows a distored line at top of the plot:
image

Second bug was pre-existing, and became apparent when zoomed into the series to observe the final pixel. 
Rotated SignalXY series were performing an incorrect interpolation for the final pixel location (another mix-up of Xs and Ys).

Code to demonstrate this issue is as follows:

```cs

        InitializeComponent();

        double[] X = new double[2];
        double[] Y = new double[2];
        X[0] = -10;
        X[1] = 10;
        Y[0] = 1;
        Y[1] = 2;

        var signal = WpfPlot1.Plot.Add.SignalXY(X, Y, color: ScottPlot.Colors.Black);
        signal.Data.Rotated = false;
        var signalrotated = WpfPlot1.Plot.Add.SignalXY(X, Y, color: ScottPlot.Colors.Black);
        signalrotated.Data.Rotated = true;
```

This plots a normal SignalXY and a rotated SignalXY on the same graph.
Before the bug fix, this becomes distorted when panning the content up or down:

image

After the bug fix the plot keeps its cross shape while scrolling off the page in any direction:
image

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -32,13 +32,14 @@ _Not yet on NuGet..._
 * Scatter: Added `GetNearestX()` to the data source to help locate the point closest to the cursor's X position (#3807) @MatKinPro
 * Controls: Disable middle-click-drag zooming on axes which have no data (#3810, #3897) @MCF
 * DataLogger: Create `Add()` overloads which accept fixed-length arrays (#3555) @h25019871990
-* SignalXY: Ensure the final point is always drawn in high density mode (#3812)
+* SignalXY: Ensure the final point is always drawn in high density mode (#3812, #3812)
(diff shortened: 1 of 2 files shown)


BrianAtZetica:
Discovered there is still a problem if Y axis is inverted. Working on this further...

PR #3922, opened 2024-06-09
Forcing range for clipping of heatmap visualization

### LINKED ISSUE #3862: SP5 Heatmap: How to calculate Heatmap not from 0, but from the minimum data value
**Question:** How to calculate Heatmap not from 0, but from the minimum data value?
The problem is that the values that come to me may not be distributed from 0, but, for example, in the interval between 80-100% (where 100% is the maximum data value). I also use a Colorbar based on this Heatmap, which would also be a good idea to start with a minimum value so as not to confuse the end user. How to do it?
Next, I will provide screenshots of graphs (two different sets of values) from Excel with my data distribution (for ease of perception, I built a graph reflecting what percentage the number is of the maximum) and with those obtained after the Heatmap + ColorBar pair.
**_Incorrect:_**
image

image
Please ignore the green bar under the "red graph", this is a piece of my form on which I place the Heatmap and ColorBar

**_Correct:_**
image
image

sdpenner:
Added ForcedRange and UseForcedRange properties to heatmap and lazy initialization of the range calculation, so the range is not calculated each time update() is called. Closes #3862

```cs
        double[,] intensityMap = SampleData.MonaLisa();

        ScottPlot.Range forcedRange = new(50, 150);
        var map = WpfPlot1.Plot.Add.Heatmap(intensityMap, forcedRange);
        map.Colormap = new ScottPlot.Colormaps.Turbo();
        WpfPlot1.Plot.Add.ColorBar(map);
```

mona_clipped


src/ScottPlot5/ScottPlot5 Sandbox/Sandbox.WPF/MainWindow.xaml.cs
```diff
@@ -8,7 +9,14 @@ public partial class MainWindow : Window
     public MainWindow()
     {
         InitializeComponent();
-        WpfPlot1.Plot.Add.Signal(Generate.Sin());
-        WpfPlot1.Plot.Add.Signal(Generate.Cos());
+        //WpfPlot1.Plot.Add.Signal(Generate.Sin());
```
bclehmann, line 12: I assume this was just for testing, but I think this file should be returned to its original state

src/ScottPlot5/ScottPlot5/Plottables/Heatmap.cs
```diff
@@ -258,7 +269,23 @@ public double GetValue(Coordinates coordinates)
...
+        {
+            // Needs update.  Get the range from intensities.
+            return Range.GetRange(Intensities); 
+        }
+        return IntensityRange;
+    }
+
+    public Range IntensityRange { get; private set; }
+    public Range ForcedRange { get; set; }
+    public bool UseForcedRange { get; set; }
```
bclehmann, line 288: I don't know if I like having a `UseForcedRange` in addition to `ForcedRange`. A user who has logic to switch the range around on the same heatmap (this is common in image processing, it's essentially a form of contrast-adjustment) will have to maintain two values that do essentially the same thing. This is especially fiddly if they're loading saved presets, which does happen a lot if they're dealing with the same kind of images which is pretty common for scientific heatmaps. This is actually how the TSA uses their X-rays, they have a bunch of preset ranges that show specific contraband better on their screens.

You could imagine how a user might set `ForcedRange` but not `UseForcedRange`, which will have no effect. Or, they could set `UseForcedRange` but not `ForcedRange`, which I believe will result in a range of [0,0], which will give a useless heatmap (if it doesn't crash from divide by zero, I haven't checked). This may sound pedantic, but I think we can eliminate this possibility by using a `Range? ForcedRange`. If it's null, it's disabled, if it's non-null, it's enabled. It becomes impossible to enable it without setting a range, and impossible to set a range without enabling it.

src/ScottPlot5/ScottPlot5/Plottables/Heatmap.cs
```diff
@@ -258,7 +269,23 @@ public double GetValue(Coordinates coordinates)
...
     public IEnumerable<LegendItem> LegendItems => Enumerable.Empty<LegendItem>();
 
-    public Range GetRange() => Range.GetRange(Intensities);
+    public Range GetRange()
+    {
+        if (UseForcedRange)
+        {
+            return ForcedRange;
+        }
+        if (IntensityRange == default(Range))
```
bclehmann, line 278: This check doesn't catch if the heatmap has had its intensities changed through an `Update` call. It will only catch if a range has never been set (i.e. on first render).

bclehmann: Worth noting that `GetRange` is only called in this class when an `Update` is requested and we need to invalidate the cached bitmap. So it's only the consumers outside of this class that would benefit, and I think the only one that would include is `ColorBar`, which provides colour axes.

If you're interested, `ColorBar` probably has some room for performance improvements since it doesn't currently even cache its image (which depends only on its colour map), nor its axis, which depends only on its range and configured edge. I imagine that caching the bitmap in particular would have bigger gains just because creating it involves some not insubstantial heap allocations. Including a pinned heap allocation at that, which I think the garbage collector doesn't like as much.

bclehmann (commented):
Thanks for this. Just one thing about your cached range and the rest is stylistic.

src/ScottPlot5/ScottPlot5/Plottables/Heatmap.cs
swharden: > but I think we can eliminate this possibility by using a `Range? ForcedRange`

I'll make this change now and push to this PR 👍

src/ScottPlot5/ScottPlot5 Sandbox/Sandbox.WPF/MainWindow.xaml.cs
swharden: I'll revert this and move this demo code into a cookbook recipe 👍

src/ScottPlot5/ScottPlot5/Plottables/Heatmap.cs
swharden: I made `ManualRange` nullable and have it automatically call `Update()` in the setter. I think this avoids the concerns mentioned above...

swharden:
I simplified this PR a good amount and will merge it in now. I think the intended functionality remains, but refinements are welcome in subsequent PRs if the behavior can still be improved.

Thanks @sdpenner for this PR and @bclehmann for your feedback along the way!

```cs

// sample data values range from 0-255
double[,] data = SampleData.MonaLisa();

// add a heatmap and colorbar to the plot
var hm = myPlot.Add.Heatmap(data);
hm.Colormap = new ScottPlot.Colormaps.Turbo();
myPlot.Add.ColorBar(hm);

// force the colormap to span a manual range of values
hm.ManualRange = new(50, 150);

myPlot.SavePng("demo.png", 400, 300);
```

image

sdpenner:
This is all good fellas.  Thanks for the feedback and simplifications.  I didn't notice a lot of nullable types in the code so I used the property but I agree using the nullable range is more elegant.  Also, creating the range during construction makes sense as the intensity array is not settable...That being said, it may make sense to make the array settable in the future as this better aligns with MVVM setting of the heatmap image.

PR #3924, opened 2024-06-10
Update Color.cs

r-j-s:
Added typecast in "public static Color FromARGB(int argb)" to prevent an infinite loop.


swharden:
Thanks @r-j-s, this is a great fix! 🚀

PR #3946, opened 2024-06-14
Implement a new DataLogger

drolevar:
There are two main goals behind this Datalogger:

- Improve performance by bringing in binary search.
- Support circular buffers, which is crucial when logging for an extended period of time.


drolevar:
@swharden Will be very nice to get your thoughts on that. The main reason for making this change for me personally was to limit memory consumption when running for very long periods. That's why I went on with adding CircularBuffer.

swharden:
Hi @drolevar, thanks for this PR! I'm starting to run out of time today and may have limited time tomorrow 😅 but this PR is definitely next up for me to review 👍 

I took a quick look in my browser and this is looking great! I like how you created a dedicated `CircularBuffer` class too. Two quick thoughts before I give it a more thorough review...

1) Where did you source `CircularBuffer.cs` from? It seems copy/pasted from something Microsofty but I'd love to include a URL to the exact source

2) Perhaps we should incorporate this behavior into the `DataStreamer` class (which already has a hacky circular buffer) instead of creating a duplicate DataLogger for this purpose? I haven't studied this PR in detail yet so there may be an obvious reason this wasn't done, but I thought I'd mention it just in case

swharden:
Hi @drolevar, I haven't heard anything in a little while and this PR is still marked "draft", however it's looking pretty good so I'm going to take a few minutes now to modify it slightly and merge it in. If you had additional work you were planning to do to these classes we can start a new PR and take a closer look.

Thanks again for this new datalogger! This is looking fantastic 🚀

swharden:
> Where did you source CircularBuffer.cs from? It seems copy/pasted from something Microsofty but I'd love to include a URL to the exact source

This looks close, but not identical
https://github.com/dotnet/aspire/blob/main/src/Shared/CircularBuffer.cs

swharden:
Hi @drolevar, thanks for the follow up! I'll throw an exception when rotation is assigned to and we can add that functionality if it's desired in the future 👍

drolevar:
@swharden Do you have an idea of what has to be done do take it out of the experimental?

swharden:
> @swharden Do you have an idea of what has to be done do take it out of the experimental?

Probably update the following sections:
* https://scottplot.net/demo/5.0/#plot-streaming-data
* https://scottplot.net/demo/5.0/#plot-growing-data

To clarify how this new class is different from the existing ones. To be honest, I'm still uncertain about this myself. 

If the new class can simply replace the existing `DataStreamer`, that's an even better option, as no changes to documentation would be required!

A few messages up you said "The reason I didn't use DataStreamer is that it didn't have the functionality I needed, however at the moment I don't remember what that was.", so maybe just clarifying that point will let us deprecate the old one and replace it with the new one

jpgarza93:
This would be amazing on a DataLogger along with the nice ViewSlide() functionality. 

I also need this functionality. I currently achieve a circular buffer by calling `DataLoger.Data.Coordinates.RemoveAt(0);` but this is not the best approach and comes with some issues: https://github.com/ScottPlot/ScottPlot/issues/3969

Issue #3956, opened 2024-06-17
SVG XML has issues with unicode characters like superscripts


aespitia:
Issue: SVG XML has issues with unicode characters like superscripts

ScottPlot Version: latest

if you use unicode characters like superscripts in labels and things, they exports as ?? on the SVG XML version.

aespitia:
another example is the "reigstered" mark

swharden:
resolved in #3957

PR #3957, opened 2024-06-17
SVG XML Updates

aespitia:
#3956

there are issues in the rendered SVG XML when you use unicode characters, such as superscripts.  This update is to keep a backward compatible SVG XML export, it's currently ascii, this introduces an optional parameters to the GetXml functions to export using UTF8 instead.

### DIFF
--- a/src/ScottPlot5/ScottPlot5/Primitives/SvgImage.cs
+++ b/src/ScottPlot5/ScottPlot5/Primitives/SvgImage.cs
@@ -1,4 +1,7 @@
-﻿namespace ScottPlot;
+﻿using SkiaSharp;
+using System.IO;
+
+namespace ScottPlot;
 
 internal class SvgImage : IDisposable
 {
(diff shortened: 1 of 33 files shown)


aespitia:
@swharden  would like feedback, if I should just make the default UTF8 and not bother with the backward compatibility?

src/ScottPlot5/ScottPlot5/Primitives/SvgImage.cs
```diff
@@ -17,9 +20,20 @@ public SvgImage(int width, int height)
         Canvas = SKSvgCanvas.Create(rect, Stream);
     }
 
-    public string GetXml()
+    public string GetXml(bool isUtf = false)
     {
-        return Encoding.ASCII.GetString(Stream.ToArray()) + "</svg>";
+        Canvas.Flush();
+
+        if (isUtf)
```
bclehmann, line 27: I think this might all be unnecessary, since the user isn't able to pass in ASCII, only unicode. Plus ASCII is forwards-compatible with utf8 as long as the top bit of each byte is zero (and if it isn't it's non-standard so there's not really a standard that we need to try and stick to).

I think we can just swap out ASCII for UTF8 across the board without needing any option to switch between them.

aespitia: updated PR to simplify the changes

Issue #3943, opened 2024-06-13
SavedImageInfo: add ConsoleWriteFIlename()


swharden:
to allow something like this to print the full file path to the console

```cs
plot1.SavePng("test.png", 600, 400).ConsoleWriteFilename();
```

right now this works

```cs
var file = plot1.SavePng(saveAs, 600, 400);
Console.WriteLine(file.Path);
```

swharden:
Also `CopyPathToClipboard()` and `CopyImageToClipboard()`

swharden:
resolved in #3965

PR #3965, opened 2024-06-18
adding console write file name function

aespitia:
adding requested functionality for writing the file path to the console, briefly looked at the clipboard request too, but it may be a little more involved for different systems, or have to introduce another library to abstract the clipboard, so, i didn't want to introduce a new dependency.  

#3943

```cs
using ScottPlot;

Plot plt = new();
plt.Add.Signal(Generate.Sin());
plt.Add.Signal(Generate.Cos());

plt.SavePng("test.png", 600, 300).ConsoleWriteFilename();
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@ _Not yet on NuGet..._
 * SVG: File encoding now supports text containing UTF8 characters (#3956, #3957) @aespitia
 * Documentation: Added a sandbox .NET API project and quickstart section to the website (#3959, #3824) @aespitia
 * Color: Added `ToColor()` and `FromColor()` to simplify conversion between `ScottPlot.Color` and `System.Drawing.Color` (#3964, ##3953) @aespitia
+* Console: Saved image path can be displayed by calling `myPlot.SavePng('demo.png', 600, 400).ConsoleWritePath()` (#3965, #3943) @aespitia
 
 ## ScottPlot 5.0.35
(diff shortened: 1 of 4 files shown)


swharden:
Thanks @aespitia! I think I'll rename `ConsoleWriteFilename` to `ConsoleWritePath` because the full path (not just the filename) will be shown. Also I think I can solve the clipboard issue by adding extension methods to the various control projects. I'll give that a quick go and merge shortly!

swharden:
Note: to get copy-to-clipboard to work in a console app you need to modify the `.csproj` file to target Windows, use Windows Forms libraries, and add the `STAThread` attribute to the main method like:

```xml
<Project Sdk="Microsoft.NET.Sdk">

    <PropertyGroup>
        <OutputType>Exe</OutputType>
        <TargetFramework>net8.0-windows</TargetFramework>
        <ImplicitUsings>enable</ImplicitUsings>
        <Nullable>enable</Nullable>
        <IsPackable>false</IsPackable>
        <UseWindowsForms>true</UseWindowsForms>
    </PropertyGroup>
...
```

```csusing ScottPlot;
using ScottPlot.WinForms;

public static class Program
{
    [STAThread]
    public static void Main()
    {
        Plot plt = new();
        plt.Add.Signal(Generate.Sin());
        plt.Add.Signal(Generate.Cos());

        plt.SavePng("test.png", 600, 300).CopyToClipboard();
    }
}
```

PR #3978, opened 2024-06-21
Signal: improve support for IReadOnlyList<T>

### LINKED ISSUE #3942: Signal: Add support for `IReadOnlyList<T>`
Presently no Add.Signal() overload supports this

sdpenner:
Added IReadOnlyList adder for creating a new signal.  Just creates a generic list from the read only one.

Not sure this makes sense, as users could easily do the ToList() themselves...

closes #3942

```cs
   public Signal Signal<T>(IReadOnlyList<T> ys, double period = 1, Color? color = null)
   {
       SignalSourceGenericList<T> source = new(ys.ToList(), period);
       return Signal(source, color);
   }
```


swharden:
Thanks @sdpenner! I refined this PR to achieve an implementation that avoids the copying associated with `ToArray()`

PR #3994, opened 2024-06-25
Fix interaction of axis panels when scale factor is more than 1

bforlgreen:
If you set the ScaleFactor to a value greater than 1, then interaction with the Axis Panels (mouse interactions) was incorrect.

To repro the issue (without these changes), set the scale factor to something like 1.5 and then try to perform a marquee select in the **YAxis only** in the area pointed to by the arrow. You'll notice it doesn't work.

image

These changes scale the mouse pos like we do with all the other functions.

I have also cleaned up some of the Scale factor code

- removed unnecessary checks of Scale factor == 1 (multiply or divide by 1 wont change anything)
- added helper methods to reduce some of the duplicate code.


swharden:
Thanks @bforlgreen, this looks great! I'll merge it in a few minutes 🚀

Issue #3985, opened 2024-06-23
CoordinateRect: add constructor overload that accepts `IAxes`


swharden:
This works:
```cs
CoordinateRect rect = new(Axes.XAxis.Range, Axes.YAxis.Range);
```

but it would be nice to be able to do this:
```cs
CoordinateRect rect = new(Axes); // where Axes is a class that implements IAxes
```

The new constructor can be added below this line
https://github.com/ScottPlot/ScottPlot/blob/fb3b1c84af3c361bf1eb8f4286ad1d4e7038afc5/src/ScottPlot4/ScottPlot/CoordinateRect.cs#L20-L27

swharden:
resolved in #4008

PR #4008, opened 2024-06-27
CoordinateRect: add constructor overload that accepts IAxes

CoderPM2011:
I added `CoordinateRect(IAxes axes)`.
It actually uses `CoordinateRect(CoordinateRange xRange, CoordinateRange yRange)`.
Therefore, I placed it below it.

#3985


swharden:
Thanks @CoderPM2011! 🚀

PR #4014, opened 2024-06-28
Rendered FPS text position

banncan:
fix rendered FPS text position


swharden:
Thanks for this fix @banncan!

### Before

image

### After

image

Issue #3968, opened 2024-06-20
Axis span: ensure rendered rectangle is at least 1px wide


swharden:
If spans are rendered as <1px they may become invisible.

Add logic so the minimum rendered axis span is always at least 1 px in size.

Lines of code to modify are here:

https://github.com/ScottPlot/ScottPlot/blob/fb3b1c84af3c361bf1eb8f4286ad1d4e7038afc5/src/ScottPlot5/ScottPlot5/Plottables/VerticalSpan.cs#L31

https://github.com/ScottPlot/ScottPlot/blob/fb3b1c84af3c361bf1eb8f4286ad1d4e7038afc5/src/ScottPlot5/ScottPlot5/Plottables/HorizontalSpan.cs#L32

PR #4017, opened 2024-06-28
Axis span: ensure rendered rectangle is at least 1px wide

CoderPM2011:
Just a simple patch. 

When the axis span is less than 1, force the mean value ±0.5 as the new range values.
like this
``` csharp
if (vert.Span < 1)
{
    float middle = (vert.Top + vert.Bottom) / 2;
    vert = new(middle - 0.5F, middle + 0.5F);
}
```

Are there other aspects to consider?

#3968


swharden:
This looks great @CoderPM2011, thanks!

```cs
[Test]
public void Test_AxisSpan_ExtremelyNarrow()
{
    ScottPlot.Plot plot = new();

    for (int i = 0; i < 10; i++)
    {
        double width = 1e-10;
        plot.Add.VerticalSpan(i, i + width);
        plot.Add.HorizontalSpan(i, i + width);
    }

    plot.HideGrid();
    plot.SaveTestImage();
}
```

## Before
image

## After
image

PR #4041, opened 2024-07-04
Fix WithZoom in AxisLimits

idotta:
`WithZoom` was using property `XRange` instead of local variable `xRange`. Also avoiding heap allocations.


swharden:
Thanks @idotta! Great catch, and we appreciate you figuring out the issue and fixing it too. This fix will be in the next release! 🚀

idotta:
No, @swharden, thank you for everything! Sorry I didn't think of creating the tests.

swharden:
> Sorry I didn't think of creating the tests.

Haha, no worries! I didn't think to create them originally which was what allowed this bug to speak into the code base in the first place 🤪

Issue #4067, opened 2024-07-12
Ignore Axis line for AutoScale


andresod:
**Question:** Is it possible to have horizontal axis line ignored by AutoScaleY ?

**ScottPlot Version:** 5.0.36

Version 4 provides a flag that allows the drawing of an horizontal axis line to be ignored by autoscale:

```cs
            var hline = WP.Plot.AddHorizontalLine(0, color: System.Drawing.Color.DimGray, width: 1, style: LineStyle.Solid);
            hline.IgnoreAxisAuto = true;
            WP.Plot.AxisAutoY();
```
V5 documentation does not mention anything like that for HorizontalLine. I checked the API and couldn't find anything obvious.

PS: performance improvement on V5 is amazing

KroMignon:
@andresod in fact, the functionality is for now not available in `ScottPlot` 5.

I have create a [pull request](https://github.com/ScottPlot/ScottPlot/pull/4069) to port the functionality from `ScottPlot` 4.1 to 5.0.

Maybe it would be available on next release 🚀

PR #4069, opened 2024-07-15
Axis line and span: Add `EnableAutoscale` property

KroMignon:
Port `IgnoreAxisAuto` feature from `ScottPlot` 4.1 to 5.0

Add support for `Crosshair`, `HorizontalLine`, `VerticalLine`, `HorizonzalSpan` and `VerticalSpan`.

Port Recipe.

This should fix issue #4067.


swharden:
Hi @KroMignon, thanks for this PR! I'm going to rename this property but otherwise leave it as it is. While doing this I had a couple thoughts for alternative implementations, so I'll list them here in case there's interest in shifting toward one of these later...

1) It may be useful to take `GetAxisLimits()` out of `IPlottable` and create `IHasAxisLimits` which would contain it but also a flag like `EnableAxisLimits`

2) Alternatively we could put a `List<IPlottale>` somewhere named something like `PlottablesExcludedFromAxisLimits`

Issue #4081, opened 2024-07-18
SignalConst not support LegendText


fanshuxian:
**Issue:** (Describe the bug here)
I have a `SignalConst` plot (not using `Signal` due to the large amount of data). Since there are multiple curves, I need to display a legend. However, after setting the `LegendText` and executing `ShowLegend`, it does not work as expected.

**ScottPlot Version:** 
5.0.36

code:
```csharp
MainPlot.Plot.Clear();
B_Root.Child = MainPlot;
var hl = MainPlot.Plot.Add.HorizontalLine(0);
hl.IsDraggable = true;
hl.LineWidth = 1;

for (int i = 0; i < OMM.SelectedTask.ChartData.Length; i++)
{
    var sign = MainPlot.Plot.Add.SignalConst(OMM.SelectedTask.ChartData[i].ToArray(), OMM.SelectedTask.XPeriod);
    sign.LegendText = OMM.SelectedTask.ChannelName?[i];
}
MainPlot.Plot.ShowLegend();
MainPlot.Plot.XLabel("Time(ms)", 15);
MainPlot.Plot.YLabel("Voltage(V)", 15);
MainPlot.Plot.Axes.Left.Min = -2;
MainPlot.Plot.Axes.Left.Max = 2;
var buff = OMM.SelectedTask.WaveData.SampleTime * 0.0625;
MainPlot.Plot.Axes.Bottom.Min = 0 - buff;
MainPlot.Plot.Axes.Bottom.Max = OMM.SelectedTask.WaveData.SampleTime + buff;
MainPlot.Plot.Font.Automatic();
MainPlot.Refresh();
```

chart:

![image](https://github.com/user-attachments/assets/c1b067f1-18f3-4cf8-83ea-fbc41b70cdcf)

KroMignon:
@fanshuxian thank you for reporting this!

In fact there is a bug in `SignalConst`, I will submit as PR to solve this. I hope this will be included in next release!

swharden:
Thanks for reporting this @fanshuxian, and for the fix @KroMignon! I merged #4082 and this will indeed be fixed in the next release 🚀

PR #4082, opened 2024-07-18
SignalConst: Enable `LegendText`, increase default marker size, and fix first point position

KroMignon:
`SignalConst` has as `LegendText` field but did not defines `LegendItems`!

Also change `LineStyle` initialization for set `Width` to 1.

This should fix issue #4081.

Doing this, I found a bug in `SignalConst` rendering for first points:
```cs

var sign = myPlot.Add.SignalConst(Generate.Sin(phase:Math.PI));
sign.LegendText = "My fabulous SignalConst LegendText";
var sign2 = myPlot.Add.Signal(Generate.Sin());
sign2.LegendText = "My fabulous Signal LegendText";
myPlot.ShowLegend();
myPlot.SavePng("bug_4081.png", 800, 600);
```
bug_4081


KroMignon:
I also fixed the `Marker` rendering based on the it was done in `Signal` and backport the fix for issue #4083 as it was done by @StendProg in #4086.

PR #4087, opened 2024-07-18
Grid: eliminate allocations for ticks of invisible axes

kebox7:
Grid rendering micro-optimization that eliminates unnecessary calculations if the vertical (X) or horizontal (Y) grid is invisible.


swharden:
Thanks @kebox7, this looks great!

PR #4089, opened 2024-07-20
 Fix : NumericFixedInterval lowest tick was not included

epegeot:
This commit fixes a bug where the first tick of a NumericalFixedInterval was not generated.

Adds a simple `- Interval` in the calculation for the lowest bound to generate the lowest tick.


swharden:
Hi @epegeot, thanks for reporting the issue, identifying a fix, and creating this PR! I'm happy to merge it now, and this fix will be in the next release 🚀

PR #4090, opened 2024-07-20
Ticks: Allow `NumericFixedInterval` interval to be a `double`

epegeot:
`NumericFixedInterval` is now able to be instanciated using doubles instead of just integers, without changing the logic behind. 
`NumericFixedInterval`  now has a property `LabelFormatter` like `NumericAutomatic`.

This PR creates a static class `DefaultFormatterHelper` to extract the `DefaultLabelFormatter` from `NumericAutomatic` so that it can be also used in `NumericFixedInterval`.

The modification from int to double in the constructor is not a breaking change, as illustrated by this code extract:

```cs
public class Test(double t = 1.0f)
{
    public double T { get; set; } = t;
}

public static class Program
{
    public static void Main()
    {
        int v = 12;  
        Test t = new(v);
        Console.WriteLine(t.T); // -> 12
    }
}
```


swharden:
Thanks @epegeot, another great improvement! 🚀

Issue #4085, opened 2024-07-18
SP5 Winforms: Not possible to invert Y-axis in on a signal plot


Jofstera:
Hi everyone, I am trying to recreate this graph in SP5 Winforms. 

![image](https://github.com/user-attachments/assets/1e1b1c89-64b5-45b1-b08f-99a1aea78f98)

As you can see it is a vertically orientated signal plot with an inverted y-axis. The data should also be updated in real-time. 
I did some digging and found some relevant documentation 

https://scottplot.net/cookbook/5.0/SignalXY/SignalXYVerticalInvertedY/

However this code does not work on Winforms

`// invert the vertical axis` 
`myPlot.Axes.SetLimitsY(5000, 0);` 

The class `ScottPlot.Winforms.FormsPlot` does not contain have the attributes `Axes.SetLimitsY`

Something like this for Winforms would be ideal
https://github.com/ScottPlot/ScottPlot/issues/1194#issuecomment-912154734

This is also related https://github.com/ScottPlot/ScottPlot/issues/3253#issuecomment-1926170698 (basically the feature request in point 1). 

I am happy to have a look at the source code myself if someone could point me in the right direction.

Thanks!

StendProg:
Hi @Jofstera,

A `Control` is just a wrapper for a `Plot`. You can always access the `Plot` property of any `Control`.
```c#
formsPlot1.Plot.Axes.SetLimitsY(5000, 0);
```

KroMignon:
Hello @Jofstera 

> The class ScottPlot.Winforms.FormsPlot does not contain have the attributes Axes.SetLimitsY

This is true, but the problem is that you are mixing-up things:
* `ScottPlot.Winforms.FormsPlot` is as graphical control, the `WinForms` backend for `ScottPlot`
* `myPlot` is a `ScottPlot.Plot` instance

To setup the plot, you have to use `Plot` member of your backend ==> `ScottPlot.Winforms.FormsPlot.Plot` => https://scottplot.net/quickstart/winforms/

for example (`formsPlot1` is supposed to be the name of the WinForms control instance):
```csharp

double[] dataX = { 1, 2, 3, 4, 5 };
double[] dataY = { 1, 4, 9, 16, 25 };

formsPlot1.Plot.Add.Scatter(dataX, dataY);
formsPlot1.Plot.Axes.SetLimitsY(5000, 0);
formsPlot1.Refresh();
```

[...]

PR #4053, opened 2024-07-07
Controls: refactor interactivity management system

swharden:
This PR explores alternative strategies for customizing user interactivity in graphical environments. 

## Primary Goals
* Ability to re-map inputs with actions (e.g., right-click-drag pan)
* Ability to add arbitrary inputs (e.g., two-finger pinch)
* Ability to customize actions (e.g., increasing sensitivity of right-click-drag zoom)
* Ability to add arbitrary actions (e.g., holding left and right mouse buttons down while dragging only pans horizontally or vertically)
* Improved ability to unit test interactive behavior
* Ability to opt-in to the new system, allowing existing users to continue using the original [Interaction.cs](https://github.com/ScottPlot/ScottPlot/blob/main/src/ScottPlot5/ScottPlot5/Control/Interaction.cs) and [PlotActions.cs](https://github.com/ScottPlot/ScottPlot/blob/main/src/ScottPlot5/ScottPlot5/Control/PlotActions.cs) system while the new system matures.
* Related: #3186

## Architecture

* User inputs are structs that inherit `IUserInput` and contain any properties that may be unique to that type of input (e.g., `LeftMouseDown` has a stores a `Pixel` and `DateTime`)

* User input actions are classes that contain logic for when and how to respond to user inputs, including whether to refresh the plot or stop processing further actions (e.g., `LeftClickDragPan` engages only after `LeftMouseDown` happens and `MouseMove` moved by more than 5 pixels, preventing other actions from being processed until `LeftMouseUp` occurs)

* Plot controls have a `UserInputProcessor` that they pass user inputs into, and it holds a collection of user input actions that decide if and how to respond to the events. Users can remove actions (like double-click benchmark), add custom actions (like holding the `A` key while left-click-dragging will only pan diagonally), and even pass custom input types through (like three-finger-drag).

## Rules to Implement
_Items are checked only when implemented and unit tests are in place_
- [x] Left-click-drag pan
- [x] Shift+ left-click-drag pan vertically
- [x] Ctrl + left-click-drag pan horizontally
- [x] Right-click-drag zoom
- [x] Shift + right-click-drag zoom
- [x] Ctrl + right-click-drag zoom
- [x] Scroll wheel zoom
- [x] Shift + scroll wheel zoom vertically
- [x] Ctrl + scroll wheel zoom horizontally
- [x] Middle-click-drag zoom rectangle
- [x] Shift + Middle-click-drag zoom rectangle horizontally
[...]

PR #4093, opened 2024-07-22
DataStreamer2: Improve support for rotated plots

drolevar:
Implemented rotation support in DataStreamer2, as mentioned in #4085.

However, I'll need some ideas from @swharden about how to approach the inversion of the Y axis to support top-down logging.


drolevar:
@swharden Is there any existing functionality to invert the X/Y axis here? I see some examples with AutoScaler.InvertY, but I don't see a good way to apply it here.

drolevar:
@KroMignon Maybe you have some good ideas how to invert the Y axis here?

KroMignon:
@drolevar I do not really spend time on this feature. But my idea would be to try to "move" this feature from plotables like `DataStreamer` up to axis manager.

To be more verbose about my idea:
Each plot contains an axis manager, which hods the horizontal and vertical axis of the plot.
Adding a rotation feature on axis manager would  allow to rotate any plotable.
Axis manager would contains a property, `Origine` to define the axis origine point:
- `BottomLeft` (default origine)
    * Horizontal axis are growing from left to rigth
    * Vertical axis are growing from bottom to top
- `BottomRight`: 
    * Horizontal axis are growing from right to left
    * Vertical axis are growing from bottom to top
- `TopLeft`: 
    * Horizontal axis are growing from left to rigth
    * Vertical axis are growing from top to bottom
- `TopRight`: 
    * Horizontal axis are growing from right to left
    * Vertical axis are growing from top to bottom

And a boolean property `FlipAxis` or `SwapAxis` to flip horizontal and vertical axis.

This way, any plotable could be drawn in any direction. But I don't know if it is good idea... What do you mean?

drolevar:
Also added support for InvertX/InvertY properties, which should help implement #4085.

@Jofstera See the demo for DataStreamer2.

swharden:
Hi @drolevar and @KroMignon, thanks for your help with this one! I'm a bit behind on messages (I got pretty deep into #4053 over the last couple weeks) but I'm just now getting to this and it looks great.

One small change I made is to replace the word "normalized" with "rectified", because I think "normalized" typically means representing a range from 0 to 1.

I'm happy this will be in an upcoming release! Eventually we can figure out how to merge the experimental data streamer with he standard one, but this is a good fix in the interim

drolevar:
@swharden Yes, I'd definitely want to figure out how to take it out of "Experimental". IMO everything possible with DataStreamer is possible with the new one, except for one property, which I will commit soon.

drolevar:
@swharden I see that DataStreamer2 is much more compatible in terms of architecture with DataLogger than DataStreamer.
So basically if one creates a DataStreamer2 with a standard List<double> the result is more or less the same as with DataLogger.
One also doesn't really need the IDataStreamer2Source interface, since the variability is provided by the actual type of the `IList<Coordinates> coordinates`.

So one can keep DataStreamer as it is, and replace DataLogger with DataStreamer2 (with minor adjustments) and I don't think it will cause any disruptions. I can try that.

Issue #4107, opened 2024-07-26
PlottableAdder: make `Plot` public


swharden:
Exposing `PlottableAdder.Plot` will make it easier for users to create custom extension methods for adding things to plots and also manipulating them

I think the best solution is to edit this line

https://github.com/ScottPlot/ScottPlot/blob/fd440273870f8dd6c2b1cf05b65a4294880015fd/src/ScottPlot5/ScottPlot5/PlottableAdder.cs#L12

to be something like this:

```cs
public Plot Plot { get; } = plot;
```

PR #4109, opened 2024-07-27
PlottablerAdder: make `Plot` public

DDiggs91:
Access modified change
Exposing PlottableAdder.Plot as requested in #4107


swharden:
Thanks @DDiggs91! This will be in the next release 🚀

PR #4110, opened 2024-07-27
AxisManager: Improve `AutoScale()` support for inverted axes

BrianAtZetica:
When axes are set to inverted and then an AutoScale is initiated (i.e. middle mouse button) the plot was restoring the axes back to normal directions. 

This was caused by the optional parameters on AutoScale being defaulted to false instead of null. 
The subsequent use of the null-coalescing operator '??' was therefore having no purpose. 

Now the default is null, so if no local parameters are provided AutoScaler parameters are used to decide on axis direction


swharden:
Great catch @BrianAtZetica, thanks! This will be in the next release 🚀

drolevar:
@BrianAtZetica Thank you very much! It was necessary.

wi-fr:
It seems that the bug fix applied here got lost again in a later release. In 5.0.43 any Mouse action restores the axes to their default orientation.

swharden:
> It seems that the bug fix applied here got lost again in a later release. In 5.0.43 any Mouse action restores the axes to their default orientation.

Thanks for pointing this out @wi-fr! I opened #4471 to track reproducing and fixing this issue in the relatively new user input processor. The new system allows automated testing, so we can add tests to ensure this behavior is implemented and does not experience regression again in the future 👍

PR #4111, opened 2024-07-27
Scatter: Added X-axis gradient fill area

CoderPM2011:
Add X-direction gradient fill to `Scatter` plot.
User can set stop points and corresponding colors.

Like Microsoft word - Create a custom gradient
https://support.microsoft.com/en-us/topic/add-a-gradient-color-to-a-shape-11cf6392-723c-4be8-840a-b2dab4b2ba3e

For example the following picture is the spectrum of an LED.
image

## How I implement it
**When `FillX` is set, it will be used instead of the setting of `FillYColor` for rendering.** 

In `Render()`, I set the `Hatch` property in `FillStyle`.
And calculate the `ColorPositions` and `Colors` of `Gradient` based on the current chart position.

- Since in `Gradient`, the range of `ColorPositions` is 0~1, so a conversion calculation has to be done on this.
- When a graph is cut off by an edge, the colors must be interpolated in order to get the correct cutoff point color.

---

## result image
After scaling or moving, the color corresponding to the X-axis can be displayed correctly.
temp

## sample code
```cs
 Coordinates[] points =
 {
     new (350, 1),
     new (380, 2),
     new (555, 3),
     new (780, 1),
     new (950, 2),
 };

 var poly = formsPlot1.Plot.Add.Scatter(points);

 poly.MarkerStyle.IsVisible = false;
 poly.FillY = true;
 poly.FillXPoints = new float[] {
     380.0F,
     440.0F,
     490.0F,
     510.0F,
     580.0F,
     650.0F,
 };
 poly.FillXColors = new Color[] {
     Color.FromHex("#610061"),
     Color.FromHex("#0000FF"),
     Color.FromHex("#00FFFF"),
     Color.FromHex("#00FF00"),
     Color.FromHex("#FFFF00"),
     Color.FromHex("#FF0000"),
 };
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,6 +24,7 @@ _Not yet on NuGet..._
 * DataStreamer: Improved support for rotated plots (#4093, #4085) @drolevar @KroMignon @Jofstera
 * Security: Removed outdated reference to `System.Text.Json` which contained CVE-2024-30105 (#4095, #4063) @SerTetora
 * Phaser: New plot type for displaying arrows to points in polar space (#4096, #3939) @CoderPM2011 @nilsakesson
+* Scatter: Added `ColorPositions` to allow placement of colors at specific X positions when using filled scatter plots (#4111) @CoderPM2011
 
 ## ScottPlot 5.0.36
 _Published on [NuGet](https://www.nuget.org/profiles/ScottPlot) on 2024-06-29_
--- a/src/ScottPlot5/ScottPlot5 Cookbook/Recipes/PlotTypes/Scatter.cs
+++ b/src/ScottPlot5/ScottPlot5 Cookbook/Recipes/PlotTypes/Scatter.cs
@@ -422,6 +422,30 @@ public class Scatter : ICategory
         }
     }
 
+    public class ScatterFillGradient : RecipeBase
+    {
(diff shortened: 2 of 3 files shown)


swharden:
@CoderPM2011 this is so cool! Thanks so much for this PR!

I'll make a few small refinements, merge this in, and publish it with the next release tomorrow 🚀

swharden:
> a few small refinements

I ended-up changing the API a bit with the goal of simplifying the Cookbook example I created. I'm going to merge this now, but if you think of improvements still you're welcome to create a new PR! 

I almost made a `PositionedColorGradient` primitive to wrap logic around accumulating colors and positions and house some of the interpolation logic. I figured it could be reused for other things (like creating custom colormaps). I noticed there is already a `Gradient` primitive, but it does something a little different, so I think there may be use for a new primitive that does exactly this. I also thought about defining position as `Coordinates` instead of `double` because that would make it easy to support vertical gradients and also diagonal gradients. I'm leaving things as they are now to keep things moving, but if you have ideas for logic to add to such a class you're welcome to give it a go in a new PR! 

Thanks again for your help on this one 😄

swharden:
Also, it would be super cool to create a cookbook demo that mimics this site

https://www.thermofisher.com/order/fluorescence-spectraviewer

image

PR #4114, opened 2024-07-29
DataLogger: replace `DataLogger` with `Experimental.DataStreamer2`

drolevar:
This should be enough to replace DataLogger with DataStreamer2.

The existing use cases should not be broken.

I've decided to put the new functionality into DataLogger and not DataStreamer because DataStreamer2 is actually based on DataLogger and the API is compatible.


drolevar:
@swharden Any comments on this PR?

swharden:
Hi @drolevar, this looks great! 

I was out of town for about a week then got behind on some other stuff so I've been away from ScottPlot for a few weeks and just now catching up 😅 

> The existing use cases should not be broken

This certainly makes things easier. Thanks for implementing this in a non-breaking way! I'm excited to see this will be in the next release 👍

src/ScottPlot5/ScottPlot5/Plottables/DataLogger.cs
```diff
@@ -1,141 +1,239 @@
...
-        Data.Add(x, y);
+        CoordinateRange rangeX = Data.GetRangeX();
+        CoordinateRange rangeY = Data.GetRangeY(rangeX);
+
+        return Rotated
+            ? new AxisLimits(rangeY, rangeX)
+            : new AxisLimits(rangeX, rangeY);
     }
 
-    public void Add(double[] xs, double[] ys)
```
jpgarza93, line 62: Why was this removed? This was supper useful when loading data from a csv file

Issue #4104, opened 2024-07-25
Invalid svg xml being generated.


aespitia:
Issue: If you try to open the resulting SVG XML, it's invalid because it's missing a "g" tag somewhere, so it will not validate because of the unclosed tag.  

ScottPlot Version: 5.0.36

Code Sample:

```cs
ScottPlot.Plot myPlot = new();

List<PieSlice> slices = new()
{
	new PieSlice() { Value = 31.6, FillColor = Colors.Red, Label = "Category 1 - 31.6%" },
	new PieSlice() { Value = 27.2, FillColor = Colors.Orange, Label = "Category 2 - 27.2%" },
	new PieSlice() { Value = 16.2, FillColor = Colors.Gold, Label = "Category 3 - 16.2%" },
	new PieSlice() { Value = 1.4, FillColor = Colors.Green, Label = "Category 4 - 1.4%" },
	new PieSlice() { Value = 3.7, FillColor = Colors.Blue, Label = "Category 5 - 3.7%" },
	new PieSlice() { Value = 14.8, FillColor = Colors.AliceBlue, Label = "Category 6 - 31.6%" },
	new PieSlice() { Value = 3.3, FillColor = Colors.Aqua, Label = "Category 7 - 3.3%" },
	new PieSlice() { Value = 1.4, FillColor = Colors.Beige, Label = "Category 8 - 1.4%" },
	new PieSlice() { Value = 99.6, FillColor = Colors.BlueViolet, Label = "Category 9 - 99.6%" },
};

var pie = myPlot.Add.Pie(slices);
myPlot.Font.Set("Arial");
pie.DonutFraction=.6;
pie.Padding = 0;
myPlot.Legend.FontSize = 10.666667f;
myPlot.HideGrid();
myPlot.Layout.Frameless();
myPlot.ShowLegend(Edge.Right);

myPlot.SaveSvg(@"c:\temp\pietest.svg",500,375);
```

![image](https://github.com/user-attachments/assets/4f40b1f4-8a8a-420a-9e55-9aa92ab9e1a2)

aespitia:
My guess is that it has something to do with how the XML is getting rendered from the memory stream of whatever is in there at the time, but doesn't know that the G path should be closed when the XML is grabbed.

DDiggs91:
Agreed, adding a closing </g> tag before the end of the svg removes the error
[...]

PR #4121, opened 2024-07-31
SvgImage: fix svg xml string generation

KroMignon:
Add possible missed closing tags in `SvgImage`.`GetXml()`.

This should fix issue #4104


src/ScottPlot5/ScottPlot5/Primitives/SvgImage.cs
```diff
@@ -17,9 +19,30 @@ public SvgImage(int width, int height)
...
+        int count = Regex.Matches(text, substring).Count;
+
+#else
+        var expression = new Regex(substring);
+        int count = expression.Matches(text).Count;
+#endif
+        return count;
+    }
+
+    private static string[] SvgTags = new string[] { "g", "svg" };
```
bclehmann, line 34: Do we know that this issue only happens for these tags? Or preferably why it happens, so we can hopefully fix it in a better way? We already cannot assume that `<svg>` is never contained within a `<g>` tag (which this code implicitly does), as you are allowed to nest SVG documents inside eachother. This would get much worse if we had to add more tags.

It would be very helpful to see some generated documents and where they go wrong, because it's entirely possible that this is exactly enough. But right now I'm seeing that it only appends two closing tags (and only appends each closing tag at most once), which seems like it's either not enough, or it indicates some very specific issue, such as not properly telling Skia that we want to terminate the document.

In fact, I'm seeing the following remark in the Skiasharp docs that makes me think I'm right and we can make Skia close the document itself: https://learn.microsoft.com/en-us/dotnet/api/skiasharp.sksvgcanvas?view=skiasharp-2.88#remarks
```
The canvas may buffer some drawing calls, so the output is not guaranteed to be valid or complete until the canvas instance is deleted.
```

I assume by deleting the canvas instance they mean disposing it, as I don't see any API that would be a better fit. Not the wording I would expect in a C# API, but it's probably just borrowed from the underlying C++ API where that's a more common term to use. Fortunately, the `SvgImage` class is marked `internal` and I believe never exposed to users, so I this wouldn't be a breaking change.

bclehmann (changes requested):
I think my comment allows a better fix for this, but if I'm wrong and it doesn't work I can take another look at the code

src/ScottPlot5/ScottPlot5/Primitives/SvgImage.cs
bclehmann: I'm seeing that the following patch properly closes the image without needing to append a `</svg>`. I think this is probably the way to go, but we should probably have some logic to throw if they call `GetXml` twice.

```
diff --git a/src/ScottPlot5/ScottPlot5/Primitives/SvgImage.cs b/src/ScottPlot5/ScottPlot5/Primitives/SvgImage.cs
index 03edcc1a3..0b53dfc64 100644
--- a/src/ScottPlot5/ScottPlot5/Primitives/SvgImage.cs
+++ b/src/ScottPlot5/ScottPlot5/Primitives/SvgImage.cs
@@ -19,7 +19,8 @@ internal class SvgImage : IDisposable

     public string GetXml()
     {
-        return Encoding.UTF8.GetString(Stream.ToArray()) + "</svg>";
+        Canvas.Dispose();
+        return Encoding.UTF8.GetString(Stream.ToArray());
     }

     public void Dispose()
```

As an aside, it's also very C++-ey to document object destruction particulars and key parts of using the API as a small remark rather than in a more conspicuous area.

KroMignon:
> I think my comment allows a better fix for this, but if I'm wrong and it doesn't work I can take another look at the code

I have changed a little bit the patch to follow your suggestion.
Calling `Canvas.Dispose()` before generating the Svg/Xml string solves the issue. I was wondering why the initial code appends the closing `</svg>` tag but had not found any relevant information in Skia/SkiaSharp project.

Thanks for your input 👍

bclehmann:
> Calling `Canvas.Dispose()` before generating the Svg/Xml string solves the issue. I was wondering why the initial code appends the closing `</svg>` tag but had not found any relevant information in Skia/SkiaSharp project.

That's great, I'm glad we could fix this and improve the code.

> I have changed a little bit the patch to follow your suggestion.

I'm not seeing it here, did you forget to push your changes? In any case, I think Scott's been pretty busy with other stuff this week, so there's no rush.

KroMignon:
> I'm not seeing it here, did you forget to push your changes?

Oups 😅

src/ScottPlot5/ScottPlot5/Primitives/SvgImage.cs
```diff
@@ -6,28 +6,42 @@ internal class SvgImage : IDisposable
     public readonly int Width;
     public readonly int Height;
     private readonly MemoryStream Stream;
-    public readonly SKCanvas Canvas;
+    public SKCanvas Canvas
```
bclehmann, line 9: This is cool, I guess I'll let Scott comment on whether we need to support reusing the `SvgImage` object or if this is too fancy.

src/ScottPlot5/ScottPlot5/Primitives/SvgImage.cs
```diff
@@ -6,28 +6,42 @@ internal class SvgImage : IDisposable
...
-        Canvas = SKSvgCanvas.Create(rect, Stream);
     }
-
     public string GetXml()
     {
-        return Encoding.UTF8.GetString(Stream.ToArray()) + "</svg>";
+        // ensure Canvas is disposed before reading stream content to avoid errors
+        _canvas?.Dispose();
+        // Canvas no more relevant
+        _canvas = null;
```
bclehmann, line 35: This looks great, the only thing I might add is a link to the docs about why the canvas needs disposing. Maybe just something like this:

```cs
// The SVG document is not guaranteed to be closed until the canvas is disposed
// See: https://learn.microsoft.com/en-us/dotnet/api/skiasharp.sksvgcanvas?view=skiasharp-2.88#remarks
```

bclehmann (approved):
looks good, mostly just documentation

swharden:
Thanks @KroMignon and @bclehmann! I was out of town for a week then took a couple weeks to catch up so it took me longer to get to this one, but I really appreciate it. Thanks @aespitia too for pointing this out over in #4104! I'm happy this fix will be in the next release 🚀

Issue #4119, opened 2024-07-30
Generate.Range()


swharden:
We have generate method that creates N consecutive values, but that's not great if we want to generate a range of numbers. For example, a range between 5 and 10 spaced by 0.05 regardless of what the final count actually is.

epegeot:
Is the goal of this to have a range function like the ones in python (start : double, stop : double, step : double = 1)?
I could handle this if that's what's wanted.

swharden:
resolved in #4122

PR #4122, opened 2024-07-31
Generate: add stepped values generator

KroMignon:
Update `Generate` to be able to generate a range of values spaces with given step.

This should fix issue #4119 

Usage example:
```cs
myPlot.Add.Signal(Generate.Range(1, 99, .5));
myPlot.SavePng("Range.png", 640, 480);
```
Range


CoderPM2011:
> This should fix issue #4122

This is the PR itself, should it be #4119?

swharden:
Thanks @KroMignon!

PR #4125, opened 2024-08-02
RenderManager: Improve scaling for high density displays during axis events

BrianAtZetica:
Repeat renders (which occur if events change the axis limits since #3952 ) was resulting in repeated calls to canvas.Scale.

If being run on a screen with a ScaleFactor larger than 1, this resulted in the plot size blowing up. 

Code to illustrate this (in WPF sandbox program) follows:

```cs
    public MainWindow()
    {
        InitializeComponent();

        WpfPlot1.Plot.ScaleFactor = WpfPlot1.DisplayScale;
        WpfPlot1.Plot.RenderManager.AxisLimitsChanged += AxesChanged;

        var size = 500;
        var cycles = 1;
        double[] xs = Generate.Consecutive(size);
        double[] ys = Generate.Sin(count: xs.Length, oscillations: cycles);

        var sig1 = WpfPlot1.Plot.Add.SignalXY(xs, ys);
    }


    private void AxesChanged(object? sender, RenderDetails e)
    {
        var leftAxisLimits = WpfPlot1.Plot.Axes.GetLimits(WpfPlot1.Plot.Axes.Bottom, WpfPlot1.Plot.Axes.Left);
        WpfPlot1.Plot.Axes.SetLimitsY(leftAxisLimits, WpfPlot1.Plot.Axes.Right);
    }
```

This initially renders correctly:
image

but if you zoom on the Left axis:
image

Moving the canvas.Scale to the Render method instead of RenderOnce (which is called multiple times) fixes this issue.


swharden:
Great catch @BrianAtZetica, thanks for identifying the problem and creating this PR!

Issue #4135, opened 2024-08-08
Heatmap keeps crashing with live data feed 


MahmoudArabied:
**Issue:** 
I implemented the heatmap to have live data feed every 100msec. It works as expected, but after a while “usually >10minutes” it crashes. 

The exception is as follows:
“An unhandled exception of type 'System.AccessViolationException' occurred in SkiaSharp.dll
Attempted to read or write protected memory. This is often an indication that other memory is corrupt.
It is related to this file SkiaSharp\SKCanvas.cs.” 

I debugged the calling stack, I found out that the exception is triggered on calling the “Render” function of Heatmap.cs. After I reviewed the source code of Heatmap.cs, I concluded that “Heatmap.cs” it is not a thread safe. This means if the Bitmap is being updated and called within very short time, it might lead to a data corruption.

I modified the Heatmap.cs source code by introducing a thread lock object. Therefore, accessing the Bitmap object is a thread safe operation. Once I applied these changes, the issue of crashing is fully resolved.

**ScottPlot Version:** 
[ScottPlot.WPF 5.0.37]
(https://www.nuget.org/packages/ScottPlot.WPF/5.0.37)

zhengqiwen1997:
Hello MahmoudArabied,

I am also using heatmap in project, but I am using version 4.1.71, could you please create a PR to fix the issue and let me know it? Thanks very much!

MahmoudArabied:
Hi @zhengqiwen1997 I pulled the changes of Heatmap.cs. and created a PR including the fix.
Have a nice day!

swharden:
Hi @MahmoudArabied and @zhengqiwen1997, thanks for the conversation here and over at #4140!

I recognize that many plot types containing "live data" may exhibit unexpected behavior (or throw an exception) if the underlying data is changed _while_ a render is in progress. This can happen with scatter plots, signal plots, heatmaps, etc. 

The recommended way to update data in async environments is to apply a "render lock" to prevent a new render from starting until you're done modifying the data. For example:

### ScottPlot5
```cs
formsPlot1.Plot.RenderManager.PreRenderLock += (s, e) =>
{
    // perform your dangerous operation here
};
```

### ScottPlot4
```cs
formsPlot1.Plot.RenderLock();
// perform your dangerous operation here
formsPlot1.Plot.RenderUnlock();
```

swharden:
[...]

PR #4140, opened 2024-08-10
#4135 Heatmap keeps crashing with live data feed

MahmoudArabied:
Added thread locker to ensure a thread-safe operation of accessing bitmap object

### DIFF
--- a/src/ScottPlot5/ScottPlot5/Plottables/Heatmap.cs
+++ b/src/ScottPlot5/ScottPlot5/Plottables/Heatmap.cs
@@ -11,6 +11,13 @@ public class Heatmap(double[,] intensities) : IPlottable, IHasColorAxis
     public bool IsVisible { get; set; } = true;
     public IAxes Axes { get; set; } = new Axes();
     private IColormap _colormap { get; set; } = new Colormaps.Viridis();
+
+    /// <summary>
+    /// A thread locaker for memory sensitive data access - thread safe
+    /// It ensures that one thread does not enter a critical section of code while another thread is in the critical section.
+    /// </summary>
+    private readonly object m_threadLock = new();
+
     public IColormap Colormap
     {
         get { return _colormap; }
@@ -314,8 +321,14 @@ public class Heatmap(double[,] intensities) : IPlottable, IHasColorAxis
     {
         DataRange = Range.GetRange(Intensities);
         uint[] argbs = GetArgbValues();
-        Bitmap?.Dispose();
-        Bitmap = Drawing.BitmapFromArgbs(argbs, Width, Height);
+
+        //Bitmap?.Dispose();
+
+        // Ensures Bitmap is accessed only by one thread at time, other threads have to wait till it is released 
+        lock (m_threadLock)
+        {
+            Bitmap = Drawing.BitmapFromArgbs(argbs, Width, Height);
+        }
     }
 
     public AxisLimits GetAxisLimits()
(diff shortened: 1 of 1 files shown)


bforlgreen:
It is the responsibly of the calling code (i.e your code) to manage threading.

More info here
https://scottplot.net/faq/async/

MahmoudArabied:
@bforlgreen Thank you for your feedback. It is understandable that I've to be manage threading in my code. I'm sorry I couldn't share any kind of source code sample from my side.  I described the issue within this ticket https://github.com/ScottPlot/ScottPlot/issues/4135

Dealing with bitmap objects within protect memory is very critical. In my case, my source code generates the heatmap every 100msec, this might lead to memory corrupted issue if the object is updated, deleted or being accessed from another thread/class even if the access within the same thread! However, I don't know what exactly occur in SkiaSharp.dll.

In the example you shared under this page
https://scottplot.net/faq/async/

In my opinion it is a risky operation to include a waiting while loop without timeout! What will happen if there is an internal error within "IsRendering", dealing with UI, this could lead to an unresponsive UI or even in worse case a crash.

What I shared in my 'Pull Request' is to include a thread lock, this is very safe in C# as the thread pool will manage the thread access and release it once the operation is completed. It does not have any performance issue. Using this approach, I couldn't reproduce the issue of  'System.AccessViolationException'  anymore.

bforlgreen:
@MahmoudArabied these are just some of my thoughts, Scott might disagree, I am not sure what his intentions for threading are. So he might have other thoughts.

The problem is that you have a added a lock that only works in your specific scenario and for only one plot type. There are many other cases just in that one `HeatMap` class where it could still break. Aside from the inconsistency with other plots, its now not clear (or documented) which parts are actually thread-safe.

The other problem is that locking does incur a performance penalty. So if you were to add it to all the necessary places in all plot types, it would most certainly have some impact on performance, probably only very small, but it would be there. It also makes development much harder, as now you need to consider all the potential places state could be updated from another thread. So maintenance is more involved.

It's simply much easier for users of the library to stop rendering, perform their update, and then continue rendering. I am not sure why you mention you would need a timeout. You'll only be waiting for the the last render to complete. As for blocking the UI, this is no different to your lock in the render.

The PreRenderLock might be a more suitable place for you also. Which is even easier to use.

The benefits of both these options is that you're only *optionally* locking once per frame. Not in a million places all over the code base. So it's the best of both worlds really.

Again, just my opinion.

swharden:
Hi @MahmoudArabied and @bforlgreen, thanks for this PR and useful discussion! 

While there's nothing wrong with the implementation put forth in this PR, it is true that "modifying data while rendering" can be a problem that affects many plot types, but this fix only adds a solution for heatmaps.

Based on this PR and discussion I clarified the recommended strategy for updating data in a way that won't cause unexpected behavior when rendering https://github.com/ScottPlot/ScottPlot/issues/4135#issuecomment-2309004190

If the strategy there isn't sufficient to meet your needs, let me know and we can take a closer look and figure out a better solution! 

Thanks again for both of your input along the way 👍

swharden:
Recognizing ScottPlot 4's API seems to be much simpler than ScottPlot 5, I created an issue #4174 aimed at making the recommended locking strategy more discoverable. I welcome suggestions!

MahmoudArabied:
@swharden @bforlgreen I really appreciate your input on this issue. Engaging in meaningful discussions can help clarify the ideas behind multithreading and locking. I find the way you introduced thread locking is clear and easy to implement. I will follow the issue in https://github.com/ScottPlot/ScottPlot/issues/4174.

Issue #4165, opened 2024-08-19
Color.New() overload for accepting hex string


swharden:
Color.FromHex() works but is less discoverable

swharden:
resolved in #4170

PR #4170, opened 2024-08-21
Color: add overload to accept hex string

aespitia:
Small update to add an new Color overload described in this ticket

https://github.com/ScottPlot/ScottPlot/issues/4165

sample usage:

var clr = new ScottPlot.Color("#FF0000");


aespitia:
> @aespitia IMHO, this is not the best way to create this constructor. Create a first instance of the class to copy it to create the class itself is not very clean for me.
> 
> Perhaps you could made a small change to this PR:
> 
>     * split static method `Color.FromHex()` into 2 parts:
>       
>       * create a private static method `Hex2Uint` to convert hex string into an unsigned int value
>       * create a `Color` constructor from unsigned int `Color (uint value)`
> 
>     * call this constructor from hex string constructor `public Color(string hex):this(Hex2Uint(hex)) {}`
> 
>     * also use this constructor in static method `Color.FromHex()` ==> `public static Color FromHex(string hex) => new Color(hex);`
> 
> 
> my 2 cts.

i like it, can make updates

aespitia:
@KroMignon updated PR with your suggestions, lmk if you had other thoughts.

src/ScottPlot5/ScottPlot5/Primitives/Color.cs
```diff
@@ -116,32 +126,38 @@ public static Color FromARGB(int argb)
...
-        byte alpha = (byte)(argb >> 24);
-        byte red = (byte)(argb >> 16);
-        byte green = (byte)(argb >> 8);
-        byte blue = (byte)(argb >> 0);
-        return new Color(red, green, blue, alpha);
+        return new Color(argb);
     }
 
-    public static Color FromHex(string hex)
+    public static Color FromHex(string hex) => new(hex);
```
kebox7, line 132: Maybe it's better to make this method private?

src/ScottPlot5/ScottPlot5/Primitives/Color.cs
```diff
@@ -116,32 +126,38 @@ public static Color FromARGB(int argb)
...
+
+    //Returns ARGB value from hex string
+    public static uint Hex2Argb(string hex)
```
kebox7, line 135: Maybe it's better to make this method private?

aespitia: oh ya he did mention private in the original post, can make the update.

kebox7: LGTM!

swharden:
This looks great! Thanks @aespitia @KroMignon and @kebox7 - I look forward to using this myself after the next release! 🚀

PR #4203, opened 2024-09-01
Make IPlotMenu in IPlotControl nullable

kebox7:
In some cases, it may be necessary to create your own control inherited from `IPlotControl`.  There is no need to use the context menu in this control. Unfortunately, the `Menu` property of `IPlotControl` always implies the presence of an instance of an object of the `IPlotMenu` type.  You have to create a stub, for example

```csharp
private sealed class PlotMenu : IPlotMenu {
    public void Reset()
    {
    }

    public void Clear()
    {
    }

    public void AddSeparator()
    {
    }

    public void Add(string Label, Action<IPlotControl> action)
    {
    }

    public void ShowContextMenu(Pixel pixel)
    {
    }
}
```

This PR modifies the `IPlotControl` interface and makes its `Menu` property nullable. Which in turn expands the possibilities of customization.

```diff
-IPlotMenu Menu { get; set; }
+IPlotMenu? Menu { get; set; }
```


swharden:
Thanks for this PR @kebox7! I agree with your assessment and am happy to merge this in so it makes it into the next release.

It looks like your last commit was less than an hour ago, so I want to double-check with you to make sure this is ready before I merge it 👍

PR #4217, opened 2024-09-04
Generate: Use `Random.Shared` if possible

LeaFrock:
Just as the title tells.


swharden:
Thanks @LeaFrock! This will be in the next release.

I'll modify this PR before merging to add the original function back to the "else" case if users aren't running .NET 5 or newer...

swharden:
@KroMignon I see you last worked on the random seed strategy a while back in #3022 so I wanted to ping you here just to make sure this is consistent with what you'd expect to see. I'm pretty sure this looks good, but would appreciate one more set of eyes!

Thanks,
Scott

KroMignon:
Looks also good to me, actually the old code simply mimic the way `Random.Shared` works.

PR #4219, opened 2024-09-05
Axes: Add `ClipLabel` option to opt-in to axis label clipping

drolevar:
It's similar to my PR about legends.

When axis titles are rather long, and the plot window is downsized, the titles start overlapping each other, which is not too esthetic. I have added clipping the titles to the axis panel rectangle, which solves the issue.

| Before | After |
| ------- | ----- |
| image | image |

I understand, that on the example it's rather ugly both ways, but on bigger plots with longer axis names that makes a difference.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@ _Not yet on NuGet..._
 * Pie: Improved label alignment (#4211) @CoderPM2011
 * Pie: First slice now starts vertically (-90 degrees) instead of to the right (0 degrees) (#4211) @CoderPM2011
 * Generate: Random seed uses `System.Random.Shared` on .NET platforms where it is available (#4217) @LeaFrock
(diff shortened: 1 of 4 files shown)


drolevar:
@swharden Something is wrong about the pipeline. There's nothing in my PR which can make it fail this way.

swharden:
Thanks @drolevar! I'll refine this PR a bit to make it an opt-in feature, then merge tonight

My thinking is that "extremely small" plots like the ones in your screenshot will always look bad, but "very small" plots look all right currently but have labels that would get clipped if this were turned on by default (example below)

image

swharden:
> Something is wrong about the pipeline

Something about the android build is raising a warning, and CICD treats warnings as errors. I'll use this opportunity to simplify the pipeline. Thanks for mentioning it! #4223

Issue #4223, opened 2024-09-06
CICD: Reduce frequency of heavy workload builds


swharden:
Some projects are BRUTAL to build, especially in the CI system. Android, Blazor, and WinUI come to mind. Many of the slowest-to-build projects are in the sandbox folder.

This issue will track updating the CICD pipeline to only build/test core projects (libraries, tests, demo apps, and nuget packages) which will significantly speed-up actions that run in the cloud, reduce the energy footprint of our automated testing, and prevent nuanced platform-related build errors from derailing the CI system.

swharden:
... unfortunately many of the sandbox apps are "required" to adequately test the controls

I'll keep working to figure out a compromise.

swharden:
Putting some hard numbers here, Maui, WinUI, and Blazor sandbox projects are the big offenders.

Ironically those are likely the 3 probably least used.

### NuGet Packages

Project | Build Time (sec)
--- | ---
ScottPlot5 | 2.82
ScottPlot.Avalonia | 2.76
ScottPlot.Blazor | 2.22
ScottPlot.Eto | 2.77
ScottPlot.Maui | 6.01 ⚠️
ScottPlot.OpenGL | 2.21
ScottPlot.WinForms | 2.48
ScottPlot.WinUI | 5.31 ⚠️
ScottPlot.WPF | 2.59

### Sandbox Projects

Project | Build Time (sec)
--- | ---
Sandbox.Avalonia.Desktop | 2.96
Sandbox.Blazor.WebAssembly | 8.97 ⚠️
Sandbox.Console | 2.20
Sandbox.ConsoleFramework | 1.77
Sandbox.Eto | 2.35
Sandbox.Maui | 51.31 ⚠️
Sandbox.UnoPlatform | 0.18
Sandbox.WebApi | 1.38
Sandbox.WinForms | 2.13
Sandbox.WinForms3D | 3.70
Sandbox.WinFormsFramework | 1.86
Sandbox.WinUI.Desktop | 9.08 ⚠️
Sandbox.WPF | 3.87
Sandbox.WPFFramework | 0.17

PR #4231, opened 2024-09-08
Multiplatform Sandbox: move projects into their own solution file

swharden:
#4223


CoderPM2011:
`SkiaSharp` uses a "solution filter" to avoid solutions that are too large.
Maybe you can refer to it.

reference: 
1. https://learn.microsoft.com/en-us/visualstudio/msbuild/solution-filters?view=vs-2022
2. https://github.com/mono/SkiaSharp/tree/main/source

swharden:
Good suggestion!

I did that a little while ago but it got really complicated. Some build tools, auto formatters, etc don't support that format. Also older versions of VS don't support that format.

Is there a way to have a solution where certain projects are "unloaded" by default? That could be a solution 🤔

swharden:
Fundamentally, I think the sandbox projects are more "dev" or "doc" than "src" (from the perspective of the NuGet packages) so I'd like to think of a way to more elegantly separate them out...

CoderPM2011:
> Fundamentally, I think the sandbox projects are more "dev" or "doc" than "src" (from the perspective of the NuGet packages) so I'd like to think of a way to more elegantly separate them out...

In fact, when I am developing, I will directly set the corresponding sandbox to the startup project.
After I modified the `ScottPlot5` code, I could directly execute the sandbox project and observe the results of the real-time modifications.😅

However, after splitting into two separate solutions, the `ScottPlot5` code cannot be viewed directly in the sandbox solution. 
Instead, one needs to switch between the two solutions.😥

---

There is another way.

Maybe you can create a dedicated `build configuration` for a specific target.
For example, when testing, only build some projects in the solution.😀

reference:
1. https://learn.microsoft.com/en-us/visualstudio/ide/understanding-build-configurations?view=vs-2022

swharden:
> build configuration

Thanks for this suggestion! I'll try this today 👍

swharden:
Great idea! Obvious in hindsight 😅 

image

Also, the .NET 9 preview terminal output is awesome

image

PR #4243, opened 2024-09-09
Bring back Add method overloads to DataLogger

drolevar:
Due to the reimplementation of the DataLogger two of the Add overloads were dropped.
But as @jpgarza93 has mentioned in #4114, they can be useful. Bringing them back.


swharden:
Thanks for this PR @drolevar, and thanks @jpgarza93 for pointing out the change! This will be in the next release 🚀

PR #4244, opened 2024-09-09
WIP Add finer steps to DateTime axis

drolevar:
@swharden Scott, I find it useful and better looking to add more divisors to Decimal and Sexagesimal lists. Sexagesimal actually could have also 3 in addition to 2. 
What do you think?

### DIFF
--- a/src/ScottPlot5/ScottPlot5/TickGenerators/TimeUnits/StandardDivisors.cs
+++ b/src/ScottPlot5/ScottPlot5/TickGenerators/TimeUnits/StandardDivisors.cs
@@ -2,8 +2,8 @@
 
 internal static class StandardDivisors
 {
-    public static readonly IReadOnlyList<int> Decimal = new int[] { 1, 5, 10 };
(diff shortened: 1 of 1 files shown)


swharden:
Hi @drolevar, thanks for this PR!

It would help me get a better understand of the problem this change solves if you can share a screenshot and/or sample code to help demonstrate the change.

I'm curious, is there a way where users can customize their own list of divisors? If so, this may be useful to document in a cookbook recipe so we don't have to rely on these hard-coded arrays.

swharden:
> It would help me get a better understand of the problem this change solves if you can share a screenshot and/or sample code to help demonstrate the change.

Hi @drolevar, thanks again for this PR! It's not super clear to me the issue this fixes though, and I'm going to focus my time toward closing other open PRs and issues for the next few days, so I probably won't invest the time here figuring out how to reproduce the issue this PR solves to evaluate it properly.

If you or someone else wants to pick this back up, let me know and we can open this up again and take a closer look! 🚀

PR #4245, opened 2024-09-10
WPF: Implement `IMultiThreadedPlotControl` with demo

### LINKED ISSUE #4174: Async: add simpler option for render locking
Sometimes it's necessary to prevent renders while data inside plottables is being modified. Following a discussion with @MahmoudArabied and @zhengqiwen1997 in #4135 and #4140, I noticed that this is much easier in ScottPlot 4 than ScottPlot 5

Could we improve the ScottPlot 5 API to make it simpler? We probably don't need to execute our dangerous code before rendering every frame, as currently suggested in https://scottplot.net/faq/async/

The lock object itself seems to be exposed, but should we encourage people to use it?

https://github.com/ScottPlot/ScottPlot/blob/17860d4b3fe761c9e122917d8fcd3f274266ccb0/src/ScottPlot5/ScottPlot5/Plot.cs#L44-L49

Alternatively, we could put the simpler `RenderLock()` and `RenderUnlock()` back in place.

swharden:
This PR creates an `IMultiThreadedPlotControl` interface which may be implemented by controls that support asynchronous rendering. It provides methods to help users modify plot data while the plot is not actively rendering to avoid race conditions. Resolves #4174

### DIFF
--- a/src/ScottPlot5/ScottPlot5 Demos/ScottPlot5 WPF Demo/DemoWindows/MultiThreading.xaml
+++ b/src/ScottPlot5/ScottPlot5 Demos/ScottPlot5 WPF Demo/DemoWindows/MultiThreading.xaml
@@ -1,0 +1,29 @@
+﻿<Window x:Class="WPF_Demo.DemoWindows.MultiThreading"
+        xmlns="http://schemas.microsoft.com/winfx/2006/xaml/presentation"
+        xmlns:x="http://schemas.microsoft.com/winfx/2006/xaml"
+        xmlns:d="http://schemas.microsoft.com/expression/blend/2008"
+        xmlns:mc="http://schemas.openxmlformats.org/markup-compatibility/2006"
+        xmlns:local="clr-namespace:WPF_Demo.DemoWindows" xmlns:wpf="clr-namespace:ScottPlot.WPF;assembly=ScottPlot.WPF"
+        mc:Ignorable="d"
+        Title="MultiThreading" Height="450" Width="800">
+    <Grid>
+        <Grid.RowDefinitions>
+            <RowDefinition></RowDefinition>
+            <RowDefinition Height="auto"></RowDefinition>
+        </Grid.RowDefinitions>
+        <wpf:WpfPlot Grid.Row="0" Name="WpfPlot1" />
+        <StackPanel Grid.Row="1" Orientation="Horizontal">
+            <CheckBox
+                FontSize="18" 
+                Padding="5 0"
+                Margin="5"
+                VerticalContentAlignment="Center"
+                Content="Continuously Render"/>
+            <Button
+                FontSize="18" 
+                Padding="5 0"
+                Margin="5"
+                Content="Change Data"/>
+        </StackPanel>
+    </Grid>
+</Window>
--- a/src/ScottPlot5/ScottPlot5 Demos/ScottPlot5 WPF Demo/DemoWindows/MultiThreading.xaml.cs
+++ b/src/ScottPlot5/ScottPlot5 Demos/ScottPlot5 WPF Demo/DemoWindows/MultiThreading.xaml.cs
@@ -1,0 +1,14 @@
+﻿using System.Windows;
+
+namespace WPF_Demo.DemoWindows;
+
+public partial class MultiThreading : Window, IDemoWindow
+{
+    public string DemoTitle => "WPF Multi-Threading";
+    public string Description => "Demonstrate how to safely change data while rendering asynchronously.";
+
+    public MultiThreading()
+    {
+        InitializeComponent();
+    }
+}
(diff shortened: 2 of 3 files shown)


swharden:
Hi @KroMignon, I'm having a very difficult time getting the WPF sandbox to run on my local machine and I was hoping you may be able to provide some insight. I'm not sure what changed - it could be a dependency version changed, or could be that I have a different combination of .NET SDKs / visual studios installed.

I suspect the issue is related to the themed custom control which made me want to ask you for your thoughts (#2526). I can create a new user control in the `ScottPlot.WPF` project named `ScottPlot.WPF.TestControl` and reference that in a sandbox app and it works as expected. However, if I try to use the `ScottPlot.WPF.WpfPlot` in the sandbox project I get an error on startup:

```
System.Windows.Markup.XamlParseException
  HResult=0x80131501
  Message='Initialization of 'ScottPlot.WPF.WpfPlot' threw an exception.' Line number '10' and line position '18'.
  Source=PresentationFramework
  StackTrace:
   at System.Windows.Markup.XamlReader.RewrapException(Exception e, IXamlLineInfo lineInfo, Uri baseUri)
   at System.Windows.Markup.WpfXamlLoader.Load(XamlReader xamlReader, IXamlObjectWriterFactory writerFactory, Boolean skipJournaledProperties, Object rootObject, XamlObjectWriterSettings settings, Uri baseUri)
   at System.Windows.Markup.WpfXamlLoader.LoadBaml(XamlReader xamlReader, Boolean skipJournaledProperties, Object rootObject, XamlAccessLevel accessLevel, Uri baseUri)
   at System.Windows.Markup.XamlReader.LoadBaml(Stream stream, ParserContext parserContext, Object parent, Boolean closeStream)
   at System.Windows.Application.LoadComponent(Object component, Uri resourceLocator)
   at WPF_Demo.DemoWindows.Quickstart.InitializeComponent() in C:\Users\scott\Documents\GitHub\ScottPlot\src\ScottPlot5\ScottPlot5 Demos\ScottPlot5 WPF Demo\DemoWindows\Quickstart.xaml:line 1

  This exception was originally thrown at this call stack:
    [External Code]

Inner Exception 1:
IndexOutOfRangeException: Index was outside the bounds of the array.
```

It's really confusing me because it claims XAML parsing errors, although the `ScottPlot.WPF` project contains almost no XAML. Any insight you can provide would be really appreciated!

Thanks,
Scott

swharden:
... to clarify, the project that doesn't run as expected is the ScottPlot5 WPF sandbox

https://github.com/ScottPlot/ScottPlot/tree/main/src/ScottPlot5/ScottPlot5%20Sandbox/Sandbox.WPF

the csproj file references the source code like this:

```xml
    <ItemGroup>
        <ProjectReference Include="..\..\ScottPlot5 Controls\ScottPlot.WPF\ScottPlot.WPF.csproj" />
    </ItemGroup>
```

but if I change it to use the package on NuGet, it works!

```xml
    <ItemGroup>
      <PackageReference Include="ScottPlot.WPF" Version="5.0.39" />
    </ItemGroup>
```

I'm so confused 😅

BHandle:
I don't have specific knowledge about this PR, but I do have a lot of practice beating WPF into submission.  What you bumped into here wasn't actually a XAML issue, the inner exception was thrown while it was initializing the control from the XAML.  So in this case, you hit an IndexOutOfRangeException inside the constructor for the WpfPlot.  Unfortunately, the stacktrace leading up to the exception was lost.  :-(   (WPF is helpful!)

I'm still new to ScottPlot, so I'm not sure what the sandbox is yet, just wanted to provide insight on why the XAML seemed to be crying.

swharden:
> the inner exception was thrown while it was initializing the control from the XAML ... Unfortunately, the stack trace leading up to the exception was lost

@BHandle, thanks for your input! Whenever I start to feel like I'm getting the hang of .NET development, WPF comes out of nowhere and slaps me in the face 🤪 I have a lot of respect for people who are good at WPF development!

> I'm not sure what the sandbox is yet

For those new to ScottPlot development, the `ScottPlot5.sln` file has minimal "sandbox" apps for testing ScottPlot controls on all the major targets it supports.  The WPF sandbox is a minimal case WPF app for testing the WPF ScottPlot control locally.

image


## Solution (for getting WPF apps to build again)

Finally got this working and the solution was simple, but it took weeks to figure out!

The csproj file had a reference to `ScottPlot.WPF.csproj` but for some reason it now also requires a separate reference to `ScottPlot.csproj`

I'm not sure why this changed suddenly. It didn't used to require it. The sandbox csproj file in the main codebase looks like this, and it used to work. Maybe it has to do with a SkiaSharp dependency version bump?
https://github.com/ScottPlot/ScottPlot/blob/743c9427e63692d31920cc424a0e62b12d6eec4a/src/ScottPlot5/ScottPlot5%20Sandbox/Sandbox.WPF/Sandbox.WPF.csproj#L1-L15

drolevar:
@swharden Some XAML parsing exceptions were related to the presence of .NET 9 preview 7. Now it's resolved with .NET 9 RC1.

swharden:
I'll take the less-heavy handed approach for the time being and let advanced users lock how they wish.

I adapted this PR to add a WPF demo that shows how to update data using `Timer` (which requires locking the `sync` object) and also using `DispatcherTimer` (which doesn't seem to require locking)

I'll update https://scottplot.net/faq/async/ to use the simpler lock example, because what's there now is pretty complicated

swharden:
https://scottplot.net/faq/async/

Issue #4251, opened 2024-09-11
Generate.RandomNumbers does not respect min and max values


ArchieCoder:
**Issue:**
Using RandomNumbers(10000, 34, 38) will only produce values that are over 38

![image](https://github.com/user-attachments/assets/9474d3c4-fa05-407f-b6f5-23f8ef97cf59)

**ScottPlot Version:** (What NuGet package are you using?)
5.0.39

```cs
var numbers = Generate.RandomNumbers(10000, 34, 38);

aespitia:
I've submitted a PR

swharden:
resolved in #4252

PR #4252, opened 2024-09-11
Feature/aespitia/random number fix

aespitia:
Fixes issue #4251

Old code was pointing to RandomSample which uses an offset and was not using the min/max that was expected. Made new generate function that uses existing RandomNumber functionality.  Also included Unit Test.


aespitia:
@swharden do you have any suggestions for this build?  it builds locally, but I think the automated build is broken?  the error message looks related to .net 9?

LeaFrock:
I think it just requires a typo fix.

```diff
    /// <summary>
    /// Random numbers between zero (inclusive) and <paramref name="max"/> (exclusive)
    /// </summary>
    public static double[] RandomNumbers(int count, double max)
    {
-         return RandomData.RandomSample(count, 0, max);
+        return RandomNumbers(count, 0, max);
    }

    /// <summary>
    /// Random numbers between <paramref name="min"/> (inclusive) and <paramref name="max"/> (exclusive)
    /// </summary>
    public static double[] RandomNumbers(int count, double min, double max)
    {
-         return RandomData.RandomSample(count, min, max);
+        return RandomSample(count, min, max);
    }
```

aespitia:
what was the typo?  RandomSample does not return what was expected in the original request.  

> I think it just requires a typo fix.
> 
> ```diff
>     /// <summary>
>     /// Random numbers between zero (inclusive) and <paramref name="max"/> (exclusive)
>     /// </summary>
>     public static double[] RandomNumbers(int count, double max)
>     {
> -         return RandomData.RandomSample(count, 0, max);
> +        return RandomNumbers(count, 0, max);
>     }
> 
>     /// <summary>
>     /// Random numbers between <paramref name="min"/> (inclusive) and <paramref name="max"/> (exclusive)
>     /// </summary>
>     public static double[] RandomNumbers(int count, double min, double max)
>     {
> -         return RandomData.RandomSample(count, min, max);
> +        return RandomSample(count, min, max);
>     }
> ```

LeaFrock:
[Generate.RandomSample](https://github.com/ScottPlot/ScottPlot/blob/5.0.39/src/ScottPlot5/ScottPlot5/Generate.cs#L306) is different from [RandomDataGenerator.RandomSample](https://github.com/ScottPlot/ScottPlot/blob/5.0.39/src/ScottPlot5/ScottPlot5/RandomDataGenerator.cs#L170). I think the original request is just what the former does. Do I misunderstand anything?

LeaFrock:
Also, for several overload methods, we should make all of them call in the same way finally. So I change `public static double[] RandomNumbers(int count, double max)` to call the longest overload `RandomNumbers(count, 0, max)`.

aespitia:
i've repointed to existing functionality as @LeaFrock  pointed out.

PR #4265, opened 2024-09-15
WIP. SignalXY.GetNearest() improvements

StendProg:
The relevance of this PR has been completely lost. PR #4270 brings all the changes implemented here.

1. Fix additional axes support for SignalXY.GetNearest() and SignalXY.GetNearestX()

Allows user to search for nearest points on `SignalXY` that use additional axes.

Added methods to `RenderDetails` to get current resolution of additional axes.

`ISignalXYSource` interface has been slightly modified. It doesn't need to know about RenderDetails because it doesn't know about axes used by Plottable. Therefore, it is passed the axis resolutions that are calculated in Plottable.

The ShowValueUnderMouseSignalXY Demo has been significantly modified to show how additional axes can be used. It was largely used for debugging purposes. So you can safely undo/modify according to your vision.

It's not hard to apply a similar fix to `ScatterSources`, but there are a lot of implementations out there and I'd rather wait for the current design to be approved.

2. The performance of the `SignalXY.GetNearest()` method has been significantly improved.

Instead of a complete search of all points, we start with the point closest to the X coordinate. Then we move to the left or right until the X distance exceeds maxDistance. At the same time, if we have found points matching the criterion (distance less than maxDistance), the threshold for stopping (maxDistance) is also updated (decreased).

In my tests, finding the nearest point for a 10 million signal stays within the interactive performance limits.

It is not difficult to think of examples when this algorithm degenerates into a complete search of all points, consuming large computations due to more complex checks. But as it seems to me, in most cases when maxDistance is small and the density of points per pixel is relatively small, this algorithm gives a significant performance gain.

### DIFF
--- a/src/ScottPlot5/ScottPlot5 Demos/ScottPlot5 WinForms Demo/Demos/ShowValueUnderMouseSignalXY.cs
+++ b/src/ScottPlot5/ScottPlot5 Demos/ScottPlot5 WinForms Demo/Demos/ShowValueUnderMouseSignalXY.cs
@@ -1,5 +1,4 @@
 ﻿using ScottPlot;
-using System.Diagnostics;
 
 namespace WinForms_Demo.Demos;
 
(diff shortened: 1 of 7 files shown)


RFBomb:
@StendProg I took a peek here, and I like these changes. Should this be accepted, my functionality over in #4270  could be adapted to use this strategy fairly easily I think.

PR #4275, opened 2024-09-18
LinePattern: add support for custom patterns

CoderPM2011:
Under the existing code, I cannot implement the following styles by myself.
image

So I changed the `LinePattern` from an `enum` to a `readonly struct` to allow the user to set the desired style.

The API is the same as before, except for the method parameter defaults.
Since the method's parameter defaults need to be determined at compile time, I've switched to using `default` instead of `LinePattern.Solid`.

#### How it's done
When `struct` is assigned a value of `default`, the internal real-value type member is set to the default value, so I declared a private `bool` flag to detect if the object is generated by `default`.

#### References
1. linestyles
  https://matplotlib.org/stable/gallery/lines_bars_and_markers/linestyles.html
1. Default values ​​of struct
  https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/builtin-types/default-values .

---

The following is an sample I wrote based on the reference.
### result image
image

### sample code
```cs
var plot = formsPlot1.Plot.Add.Line(-1, 0, 1, 0);
plot.LineWidth = 5;
plot.LinePattern = new LinePattern(new float[] { 10, 3 }, 5);
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@ _Not yet on NuGet..._
 * DataLogger: Added `Add()` overloads to be consistent the original DataLogger API (#4243, #4114) @drolevar @jpgarza93
 * Fonts: Improve typeface caching to significantly improve Avalonia performance on Linux (#3439, #4250) @kebox7
 * Generate: Improved `RandomNumbers()` to include lower boundary as described in XML docs (#4251, #4252) @aespitia @LeaFrock @ArchieCoder
+* LinePattern: Added `Name` property and added support for custom patterns (#4275, #4289) @CoderPM2011
 
 ## ScottPlot 5.0.39
 _Published on [NuGet](https://www.nuget.org/profiles/ScottPlot) on 2024-08-02_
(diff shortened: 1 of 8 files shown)


swharden:
@CoderPM2011, thanks so much for this PR! This is fantastic! I'm going to extend it a bit to add some cookbook examples and use reflection to add a `GetAllPatterns()` method, then merge it in shortly 🚀🚀🚀

swharden:
```cs
List<LinePattern> patterns = [];
patterns.AddRange(LinePattern.GetAllPatterns());
patterns.Add(new([2, 2, 5, 10], 0, "Custom"));

Plot plot = new();
for (int i = 0; i < patterns.Count; i++)
{
    var line = plot.Add.Line(0, i - 1, 10, i);
    line.LinePattern = patterns[i];
    line.LineWidth = 2;
    plot.Add.Text(patterns[i].Name, 10, i);
}
```

image

swharden:
> When struct is assigned a value of default, the internal real-value type member is set to the default value, so I declared a private bool flag to detect if the object is generated by default.

I think I simplified this in a way that still makes sense, but one more set of eyes on this bit wouldn't hurt 😅 

```cs
public SKPathEffect GetPathEffect()
{
    return Intervals is null
        ? Solid.GetPathEffect()
        : SKPathEffect.CreateDash(Intervals, Phase);
}
```

PR #4278, opened 2024-09-20
Pie: respect `LineStyle`, fix automatic axis limits, and demonstrate percent labels

LeaFrock:
Bring the feature of Pie plot which ScottPlot 4 owns to 5.

image

### DIFF
--- a/src/ScottPlot5/ScottPlot5/Plottables/Pie.cs
+++ b/src/ScottPlot5/ScottPlot5/Plottables/Pie.cs
@@ -102,22 +102,6 @@ public class Pie : PieBase
                 slice.LabelStyle.Render(rp.Canvas, px, paint);
             }
 
-            var slicePercentFormat = SlicePercentFormat;
-            if (slicePercentFormat != null && slicePercentFormat.StartsWith("P", StringComparison.OrdinalIgnoreCase))
-            {
-                Coordinates polar = new PolarCoordinates(1.0 * SlicePercentDistance, centerAngle).ToCartesian();
-                polar.Y = -polar.Y;
-                Pixel px = Axes.GetPixel(polar) - origin;
-                var percentLabelStyle = slice.PercentLabelStyle;
-                // Let it be if we've already set the percent label text manually.
-                if (string.IsNullOrEmpty(percentLabelStyle.Text))
-                {
-                    // TODO: maybe someone wants to specify the CultureInfo too
-                    percentLabelStyle.Text = percentage.ToString(slicePercentFormat, CultureInfo.InvariantCulture);
-                }
-                percentLabelStyle.Render(rp.Canvas, px, paint);
-            }
-
             totalAngle += sliceAngle;
             path.Reset();
         }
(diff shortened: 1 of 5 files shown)


src/ScottPlot5/ScottPlot5/Plottables/Pie.cs
```diff
@@ -105,6 +106,28 @@ public override void Render(RenderPack rp)
                 slice.LabelStyle.Render(rp.Canvas, px, paint);
             }
 
+            var slicePercentFormat = SlicePercentFormat;
+            if (slicePercentFormat != null &&
+#if NET5_0_OR_GREATER
+                slicePercentFormat.StartsWith('P') // Just follow the code quality rule CA1865
+#else
+                slicePercentFormat.StartsWith("P", StringComparison.Ordinal)
```
aespitia, line 114: maybe this can be OrdinalIgnoreCase?

LeaFrock: No. See [https://learn.microsoft.com/en-us/dotnet/standard/base-types/standard-numeric-format-strings#percent-format-specifier-p](https://learn.microsoft.com/en-us/dotnet/standard/base-types/standard-numeric-format-strings#percent-format-specifier-p).

aespitia:
sorry, i was talking about this part:
  slicePercentFormat.StartsWith('P') // Just follow the code quality rule CA1865
#else
                slicePercentFormat.StartsWith("P", StringComparison.Ordinal)
                
                
in common usage, lowercase "p" actually works too.

                double number = .2468013;
Console.WriteLine(number.ToString("p", CultureInfo.InvariantCulture));
// Displays 24.68 %
Console.WriteLine(number.ToString("p",
				  CultureInfo.CreateSpecificCulture("hr-HR")));
// Displays 24,68%
Console.WriteLine(number.ToString("p1", CultureInfo.InvariantCulture));
// Displays 24.7 %

LeaFrock:
Well, the document doesn't mention the lower case. But after I check the [source codes](https://source.dot.net/#System.Private.CoreLib/src/libraries/System.Private.CoreLib/src/System/Number.Formatting.cs,461) of `double.ToString` , I admit that you're right. I'll improve it then.

swharden:
Hi @LeaFrock and @aespitia, 

I believe this functionality is possible with the existing library and a code change is not required. I'll adapt this PR to add a cookbook recipe that demonstrates this 👍 

```cs
// create a pie chart
double[] values = [6, 8, 10];
var pie = myPlot.Add.Pie(values);
pie.ExplodeFraction = .1;
pie.ShowSliceLabels = true;
pie.SliceLabelDistance = 0.5;

// determine percentages for each slice
double total = pie.Slices.Select(x => x.Value).Sum();
double[] percentages = pie.Slices.Select(x => x.Value / total * 100).ToArray();

// set each slice label to its percentage
for (int i = 0; i < pie.Slices.Count; i++)
{
    pie.Slices[i].Label = $"{percentages[i]:0.0}%";
    pie.Slices[i].LabelFontSize = 20;
    pie.Slices[i].LabelBold = true;
    pie.Slices[i].LabelFontColor = Colors.Black.WithAlpha(.5);
}

// hide unnecessary plot components
myPlot.Axes.Frameless();
myPlot.HideGrid();
myPlot.HideLegend();
```

image

LeaFrock:
@swharden The problem is that sometimes we want two labels: one is inside(as a percent) and the other is outside(as a tag or other information). It's the original feature request of this PR.

The current `SliceLabel` does not satisfy the requirement.

PR #4280, opened 2024-09-20
Pie:  Make the label and legend text settings independent

### LINKED ISSUE #3295: SP5: Slice Labels
Pie charts currently don't support slice labels just like SP4 has. A pie chart without value/% labels is difficult to interpret. SP4 has the following available, perhaps bring SP5 to parity on this is a good first shot:
 .SliceLabels = labels
 .ShowPercentages = True
 .SliceLabelPosition = 0.62

LeaFrock:
Fix #3295 . It should be compatible to old codes.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ _Not yet on NuGet..._
 * LinePlot: Added `CoordinateLine` property (#4277, #4274) @aespitia
 * Pie: Improved respect for `LineStyle` customizations (#7278)
 * Pie: Improved automatic axis limit behavior for pie charts with internal slice labels (#7278)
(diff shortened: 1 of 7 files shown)


swharden:
Thanks for this PR @LeaFrock!

The `Label` now only controls the slice label, and `LegendText` only controls what appears in the legend. They are now totally separate, and the cookbook recipes have been updated to reflect this change 👍

LeaFrock:
A reminder: the [commit](https://github.com/ScottPlot/ScottPlot/pull/4280/commits/5e99dc3141830506539964a6882785c844cdffaa) imports a break change which might affect the developers who use old versions.

PR #4286, opened 2024-09-23
SignalXY: fix exception thrown when data runs outside the view window

### LINKED ISSUE #4261: Index was outside bounds of the array --  Occurs if zooming into an area where no SignalXY values exist
**Issue:** 
I encountered a hard crash when attempting to zoom into a WPF plot using the native functionality (middle click + drag -> zoom). 
The issue was 'Index was outside the bounds of the array', being generated from the SignalXYSourceDoubleArray class.

This first photo occurs at an interesting spot, I think reached that point due to JIT doing its job, because I successfully zoomed a few times before throwing the error
image

This exception seems to occur as soon as entering the method (which blows my mind a bit). This was taken on the first time zooming after loading in the file.
image

For the record here, this library is great and my application works successfully on the various files I throw at it, except for (so far) one file. The files I'm working with are roughly 20-40MB xml files filled out in the following format : 

```
<slot PollingMode=Edge/Poll>
     <item><timestamp>1123123</timestamp><value>0</value></item>
     <item><timestamp>1123423</timestamp><value>1</value></item>
</slot>
```

When in PollingMode=Edge, I manipulate the values to create a square wave using a scatter plot (if there is a better plot let me know!)
When in PollingMode=Poll, I create an array of X values and an array of Y values, and create a SignalXY plot.

Now, I've determined that the root of this is that I was zooming in on a section of the plot where the SignalXY does not exist, as shown below. Note that the blue and orange lines on right are the only data available to those signal plots, and exist outside the zoom range.

image

**ScottPlot Version:** 5.0.39

**Code Sample:** See http://scottplot.net/faq/repro/ for tips about creating reproducible code samples

```cs
private void CreateTheProblem()
{
    var scatterX = new double[20] { 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15, 16, 17, 18, 19, 20 };
    var scatterY = new double[20] { 0, 0, 1, 1, 0, 0, 1, 1, 0, 0, 1, 1, 0, 0, 1, 1, 0, 0, 1, 1 };

    var signalX = new double[5] { 16, 17, 18, 19, 20 };
    var signalY = new double[5] { -20, 1, 15, 15, 10};

    var sig = MyPlot.Plot.Add.SignalXY(signalX, signalY, ScottPlot.Colors.Blue);
    var scat = MyPlot.Plot.Add.Scatter(scatterX, scatterY, ScottPlot.Colors.Red);

    sig.Axes.YAxis = MyPlot.Plot.Axes.Left;
    scat.Axes.YAxis = MyPlot.Plot.Axes.Right;
    scat.Axes.XAxis = MyPlot.Plot.Axes.Bottom;
    sig.Axes.XAxis = MyPlot.Plot.Axes.Bottom;
    MyPlot.Refresh();
// once view is loaded, simply zoom into an area of the plot that does not have any SignalXY drawn
}
```

Note : Adjusting so they have different X axes has no effect on this issue. BUT, Scatter does not produce the issue, only SignalXY does, which I need to use for both performance reasons and inconsistency of raw data x values.

Secondary Note : Simply zooming in and out with your scroll wheel will also cause this problem, but it can also trigger the  fault on line 77, because `GetLastPointX` incorrectly returned an index value less than `GetFirstPointX`
```cs
throw new InvalidDataException("Xs must contain only ascending values. " + $"The value at index {num} ({Xs[num]}) is greater than the value at index {num2} ({Xs[num2]})");
```

<Details>
<summary>Relevant Stack Trace </summary>

```
       at ScottPlot.DataSources.SignalXYSourceDoubleArray.GetPixelsToDrawHorizontally(RenderPack rp, IAxes axes, ConnectStyle connectStyle) in ScottPlot.DataSources\SignalXYSourceDoubleArray.cs:line 75
       at ScottPlot.Plottables.SignalXY.Render(RenderPack rp) in ScottPlot.Plottables\SignalXY.cs:line 232
       at ScottPlot.Rendering.RenderActions.RenderPlottables.Render(RenderPack rp) in ScottPlot.Rendering.RenderActions\RenderPlottables.cs:line 21
       at ScottPlot.Rendering.RenderManager.RenderOnce(SKCanvas canvas, PixelRect rect)
       at ScottPlot.Plot.Render(SKCanvas canvas, PixelRect rect) in ScottPlot\Plot.cs:line 207
       at SkiaSharp.Views.WPF.SKElement.OnRender(DrawingContext drawingContext)
       at System.Windows.UIElement.Arrange(Rect finalRect)
       at System.Windows.ContextLayoutManager.UpdateLayout()
       at System.Windows.ContextLayoutManager.UpdateLayoutCallback(Object arg)
       at System.Windows.Media.MediaContext.FireInvokeOnRenderCallbacks()
       at System.Windows.Media.MediaContext.RenderMessageHandlerCore(Object resizedCompositionTarget)
       at System.Windows.Media.MediaContext.RenderMessageHandler(Object resizedCompositionTarget)
       at System.Windows.Threading.ExceptionWrapper.InternalRealCall(Delegate callback, Object args, Int32 numArgs)
       at System.Windows.Threading.ExceptionWrapper.TryCatchWhen(Object source, Delegate callback, Object args, Int32 numArgs, Delegate catchHandler)
       at System.Windows.Threading.DispatcherOperation.InvokeImpl()
       at MS.Internal.CulturePreservingExecutionContext.CallbackWrapper(Object obj)
       at System.Threading.ExecutionContext.RunInternal(ExecutionContext executionContext, ContextCallback callback, Object state) in /_/src/libraries/System.Private.CoreLib/src/System/Threading/ExecutionContext.cs:line 179
    --- End of stack trace from previous location ---
       at System.Threading.ExecutionContext.RunInternal(ExecutionContext executionContext, ContextCallback callback, Object state) in /_/src/libraries/System.Private.CoreLib/src/System/Threading/ExecutionContext.cs:line 203
       at MS.Internal.CulturePreservingExecutionContext.Run(CulturePreservingExecutionContext executionContext, ContextCallback callback, Object state)
       at System.Windows.Threading.DispatcherOperation.Invoke()
       at System.Windows.Threading.Dispatcher.ProcessQueue()
       at System.Windows.Threading.Dispatcher.WndProcHook(IntPtr hwnd, Int32 msg, IntPtr wParam, IntPtr lParam, Boolean& handled)
       at MS.Win32.HwndWrapper.WndProc(IntPtr hwnd, Int32 msg, IntPtr wParam, IntPtr lParam, Boolean& handled)
       at System.Windows.Threading.ExceptionWrapper.InternalRealCall(Delegate callback, Object args, Int32 numArgs)
       at System.Windows.Threading.ExceptionWrapper.TryCatchWhen(Object source, Delegate callback, Object args, Int32 numArgs, Delegate catchHandler)
       at System.Windows.Threading.Dispatcher.LegacyInvokeImpl(DispatcherPriority priority, TimeSpan timeout, Delegate method, Object args, Int32 numArgs)
       at MS.Win32.HwndSubclass.SubclassWndProc(IntPtr hwnd, Int32 msg, IntPtr wParam, IntPtr lParam)
       at MS.Win32.UnsafeNativeMethods.DispatchMessage(MSG& msg)
       at System.Windows.Threading.Dispatcher.PushFrameImpl(DispatcherFrame frame)
       at System.Windows.Application.RunDispatcher(Object ignore)
       at System.Windows.Application.RunInternal(Window window)
       at App.Main()
```
</Details>

RFBomb:
Resolves #4261 by adding a validation check to check results of GetFirstPoint and GetLastPoint.

For some reason, I was only able to reproduce this bug when a plot had both a scatter and a signalXY.
- When zooming into the plot, and the signalXY was outside of the zoomed area, the application would crash because the first index was greater than the last index when attempting to draw the SignalXY (which had nothing to draw because it was outside the viewing window)

Code to reproduce ( I pasted this into the SignalXY tester in the WPF demo and called it during constructor)

```cs
private void CreateTheProblem()
    {
        var scatterX = Generate.Consecutive(20);
        var scatterY = Generate.SquareWave(20, 2);

        var signalX = new double[5] { 16, 17, 18, 19, 20 };
        var signalY = new double[5] { -20, 1, 15, 15, 10 };

        var sig = WpfPlot1.Plot.Add.SignalXY(signalX, signalY, ScottPlot.Colors.Blue);
        sig.Axes.YAxis = WpfPlot1.Plot.Axes.Left;
        sig.Axes.XAxis = WpfPlot1.Plot.Axes.Bottom;

        var scat = WpfPlot1.Plot.Add.Scatter(scatterX, scatterY, ScottPlot.Colors.Red);
        scat.ConnectStyle = ConnectStyle.StepHorizontal;
        scat.Axes.YAxis = WpfPlot1.Plot.Axes.Right;
        scat.Axes.XAxis = WpfPlot1.Plot.Axes.Bottom;

        WpfPlot1.Refresh();
        // once view is loaded, simply zoom into an area of the plot that does not have any SignalXY drawn
    }
```


RFBomb:
`DataLoggerSource` may also have this bug present! As such, it may be wise to move the validation logic into a static utility class (such as the one I provided over in #4270 )

PTA00:
我也发现了这个bug，希望快点发布新版

StendProg:
https://github.com/ScottPlot/ScottPlot/blob/743c9427e63692d31920cc424a0e62b12d6eec4a/src/ScottPlot5/ScottPlot5/DataSources/SignalXYSourceDoubleArray.cs#L100-L102
That's the whole problem with this check. It is really unnecessary and does not check anything completely. We already used binary search before it, and only after it we check that the array is sorted (which allows us to use binary search), and only by 2 points, which are relatively randomly selected depending on the Plot display settings. So we may realize that the array is not sorted, or we may not realize it, it depends on how lucky we are.

The further code is resistant to wrong index order and correctly returns an empty list, so you don't need to make an early return.

To summarize: you have added additional checks to prevent errors in the check, which is not really necessary (IMHO). 

@swharden had some considerations when he added this check. But personally I'm in favor of just removing it.

RFBomb:
This PR fixed two issues though:

it was crashing applications (including demo) if scatter and SignalXY exist on same plot and you zoom in where no signalXY is displayed. The crash was due to one of two scenarios:

- incorrectly throwing an out of range exception (how is index 0 out of range when collection has 10 items?!)
- the functions to determine start/stop indexes were throwing the exception you want removed. Not because collection is out of order, but because the first was a larger index than the second somehow. I don't understand how though. 

Either way, some validation needs to occur here to squash the bug, as I don't believe it has to do with the index search.

StendProg:
Here is you reproduce code (but in Winforms)
I fixed only this:
```c#
var scatterY = Generate.SquareWave(/*20*/ 10, 2);
```
This PR:

https://github.com/user-attachments/assets/9fa67f4a-b031-408e-b0bd-e8588f64e17a

This PR, but `if (ValidateVisibleRange(dataIndexFirst, dataIndexLast) is false) return [];` commented out:

https://github.com/user-attachments/assets/80a83720-a421-4b76-ac86-60deca448a12

It looks like your fix breaks the interpolation on the outer points due to the early return.

> Not because collection is out of order, but because the first was a larger index than the second somehow. I don't understand how though.

If the start index is larger than the end index, it is an indicator that the number of points inside is empty. Further code handles this correctly. Moreover, it correctly handles external points and interpolation by them.
The accending check itself is incorrect, it does not take into account the fact that the start index may be larger.

> incorrectly throwing an out of range exception (how is index 0 out of range when collection has 10 items?!). When it throws here, both first and last are equal to 0.

I have not been able to reproduce this.

RFBomb:
I will retest and upload a video of it crashing, along with a demo in winforms and wpf that is reproducible. I was able to reproduce this on two separate computers with fresh pulls of the library. 

You also zoom in on the signal area, that may be why yours is successful.  I was zooming to the left of the start of the area, where the scatter exists but there are no values for the signalXY.

RFBomb:
https://github.com/user-attachments/assets/c7f1a108-09da-401c-b862-aebf64dddd02

StendProg:
A very handy example, thank you.

I'm not arguing that there is a bug that throws an excepcion in the original version.
I'm just saying that it's caused by incorrect ascending check.
If you comment out the check in the original version, everything will work fine.
If you comment out ValidateVisibleRange from this PR, which also contains this check, you will also get error-free results.

If we decide to keep this check, then your PR should be adjusted to bypass the accending check in case of “invalid” indexes, but still skip them further down the code. (the missing lines when zooming deep inside the interval are present in your PR example). 
Although personally I am in favor of just removing this accending check.

> incorrectly throwing an out of range exception (how is index 0 out of range when collection has 10 items?!). When it throws here, both first and last are equal to 0

I was able to reproduce the Exception from the first case as you describe. I don't know why it displays dataIndexLast == 0 in the debugger. In fact dataIndexLast == -1 if you catch that point before the Exception. (Which is also correct for the code that follows, but the check doesn't take this into account)

RFBomb:
@StendProg I understand now, sorry for being abrasive/defensive there. Removing the exception does resolve the issue, and I was able to see after some additional testing that the `lastIndex = -1` under that first scenario. (I did check for <0 in my validation)

I was also able to reproduce the interpolation issue you mentioned while my early escape was included.

I'm uploading a new commit that removes the check altogether. Additionally, if the visible range length is 0, the binary search will always return index 0, which results in `GetColumnPixels` returning an array of 0. The `VisiblePoints` value will never be populated while `visibleRange.Length == 0`, so I put in a check for that.

Finally, I suspected that DataLogger may have the same problem because it is so similar, so I tested it. DataLogger does not have the same issue, because (unsure why this occurs) it frames the plot and prevents adjustment of X/Y. (datalogger is green in video)


https://github.com/user-attachments/assets/ee5b02f3-98aa-4a97-bfc4-76e40362d332

swharden:
Hi @RFBomb and @StendProg, thanks for your feedback!

This issue/PR reveals that the ascending value checker could fail if SignalXY data ran outside the data area.

The original solution recommended by this PR was to remove the ascending value checker. Indeed, it's not a thorough check (because inspecting every X value in the entire dataset on every frame render would be to costly). While not completely thorough, it catches many cases. Without such a check, users who feed non-ascending X values into a SignalXY plot will have no negative feedback indicating there was a problem, and the plot may have confusing artifacts without a clear explanation. I found this from personal use when I accidentally used a SignalXY instead of a Scatter plot.

If there's a better way to check for ascending values that isn't a negative performance impact, I'm happy to implement such an improvement! To correct the short term issue, I added tests that replicated the bug and added a check to avoid it, while preserving the original exception for users who accidentally use SignalXY to plot non-ascending X values.

RFBomb:
Yea, other than brute-forcing a check across the collection, which is really inefficient, I don't see a better check than the one that was originally implemented, or the one I had put in with the validation method initially (which was non-exhaustive). 

But i like the IsValid property being added, it's an elegant solution to various issues.

PR #4295, opened 2024-09-26
StrongName Sign all Sandbox and Control Assemblies

RFBomb:
Some of the assemblies were signed, and others (such as WPF Controls) were not signed. This adds the signing key to all assemblies.

- This can not be accepted unless external assemblies (like OpenTK) are also strongly named. 

Workaround:  https://brutaldev.com/post/net-assembly-strong-name-signer


swharden:
Thanks for pointing this out @RFBomb! 

I extended this PR to add a unit test that scans all .csproj files in the repository and assert they have `<SignAssembly>` defined

swharden:
Although controls all now have strong names, note that they do not have strongly named dependencies. This results in a CS8002 warning (which is now ignored in the .csproj files)

> Strong naming has no benefits on .NET Core/5+. C# compiler produces CS8002 warning for strong-named assemblies referencing non-strong named assemblies. It is fine to suppress this warning for libraries that target .NET Core/5+ only.
> 
> https://learn.microsoft.com/en-us/dotnet/standard/library-guidance/strong-naming

Issue #3795, opened 2024-05-07
Contour line (aka isoline), as new plot type or additionally to heat map


CodeBehemoth:
**Idea 1.**
Contour line additionally to heat map can make the plot more clearer. E.g. :
![image](https://github.com/ScottPlot/ScottPlot/assets/31969971/a028288e-db4d-4dad-98f9-01399444da51)
or
![image](https://github.com/ScottPlot/ScottPlot/assets/31969971/1a37ac1d-3d8f-423f-a014-e3f86fb51794)

In the first version only the color of the contour line should be configurable.

**Idea 2.**
Isoline as new plot type, such as:

![image](https://github.com/ScottPlot/ScottPlot/assets/31969971/0f0cf885-86f1-4fa8-b375-e048179fdf85)

In this case the values (heights) can be displayed on the lines.

jon-rizzo:
Putting these comments here, so they don't get lost on Discord.  Many (most?) engineering and mapping applications solve this problem using Delaunay Triangulation.  That is, given a set of 2D points, create a Triangular Irregular Network (TIN) that establishes a straight line interpolation between every point and neighboring points.  Attached to these 2D points can be any additional attribute, which might represent an elevation, temperature, light intensity, or any other value that needs to be contoured.  Building the TIN is the hard part.  The edges of each triangle represents a straight-line interpolation between values.  Once the TIN is built, contours can be drawn by simply looking at each triangle, and iterating from the lowest to the highest (elevation/temperature, etc) value, and using the edges of the triangle to interpolate the locations of the contours (and draw them) within that triangle.  Repeat for each triangle, and all contours have been drawn.

There is an open-source project, DelaunatorSharp, (here: https://github.com/nol1fe/delaunator-sharp) which is a C# implementation of the original Delaunator project.  DelaunatorSharp uses the MIT license, so as long as its copyright notice remains intact, the code can be included, adapted, modified, etc in other projects.  That is perfect, since it would likely be necessary to add a Z value to the points (the value to be contoured).  

[...]

Issue #2230, opened 2022-11-06
Contour plot: new plot type?


xichaoqiang:
@swharden 

Dear Sir,

Contourf type figure is often used in our geology community. 

And the below code and figure is contourf type figure in matplotlib.

Best regards,
Xi

```
import numpy as np
import matplotlib.pyplot as plt
fig = plt.figure()
ax = fig.add_subplot(111)
x = np.linspace(0, 3, 20) 
y = np.linspace(0, 3, 20) 
x, y = np.meshgrid(x, y) 
def fz(x, y):      
    return np.sin(x)+np.cos(y)+1  
z = fz(x, y)       

ax = plt.contour(x,y,z,colors='black')   
ax.clabel(fmt='%1.1f', fontsize=16)      
ax = plt.contourf(x,y,z,cmap='rainbow')   
ax = plt.colorbar(label="contour level")  
plt.show()
```

![image](https://user-images.githubusercontent.com/30900789/200165580-6c1b65dc-d148-4eb6-abfe-04fd19f55b99.png)

jon-rizzo:
Just to connect the dots in this thread - my [June 25th and 26 comments here](https://github.com/ScottPlot/ScottPlot/issues/3795) outline a naive (but functional) implementation for drawing contours on a Delaunay TIN of 2D points.  The TIN can be constructed by adding the code directly from DelaunatorSharp, with a small modification for providing an attribute representing the value to be contoured (Z value).  DelaunatorSharp uses the MIT license, so this use should be allowable, as long as there is proper attribution as outlined in that license.  The pieces are all here to create a TINPlot plot type, which can have various properties such settings to control whether or not to display the triangles (and what color to use), and settings to control the contour interval, color, and even heatmap options as shown above. I suggest TINPlot as the plot type instead of ContourPlot, because sometimes it might be useful to display and stylize the TIN without the contours.  Also, a TIN is only one way to contour a set of 3d points, but it is a very common solution to this problem, especially in the civil engineering world.  By naming this TINPlot, other implementations could be added in the future without causing a conflict over the name ContourPlot.

PR #4296, opened 2024-09-27
Contour Plot

jon-rizzo:
Added DelaunatorSharp code with license agreement on each file, with a few minor changes: the addition of a Z attribute to Point, and changes that were necessary to support nullable references types.  Added TINPlot and Coordinate3d classes to create TIN diagrams, Voronoi Diagrams, and draw contours.

A TINPlot is a class representing a Triangular Irregular Network, or TIN.  A TIN is a 2d triangulation of points that will allow you to interpolate the value of a third dimension on a straight line between neighboring points.  This class also allows you to draw a Voronoi diagram as well as the contours resulting from this interpolation.  Each diagram has its own point and line style settings.  Future work could include smoothing and could also be used to render a 3d surface, since the result is a set of 3d triangles.  

#2230 #3795 
Example: [surface.csv](https://github.com/user-attachments/files/17169514/surface.csv)

TINPlot

Data taken from this location:
https://cmgds.marine.usgs.gov/data/field-activity-data/2016-030-FA/

```cs
List<Coordinates3d> pts = new();
using (StreamReader sr = new(".\surface.csv")) // csv file with x,y,z points
{
    //sr.ReadLine(); // skip the header
    while (!sr.EndOfStream)
    {
        string[] cols = sr.ReadLine().Split(',');
        pts.Add(new Coordinates3d(double.Parse(cols[0]), double.Parse(cols[1]), double.Parse(cols[2])));
    }
}

TINSourceCoordinates3dArray src = new(pts.ToArray());
TINPlot tin = new(src);
myPlot.Add.Plottable(tin);
myPlot.SavePng("tin.png");
```


jon-rizzo:
Ugh.  It looks like DelaunatorSharp doesn't do any null handling.  Not sure how to resolve.

StendProg:
Hi @jon-rizzo,

1. Please add screenshots of Plottables examples in the description. It will be illustrative.
2. DelavourSharp has nuget packet. It is not clear why you need to copy the code. Is there any way to make your code modifications on top of the installed package?
3. ScottPlot tries to have a minimum number of dependencies. Maybe it would be better to make a separate project (inside the solution, for example, as Controls are implemented) that will have this dependency and plug it in as needed. Without adding the dependency to the main ScottPlot project. (Not in a hurry to redesign just thoughts for discussion.)

jon-rizzo:
> 1. Please add screenshots of Plottables examples in the description. It will be illustrative.
was hoping that someone could contribute a meaningful example.  my data is all proprietary and cannot be shared.  random points with random elevations do not make a meaningful example.  

> 2. DelavourSharp [sic] has nuget packet. It is not clear why you need to copy the code. Is there any way to make your code modifications on top of the installed package?
scott expressed a desire not to make a dependency. I included the code to avoid a dependency & added the license agreement.

> 3. ScottPlot tries to have a minimum number of dependencies. Maybe it would be better to make a separate project (inside the solution, for example, as Controls are implemented) that will have this dependency and plug it in as needed. Without adding the dependency to the main ScottPlot project. (Not in a hurry to redesign just thoughts for discussion.)
I had to make changes to DelaunatorSharp so that it would build against the null handling requirements.

Not sure how to resolve the build failure.  it is saying that I need to install various workloads.  not sure how this is done or why it is required.

StendProg:
> Not sure how to resolve the build failure. it is saying that I need to install various workloads. not sure how this is done or why it is required.

All PR's are susceptible to this error for now. It appears to be a broken pipeline. You don't have to worry about this particular error, it is not up to you.

jon-rizzo:
here is an example.  
[surface.csv](https://github.com/user-attachments/files/17169514/surface.csv)

TINPlot

Data taken from this location:
https://cmgds.marine.usgs.gov/data/field-activity-data/2016-030-FA/

swharden:
Hi @jon-rizzo, thanks for this PR! I'm just now starting to take a closer look at it.

However, I think @StendProg and I are having trouble evaluating this PR because it's not clear what the intended visual output is supposed to be. Proprietary sample data is not required to demonstrate how this plot type works if we can communicate the main idea using generated sample data.

I modified this PR to add a cookbook page demonstrating how this plot type works. When the test runs is generates the following output. Is this the output you would expect given the sample data generated here? Do you have a recommendation for a better way to generate sample data? I'm a bit confused because the output looks very different from what I'd expect to see (plots like the demonstration ones at the top of #2230 and #3795)

```cs
Coordinates3d[] cs = new Coordinates3d[500];
for (int i = 0; i < 500; i++)
{
    double x = Generate.RandomNumber(0, Math.PI * 2);
    double y = Generate.RandomNumber(0, Math.PI * 2);
    double z = Math.Sin(x) + Math.Cos(y);
    cs[i] = new(x, y, z);
}

TINSourceCoordinates3dArray data = new(cs.ToArray());
ScottPlot.Plottables.TINPlot contour = new(data);
myPlot.PlottableList.Add(contour);
```

image

Thanks for your feedback!
Scott

swharden:
I probably won't put too much more time into researching this at the moment, but highly relevant information is that Python's matplotlib has `tripcolor` and `tricontourf` functions which seem to use these data as a starting point for creating colored contour plots, so I think we are off to a good start here...

image

jon-rizzo:
So, the contouring algorithm that I implemented is admittedly very naiive.  It draws straight line contours on each triangle.  As a result, large data sets (such as the one that I had originally cited) will show more "pleasing" contours than small ones.  there are other contouring algorithms that will result in smoother (more visually pleasing, but technically less precise) contours.  there are both contour smoothing algorithms and surface (TIN) smoothing algorithms that can do this so I don't think there is one "right answer".  I think your data set did not generate contours because all of the values would be between 0 and 1.  try changing the contouring interval to something smaller, like 0,1.  Contours currently render in a single color.  the colorized triangles that you showed (or other colorization options) should be easy enough to implement in the future, but I may have to leave that to others.  Mostly, I wanted to get more eyes on this before I went too far.

jon-rizzo:
Sorry for the confusion - I have the defaults set differently than I was expecting.  By default, the contours are not shown.  you have to set the ContourLineWidth to at least 1, and your data set works best with the MinorInterval set to 0.1.   If you modify your example as follows. you should get the results below.  

`Coordinates3d[] cs = new Coordinates3d[500];
for (int i = 0; i < 500; i++)
{
    double x = Generate.RandomNumber(0, Math.PI * 2);
    double y = Generate.RandomNumber(0, Math.PI * 2);
    double z = Math.Sin(x) + Math.Cos(y);
    cs[i] = new(x, y, z);
}

TINSourceCoordinates3dArray data = new(cs.ToArray());
ScottPlot.Plottables.TINPlot contour = new(data);
contour.MinorInterval = 0.1;
contour.ContourLineWidth = 1;
myPlot.PlottableList.Add(contour);`

Contours

jon-rizzo:
The first example that I posted is an elevation contour map of a real place on the earth.  these are rarely, if ever, described as beautiful, but this is the context with which I am most familiar.  I think the triangles were making it difficult to see the contours, but can have other uses, as scott pointed out in his images.  my previous attempts with random points used random elevations, and the results were far from 'beautiful' because that type of data is not very meaningful.  It did not occur to me to use a trig function based on X & Y to calculate Z.  

There are many more features that can be added in the future - the triangles themselves could be colored based a heatmap (like scott is showing on the left) or the space between the contours can be colored based on a heatmap (like scott is showing on the right).  the contours could also be colored based on a heatmap. the most useful feature to most folks, however, would probably be adding labels to each index contour, each intermediate contour, or both.  The plot control can also draw Voronoi diagrams of the resulting TIN, but I do not know the practical applications of those diagrams, or if any such applications even exist.

swharden:
Hi @jon-rizzo, thanks for the clarification - that helps a lot! The path forward is now a lot easier to see.

Also @StendProg I **love** your example showing the contour lines over the heatmap!

I'll continue to work on this PR and create cookbook examples as I go. I'll probably work toward moving as much logic as possible out of the `TINPlot` plottable and put it in static classes so the same logic can be used by other plot types. 

Presently a lot of math is performed in the render loop, but it seems unlikely that users will be creating plots like this from "live data", so that work could probably be done up front. I'm even thinking that, when given a collection of 3d points, users may want to get different things to create different types of plots (perhaps leveraging existing plottables):
* A collection of lines (to show show the triangulated network) using `AddLine()` or `AddLines()`
* A collection of triangles (to create a Voronoi diagram) using `AddPolygon()`
* A collection of paths (to show contour lines) using `AddPolygon` or perhaps a new `ContourLine` plot type that has logic built in to place labels inline

I'll keep experimenting and commit changes to this PR as I go. 
Thanks again everyone for your input along the way!

jon-rizzo:
> but it seems unlikely that users will be creating plots like this from "live data", 

unlikely, but not unreasonable or impossible.  I have seen several dynamic "grading optimizers" (in the civil engineering world) that adjusts elevations of all of the points and re-renders the tin and contours continuously while the optimizer converges on a solution . There are also simple widgets like this one: https://www.geogebra.org/m/Bkn8sw7v

swharden:
> but it seems unlikely that users will be creating plots like this from "live data"

> unlikely, but not unreasonable or impossible

Still working on this, but I'll precalculate all the lines when the plottable is constructed, but expose an `Update` method to allow users to recalculate the network if/when data changes.

I greatly simplified `TINPlot` (and the `Delaunator` backend), and am continuing to work toward separating calculation logic from display logic. I may set this down soon today and pick it up again tomorrow 👍

StendProg:
Very strange algorithm for finding contours. In fact, it just hatch each triangle with a given step, starting from the minimum value. Hoping that the hatchings on different triangles will add up to a single path. This works in some form, but it's not hard to come up with exceptions breaking this. The only advantage is speed. In a good way I think it is necessary to traverse neighboring triangles making a single path, but this is probably a very difficult and computationally expensive task.

For example: here the vertices of the triangle are deliberately made a little less than 0.5. As a result, we have a highly irregular pitch of contour lines on the square.
IrregularContour

After some thought I came to the conclusion that if we want to build a contour on the heatmap, we do not have to triangulate it, we can do bilinear filtering directly on the rectangles that make up the heatmap.

jon-rizzo:
As you mentioned, the benefit is speed.  That, and simplicity of implementation.  a spatial index could easily be created if the desire is for continuous, connected paths.  There are more advanced algorithms for contouring to achieve smoother contours.  There are many established algorithms for doing all of these things.  Delaunay Triangulation is one very common solution, particularly with points that have irregular or random distribution.  as a bonus, the resulting triangles can also be rendered quickly in 3D or displayed with different colors based on the elevation or slope to perform various analyses.  if the points are on a grid, other solutions might be faster and/or more elegant.  For example, the CONREC algorithm. but, my need is specifically for random points.  I don't think there is one "best" way to draw contours for all situations, so I don't view this as an either/or choice.  more solutions = more better.

swharden:
Okay, I brought this forward to a place where I'm happy to merge. This works now:

```cs
myPlot.Add.ContourLines(coordinates);
```

### Limitations / Improvements

The `ContourLines` plottable constructor has an overload which accepts `Coordinates[,]` for displaying contour lines from rectangular datasets. As suggested earlier, a better algorithm (bilinear filtering?) could be used to generate contour lines for this case.

Also every `ContourLine` presently only has a `Path` with two points. In other words, the contour lines are a bunch of short segments and aren't actually "connected" into polygons. This would have to be fixed to enable filled contour plots.

I don't intend to put time into these improvements right now (I wish to focus on some other open high priority issues) but if one of you wants to add this enhancement and create a PR I would be happy to review it!

## New Cookbook Recipes

### Contour Lines from a Irregular Points (`Coordinates`)
image

### Contour Lines from a Rectangular Grid (`Coordinates[,]`)
image

### Contour Lines over Heatmap
image

### Contour Lines with Colormap
image

Issue #4304, opened 2024-10-07
coxcomb positon of inital slice moves depending on total number of slices


johndoh:
**Issue:**

The position of the first slice on a coxcomb chart moves depending on the number of slices in the chart. Even if the Rotation property is set to 0. This is different to the pie chart where the first slice is always in the same place.

| # | Pie | Coxcomb |
| - | --- | ------- |
| 3 | ![pie3](https://github.com/user-attachments/assets/7f03b014-ac54-4085-8919-270a67cbaf8f) | ![coxcomb3](https://github.com/user-attachments/assets/ed825923-3ab2-4e71-a47a-53b09c27dbb5) |
| 4 | ![pie4](https://github.com/user-attachments/assets/f7186bf8-85d8-4290-8f11-11f11effe40a) | ![coxcomb4](https://github.com/user-attachments/assets/01a057e8-e02c-4648-9018-b88b82e05164) |
| 5 | ![pie5](https://github.com/user-attachments/assets/35c10ca6-4635-405d-a0ea-9d81a54ec657) | ![coxcomb5](https://github.com/user-attachments/assets/62be63a6-b2eb-4584-b2b3-3f5c2cc27e5f) |

Expected behavior: the position of the first slice is always in the same page.

**ScottPlot Version:** 5.0.39

**Code Sample:**

```cs
double[] values = { 5, 2, 8, 4, 8 };
for (int j = 3; j <= 5; j++)
{
    ScottPlot.Plot myPlot1 = new();
    myPlot1.Add.Pie(values.Take(j));
    myPlot1.SavePng("pie" + j + ".png", 400, 300);

    ScottPlot.Plot myPlot2 = new();
    myPlot2.Add.Coxcomb(values.Take(j));
    myPlot2.SavePng("coxcomb" + j + ".png", 400, 300);
}
```

I think the problem might be the first canvas rotation is done twice:

https://github.com/ScottPlot/ScottPlot/blob/743c9427e63692d31920cc424a0e62b12d6eec4a/src/ScottPlot5/ScottPlot5/Plottables/Coxcomb.cs#L51-L55

so moving line 55 to the end of the `for` block might fix it

StendProg:
Hi @johndoh,

Looks like you've got it all figured out pretty well. I've tested your solution, everything works as it should. Would you like to do a PullRequest? 
The alternative would have been to skip the first rotation more explicitly:
```c#
  if (i != 0)
[...]

PR #4305, opened 2024-10-07
Coxcomb: fix position of intiial slice to match rotation setting (#4304)

johndoh:
fix for the initial double rotation issue described in #4304


swharden:
Thanks @johndoh, this looks great! 

I refactored it a bit to overcome merge conflicts caused by #4278 and #4280 but it's looking good now and will be in the next release 👍

PR #4316, opened 2024-10-10
CoordinateRange: refactor and improve tests

CoderPM2011:
I reorganized `CoordinateRange` and added new properties and methods.
If something is wrong, please tell me. 😉

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,7 @@ _Not yet on NuGet..._
 * Coxcomb: Improved first slice orientation consistency (#4304, #4305) @johndoh
 * Grid: Improved alternating fill grid appearance for live and interactive graphs (#4302, #4307) @StendProg @bradcb212
 * SignalXY: Fixed bug causing plots with inverted horizontal axes to crash under specific conditions (#4313, #4315) @StendProg @lguelat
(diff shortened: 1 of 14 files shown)


swharden:
Hey @CoderPM2011, thanks for this PR! Wow, `CoordinateRange` was pretty sloppy 🤪 You did a great job making it better!

There's nothing you did that's incorrect, but for good measure I'll review this file line-by-line (and add tests as I go) because I want to make sure this looks nice and stays working as expected moving forward.

The biggest potential issue is that it's not clear whether `CoordinateRange` supports inverted ranges. The constructor arguments suggest that the first value should be less than the second, but I don't think it's enforced, and things like `Span` assume this is the case or a negative value could be returned (but maybe that's the desired behavior?). It's questions like this I'll think about as I review `CoordinateRange` today. I'll commit changes to this PR as I go and merge shortly.

Thanks again for your work on this PR, and all the other ones up until now! You've done a lot of great work to make this library awesome 🚀 ✨

idotta:
Is there a reason for `CoordinateRange.Center` being removed or was it a mistake while refactoring? I was using it before.

swharden:
Seems like a mistake! It should definitely be there 😅 

Ideally with a test to make sure it doesn't get removed accidentally again 🤪

idotta:
I'll try to make a PR to fix it asap then. Thanks!

PR #4321, opened 2024-10-12
Modify PolarAxis

### LINKED ISSUE #4310: Radar plot labels do not match data
**Issue:** 
Radar plot data points are populated in the inverse direction of the labels. This is true regardless if clockwise is true or false when passed into `SetSpokes()`. 

**ScottPlot Version:** 5.0.39

```cs
var plot = new ScottPlot.Plot();

double[] values = [0, 1, 2, 3, 4, 5];
var radar = plot.Add.Radar(values);

double[] tickPositions = [1, 2, 3, 4, 5];
radar.PolarAxis.SetCircles(tickPositions);

var labels = values.Select(v => v.ToString() ).ToArray();
radar.PolarAxis.SetSpokes(labels, values.Max() + 1);

var path = new FileInfo(Path.Combine(Environment.SpecialFolder.Desktop.GetPath(), "radar.png"));
plot.SavePng(path.FullName, 400, 300);
```

radar

Setting a breakpoint right before `SavePng()` and inspecting `plot.PolarAxis.Spokes` shows that each spoke has the correct relative angle (assuming negative angles are clockwise rotation); however `PolarAxis.RotationDegrees` seems to have the wrong sign relative to the direction of rotation.
image
Unless my interpretation of these properties is incorrect (which could definitely be the case), the entire plot should be rotated clockwise 90 degrees, and then each spoke should be rotated 60 degrees clockwise from there instead of the 90 degree counter-clockwise for the plot and then 60 degree clockwise rotation for each spoke in the series.

CoderPM2011:
- fix clockwise and counterclockwise results of `GetCoordinates()`
  fixed #4310
- rename `RotationDegrees` to `Rotation` and change type to `Angle`
- refactor `GetCoordinates()` with `PolarCoordinates`
- synchronizes `Rotation` with the angular direction of the Cartesian coordinate system
  please refer to the following for differences

## Default
``` cs
var polarAxis = plot.Add.PolarAxis();
```
image

## Rotating polar axis
``` cs
polarAxis.Rotation = Angle.FromDegrees(90);
```
### Before
image

### After
image

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,7 @@ _Not yet on NuGet..._
 * Axes: Added a `Plot.Axes.TightMargins()` shortcut for setting autoscale margins to tightly fit the data
 * ContourLines: New plot type for displaying lines that mark points of equal elevation given a collection of 3D points (#4296, #2330, #3795) @jon-rizzo @StendProg
 * Maui: Improved the .NET MAUI ScottPlot control and added quickstart documentation to the website (#4320, #4023, #4013) @KosmosWerner @ByteSore
+* Radar: Improved rotational direction of labels (#4321, #4310) @CoderPM2011 @bry-decelles
 
 ## ScottPlot 5.0.39
 _Published on [NuGet](https://www.nuget.org/profiles/ScottPlot) on 2024-08-02_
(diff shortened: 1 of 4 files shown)


swharden:
Thanks!

This also changes fixes the existing cookbook recipe

```cs
double[,] values = {
    { 78,  83, 84, 76, 43 },
    { 100, 50, 70, 60, 90 }
};

var radar = myPlot.Add.Radar(values);

string[] spokeLabels = { "Wins", "Poles", "Podiums", "Points", "DNFs" };
```

before | after
---|---
image|image

PR #4326, opened 2024-10-12
Contour lines on rectangular grid

StendProg:
Contour lines are now calculated without triangulation of the original grid.

The MarchingSquares algorithm is used to find the contour.
The resulting lines are merged into a common path. There can be several of them for one elevation level.

Since our lines are merged into a path, we can add a label on top of the drawn path.

Grid primitives use indexes instead of real values. This is useful for accurate comparison of integer values, and reduces the number of interpolations by about half (we discard half of the points in the merge process)
ContoursWithLabels

It looks like the Heatmap was originally flipped by Y. I did the reverse flip manually in the cookBook example.


swharden:
Thanks so much @StendProg! This is looking _fantastic_!

image

Issue #4353, opened 2024-10-17
LogMinorTickGenerator infinite loop


Cassar17:
**Question:** 

I am having an issue with the LogMinorTickGenerator function and I'm not sure whether it's a bug or if I'm not using it correctly. In my application, I have multiple charts that utilize the logarithmic scale, but for some reason the program completely crashes when it tries to resize the chart, which can happen if the user resizes the entire application window, or only the section in which the chart is plotted.

After some inspection, I noticed that in the LogMinorTickGenerator class, there's a function GetMinorTicks which is always triggered whenever the chart is resized. This function receives a list of doubles named majorPositions, and there's a for loop iterator that increments by a deltaMajor which is calculated by the difference between the first and second values of majorPositions. 
Problem is, sometimes the majorPositions list comes with two identical values, and the deltaMajor value will be 0, resulting in an infinite loop and a crash.

![image](https://github.com/user-attachments/assets/2bfbc67e-f910-42c8-b352-9f03dca09fc3)
![image](https://github.com/user-attachments/assets/11dfa05d-eeb9-4049-9cff-7d9b9edd2cb6)

I couldn't figure out where this majorPositions come from, and I don't kow if I have any way of defending it against sending identical values. 

The code below shows the way I'm building the plot with the logarithmic scale.

Am I doing something wrong, or is this a bug that's beyond my control?

Thank you for your time

**ScottPlot Version:**
5.0.40

**Code Sample:** 

```cs
ScottPlot.TickGenerators.LogMinorTickGenerator minorTickGen = new();
ScottPlot.TickGenerators.EvenlySpacedMinorTickGenerator bottomGen = new(5);
ScottPlot.TickGenerators.NumericAutomatic tickGen = new();
ScottPlot.TickGenerators.NumericAutomatic bottomTicks = new();

//minorTickGen.Divisions = 5;
tickGen.MinorTickGenerator = minorTickGen;
bottomTicks.MinorTickGenerator = bottomGen;
bottomTicks.IntegerTicksOnly = true;

//bottomTicks.TargetTickCount = Convert.ToInt32(logXs[^1]);
static string LogTickLabelFormatter(double y) => $"x10^{y:F1}";

//bottomTicks.IntegerTicksOnly = true;

tickGen.IntegerTicksOnly = true;

tickGen.LabelFormatter = LogTickLabelFormatter;
//bottomTicks.LabelFormatter = LogTickLabelFormatter;
[...]

PR #4354, opened 2024-10-17
Fix typo in GenerateTickPosition

StendProg:
With very small Plot sizes, MajorTickGenerator sometimes returned only 2 absolutely identical positions. This caused problems when calculating logarithmic axes. #4353
This may be due to a typo corrected in this PR.


swharden:
Thanks @Cassar17 for reporting this issue, and @StendProg for the quick fix! 

I'm merging this now and it will be in the next release, likely published on NuGet near the end of this weekend 👍

PR #4357, opened 2024-10-18
Added Center property back to CoordinateRange

idotta:
Added Center property back to CoordinateRange, removed unintentionally in #4316.

Also added tests for properties `Span`, `Center` and `Length`, to make sure they're not removed unintentionally.

```cs
CoordinateRange range = new(0, 10);
double center = range.Center;
```


swharden:
Awesome, @idotta, thanks so much! Tests and everything 🚀 ✨ 

This will be in the next release 👍

PR #4374, opened 2024-10-22
WinUI: fix conditional compilation for .NET 8

### LINKED ISSUE #4288: WinUIPlotMenu.OpenSaveImageDialog() is broken
**Issue:** This condition is no longer enabled after https://github.com/ScottPlot/ScottPlot/commit/e44aee1ff98d7f96f5493d01f78f908c3ca80d08 bumped the target framework, causing an exception.

https://github.com/ScottPlot/ScottPlot/blob/743c9427e63692d31920cc424a0e62b12d6eec4a/src/ScottPlot5/ScottPlot5%20Controls/ScottPlot.WinUI/WinUIPlotMenu.cs#L84-L88

```
Invalid window handle. (0x80070578)
Consider WindowNative, InitializeWithWindow
See https://aka.ms/cswinrt/interop#windows-sdk
```

**ScottPlot Version:** 5.0.39

**Code Sample:** N/A

vilgotf:
.NET 8 was recently added as a target framework which exposed this condition as too limited as it is only true for .NET 6.

Since WinUI 3 requires .NET 6 and Windows 10 1809, this condition can be wholly removed; dependents will fail the WASDK conditions regardless.

Closes #4288

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ _Not yet on NuGet..._
 * Bar: Use `LineStyle` and `FillStyle` similar to other plottables and deprecate old property names
 * Histogram: Refactored to simplify creation of creating probability curves and cumulative probability histograms (#4287, #4367)
 * Colormap: Refactor all `ScottPlot.Colormaps` to favor composition over inheritance and add extension methods to `IColormap` (#4248)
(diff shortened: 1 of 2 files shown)


swharden:
Thanks @vilgotf! I'll publish a new package on Nuget within the next few days and I'm excited that this will be in it 🚀

PR #4394, opened 2024-10-27
CoordinateRangeMutable: Adjust based on CoordinationRange

CoderPM2011:
Since the `CoordinationRange` was adjusted in #4316, 
`CoordinateRangeMutable` should also try to maintain the same architecture as `CoordinationRange`.

### DIFF
--- a/src/ScottPlot5/ScottPlot5 Cookbook/Recipes/General/AxisRecipes.cs
+++ b/src/ScottPlot5/ScottPlot5 Cookbook/Recipes/General/AxisRecipes.cs
@@ -293,7 +293,7 @@ public class AxisAndTicks : ICategory
                 new ScottPlot.TickGenerators.TimeUnits.Hour(), 1,
                 // Here we provide a delegate to override when the ticks start. In this case, we want the majors to be
                 // 00:00, 06:00, 12:00, etc. and the minors to be on the hour, every hour, so we start at midnight.
-                // If you do not provide this delegate, the ticks will start at whatever the Min on the x-axis is.
+                // If you do not provide this delegate, the ticks will start at whatever the Value1 on the x-axis is.
                 // The major ticks might end up as 1:30am, 7:30am, etc, and the tick positions will be fixed on the plot
                 // when it is panned around.
                 dt => new DateTime(dt.Year, dt.Month, dt.Day));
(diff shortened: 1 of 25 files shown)


swharden:
Hi @CoderPM2011, thanks so much for this PR!

I'll be honest, I'm having an extremely difficult time reviewing this because it appears to be very breaking and touches so many files 💀 

Could this have been less breaking and affected only one file if `CoordinateRangeMutable.Min` and `CoordinateRangeMutable.Max` remained as properties? I'm 100% behind adding `Value1` and `Value2` as new names and allowing them to be inverted, but it's not super clear whether all these modified files and breaking changes are necessary to allow for that feature.

I'm trying to get through about 200 issues over the next [however many days that takes], so it's not really time effective for me to put too much effort into this one at the moment. Any insight you have about how to simplify this PR would be really helpful! 

Thanks!

CoderPM2011:
Since the original `Min` and `Max` could represent inversions by themselves, and since many classes rely on this feature, they were renamed `Value1` and `Value2` to preserve this feature.

---

This class may not be so well adjusted.😣

In fact, I have tried other methods before, but then gave up. 😵
1. Change it to struct, like `CoordinateRange`
  This will lead to changes in the way `Min` and `Max` assignments are written.
2. Contains `CoordinateRange` and replaces internal logic with `CoordinateRange`
  It's a bit complicated to implement, and I forgot if there were any other reasons.😓
3. Adjust the getters and setters of the `Min` and `Max` attributes to automatically set `Value1` and `Value2`, but return the real `Min` and `Max` like `CoordinateRange`.
  After implementation, this logic seems to be untenable.

---

Maybe use the previous version of the `CoordinateRange` schema instead.🤔
But it will still be different from the current version.😥
https://github.com/ScottPlot/ScottPlot/blob/f0279f28c203218415f6f7b0a51606dab76bc36d/src/ScottPlot5/ScottPlot5/Primitives/CoordinateRange.cs#L12-L14

swharden:
> After implementation, this logic seems to be untenable.

It seems like you put a lot of time into looking into this, and the problem is more complex than it seems at first!

I notice there's a `Range`, `IndexRange`, `CoordinateRange`, and I was reminded today we even have `PixelRangeX` and `PixelRangeY` 🤪 Maybe some day in the future we can redo all the range functions at once, and lean on a lot of shared logic and an extensive test system?

That seems like a lot of work, but maybe one day when we get through the backlog of open issues that may be a good target for refinement 🤔

I'm inclined to close this issue for now, and reference it when this topic gets picked up again. What do you think? I feel bad suggesting closing PRs that people have put a lot of effort into, but it seems I'm unable to give this topic the attention it deserves presently and it may be a better experience for both of us if we pick it up again when after the larger open issues are resolved and we can take as long as we want to work on it.

swharden:
Following-up, working on #4471 this is a minefield of really hard to predict side effects

Automatically rectifying ranges is required for things like axis rules

Not automatically rectifying ranges is required for things like saving and recalling axis limits for inverted axes

Let's not take any action right now, but I'm thinking a better solution to this overall problem is to leverage the type system to split out an `AxisRange` type specifically for working with axis limits and giving it some strict rules and features that don't apply to more general purpose types like `CoordinateRange`

Issue #4416, opened 2024-11-01
Graphics glitch in Maui windows and iOS


King-Taz:
ScottPlot 5.0.42 in Maui

Has anyone else observed a graphics glitch when panning in Windows and iOS? Android does not appear affected. When panning is started, the plot momentarily jumps off screen and comes back on screen again. The only method that corrects the behavior is to "autoscale" after app loads using context menu. Autoscale and Refresh in the program will not correct the behavior, it has to be autoscale in the plot window.

The maui app is simple. In the XAML 

`    <Grid ColumnDefinitions="*" RowDefinitions="*" Padding="30,20,30,20">
        <ScottPlot:MauiPlot x:Name="MauiPlot1" />
    </Grid>
`
In the code behind, call this method after app initialization

`        private void SimpleDrawPhasorPlot()
        {

            //Create arrow base coordinates used for all vectors
            Coordinates arrowBase = new(0, 0);

            // create arrow 1
            Coordinates arrow1Tip = new(5, 0);
            CoordinateLine arrow1Line = new(arrowBase, arrow1Tip);
            // add a simple arrow
            var arrow1 = MauiPlot1.Plot.Add.Arrow(arrow1Line);
            arrow1.ArrowFillColor = Colors.Red;
            arrow1.ArrowLineWidth = 0;
            arrow1.ArrowWidth = 5;
            arrow1.ArrowheadLength = 20;
            arrow1.ArrowheadAxisLength = 20;
            arrow1.ArrowheadWidth = 20;
            var dl1 = MauiPlot1.Plot.Add.Text("T(1)", arrow1Tip);
            dl1.LabelFontSize = 36;
            dl1.LabelFontColor = Colors.Black;

            //Enable touch
            MauiPlot1.EnableTouchEvents = true;

            //Autoscale
            MauiPlot1.Plot.Axes.AutoScale();

        }`

Without Touch enabled, the behavior is even more erratic. Feel free to add Refresh, it doesn't change anything.

KosmosWerner:
Hi! I just published a PR [here](https://github.com/ScottPlot/ScottPlot/pull/4417), fixing the glitchs you mentioned. I have corrected them, and now everything should work well.

[...]

PR #4417, opened 2024-11-01
MAUI: improvements in pan and zoom

KosmosWerner:
Hi! In https://github.com/ScottPlot/ScottPlot/issues/4416, errors were reported in the desktop version. These errors have been fixed, and some unnecessary parts of the code have also been removed.
These errors were caused because two events were responsible for panning the plot. Mouse buttons as well as the scroll wheel can now be used in desktop versions 😀
There is still a minor issue: when holding down the screen and then adding another finger to zoom, the plot jumps. This is equivalent to holding down the mouse click and then moving the scroll wheel. However, for now, these are the most critical fixes


swharden:
Great fix @KosmosWerner, thanks so much! I'm excited this will be in the next release, likely published near the end of this weekend 🚀

PR #4425, opened 2024-11-03
WinForms: improve the designer experience for .NET Framework apps

### LINKED ISSUE #4362: WinForms: improve the designer experience for .NET Framework apps
SkiaSharp can't load in the designer view of .NET Framework WinForms apps. The control appears in the toolbox, but drag-dropping it onto the window causes a spectacularly confusing error (screenshot below).

The presently recommended workaround is to add a `FormsPlot` programmatically (https://scottplot.net/faq/add-winforms-programmatically/) but some users may find this confusing (e.g., #4312)

This issue will track using preprocessor directives to create a version of `FormsPlot` specifically for .NET Framework projects that does not use SkiaSharp and will not crash in this way, but that will replace itself at runtime with a regular `FormsPlot` containing the SkiaSharp surface.

image

CoderPM2011:
resolve #4362

After investigation, it was found that `FormsPlot` inherits `FormsPlotBase`, and its member `Plot` will call the method in `SkiaSharp`, so `FormsPlotBase` also needs to be adjusted.

Adjusted to detect whether `FormsPlotBase` throws an exception that `libSkiaSharp.dll` cannot be found, and display the current version and error message on the control.

like this
image

--- 

I found that I had to be careful whether `FormsPlot` properties were set in `Form.Designer.cs`.
If the property calls `SkiaSharp`, it may also affect the control and cause an error.🤔

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ _Not yet on NuGet..._
 * Financial Charting: Added experimental `FinancialTimeAxis` plottable as an alternative to using custom axes or tick generators (#4385) @quantfreedom @vladislavpweetsoft
 * Triangular Axis: Users may now `Add.TriangularAxis()` and use its methods to get Cartesian coordinates from points in triangular space (#4421, #4413, #4424) @manaruto
 * Bar: Exposed `Rect`, `ErrorLines`, and `AxisLimits` properties (#4423) @tiger2014
+* FormsPlot: Improved the Visual Studio design time experience for users working on .NET Framework projects (#4425, #4362) @CoderPM2011
 
 ## ScottPlot 5.0.42
 _Published on [NuGet](https://www.nuget.org/profiles/ScottPlot) on 2024-10-29_
--- a/src/ScottPlot5/ScottPlot5 Controls/ScottPlot.WinForms/FormsPlot.cs
+++ b/src/ScottPlot5/ScottPlot5 Controls/ScottPlot.WinForms/FormsPlot.cs
@@ -19,7 +19,7 @@ public class FormsPlot : FormsPlotBase
 
     public FormsPlot() : base()
     {
-        if (NotFoundSkiaDll)
+        if (IsDesignerAlternative)
         {
             return;
         }
--- a/src/ScottPlot5/ScottPlot5 Controls/ScottPlot.WinForms/FormsPlot.cs
+++ b/src/ScottPlot5/ScottPlot5 Controls/ScottPlot.WinForms/FormsPlot.cs
@@ -9,18 +9,20 @@ namespace ScottPlot.WinForms;
 [ToolboxItem(true)]
 public class FormsPlot : FormsPlotBase
 {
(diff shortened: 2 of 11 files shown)


swharden:
Wow @CoderPM2011, great find! This is fantastic! 🚀

swharden:
Hi @CoderPM2011 - I'm curious, how are you testing this?

I have a SDK project that I set to target `net48` and I'm getting this 😅 

image

swharden:
Oof, it seems to be failing for me the same way when I try to load any control (including a blank one) from the `ScottPlot.WinForms` assembly 💀

CoderPM2011:
> Hi @CoderPM2011 - I'm curious, how are you testing this?

I first have the `ScottPlot.WinForms` project generate the Nuget package.
Then in the newly created `.Net Framework 4.6.2` project, install the locally generated Nuget package for testing.

swharden:
> > Hi @CoderPM2011 - I'm curious, how are you testing this?
> 
> I first have the `ScottPlot.WinForms` project generate the Nuget package. Then in the newly created `.Net Framework 4.6.2` project, install the locally generated Nuget package for testing.

Ah, interesting! I've only tested this running from source which may explain why I've had difficulty reaching a solution.

Thanks again for your feedback and this PR! I'm excited to give this a try soon 👍

CoderPM2011:
> Ah, interesting! I've only tested this running from source which may explain why I've had difficulty reaching a solution.

I'm guessing that the reason for the failure to work from the source code may be related to multiple target frameworks or whether VS itself is running properly. 🤔

CoderPM2011:
In my repository branch [test-#4362](https://github.com/CoderPM2011/ScottPlot/tree/test-%234362).

I tested the pattern directly referenced in the source project.
I found that the multi-target framework does affect the toolbox.
Maybe it has something to do with this https://github.com/dotnet/msbuild/issues/7856 .

If I change `ScottPlot.WinForms` to only target `net462` then I get it working.
Sometimes you encounter an exception. After restarting `Visual Studio`, try again and it will work normally.

My test result
image

swharden:
> I found that the multi-target framework does affect the toolbox.

Wow, that was the magic piece of information! I got it working now and will refine the error screen a bit. This is what I'm thinking of having it show when the DLL fails to load:

image

... but if this is successful after the next release, I'm tempted to use `System.Drawing.Graphics` to manually draw a title, grid, and axes, and display it as a `Bitmap` in a `PictureBox` and I bet most people will have no idea there was a failure! 🚀

swharden:
I'll merge this now, but note that it has been modified to only show this message if design time mode is detected. The reason is I don't want that message accidentally showing up on client machines of a third party publishes an app using ScottPlot. Instead, we can explore adding a new PR to display a user-facing message (not aimed at developers necessarily) for when the DLL cannot be located. 

This is a semi-common issue actually, unrelated to running inside visual studio, because sometimes people build an app but forget to include all those platform specific subfolders (e.g., I once published a WinForms app and omitted the arm folders but someone with one of those new windows arm tablet PCs couldn't load the app)

Issue #4253, opened 2024-09-12
SignalConst: Expose `Data` to allow users to set X and Y offsets


RR-Git-Projects:
**Question:** (Ask your question here)
I can plot Signal plots just fine using a date axis and offset, but doing the same process on a SignalConst plot seems to be impossible?

For a Signal plot (plotting data by the minute) this works great:

```
List<Double> ys = new List<Double>();

for (int row = 0; row < test_data.Rows.Count; row++)
{
    ys.Add(Convert.ToDouble(test_data.Rows[row][1]));
}

ScottPlot.Plottables.Signal line = SignalPlot.Plot.Add.Signal(ys: ys.ToArray());
line.Data.XOffset = System.DateTime.Now.ToOADate();
line.Data.Period = 1.0 / (60.0 * 24.0);
SignalPlot.Plot.Axes.DateTimeTicksBottom();
SignalPlot.Plot.Axes.AutoScale();
SignalPlot.Refresh();
```

But the Data property is obviously inaccessible for SignalConst plots, so dates always start at 1900-01-01. I tried various ways of trying to add the offset *before* plotting, but none of them seem to work. Am I missing something here?

**ScottPlot Version:** 
Scottplot.Blazor 5.0.39

matej-mnoucek:
I miss the same functionality 🙏

swharden:
It sounds like the core of this issue is that you cannot set the X offset when using SignalConst

> But the Data property is obviously inaccessible

I suspect this is an oversight. It looks like it is private readonly.

https://github.com/ScottPlot/ScottPlot/blob/ede0a913b6cedc116392a3f57c1764554a07ab90/src/ScottPlot5/ScottPlot5/Plottables/SignalConst.cs#L8

If someone wants to make a pull request to make that public, I'd be happy to merge it in quickly:

```cs
public SignalConstSource<T> Data { get; } = new(ys, period);
```

matej-mnoucek:
PR: #4440

RR-Git-Projects:
Thanks for this! (and thanks [matej-mnoucek](https://github.com/matej-mnoucek) for doing the pr, I have never done one on somebody else's repo and I was nervous about trying.) It's neat to see people actually upkeeping their libraries. Your tool is honestly amazing, it's found its way into a number of projects I'm working on. Keep up the wonderful work!

PR #4440, opened 2024-11-04
fix: make `Data` property in SignalConst public

matej-mnoucek:
Make `Data` property public, so users can e.g. control `XOffset` when using SignalConst.

Ref: #4253


swharden:
Thanks @matej-mnoucek! This will be in the next release 🚀

PR #4441, opened 2024-11-04
SVG: XML contains duplicated content due to uninitialized axis

### LINKED ISSUE #4436: SVG: XML contains duplicated content
**Issue:**

I am using ScottPlot to render plots in PDFs using QuestPdf. I was confused why the render time in QuestPdf was so much higher than expected - until I noticed that the SVG produced by `GetSvgXml` and `GetSvgHtml` is duplicated fives times! That is, it contains the initial `svg` element and then the content is repeated five times 

```xml
<?xml version="1.0" encoding="utf-8" ?>
<svg xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink" width="400" height="300">
    <rect fill="white" width="400" height="300"/>
    ...
    <rect fill="white" width="400" height="300"/>
    ...
    <rect fill="white" width="400" height="300"/>
    ...
    <rect fill="white" width="400" height="300"/>
    ...
    <rect fill="white" width="400" height="300"/>
    ...
</svg>
```

**ScottPlot Version:**

5.0.43

**Code Sample:**

```cs
using ScottPlot;

var plt = new Plot();
var svgString = plt.GetSvgXml(400, 300);

Console.WriteLine(svgString);
```

The repeated content overlays each other perfectly, so it is not visually noticeable. You have to inspect the SVG output.

sunefred:
Partly fixes #4436 

This fixes an issue where an uninitialized axis (such as the `Right` axis for a default `Plot`) would result in the render counter to max out at five (5) renders. The condition that checks weather an axis has changed or not was updated to correctly handle such an axis where range is `[Infinity, -Infinity]` (reversed).

Additionally the equality implementation for `CoordinateRange` was updated. It will now correctly differentiate between `[-Infinity, Infinity]` and `[Infinity, -Infinity]`.

I am sure there are other reasons that multiple render-passes could happen, in which case the SVG output would still contain duplicated content. This PR only fixes the particular issue with uninitialized axis.


swharden:
Thanks @sunefred, I appreciate you diving down to find the surprisingly nuanced bug underlying this issue! I'll merge this now and it will be in the next release, likely published to NuGet within the next few days or by the end of this weekend at the latest.

Thanks again! 🚀

PR #4460, opened 2024-11-08
DataLogger: use `List<Coordinates>` instead of `IList<Coordinates>` and document `Data.RemoveRange()`

Fruchtzwerg94:
DataLogger does not allow to limit its points which means its points will increase forever.
Added two functions allowing to limit by X or length manually.


swharden:
Hi @Fruchtzwerg94, thanks for this suggestion! I will modify this PR to adapt a similar but alternative implementation of this functionality, and will also document its use in the cookbook 👍 

```cs
// setup a logger that will grow as data is added
var logger = myPlot.Add.DataLogger();

// add ten values
logger.Add(Generate.RandomSample(10));

// remove the oldest five values
for (int i=0; i<5; i++)
{
    logger.Data.Coordinates.RemoveAt(0);
}
```

image

Fruchtzwerg94:
Hi @swharden , 

awesome, thanks for bringing it in so fast. A minor thing: Since `DataLogger.HasNewData` is private, it can't be set after items changed. Adding a way to set it would be nice as well. I can prepare a PR for it but I assume its faster for you just to bring in a minor change here without a PR.

Thanks,
Philipp

swharden:
> I can prepare a PR for it but I assume its faster for you just to bring in a minor change here without a PR

Actually if you're up for making a PR, I'd appreciate it! I can merge that from the browser (or even my phone) and don't have to keep switching branches on my local desktop setup 😄

Fruchtzwerg94:
See: #4470

dlampa:
How is one supposed to initialize a DataLogger with a CircularBuffer after this change? Is the CircularBuffer now obsolete?

swharden:
> How is one supposed to initialize a DataLogger with a CircularBuffer after this change? Is the CircularBuffer now obsolete?

This is a good question! I opened up a new issue #4507 to track discussion of this topic

Issue #4380, opened 2024-10-23
Cookbook: Collapse unnecessary chapters


swharden:
* These 3 chapters should probably be merged into a single one.
* A new page for "Plottable Management" should describe `Add()`, `Clear()`, and `GetPlottables<T>()`

---

![image](https://github.com/user-attachments/assets/3c3e1c86-a863-4fd6-b89d-b0752d291b3e)

PR #4491, opened 2024-11-11
Blazor Cookbook: move into Demo folder

swharden:
... and make a small code change to fix broken chapter filtering introduced when the website chapters were recently refactored #4380


jamsheed101:
Hello

Issue #4306, opened 2024-10-07
Legend text rtl support


moranmono:
**Question:** Hi, I have searched through the open and closed issues as well as discussions, but I couldn't find any mention of right-to-left (RTL) support for legend text. I understand that ScottPlot currently does not support RTL for legend text. Is this something that is planned?

**ScottPlot Version:** version 5.0.39

**Code Sample:** See http://scottplot.net/faq/repro/ for tips about creating reproducible code samples
```cs
private LegendItem CreateLegendItem(string legendText, Color color)
        {
            return new LegendItem()
            {
                LabelText = legendText,
                MarkerLineColor = color,
                LineColor = color,
                LineWidth = 3,
            };
        }
```

StendProg:
Hi @moranmono,

I am totally unfamiliar with RTL languages, please forgive my ignorance.
As far as I understand, Arabic is displayed correctly, but in reverse left to right form?
And you want this reversal to be done automatically?
A little research led me to the conclusion that it is not that difficult to do in SkiaSharp.

To do this we need to add a dependency of the main project on the `SkiaSharp.HarfBuzz` package.
After that, we need to replace all calls to` Canvas.DrawText(....)` with something like this 
```c#
using (var tf = SKFontManager.Default.MatchCharacter('א'))
using (var shaper = new SKShaper(tf))
canvas.DrawShapedText(shaper, ....)
```
There could still be issues with MeasureText methods, maybe they would require a special wrapper too, I haven't investigated here.

Anyway, this is a big enough architectural change that it may not be accepted, I just shared what I found.

moranmono:
Hi. 
Thank you for the quick response. Yes, in my case, the Hebrew letters appear OK, but the order is reversed. I am not looking for an automatic solution; I know it could lead to a major change and will take a lot of time. 
I already know the culture that my client is using; I have several clients in several different languages, so maybe you could add a flag of some sort, like this ex: 
```cs
sig1.LegendText = "אמת";
WpfPlot1.Plot.Legend.FontSize = 40;
WpfPlot1.Plot.Legend.IsRTL = true;
WpfPlot1.Plot.ShowLegend();
```

StendProg:
In fact, the solution turned out to be quite simple. The static field looks ridiculous, but it does what it's supposed to do. The problem is an additional dependency on `SkiaSharp.HarfBuzz`. `ScottPlot` really avoids adding dependencies.

[...]

PR #4500, opened 2024-11-12
RTL support

StendProg:
Adds support for Right to Left languages. #4306
This change affects the output of absolutely all texts, not just legends as suggested by #4306.

RTL support is disabled by default, and everything works as before, unchanged.
Setting the `LabelStyle.RTLSupport` static property results in additional wrapping of the output text in `Shaper`.
Actually this change may include a lot more than just RTL support, since there's a whole separate engine under the hood for rendering characters, but I'm not good at that.

I really don't like customizing the behavior with a static property, but I don't see any other easy ways to solve this problem. `LabelStyle` constructors are spread all over the project in large numbers.

Added a demo of RTL usage to be displayed in the legend. I'm sure you can fix/extend for better user experience.

```c#
LabelStyle.RTLSupport = true;
var sig1 = myPlot.Add.Signal(Generate.Sin(51));
sig1.LegendText = "אמת";
```

### DIFF
--- a/.github/workflows/pr1.yaml
+++ b/.github/workflows/pr1.yaml
@@ -21,8 +21,8 @@ jobs:
       - name: Build
         run: bash .github/workflows/scripts/check-build.sh
       - name: ScottPlot5 Unit Tests
-        run: dotnet test "src/ScottPlot5/ScottPlot5 Tests/Unit Tests/ScottPlot Unit Tests.csproj" --configuration Release --no-build --verbosity minimal
+        run: dotnet test "src/ScottPlot5/ScottPlot5 Tests/Unit Tests/ScottPlot Unit Tests.csproj" --configuration Release --verbosity minimal
       - name: ScottPlot5 Cookbook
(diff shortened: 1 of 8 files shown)


StendProg:
Perhaps the pipeline error is related to this flag:

dotnet test "src/ScottPlot5/ScottPlot5 Cookbook/ScottPlot Cookbook.csproj" --configuration Release ***--no-build*** --verbosity minimal

swharden:
I'll remove the `--no-build` flag and if that doesn't work I'll try adding HarfBuzz as a dependency in the Cookbook project. I think the act of installing it as a NuGet package in a project ensures the DLLs it requires get copied to the correct place. We'll see 🤞

swharden:
... nope, that didn't work. I'm now suspecting the build only fails on Linux systems.

swharden:
That worked! For others having a similar issue running an app that depends on SkiaSharp.HarfBuzz on a Linux system:

```
System.DllNotFoundException : Unable to load shared library 'libHarfBuzzSharp' or one of its dependencies. 
```

I solved it by adding the following to my csproj:
```xml
<PackageReference Include="HarfBuzzSharp.NativeAssets.Linux" Version="7.3.0.3" />
```

Issue #4504, opened 2024-11-13
Controls: axis limits should not "jump" mid-drag when lock keys are pressed


erik-burger:
**Suggestion:** Something I’ve noticed that I’m unsure if its expected behavior is how a plot reacts when the axis lock button is pressed/released when still having the mouse button pressed. For example, if I in a plot press down the mouse key and moves the plot and then press the X axis lock key then the plot moves to the location on the Y axis where the mouse initially was pressed.

Another example is if I initially in a plot press down both the mouse button and an Y axis lock key and then move my cursor bot in X and Y direction. If then the axis lock key is released, then the plot moves to the location where the plot would have been if I never pressed the axis lock. 

The point I’m trying to make is that the plot can move in maybe unexpected ways if the x or y axis lock is pressed and released while the mouse button is still pressed, and the cursor is moved.

My suggestion is that the plot should stay in its last position if the x or y axis lock is pressed or released leading to no unexpected movements of the plot. I will created a PR that fixes this, will link this issue in the pr

swharden:
Hi @erik-burger, thanks for this issue and PR #4505! I saw it when it came in but I'm taking a little time before I start working on it because the "jump" in the present behavior was actually intentional, but if it's surprising to most users then perhaps the strategy proposed here is a better default setting. Either way, I like the idea of implementing this and giving the user a flag to switch between the two behaviors. 

I'll spend a little time researching other similar charting libraries handle this behavior (matplotlib, matlab, and pyqtgraph come to mind) and try to make ScottPlot behave most consistently with the industry standard (if there is a consistent behavior).

If you get to it before me and want to place this behavior behind a `bool` to enable/disable it, that would be great! The only remaining decision is for me to decide whether to make it `true` or `false` by default.

Thanks again for this contribution!

swharden:
Hi @erik-burger,

One of the most commonly used plotting libraries in scientific computing spaces is Python's matplotlib. I used the following code to investigate their behavior. Left-click-drag pans, and pressing `x` locks the horizontal axis and pressing `y` locks the vertical axis. In both cases a "jump" is observed (consistent with ScottPlot's behavior) so I'm starting to think that this is the typical/expected behavior 🤔 

```py
import matplotlib.pyplot as plt
plt.plot([1, 2, 3, 4], [1, 4, 9, 16])
plt.show()
```

![python-lock](https://github.com/user-attachments/assets/c34fc4cc-001c-47ae-a78c-81a40205f357)

[...]

PR #4505, opened 2024-11-13
Controls: do not change axis limits mid-drag when lock keys are pressed

erik-burger:
…or MouseDragPan

This pull request removes the move a plot can do I the x or y axis lock is pressed while moving the view with the cursor. Instead, the plot moves from it’s last position if the axis lock key is pressed.

https://github.com/ScottPlot/ScottPlot/issues/4504


swharden:
https://github.com/ScottPlot/ScottPlot/issues/4504#issuecomment-2480684117

PR #4519, opened 2024-11-15
OpenGL: update version line to fix bug causing lines not to be rendered

onur-akaydin:
ScatterGLCustom is not drawing any line due to a small mistake in LinesProgramCustom.cs.

I'm not an expert in OpenGL, but when we write:
"#version 430 core"

instead of:
"# version 430 core"

ScatterGLCustom starts drawing the lines.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@ _Not yet on NuGet..._
 * Controls: Updated mouse wheel scroll fractions so zoom-out wheel events more accurately reverse zoom-in wheel events (#4516) @quantfreedom
 * Candlestick: Updated `FallingColor` property to change both the fill and line colors with one assignment (#4521) @czastack
 * Interactivity: Mouse click times in double-click events no longer resets after losing focus, improving double-click behavior in WPF controls (#4524) @onur-akaydin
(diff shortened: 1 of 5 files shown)


onur-akaydin:
I just noticed that the same mistake exists also in MarkerFillCircleProgram.cs file. Shall I create a new pull request for it?

swharden:
Hi @onur-akaydin, you can make all the changes in this PR.

@StendProg is more familiar with the OpenGL controls so I'll invite their comment if they're free to give feedback on this PR.

Thanks to both of you for your help on this!

StendProg:
Hey @onur-akaydin, @swharden

> I'm not an expert in OpenGL, but when we write:

I'm not an expert either, this project is my only experience.
But you are probably right, you should specify `#version`  
for example according to this http://www.c-jump.com/bcc/common/Talk3/OpenGL/Wk03_shader_intro/W03_0090_glsl_version_number.htm

I have tested, and both work correctly on my machine `# version` and `#version`

Apparently there are different versions of OpenGl drivers, some of them forgive such inaccuracies as I have. Some of them require a more strict specification, like yours.
Let's just fix it as you suggest, it shouldn't be worse.

swharden:
Thanks for your insightful feedback! I'll merge this and push it in the next release later today 👍

Issue #4147, opened 2024-08-12
When click legend name can invert the visibility of curve.


ZhuYajunFly:
When click legend name can invert the visibility of curve.
If the curve is hide, make the legend name's text color look more gray, or add a line in the middle on legend name's text.
When a chart contains multiple curves, comparing the differences between individual curves can be particularly useful.

ris-work:
Yes, I would like that as I do have quite a lot of things to display. I think colours repeat every 9-10 entries and I had to randomize the LinePattern or the MarkerStyle. It would be great if I could interactively choose/exclude certain entries.

swharden:
Hi @ZhuYajunFly and @ris-work, thanks for this question! It seems @nilsakesson and @BambOoxX from #3842 may like to see this feature as well. I'm not presently working on this, but if someone wants to implement it I put some notes here to get started: https://github.com/ScottPlot/ScottPlot/issues/3842#issuecomment-2458477860

Issue #3842, opened 2024-05-17
Legend: demonstrate how to toggle visibility when items are clicked


nilsakesson:
**Suggestion:** (Add functionality to show/hide a plottable with clicking the corresponding legend)

There are similar closed issues to this but there is nothing in SP 5.

Feature: click legend items to toggle visibility #1326
Legend: how to toggle visibility for each plottable #1482
Preferred way to plot numerous SignalPlot #1573
Legend: detach and toggle visibility of plottables (WinForms) #1589

In SP4 winforms there was a possibility to detach the legend from the plot and get the functionality to toggle the visibility of each plottable. It's best described in #1589.
I'm looking for a feature where you can click the legend within the plot to toggle it. 

There is a way to do this in matlab showed below
![ClickableLegend](https://github.com/ScottPlot/ScottPlot/assets/87042707/e081b3b1-2453-4a99-8a19-0e5a30c81227)

BambOoxX:
@nilsakesson Yes, it was a super nice addition, I thought it was already ported in SP5 but I'm missing it there too ! 
@swharden  If you want some help, I might have a bit of time there.

swharden:
I took a look at this tonight with the goal of implementing a solution, but called it off because it will take more time than I'm interested in investing at this time.

I'd be happy to review a pull request if someone decides to give this a go though!

I think the best implementation strategy would be as follows:

Edit somewhere around here:
https://github.com/ScottPlot/ScottPlot/blob/f84ee27e0956cf3dc17f8b8bab807faadcc25a57/src/ScottPlot5/ScottPlot5/Primitives/Legend.cs#L272-L280

...and save a list of `PixelRect` objects, one per rendered legend item.

Then add a method to this class like `GetItemIndexUnderMouse()` so users can add code to their OnClick events like:

```cs
int legendItemIndex = formsPlot1.Plot.Legend.GetItemIndexUnderMouse(pixel);
```

Perhaps return -1 if no item is under the mouse?

Yeah, this doesn't return the _plottable_ under the mouse, but many legend items are not tied to plottables and manual legends exist too, so I'm thinking this is the closest we're going to get without modifying a lot of code to add a nullable plottable to the `LegendItem` class then editing every plottable to tie itself to all the legend items it creates

BambOoxX:
@swharden I'll try and take a shot at this. One question though, is there any obvious link between the legend item and the plottable when a legend is built this way ? 
[...]

PR #4523, opened 2024-11-16
LegendItem: add `Parent` property to pair items with plottables

BambOoxX:
Add legend viewer functionnality to `ScottPlot.WinForms` as in SP4. See #3842 and #4147

@swharden there are some points I'd need help with :

- I could not reproduce the "graying" effect that was present in SP4 to show that a plot is hidden after clicking on the `LegendItem`. There is probably some additional SK painting to perform, but I don't really know how to do this properly
- There is a strange behavior when trying to change marker sizes, the behavior feels unnatural (marker size vs marker line width) 

There are probably other things to improve. Let me know.


swharden:
Hey @BambOoxX, thanks so much for your work on this!

I know we can work toward implementing a detachable legend viewer. However, my recommendation is to do this in two pull requests to keep things simple and moving fast.

Can we adapt this PR to not modify anything in the WinForms namespace? Adding that pop-out control is something we should probably put in the demo app (not ship with the ScottPlot package) and we can discuss that in a separate PR.

Similarly, can we remove `IsDetached` and `IsToggled` from `Legend`? I'd like to do whatever we can to keep the core ScottPlot project GUI-free (this property doesn't make sense for console apps), so perhaps we can solve that problem later in that second pull request. I'm thinking a `Dictionary<Legenditem, bool>` or maybe a custom class could keep track of GUI-related modifications to legend items, but again, more discussion for that second PR...

If this PR becomes GUI-free, I can merge it today and it will definitely be released in the next release!

src/ScottPlot5/ScottPlot5/Primitives/Legend.cs
```diff
@@ -123,7 +125,7 @@ public LegendItem[] GetItems()
         if (DisplayPlottableLegendItems)
         {
             var plottableLegendItems = Plot.PlottableList
-                        .Where(item => item.IsVisible)
+                        // .Where(item => item.IsVisible)
```
swharden, line 128: Is this intentional?

BambOoxX:
I'll see to modify the implementation per your observations tomorrow.

BambOoxX:
All changes you be limited to plottables right now. I reverted all changes related to the GUI.

swharden:
Thanks @BambOoxX! I'll refine this a bit, merge, then release a new package 🚀

swharden:
Sure thing! 

I don't have it meticulously planned out in my head, but is it something that could be added to the demo app without requiring code modification of the core library?

swharden:
Sounds good! No rush, but I'm here if you hit any snags.

I don't have it all figured out in my head - I'm just assuming it won't be too hard - but it will be interesting to see how this one plays out!

PR #4529, opened 2024-11-18
Fix: restore original PlotControl after closing new window

david3951445:
This PR addresses an issue in ScottPlot 5 related to the "Open in New Window" functionality in both WinForms and WPF.

- Issue:
  - In WPF, after closing the new window, the ContextMenu on the original window would no longer appear.
  - In WinForms, the ContextMenu was shown at the wrong position because it was being displayed on the new PlotControl instead of the original one.
- Cause: The issue occurred because, during the process of moving the Plot content to the new window, the original window’s PlotControl was overwritten but not properly restored when the new window was closed.
- Solution: After the new window is closed, the PlotControl on the original window is properly restored, ensuring the ContextMenu behaves as expected.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,6 @@
 ## ScottPlot 5.0.47
 _Not yet on NuGet..._
+* Controls: Fix issue preventing the context menu from appearing after it was used to open a new window (#4529) @david3951445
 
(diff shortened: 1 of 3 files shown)


swharden:
Thanks @david3951445, this looks great! I'm happy to see this bug fixed in the next release 🚀

Issue #4496, opened 2024-11-12
Signal: how to get the vertical range (not just point) near the cursor


NiRott:
In analogy to your function  AxisLine.IsUnderMouse in the example of draggable axis lines(cookbook 5.0.36), I tried to do the same with signal plots. The only problem is that I only find points and no lines. Is there is probably a better solution?

**ScottPlot Version:**  5.0.44

```cs
private int GetIndexOfSignalPlot(Pixel mousePixel)
{
    Coordinates coord = SP.Plot.GetCoordinates(mousePixel);

    double Mindistance = double.PositiveInfinity;
    int IndexNearest = -1;
    for (int i = 0; i < SignalPlots.Count(); i++)
    {
        var Index = SignalPlots[i].GetNearestX(coord, SP.Plot.LastRender, float.MaxValue);
        if (plotViewModel.ListView_Model.Count > i) plotViewModel.ListView_Model[i].ChannelValue = Index.Y.ToString();
        if (!SignalPlots[i].IsVisible) continue;
        DataPoint nearest = SignalPlots[i].GetNearest(coord, SP.Plot.LastRender);
        if (nearest.IsReal)
        {
            double distance = Math.Sqrt(Math.Pow(nearest.X - coord.X, 2) + Math.Pow(nearest.X - coord.X, 2));
            Log.WriteDebug(i.ToString(), distance.ToString());
            if (distance < Mindistance)
            {
                Mindistance = distance;
                IndexNearest = i;
            }
        }
    }
    return IndexNearest;
}
```

NiRott:
I don't think I've explained my problem properly. The function works, it finds the signal plot that is closest to the mouse coordinate if the sample points are close enough together. But if I click on a line segment whose endpoints are too far away, I would still have to calculate the course of the signal plot.

By the way, my GetLimitsY does not take any parameters

StendProg:
Hi @NiRott,

Are you trying to find hits in linear segments of the signal?
This is not an easy task and was rejected due to complexity.
If you are still serious about it, you can look up the developments in #1111 

https://github.com/ScottPlot/ScottPlot/pull/1111/files#diff-4eb2bc05f3892fceefba0a3997cc14528d5a42d290950c57e70446e884893a2c ). 
They are badly outdated, but there they implemented snap-based intersection checks with segments.

NiRott:
[...]

PR #4531, opened 2024-11-18
Add pixel-based hit detection and drag support for plottables

StendProg:
There was a request for a Dragable Demo loosely affecting the core library. #4496

1. Added the ability to drag any `Plottable` that has coordinate offset fields. To do this, wrap the Plottable in `DragablePlottableDecorator` and add it to the Plot.Plottables list.
2. Added the ability to check if a rectangular area falls into a Plottable for all Plottables except for Plottables outside of DataArea (this can be implemented if needed). If we need hit tests only without drag capabilities, we can wrap the Plottable in a `HitablePlottableDecorator`. It is the base for `DragablePlottableDecorator`.
3. Added Dragable Plottable Winforms Demo. My goal was to test all Plottables, I'm sure you will change/extend for better user experience.

A special map is used to hitCheck on Plottable. It is obtained by separate rendering of plottable in `Offscreen Bitmap` with transparent background. This rendering is done every time together with the usual Render() request from Plot.
This approach is very versatile and allows to search for hits for absolutely any Plottable. Yes it actually doubles the rendering time for each Plottable with drag capability. But due to the high initial speed of all Plottables, this drop is not really noticeable.

GLPlottables in theory should also work, but the time will be unacceptable, so it is not recommended.
Also, this approach will not work when a single Plottable is shared by multiple Plots.

In theory there could be more Plottables supported for Drag, but I only took those that already had OffsetX and OffsetY in some form.
If we develop this idea further, it would be nice to have an interface like `ICoordinateOffset`. Then it would be easier to automate the search for offsets. But this is a significant amount of changes that can be made at any time.

```cs
        var scatter = new Scatter(new ScatterSourceDoubleArray(xs, ys2));
        formsPlot1.Plot.Add.Plottable(new DragablePlottableDecorator(scatter));
```

https://github.com/user-attachments/assets/941ba203-145f-4fd2-a33e-736eedc52562

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,6 @@
 ## ScottPlot 5.0.47
 _Not yet on NuGet..._
+* Controls: Fix issue preventing the context menu from appearing after it was used to open a new window (#4529) @david3951445
 * Interactivity: Created `HitablePlottableDecorator` and `DragablePlottableDecorator` classes that wrap any `IPlottable` to add pixel-based mouse collision detection and drag capability to any plot type (#4531, #4496) @StendProg
 
 ## ScottPlot 5.0.46
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,6 @@
 ## ScottPlot 5.0.47
 _Not yet on NuGet..._
+* Interactivity: Created `HitablePlottableDecorator` and `DragablePlottableDecorator` classes that wrap any `IPlottable` to add pixel-based mouse collision detection and drag capability to any plot type (#4531, #4496) @StendProg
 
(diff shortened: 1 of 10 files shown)


swharden:
Thank you for this @StendProg! This has been one of my favorite pull requests 😄 It is a very clever solution, and I am excited to get this merged so it can be extended in the future!

I'm beginning to work on adding more interfaces to support plottables which may be dragged from multiple "handles" (e.g., arrows with two points that could be independently dragged by their base or their tip), so I am happy that the "theme" of this upcoming release will be focused on improving hit detection and drag support in multiple ways.

Thanks for your suggestions along the way, and as always thank you for your excellent work 🚀

Issue #4483, opened 2024-11-10
Improve finance demo


swharden:
This issue tracks short term goals to improve financial charting resulting from a discussion today with @quantfreedom and @VladislavPustovarov

* Progress will be evaluated using the [Sandbox.WinFormsFinance](https://github.com/ScottPlot/ScottPlot/tree/main/src/ScottPlot5/ScottPlot5%20Sandbox/Sandbox.WinFormsFinance) project

* This work extends #4385

## Task 1: Scott
- [x] Improve cursor tracking (show labels, snap to bars, etc.)
- [x] Improve performance for very large numbers of bars (50,000)
- [x] Simplify the process of automatic vertical scaling. ~~Use strategy pattern to allow multiple scaling strategies including custom ones injected by the user.~~ _It is already a function that can be created by the user and assigned to the `ContinuousAutoscaleAction` action so creating additional classes is not required._
- [x] Add something like "GetOhlcNearest(x)" to simplify mouse snapping
- [x] Improve tick behavior when windows are resized. Mimic TradingView.com where ticks "slide" off the screen to the left but do not change in density as the window is resized

## Task 2: Vlad
- [x] Add more views by creating classes in the [TickGenerators/Financial](https://github.com/ScottPlot/ScottPlot/tree/main/src/ScottPlot5/ScottPlot5/TickGenerators/Financial) namespace (suggested views are listed below)
- [x] Improve the logic used to decide which view is best given the figure size and axis limits

```
YearOnly - Year
YearAndJuly - Year, July
YearAnd2Months - Year, May, Sep
YearAnd3Months - Year, Apr, Jul, Oct
YearAnd4Months - Year, Mar, May, Jul, Sep, Nov
Months - Year, then each month
MonthAndThirdMonday - months are short names, days are the number
MonthAndEveryMonday - months are short names, days are the number
MonthAndEveryOtherDay - months are short names, days are the number
MonthAndEveryDay - months are short names, days are the number
DayAndEvery12Hours - show the day number for midnight, then show 6, 12, 18
DayAndEvery4Hours
DayAndEvery3Hours
EveryHour
EveryHalfHour
Every15Minutes
Every5Minutes
EveryMinute
Seconds45
Seconds30
Seconds15
Seconds10
Seconds5
Seconds1
```

## Task 3: Scott
_This work will be performed after Vlad finishes the date work above_
- [ ] Add lower limit to ensure ticks don't get smaller than candles
- [x] Add volume bars to the demo
[...]

PR #4551, opened 2024-11-23
FinancialTimeAxis: improve tick generation logic

VladislavPustovarov:
Addresses topics in #4483


swharden:
Hi @VladislavPustovarov (and @quantfreedom), thanks for this PR! I'm going to merge this now so it will be in todays release on NuGet, but I'll point out that there's room for improvement:

1) The "first tick" could be chosen using better logic so it doesn't "skip around" as the axis is panned. Good first ticks are things like "Jan", "The first of the month", "The first minute of each hour", etc. This can be fixed in a future pull request.

2) Some of the scales show mixed label formats that I don't think is intended. Here are some examples from the financial demo project in `src\ScottPlot5\ScottPlot5 Sandbox\Sandbox.WinFormsFinance`

image

image

image

image

swharden:
Note that I modified this PR to include a cookbook recipe. When I publish this package the cookbook recipe will be automatically converted to HTML and added to https://scottplot.net/cookbook/5.0/Finance/

PR #4564, opened 2024-11-25
Fix Generate.AddNoiseInPlace scaling, close #4563

### LINKED ISSUE #4563: Generate.AddNoise(): fix scaling error
`ScottPlot.Generate.AddNoise` is supposed to be "symmetric" (± magnitude), per docs comments

> Return a copy of the given data with random noise added (± magnitude)

But this quick experiment (PowerShell) shows that it is not

```powershell
# all zeros
[double[]] $a = [double[]]::new(20)

# expected [-1, 1), actual seem to be [-0.5, 1.5)
[ScottPlot.Generate]::AddNoise($a, 1.0)
```

The implementation (`AddNoiseInPlace`) per ILSpy is this:

```csharp
public void AddNoiseInPlace(double[] values, double magnitude)
{
    for (int i = 0; i < values.Length; i++)
    {
        double num = (2.0 * RandomNumber() - 0.5) * magnitude;
        values[i] += num;
    }
}
```

Is `- 0.5` a mistake? Should not it be `- 1.0`?

**The question**: Is this a bug or I misinterpret or miss something?

**ScottPlot Version:** 5.0.46

nightroman:
This PR fixes the tiny bug reported and confirmed in #4563


bclehmann (approved):
LGTM

PR #4592, opened 2024-12-04
Release disposable resources after render cycle

vadimart92:
Fixed an issue where SKPaint instances were not being disposed of properly in the RenderPack class. This oversight caused the instances to be disposed of in the destructor, potentially leading to Skia exceptions.


bclehmann (approved):
LGTM

swharden:
Thanks @vadimart92, this looks great!

Issue #4596, opened 2024-12-07
CoordinateRect.Contains doesn't work for 'inverted' objects


SprinterDave:
**Issue:**
The `CoordinateRect.Contains` class does not act logically for `CoordinateRect`s where `IsInvertedX` or `IsInvertedY` returns true.

**ScottPlot Version:**
5.0.47

**Code Sample:**
```cs
using System.Diagnostics;

namespace ScottPlotInvertedCoordinateRect
{
  internal class Program
  {
    static void Main(string[] args)
    {
      double value = 5;
      ScottPlot.Plot myPlot = new();
      ScottPlot.Bar myNegativeBar = new() { Value = -value };
      ScottPlot.Bar myPositiveBar = new() { Value = value };
      myPlot.Add.Bars(new[] { myNegativeBar, myPositiveBar });

      myPlot.SavePng("repro.png", 400, 300);
      myPlot.ZoomRectangle.VerticalSpan = true;

      // Should not assert
      Debug.Assert(myPositiveBar.Rect.Contains(0, value / 2),
        $"Rect {myPositiveBar.Rect} should contain the point (0, {value/2})");

      // Should not assert, but does
      Debug.Assert(myNegativeBar.Rect.Contains(0, -value / 2),
        $"Rect {myNegativeBar.Rect} should contain the point (0, {-value / 2})");
    }
  }
}
```

swharden:
Resolved in #4597

PR #4597, opened 2024-12-07
Fix Contains method for inverted rects

SprinterDave:
Fix for https://github.com/ScottPlot/ScottPlot/issues/4596


src/ScottPlot5/ScottPlot5 Tests/Unit Tests/UnitTests/PrimitiveTests/CoordinateRectTests.cs
```diff
@@ -38,4 +38,20 @@ public void Test_CoordinateRect_Expanded()
         CoordinateRect cRect2 = cRect.Expanded(coord);
         cRect2.Contains(coord).Should().BeTrue();
     }
+
+    [TestCase(0, 0, 1, 1, 0.5, 0.5, true, TestName = "Test Contains with no inverted")]
+    [TestCase(0, 0, 1, -1, 0.5, -0.5, true, TestName = "Test Contains with inverted X")]
+    [TestCase(0, 0, -1, 1, -0.5, 0.5, true, TestName = "Test Contains with inverted Y")]
+    [TestCase(0, 0, -1, -1, -0.5, -0.5, true, TestName = "Test Contains with inverted X and Y")]
```
bclehmann, line 45: These test cases are a little hard to parse, just because there's 6 numbers in there, I'd prefer to pass them in as a struct or something, just so that it's more clear to read.

I might suggest something like this:
```cs
private static readonly (CoordinateRect rect, Coordinates pt)[] ContainsTestCases = {
    (new(0, 1, 0, 1), new(0.5, 0.5)), // Uninverted Rect
    (new(0, 1, 0, -1), new(0.5, -0.5)), // inverted X
    (new(0, -1, 0, 1), new(-0.5, 0.5)), // inverted Y
    (new(0, -1, 0, -1), new(-0.5, -0.5)) // inverted X and Y
};

[Test, TestCaseSource(nameof(ContainsTestCases))]
public void Test_CoordinateRect_Contains((CoordinateRect rect, Coordinates pt) testCase)
{
    Assert.That(testCase.rect.Contains(testCase.pt), Is.True);
}
```

Note that it's equivalent, the order of the numbers is just different because the `CordinateRect` constructor takes `left, right, bot, top` and your code lists them in the order `left, bot, right, top`

bclehmann (approved):
Looks good, thanks for including tests.

I just had one comment to make the tests more readable, I'd rather pass the test params in as structs just because my eyes start to glaze over trying to keep 6 params in the right order.

src/ScottPlot5/ScottPlot5 Tests/Unit Tests/UnitTests/PrimitiveTests/CoordinateRectTests.cs
swharden: Thanks for this feedback Ben! While I do agree, I'm okay merging this as it is given the situation... I'm grateful when anonymous internet contributions with tests at all 🙃

swharden:
Thanks @SprinterDave! BTW I was really inspired when I participated in FIRST as a teen in the early 2000s and it's great to see others mentoring in that program today. 

Hope your 2025 is off to a great start,
Scott

PR #4686, opened 2025-01-10
Axis Rules: Improve support for boundaries on plots with inverted axes

kebox7:
PR #4609 (Axis Rules: limit rules do not function property on inverted axes) adds support for inner/outer bounds rules for inverted axes. But it has some peculiarities, when panning the chart to the specified boundary values, zoomming occurs. In my opinion, this is not very correct. This PR fixes this behavior.

`MinimumBoundary` вefore:

https://github.com/user-attachments/assets/c84070d2-a169-495b-94c4-4c7fd42f98ed

`MinimumBoundary` after:

https://github.com/user-attachments/assets/4eb02416-690b-4bde-9610-9b344e3af921

`MaximumBoundary` вefore:

https://github.com/user-attachments/assets/8803d07f-0e41-4836-8ae8-a649556084f7

`MaximumBoundary` after:

https://github.com/user-attachments/assets/0c0fed50-e548-413c-85bc-a8aa35a9e4da


swharden:
Thanks @kebox7, this looks great! 🚀

PR #4693, opened 2025-01-12
Added CleanupGL() method into ScatterGL class

onur-akaydin:
Although the InitializeGL method is intended to be called only once, I had to call it many times in a project. Especially in cases where Vertices were recalculated. InitializeGL() causes memory leaks when called again. To prevent this, I added the CleanupGL() method to cleanup OpenGL resources and call it from the beginning of InitializeGL().

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 ## ScottPlot 5.0.53
 _Not yet on NuGet..._
 * Axis Rules: Improved support for inner and boundaries on plots with inverted axis limits (#4686, #4609) @kebox7
+* OpenGL: Improved behavior for applications which have multiple OpenGL instances (#4693) @onur-akaydin
 
 ## ScottPlot 5.0.52
 _Published on [NuGet](https://www.nuget.org/profiles/ScottPlot) on 2025-01-08_
--- a/src/ScottPlot5/ScottPlot5 Controls/ScottPlot.OpenGL/Plottables/ScatterGL.cs
+++ b/src/ScottPlot5/ScottPlot5 Controls/ScottPlot.OpenGL/Plottables/ScatterGL.cs
@@ -61,11 +61,11 @@ public class ScatterGL : Scatter, IPlottableGL
     {
         if (GLHasBeenInitialized)
         {
-            // Delete vertex arrays and buffers
             GL.DeleteVertexArray(VertexArrayObject);
             GL.DeleteBuffer(VertexBufferObject);
 
-            // Dispose shader programs
+            LinesProgram?.GLFinish();
+            MarkerProgram?.GLFinish();
             LinesProgram?.Dispose();
             MarkerProgram?.Dispose();
 
--- a/src/ScottPlot5/ScottPlot5 Controls/ScottPlot.OpenGL/Plottables/ScatterGL.cs
+++ b/src/ScottPlot5/ScottPlot5 Controls/ScottPlot.OpenGL/Plottables/ScatterGL.cs
@@ -42,6 +42,7 @@ public class ScatterGL : Scatter, IPlottableGL
 
     protected virtual void InitializeGL()
     {
+        CleanupGL();
         LinesProgram = new LinesProgram();
         MarkerProgram = new MarkerFillCircleProgram();
 
@@ -54,6 +55,22 @@ public class ScatterGL : Scatter, IPlottableGL
         GL.EnableVertexAttribArray(0);
         Vertices = Array.Empty<double>();
         GLHasBeenInitialized = true;
+    }
+
+    protected virtual void CleanupGL()
+    {
+        if (GLHasBeenInitialized)
+        {
+            // Delete vertex arrays and buffers
+            GL.DeleteVertexArray(VertexArrayObject);
+            GL.DeleteBuffer(VertexBufferObject);
+
+            // Dispose shader programs
+            LinesProgram?.Dispose();
+            MarkerProgram?.Dispose();
+
+            GLHasBeenInitialized = false;
+        }
     }
 
     protected Matrix4d CalcTransform()


swharden:
Hi @onur-akaydin, thanks for this PR! I'm going to merge it now and publish a new package in the next few hours and this will be included. 

I'll ping @StendProg (the creator of the OpenGL controls) in case they wish to provide any feedback about this change.

StendProg:
Hi @onur-akaydin, @swharden thx for pinging.

It's not really clear to me what this change does. You write that you call `InitializeGL()` many times, but it is not clear how you do it for a `protected` method. If you created a derived class, you could override the logic and add the cleanup directly in it. It's also not clear how you modify vertices, the array of vertices is formed in the constructor.

In fact your solution in this PR changes absolutely nothing for `ScatterGL`, `InitializeGL` will be called only 1 time as it was.

If we think about extending the functionality, it would be nice to add `Dispose()` for `ScatterGL`, which cleans up resources in the way you suggested.
And for changing data, a method like `Update(double[] newX, double[] newY)` would be suitable, which would clean up resources and recreate arrays of vertices.

onur-akaydin:
Hi @StendProg,

Thank you for your feedbacks & you are completely right about your questions and concerns.

Yes, I have another UpdateableScatterGLCustom class derived from ScatterGLCustom. Even though InitializeGL() is virtual, I avoided overriding it because I needed the OpenGL methods like GL.BindBuffer() inside it, and I didn't want to reference OpenTK in my UpdateableScatterGLCustom project.

But from your perspective as a project maintainer, I understand that I made a premature pull request in order to act quickly, please excuse me.

In the meantime, yesterday I found a better way to update the Vertices and update GL accordingly (without calling InitializeGL over and over again). So here's what my UpdateableScatterGLCustom class looks like now:

`public class UpdateableScatterGLCustom : ScatterGLCustom
{
    public UpdateableScatterGLCustom(UpdateableScatterSourceDoubleArray data, IPlotControl control) : base(data, control)
    {
        UpdateableData.DataUpdated += UpdateableData_DataUpdated;
    }

    private void UpdateableData_DataUpdated(object? sender, EventArgs e)
    {
        UpdateVertices();
    }

    public UpdateableScatterSourceDoubleArray UpdateableData => Data as UpdateableScatterSourceDoubleArray;
   
    public virtual void UpdateVertices()
    {
        var dataPoints = Data.GetScatterPoints();
        double[] newVertices = new double[dataPoints.Count * 2];
        for (int i = 0; i < dataPoints.Count; i++)
        {
            newVertices[i * 2] = dataPoints[i].X;
            newVertices[i * 2 + 1] = dataPoints[i].Y;
        }

        Vertices = newVertices;
        //VerticesCount = Vertices.Length / 2;    // Cannot change VerticesCount because it is readonly

        if (GLHasBeenInitialized)
        {
            GL.BindBuffer(BufferTarget.ArrayBuffer, VertexBufferObject);
            GL.BufferData(BufferTarget.ArrayBuffer, Vertices.Length * sizeof(double),
                Vertices, BufferUsageHint.DynamicDraw); // Note: Changed to DynamicDraw
        }
    }
}
`

I hope I could explain my problem and purpose better. So, how do you suggest we fix the situation? I don't feel confident in Git, but if you'd like I can try to create a new PR with the following changes:

* Rollback from #4693
* Implement IDisposable for ScatterGL and call CleanupGL() inside Dispose(). CleanupGL might be improved.
* Change protected readonly int VerticesCount >> protected int VerticesCount
* Call Dispose() of IPlottables inside from Plot.Dispose()

Would you like me to make these and push a new PR?

StendProg:
Hi @onur-akaydin,

>Yes, I have another UpdateableScatterGLCustom class derived from ScatterGLCustom. Even though InitializeGL() is virtual, I avoided overriding it because I needed the OpenGL methods like GL.BindBuffer() inside it, and I didn't want to reference OpenTK in my UpdateableScatterGLCustom project.

I meant that you could do it that way, without having to make changes to the base `ScatterGL` class:
```c#
    protected override void InitializeGL()
    {
        CleanupGL();
        base.InitializeGL();
    }
```

> I hope I could explain my problem and purpose better. So, how do you suggest we fix the situation? I don't feel confident in Git, but if you'd like I can try to create a new PR with the following changes:
Rollback from https://github.com/ScottPlot/ScottPlot/pull/4693
Implement IDisposable for ScatterGL and call CleanupGL() inside Dispose(). CleanupGL might be improved.
Change protected readonly int VerticesCount >> protected int VerticesCount
Call Dispose() of IPlottables inside from Plot.Dispose()

If you have no objections to sharing your developments on `UpdatableScatterGL`, let's just add them directly to the `ScatterGL` base class. This doesn't conflict with its design, and the functionality to update points will be available out of the box for you and others.

>Would you like me to make these and push a new PR?

If you don't mind doing such work, I support it in every possible way, and I will try to make a review of your PR.

onur-akaydin:
Okay, I'll do the revisions regarding Dispose() now.

> If you have no objections to sharing your developments on UpdatableScatterGL, let's just add them directly to the ScatterGL base class. This doesn't conflict with its design, and the functionality to update points will be available out of the box for you and others.

I would like to do this. But I'm scared because it has a potential to touch too many places in the codebase and I don't feel so confident about it. For example it may require defining new UpdatableScatterSource variants: UpdatableScatterSourceDoubleArray, UpdatableScatterSourceCoordinatesArray, UpdatableScatterSourceGenericArray and so on. My own implementation works for me but I'm not sure how it would behave on a larger scale:

```csharp
public class UpdatableScatterSourceDoubleArray : IScatterSource, IDataSource, IGetNearest
{
    public event EventHandler DataUpdated;

    public UpdatableScatterSourceDoubleArray(double[] xs, double[] ys)
    {
        Xs = xs;
        Ys = ys;
        MaxRenderIndex = Math.Min(xs.Length, ys.Length) - 1;
    }

    public void Update(double[] newXs, double[] newYs, bool alwaysChangeArrayReference = false)
    {
        if (alwaysChangeArrayReference)
        {
            xs = newXs;
            ys = newYs;
        }
        else
        {
            // decides whether to create a new instance or updating values by Array.Copy:
            HelperMethods.UpdateArray(newXs, ref xs); 
            HelperMethods.UpdateArray(newYs, ref ys);
        }

        MaxRenderIndex = Math.Min(xs.Length, ys.Length) - 1;

        OnUpdate();
    }

    public virtual void OnUpdate()
    {
        DataUpdated?.Invoke(this, EventArgs.Empty);
    }

    private double[] xs;
    private double[] ys;
    private int maxRenderIndex;

    public double[] Xs { get => xs; protected set => xs = value; }
    public double[] Ys { get => ys; protected set => ys = value; }
    
   ........ The rest of the code is almost the same of the ScatterSourceDoubleArray .....
}
```

At a later time, I will create a separate PR for UpdatableScatterSource... variants so you can check it out.

Issue #4664, opened 2025-01-06
Inconsistency: KeyboardAutoscale is internal while other user action responses are public


manaruto:
I am using ScottPlot 5’s user action response system to enable or disable specific default interactions. Most user action responses (like SingleClickContextMenu, KeyboardPanAndZoom, etc.) are public, so I can remove them by type:

```
Chart.UserInputProcessor.RemoveAll<ScottPlot.Interactivity.UserActionResponses.SingleClickContextMenu>(); 
Chart.UserInputProcessor.RemoveAll<ScottPlot.Interactivity.UserActionResponses.KeyboardPanAndZoom>();
```

However, KeyboardAutoscale is declared internal in 

```
internal class KeyboardAutoscale(Key key) : KeyPressResponse(key, AutoScale)
{
    public static void AutoScale(Plot plot, Pixel pixel)
    {
        plot.Axes.AutoScale();
    }
}
```
```

// This line fails to compile because KeyboardAutoscale is internal
Chart.UserInputProcessor.RemoveAll<ScottPlot.Interactivity.UserActionResponses.KeyboardAutoscale>();
```

Instead, I must remove it using its string name:

```
Chart.UserInputProcessor.UserActionResponses.RemoveAll(
    x => x.GetType().Name == "KeyboardAutoscale");
```

This is inconsistent with other user action responses and makes it more awkward to selectively remove the KeyboardAutoscale behavior.

Suggestion:

    Make KeyboardAutoscale public like the other user action responses, 
    or
    Provide a built-in helper method in UserInputProcessor (similar to DoubleLeftClickBenchmark(bool)) 
    
ScottPlot 5.0.47

manaruto:
I've submitted a pull request to address this issue: [#4743](https://github.com/ScottPlot/ScottPlot/pull/4743). Please review it when you have time.
Thanks!

swharden:
#4743

PR #4743, opened 2025-01-27
Make KeyboardAutoscale.cs public to allow removal from UserInputProcessor

manaruto:
This pull request updates the KeyboardAutoscale class  by changing its access modifier from internal to public. This change ensures consistency with other user action response classes and allows users to remove the KeyboardAutoscale action using the UserInputProcessor.RemoveAll<T> method.

[#4664](https://github.com/ScottPlot/ScottPlot/issues/4664)

### DIFF
--- a/src/ScottPlot5/ScottPlot5/Interactivity/UserActionResponses/KeyboardAutoscale.cs
+++ b/src/ScottPlot5/ScottPlot5/Interactivity/UserActionResponses/KeyboardAutoscale.cs
@@ -1,6 +1,6 @@
 ﻿namespace ScottPlot.Interactivity.UserActionResponses;
 
-internal class KeyboardAutoscale(Key key) : KeyPressResponse(key, AutoScale)
+public class KeyboardAutoscale(Key key) : KeyPressResponse(key, AutoScale)
 {
     public static void AutoScale(IPlotControl plotControl, Pixel pixel)
     {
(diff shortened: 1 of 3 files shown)


src/ScottPlot5/ScottPlot5/Interactivity/UserActionResponses/KeyboardAutoscale.cs
```diff
@@ -1,6 +1,6 @@
 ﻿namespace ScottPlot.Interactivity.UserActionResponses;
 
-internal class KeyboardAutoscale(Key key) : KeyPressResponse(key, AutoScale)
+public class KeyboardAutoscale(Key key) : KeyPressResponse(key, AutoScale)
```
bwedding, line 3: I started to do this and saw yours. I'd rather make it public sealed than just public.

swharden: Hi @bwedding, thanks for this suggestion! I modified this PR to include tests which can ensure that all user action responses have the same access modifiers. Can you clarify why you make this suggestion? I understand the function of sealed classes, but it's not obvious why it may be desirable in this case. Depending on your response we can modify the test as needed to ensure all all sealed if that ends up being a goal. Thanks for the clarification!

Issue #4746, opened 2025-01-27
Use List<double> as datasource for SignalXY?


JoeStoneAT:
**Problem:**
For a visualisation of data from a DAQ-system I have my data in some classes doing the daq, calculation, datahandling, saving,... There my data are in `List<double>` as this is for me much easier than using `double[]` and increasing and copying the array from time to time. I had it this way, before I found the possibility to use `List<double>` in Scottplot5.
The lists for X and in the case of not equidistant signal sources timestamp Ys are growing by time. Existing data are not changed.
 
And then I'm using a SignalPlot for displaying
`thePlot = plot.Add.Signal(ys:dHandler.data[fqVarName].valList, period: ((clsDataPeriodic)dHandler.data[fqVarName]).dt);`

Now I wanted to extend my possibilities with SignalXY  for non-equidistant x-values but there as datasource only `double[]` or `T[]` is possible.
(SignalXYSourceDoubleArray and SignalXYSourceGenericArray)
I would need a
`plot.Add.SignalXY(ys: dHandler.data[fqVarName].valList,  xs: ((clsDataUnevenly)dHandler.data[fqVarName]).timestampList);` to use my existing Lists of doubles.

**Question:** 
Is there a way to use "external" `List<double>` as the x and y data of SignalXY? The effort to rework everything to `double[]` would be pretty high.

_Possible Alternatives?_
Doing a `List<double>.toArray()` 3-4 times a second before a plot refresh should work, but this will most probably bring me into performance issues I think. Depending of the measurement time and sample rate I have from 0 to some 100 Mio doubles in the `List<double>` per signal.

1. `mySignalXYPlot.Data = new SignalXYSourceDoubleArray(valueList.toArray(),timestampList.toArray())`
2. `WpfPlot.Refresh()`

DataLogger?
In some other small programs I use DataLogger. Would it work to store the public `List<Coordinates>` is my dataHandler? Until now I used only the Add() method of the data handler.

**ScottPlot Version:** 5.0.54

best regards
Josef

swharden:
Hi @JoeStoneAT, thanks for this question!

I suspect the missing support for `List<T>` was just because it was not implemented, not because it can't be done.

I'm focusing my personal efforts on some higher priority open issues, but if you'd be willing to give this a go I'd be happy to help answer questions you have along the way, or ideally merge in a pull request so this feature can be added in the next release!

[...]

PR #4753, opened 2025-01-28
SignalXY: Improve support for generic lists

JoeStoneAT:
Hello @swharden 
btw it's my first PR in total :-)

is this, what you had in mind with #4746 
I did not test it in my application but in the WinForms Demo. There it's working

In the Cookbook I wanted to refresh with modified data, but did not find a way to call the Refresh-Method

best regards
Josef

### DIFF
--- a/src/ScottPlot5/ScottPlot5 Cookbook/Recipes/PlotTypes/SignalXY.cs
+++ b/src/ScottPlot5/ScottPlot5 Cookbook/Recipes/PlotTypes/SignalXY.cs
@@ -37,8 +37,8 @@ public class SignalXY : ICategory
 
     public class SignalXYGeneric : RecipeBase
     {
-        public override string Name => "SignalXY Generic";
-        public override string Description => "SignalXY plots support generic data types, " +
+        public override string Name => "SignalXY Generic Array";
+        public override string Description => "SignalXY plots support arrays with generic data types, " +
             "although double is typically the most performant.";
 
         [Test]
@@ -52,6 +52,8 @@ public class SignalXY : ICategory
                 xList.AddRange(Generate.Consecutive(1000, first: 2000 * i).Select(x => (int)x));
                 yList.AddRange(Generate.RandomSample(1000).Select(x => (float)x));
             }
+
+            // this example converts
             int[] xs = xList.ToArray();
             float[] ys = yList.ToArray();
 
@@ -63,25 +65,21 @@ public class SignalXY : ICategory
     public class SignalXYGenericList : RecipeBase
     {
         public override string Name => "SignalXY Generic List";
-        public override string Description => "SignalXY plots are a high performance plot type " +
-                                              "for X/Y data in List<Tx,Ty> where the X values are always ascending.";
+        public override string Description => "SignalXY plots support generic lists.";
 
         [Test]
         public override void Execute()
         {
-            // generate sample data with gaps
-            List<double> xList = new();
-            List<double> yList = new();
-            xList.AddRange(Generate.Consecutive());
-            yList.AddRange(Generate.Sin(51));
+            // create X and Y list with initial data
+            List<double> xs = new(Generate.Consecutive(count: 10, first: 0));
+            List<double> ys = new(Generate.RandomSample(10));
 
-            // add a SignalXY plot
-            var plt = myPlot.Add.SignalXY(xList, yList);
-            
-            // Extend Data and refresh
-            //xList.AddRange(Generate.Consecutive());
-            //yList.AddRange(Generate.RandomWalk(51));
-            //???.Refresh
+            // create a signal plot from our X and Y lists
+            myPlot.Add.SignalXY(xs, ys);
+
+            // data can be added to lists later
+            xs.AddRange(Generate.Consecutive(count: 10, first: 50));
+            ys.AddRange(Generate.RandomSample(10));
         }
     } 
     
--- a/src/ScottPlot5/ScottPlot5 Cookbook/Recipes/PlotTypes/SignalXY.cs
+++ b/src/ScottPlot5/ScottPlot5 Cookbook/Recipes/PlotTypes/SignalXY.cs
@@ -59,7 +59,32 @@ public class SignalXY : ICategory
             myPlot.Add.SignalXY(xs, ys);
         }
     }
+    
+    public class SignalXYGenericList : RecipeBase
+    {
+        public override string Name => "SignalXY Generic List";
+        public override string Description => "SignalXY plots are a high performance plot type " +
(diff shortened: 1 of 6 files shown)


src/ScottPlot5/ScottPlot5 Cookbook/Recipes/PlotTypes/SignalXY.cs
```diff
@@ -59,7 +59,32 @@ public override void Execute()
...
+            List<double> xList = new();
+            List<double> yList = new();
+            xList.AddRange(Generate.Consecutive());
+            yList.AddRange(Generate.Sin(51));
+
+            // add a SignalXY plot
+            var plt = myPlot.Add.SignalXY(xList, yList);
+            
+            // Extend Data and refresh
+            //xList.AddRange(Generate.Consecutive());
```
bclehmann, line 82: The required logic to extend the data is like this:

```cs
xs.Add(x);
ys.Add(y);
plt.Data.MaximumIndex++;
WpfPlot1.Refresh();
WpfPlot1.Plot.Axes.AutoScale();
```

I think everything except the increment to `MaximumIndex` field is expected by the user. It might be worth having a `Update()` function, since the `MaximumIndex` field is probably not well known. The other option would be to have `MaximumIndex` become nullable, and it's ignored unless it's specifically set by the user.

We could implement this something like this:

```cs
private int? _maximumIndex = null;
public int MaximumIndex {
  get => _maximumIndex.GetValueOrDefault(Count - 1);
  set {
    _maximumIndex = value;
  }
}
```

bclehmann (commented):
This looks good, I have just a couple thoughts on performance and API design.

bclehmann:
This PR does bring me some concerns just because of how complicated the `ISignalXYSource` implementations have become, and since their nature requires them to be duplicated a fair bit. I don't think that's a reasonable thing to expect you to solve since it's a lot bigger than the original feature, but it does deserve some thought.

I think the `IDataSource` approach was a very good one for allowing plottables to take very different types of input, but it looks like for `SignalXY` it's moved a lot of complexity from the plottable (which is disturbingly simple, especially compared to its previous iteration in 4.1) into the sources. Arguably that's where the complexity belongs for this sort of plottable, since so much of the complexity is in handling different inputs, but I don't like that the complexity is what gets duplicated.

I'd like if this could be simplified or consolidated, but I'm unsure how. Maybe a base-class, or maybe `SignalXYSourceGenericArray` and `SignalXYSourceDoubleArray` get implemented in terms of `SignalXYSourceGenericList`? That would require a wrapper from `T[]` and `double[]` to `IReadOnlyList<T>`, but I don't think that would be too bad. My understanding is the only reason `T[]` doesn't implement `IReadOnlyList<T>` is some accident of history not an implementation issue, similar to how `T[]` is an `IList` but not an `IList<T>`.

Either of those options would have performance costs due to added indirection, but I'm not convinced they would be relevant or even measurable. Especially since pulling everything through an `IDataSource` hasn't had noticeable performance issues.

This does read a lot like my original idea, which was for every plottable to take an `IReadOnlyList<T>` or `IReadOnlyList<double>`, which I think still has merit, it's just not very flexible for the user, who expects to be able to throw in a `double[]` and have it Just Work™. I think implementing most/all of the data sources on an `IReadOnlyList<T>` but maintaining that simplicity for the user gives the best of both worlds, it's easy for us to reason about and maintain, but it's also easy for users to just ignore all that lies beneath the surface.

swharden:
Hi all! First off, Joe, thanks for this PR! As your first PR, this one is really complicated and tackles a big issue, and you did great work! Congratulations on your first successful pull request 😄 🎉 🚀 

---

Much of the discussion below is advanced, and since the discussion started here I'll keep it going here. My inclination is to merge this to get the feature working, then refactor the system toward simplicity and reduced code duplication in the future.

> I don't like that the complexity is what gets duplicated

> I'd like if this could be simplified or consolidated, but I'm unsure how

Haha - this is my sentiment too! A perfect solution hasn't presented itself, but I think there is room for improvement. I designed it this way to get ScottPlot 5 up and running quickly, and now that it is mature I like the idea of exploring improvements to this design. I'll merge this PR, and we can refactor data sources in subsequent PRs.

> My understanding is ... some accident of history not an implementation issue

You're not wrong 🙃 

> I think implementing most/all of the data sources on an IReadOnlyList<T> but maintaining that simplicity for the user gives the best of both worlds

My thought (consistent with this being a historical accident) is that I got ScottPlot 5 off the ground by implementing "double only" plot types, then after the main plot types were there I bolted on generics. Since we have implementations for both generics and non-generics, we should do some performance testing to evaluate how much slower the generic forms are. I think they are potentially slower because they call `ToDouble()` a lot, which may be why I favored keeping them separate. Regarding keeping everything readonly list vs. array, same thing, if performance says they're equivocal we can drop the concrete type implementations and cut our duplication and maintenance burden with one clean sweep.

I'm going to focus on clearing out the open PRs/issues over the next few weeks, but if someone is inspired to dive into refactor land and benchmark land, I'd love to hear what they discover!

src/ScottPlot5/ScottPlot5/DataSources/SignalXYSourceGenericList.cs
```diff
@@ -0,0 +1,476 @@
...
+        return SearchIndex(x, range);
+    }
+
+    /// <summary>
+    /// Search the index associated with the given X position limited to the given range
+    /// </summary>
+    private (int SearchedPosition, int LimitedIndex) SearchIndex(double x, IndexRange indexRange)
+    {
+        NumericConversion.DoubleToGeneric((x - XOffset) / XScale, out Tx x2);
+        int index = Array.BinarySearch(Xs.ToArray(), indexRange.Min, indexRange.Length, x2);
```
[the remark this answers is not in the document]
swharden: Hey @bclehmann, thanks for pointing out this potential performance issue! Consistent with my comment below, I'm inclined to merge this now to get the feature up and refactor toward performance (and reduced code duplication) if/as needed, and only in concert with actual benchmark testing. 

That's an appreciable amount of work so I don't intend to dive into it right now, but I'll add a code comment to note it's a potential site for performance optimization and someone can revisit this if they're interested in improving performance of this data source in the future 👍

src/ScottPlot5/ScottPlot5 Cookbook/Recipes/PlotTypes/SignalXY.cs
swharden: Oof, it seems like there could be a better design where maximum index is optional, nullable, initialized to infinity, or something so the user doesn't have to do this. I'll try to put something in now. Thanks for pointing this out!

Issue #4735, opened 2025-01-25
MinimumSpan rule does not handle inverted axes correctly


manaruto:
```
signal.Data.Rotated = true;
myPlot.Axes.SetLimitsY(1000, 0);
```

The MinimumSpan rule does not work as expected because the Span property in the CoordinateRangeMutable class returns a negative value in the inverted state.

### Expected Behavior
The span should be handled correctly regardless of axis inversion.

### Current Behavior
The MinimumSpan rule  in the MinimumSpan class uses:

```
if (YAxis.Range.Span < YSpan)
{
    double yMin = YAxis.Range.Center - YSpan / 2;
    double yMax = YAxis.Range.Center + YSpan / 2;
    YAxis.Range.Set(yMin, yMax);
}
```

This fails when the axis is inverted, it overrides the axis limits, the user can not zoom in or out.

### Proposed Fix

Modify the code in MinimumSpan class as follows to handle the inverted axis correctly:

```
if (YAxis.IsInverted())
{
    if (Math.Abs(YAxis.Range.Span) < YSpan)
    {
        double yMin = YAxis.Range.Center - YSpan / 2;
        double yMax = YAxis.Range.Center + YSpan / 2;
        YAxis.Range.Set(yMax, yMin);
    }
}
else
{
    if (YAxis.Range.Span < YSpan)
    {
        double yMin = YAxis.Range.Center - YSpan / 2;
        double yMax = YAxis.Range.Center + YSpan / 2;
        YAxis.Range.Set(yMin, yMax);
    }
}
```

However, I am not sure if this is the best approach, or if it would be better to address the issue by modifying the Span property in the CoordinateRangeMutable class.

### Additional Considerations

The same issue applies to both X and Y axes, meaning fixes should be applied consistently to both dimensions.
The issue also affects the MaximumSpan rule, which should be adjusted similarly.

**ScottPlot Version: 5.053

swharden:
Hi @manaruto, I think this may have been corrected in the most recent version (#4686, #4609) 

If the issue is present in the current version and your fix corrects the issue, can you create a pull request so I can merge in your code change? It looks like the proposed fix you offer will go here:

https://github.com/ScottPlot/ScottPlot/blob/fff3fca297c0e05c85e5dc19033504a3454605cf/src/ScottPlot5/ScottPlot5/AxisRules/MinimumSpan.cs#L11-L27

Thanks!
Scott

---

[...]

PR #4755, opened 2025-01-29
Fix inverted axis min max span

manaruto:
This PR fixes the behavior of MinimumSpan and MaximumSpan rules when an axis is inverted (Min > Max). Previously, these rules did not correctly account for inverted axes, leading to incorrect enforcement of span constraints.

➜ If MinimumSpan or MaximumSpan is used alongside a minimum or maximum boundary, the constraints were not being applied correctly.
    
➜ If MinimumSpan or MaximumSpan is used alone (without min/max boundary constraints), the axis ticks would not be displayed as inverted.

[#4735](https://github.com/ScottPlot/ScottPlot/issues/4735)


swharden:
Hi @manaruto, thanks so much for this fix! It took me longer than typical to get to this one, but I'm happy to merge it in now and it will be in the next release 🚀

Issue #4629, opened 2024-12-24
Pie chart legend size differences between 5.0.43 and latest, legend is getting cut off


aespitia:
**Issue:** There seems to be a difference in the way legend size was calculated between 5.0.43 and 5.0.44 (onward) for pie charts at least.  Some of the legend is getting cut off, where it used to not be cut off.  Previously, the pie chart would shrink in size to accomodate all of the legend text, now the pie chart is bigger, but losing part of the legend.  Screen shots below to show the difference, the sample code is from LinqPad
![5 0 47-sample](https://github.com/user-attachments/assets/8c2f9219-08b0-40e7-982f-e8cd7c828c2c)
![5 0 40-sample](https://github.com/user-attachments/assets/e1e7030e-a05d-4f31-8721-60b5135e1c1b)

**ScottPlot Version:**  5.0.47

```cs
ScottPlot.Plot myPlot = new();

List<PieSlice> slices =
[
			new PieSlice() { Value = 32, Label = "Category 1", LegendText = "Category 1: 32%"},	
			new PieSlice() { Value = 27, Label = "Category 2", LegendText = "Category 2: 27%"},	
			new PieSlice() { Value = 16, Label = "Category 3", LegendText = "Category 3: 16%"},	
			new PieSlice() { Value = 15, Label = "Category 4", LegendText = "Category 4: 15%"},	
			new PieSlice() { Value = 4, Label = "Category 5", LegendText = "Category 5: 4%"},	
			new PieSlice() { Value = 3, Label = "Category 6", LegendText = "Category 6: 3%"},	
			new PieSlice() { Value = 1, Label = "Category 7", LegendText = "Category 7: 1%"},	
			new PieSlice() { Value = 1, Label = "Category 8", LegendText = "Category 8: 1%"},	
			new PieSlice() { Value = 100, Label = "Category 9", LegendText = "Category 9: 100%"},	
			
];

var pie = myPlot.Add.Pie(slices);

foreach (var s in pie.Slices)
{
	s.LabelFontSize = 0;
}

pie.LineStyle.Width =1;
pie.LineStyle.Color = Color.FromHex("#000000");
myPlot.ScaleFactor=1;
pie.SliceLabelDistance = 1.2;
pie.ExplodeFraction = 0;
myPlot.Legend.Orientation = Orientation.Vertical;
myPlot.Legend.FontSize = Convert.ToSingle(10.667);
myPlot.Legend.Alignment = Alignment.UpperLeft;
[...]

PR #4790, opened 2025-02-14
AxisLimitsChanged calculation compares proper axes

### LINKED ISSUE #4783: Avalonia: Lower performance from 5.0.53 to 5.0.54
Using Avalonia with ScottPlot on Windows and Mac, even though plot benchmark reports similar rendering times, overall app responsive is visibly degraded on 5.0.54. Issue is resolved by downgrading to 5.0.53. This happens both on Windows and Mac with default Avalonia skia renderer.

Tested with rendering 32 real-time channel using Signal (with fast/regular signal source), DataStreamer on high dpi displays. Size of app from 1/4 of screen to full size does not seem to be a factor

bclehmann:
This compares the last render axes to the render before (rather than to itself). This is also robust to inverted axes, with the caveat that comparing [1, -1] to [-1, 1] will still report that the axes have changed (which is probably correct?)

Resolves #4783, see my comments in that issue for more info.

It also appears that my fix did not reintroduce the problem in #4629

## Remaining concerns

I'm not thrilled to learn that the secondary axes are not properly initialized until mouse pan, especially that they're initialized to the most edge-case-causing possible value of [infinity, -infinity].

I'd also like to fix `CoordinateRangeMutable.Min` to report the actual minimum value, not just the value that's passed in as its first parameter. This would match `CoordinateRange`'s behaviour, but I don't look forward to wading through its downstream consumers to see if this change would break them.


bclehmann:
The autoformat thing is listing files which I didn't touch, so I suspect this might be from a new `dotnet format `version?

swharden:
> I suspect this might be from a new dotnet format version?

Oof, it's worse... new version problem only on Linux but not Windows 🤪

swharden:
Sorry for the confusion... there _are_ autoformat issues in the main branch, but they're distinct from the ones reported in this PR. what a mess 😅

swharden:
> I'd also like to fix CoordinateRangeMutable

... I'll note that this has been on my radar for a while. Both CoordinateRange and CoordinateRangeMutable could both be updated to use more clearly defined (perhaps even verbose behavior), ideally leveraging the type system instead of magic floating point numbers.

Let's leave things as they are now because subtle changes in interactive behavior are very difficult to detect, and I'll work toward refactoring this system when I have a bigger block of time to devote 100% of my effort into doing it right.

Thanks again for this PR!

PR #4794, opened 2025-02-15
SignalXYSourceGenericList: Reduce allocations and copying to improve performance

bclehmann:
This requires a binary search implementation for IReadOnlyList. I tested this by calling both Array.BinarySearch on Xs.ToArray() and my implementation on every received input and throwing if they didn't match.

The performance is much improved, the previous implementation dropped below 60fps for relatively modest input lengths (much past 10000 points) At this point `System.Linq.Enumerable.ToArray<T>` takes up ~75% of the CPU time in the profiler. On the new one I left it running by accident and it's still going strong at 1 million+ points, even on a debug build.

Discussed in my review comment on #4753

I tested it with this Avalonia app:

```cs
using Avalonia.Controls;
using ScottPlot.Avalonia;
using ScottPlot;
using System.Collections.Generic;
using Avalonia.Threading;
using System.Linq;
using System;

namespace Sandbox.Avalonia;

public partial class MainView : UserControl
{
    readonly ScottPlot.DataGenerators.RandomWalker Walker1 = new(0);

    readonly DispatcherTimer updateTimer = new();

    public MainView()
    {
        InitializeComponent();

        AvaPlot.UserInputProcessor.IsEnabled = true;

        List<double> xList = new();
        List<double> yList = new();

        var plt = AvaPlot.Plot.Add.SignalXY(xList, yList);

        updateTimer.Interval = TimeSpan.FromMilliseconds(200);
        updateTimer.Start();
        updateTimer.Tick += (s, e) =>
        {
            int count = 50;

            // add new sample data
            xList.AddRange(Generate.Consecutive(count, 1, xList.LastOrDefault(0) + 1));
            yList.AddRange(Walker1.Next(count));
            AvaPlot.Plot.Axes.AutoScale();
            AvaPlot.Refresh();
        };
    }
}

```

PR #4795, opened 2025-02-16
CoordinateRangeMutable: Make Min, Max inverted-aware

bclehmann:
This was brought up in #4794 

## Process

I renamed `Min` and `Max` to Value1 and Value2 and went through the changes, swapping them back to Min/Max when I judged that that was the intention of the code (and was not a bug).

I collapsed some manual checks on `IsInverted` that were no longer needed, and I also replaced (presumably) accidental uses of bitwise and `&` with logical and `&&`. To my surprise, `bool1 & bool2` returns a boolean in C#, and is thus legal in if conditions. The only point of this with booleans would be to prevent short-circuit evaluation and the potential branch prediction mishaps it can sometimes cause.

Finally, the `Equals` implementation for `CoordinateRangeMutable` was incorrect. I fixed that, though maybe it should be made in a separate PR since I have the feeling this one could take some time to get right.

## Remaining ambiguity

I have a fair number of questions, they're mostly in TODO comments.

I'm unsure of how to handle setters here. Maybe the right answer is to require these to be readonly properties, and I'm definitely warming to that viewpoint. My first pass was to have setters on Min/Max change whichever of Value1/Value2 was actually the min or max. Which is a relatively practical choice for simplifying code that was already consciously deciding to preserve `IsInverted` when writing to these properties. But it is also ambiguous in the case when Value1 = Value2, and so you can get very different ranges depending on what is fundamentally an arbitrary choice.

There's a lot of similar choices, like for the range `[1, -1]` there's two ways to expand that range to include 2. You can turn it into `[-1, 2]`, and you can turn it into `[2, -1]`. Currently my code uninverts them as did the previous implementation, which I feel is most reasonable, but I do also get the feeling that `IsInverted` should maybe be preserved?

Relatedly, should `[1, -1].Contains(0)` return true? Currently it does (the old code did not), but is this correct? Formally this is saying that the following includes the element 0, which is clearly false.

 $$ \\{ x \\: : \\: 1 \leq x \leq -1 \\} $$

And finally, this is inevitably a breaking change. If we argue that it's a bug fix then that's one thing, but if there's user code out there that is already handling inverted ranges then I don't think this is going to be pleasant for them.

All of this makes me wonder if this change is too ambitious and we should instead be trying to find a more modest way to manage this misnomer. Maybe that's just doc comments on `Min` and `Max`.

## Testing

Since this touches so much it'll be hard for me to adequately test this. At the very least I want to ensure that #4783 and #4629 don't regress, but I haven't tested that yet.

### DIFF
--- a/src/ScottPlot5/ScottPlot5/AxisRules/MaximumBoundary.cs
+++ b/src/ScottPlot5/ScottPlot5/AxisRules/MaximumBoundary.cs
@@ -11,62 +11,27 @@ public class MaximumBoundary(IXAxis xAxis, IYAxis yAxis, AxisLimits limits) : IA
         double horizontalSpan = Math.Min(Math.Abs(XAxis.Range.Span), Limits.XRange.Span);
         double verticalSpan = Math.Min(Math.Abs(YAxis.Range.Span), Limits.YRange.Span);
 
-        if (XAxis.IsInverted())
+        if (XAxis.Range.Max > Limits.XRange.Max)
         {
-            if (XAxis.Range.Min > Limits.XRange.Max)
-            {
-                XAxis.Range.Min = Limits.XRange.Max;
-                XAxis.Range.Max = Limits.XRange.Max - horizontalSpan;
-            }
-
-            if (XAxis.Range.Max < Limits.XRange.Min)
-            {
-                XAxis.Range.Max = Limits.XRange.Min;
-                XAxis.Range.Min = Limits.XRange.Min + horizontalSpan;
-            }
+            XAxis.Range.Max = Limits.XRange.Max;
+            XAxis.Range.Min = Limits.XRange.Max - horizontalSpan;
         }
-        else
+        if (XAxis.Range.Min < Limits.XRange.Min)
         {
-            if (XAxis.Range.Max > Limits.XRange.Max)
-            {
-                XAxis.Range.Max = Limits.XRange.Max;
-                XAxis.Range.Min = Limits.XRange.Max - horizontalSpan;
-            }
-
-            if (XAxis.Range.Min < Limits.XRange.Min)
-            {
-                XAxis.Range.Min = Limits.XRange.Min;
-                XAxis.Range.Max = Limits.XRange.Min + horizontalSpan;
-            }
+            XAxis.Range.Min = Limits.XRange.Min;
+            XAxis.Range.Max = Limits.XRange.Min + horizontalSpan;
         }
 
-        if (YAxis.IsInverted())
+        if (YAxis.Range.Max > Limits.YRange.Max)
         {
-            if (YAxis.Range.Min > Limits.YRange.Max)
-            {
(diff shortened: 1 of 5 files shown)


swharden:
> accidental uses of bitwise and & with logical and &&

Oof! That's a danger. I often do `bool || bool` and even `bool |= bool`, but I can see where `bool & bool` can be a bug that compiles only to cause hard to locate runtime bugs...

> Finally, the Equals implementation for CoordinateRangeMutable was incorrect

Oof! And thanks 😅 

> ambiguity

To be honest, I'm inclined _not_ to solve these issues right now. They are very far reaching. My plan for a while has been to complete delete `CoordinateRange` and `CoordinateRangeMutable`, and make something simpler with more defined behavior.

From experience that any modifications to this class, however simple they may seem, cause very difficult to predict regression bugs. Often they only affect obscure interactive use cases, like inverted non-primary axes, making it very difficult to effectively test...

For this reason I'm inclined to touch this file as little as possible (well, in this case let's merge and see what happens lol - I'll test it carefully before the release), then give this some serious time and attention when I have more of both to commit to this topic to solve it well.

PS: English is hard, but we should consider that a range of values on the horizontal and vertical axes extends from left to right and bottom to top, even if they're inverted. Maybe instead of "min" and "max" we should say "bottom" and "top", allowing that the _value_ of the bottom may be greater than the value at the top (an inverted axis situation). It's ambiguities in language and variable names that cause a lot of confusion around these classes. 

Thanks again for your work on this!

swharden:
> let's merge and see what happens

> any modifications to this class, however simple they may seem, cause very difficult to predict regression bugs

Yes, the autoformatter is having issues due to line endings (I'm working on it, but we can merge even if the autoformatter fails)

... but the unit tests are failing, so I'll not merge this one and put it down until we can figure out what the issue is 😅

bclehmann:
Yeah I think the move here is to scale this back a lot. Make `Min` and `Max` either not settable, or they assume that they aren't on an inverted range and they set `Value1` and `Value2` respectively. Effectively then the only change here would be to ensure that `Min`/`Max` reports the actual min/max.

I don't like my current implementation, I think it's cute, but it isn't intuitive or good to use. It's already addressing an edge case, we should at least keep the edge case easy to reason about. Honestly, I think it would be better to just add a doc comment about the issue and leave it be. I think replacing a rare issue with a different (admittedly more) rare issue that makes significantly less sense is not ideal. If this is already a breaking change we should at least break it to be right.

If making Min and Max return the correct values is still too weird then I think a doc comment is the right move.

swharden:
> I don't think it should be merged

I'll close this PR 👍 

> I think it would be better to just add a doc comment about the issue and leave it be

This topic is very much on my mind, and I'll use this PR as a more extensive source of nuanced documentation than a line comment could provide. Thanks again for your work on this, and I look forward to improving this system in the future!

PR #4801, opened 2025-02-18
Creates Avalonia Demo

bclehmann:
I'd like to get a lot more into this, I hadn't realized that the WPF demo was so incomplete. At least getting the cookbook running I think is important.

You'll notice that this isn't quite a copy-paste exercise from the WPF demo like it was when I ported the 4.0 demo to Avalonia. I figured since I know Avalonia a little better now (and the WPF demo is so small that copy-paste wouldn't save much time) that I would do it the MVVM way.

I may well decide to rewrite parts of the WPF demo like this too if that makes sense, but I think that's a consideration for later.

image


swharden:
> /home/runner/work/ScottPlot/ScottPlot/src/ScottPlot5/ScottPlot5 Demos/Avalonia Demo/Avalonia Demo.csproj requires 'SignAssembly' to be defined.

Unit tests are failing because all .csproj files require:

```xml
<SignAssembly>True</SignAssembly>
<AssemblyOriginatorKeyFile>../Key.snk</AssemblyOriginatorKeyFile>
```

bclehmann:
I'm more or less happy with it, I've got the cookbook running and looking reasonable, I just have to debate with myself about how many demos I'm going to port over.

That means if you want to take it off my hands now you're welcome to.

src/ScottPlot5/ScottPlot5 Demos/Avalonia Demo/Views/MainWindow.axaml
```diff
@@ -0,0 +1,80 @@
...
+				<Label
+                    Content="ScottPlot Avalonia Demo"
+                    Foreground="WhiteSmoke"
+                    FontSize="26"
+                    FontWeight="SemiBold"
+                    Padding="0 5 0 0"
+                    Margin="0"/>
+				<Label
+                    Name="Subtitle"
+                    Content="{Binding VersionStringFull}"
```
heartacker, line 37: could it be space or tab only,
maybe you can use `xamlstyler.console` to format this axaml file while in dotnet cli,
you can also install the vsix in vs 2022
and also ,you can find ext in vscode name:` dabbinavo.xamlstyler`

bclehmann: I went with `xstyler -d "path"`. It's a shame that this can't be configured with a `.editorconfig` file like our C# code.

heartacker:
yes. xstyler has their own config

swharden:
This is an ***enromous*** amount of effort @bclehmann, thank you so much!

I recognize there are a few other demo windows you may want to add later. I'll go ahead and merge this today so adding those new features won't involve such a large PR 👍

PR #4829, opened 2025-02-27
Tooltip: new plot type

### LINKED ISSUE #4751: Tooltip: new plot type (copy from ScottPlot 4)
A tooltip is like `Text` except it has a little pointer and is anchored to the coordinate system of the plot. It would be nice to bring this ScottPlot 4 plot type to ScottPlot 5, including the option to customize the orientation.

* ScottPlot 4 tooltip: https://github.com/ScottPlot/ScottPlot4-obsolete/blob/main/src/ScottPlot4/ScottPlot/Plottable/Tooltip.cs

* Related: #4706

CoderPM2011:
resolve #4751.

I inherited class of `Text` Plot and added the graphic rendering part of `Tooltip`.

Since `RenderTooltipShape()` is designed to be `protected virtual`, users can implement custom subclasses if they have different ideas about the current design. 😉

---
The current design requires the tail target coordinates to be specified and will automatically rotate in the target direction.

Customizable tooltip tail width percentage.
The actual width of the tail is the lesser of the length or width of the tooltip body.

#### The image below shows the actual shape of the tail profile.
image

---

### demo image
image

### demo code
```cs
Coordinates origin = new(0, 0);
for (int i = 360; i > 0; i -= 30)
{
    PolarCoordinates polar = new(i + 1, Angle.FromDegrees(i));
    var tooltip = formsPlot1.Plot.Add.Tooltip($"{i}", polar.ToCartesian(), origin);
    tooltip.Alignment = Alignment.MiddleCenter;
    tooltip.FillColor = Colors.Red.WithAlpha(0.5);
    tooltip.LineColor = Colors.Black.WithAlpha(0.5);
    tooltip.LineWidth = 5;
}
```


src/ScottPlot5/ScottPlot5/Plottables/Tooltip.cs
```diff
@@ -0,0 +1,98 @@
+namespace ScottPlot.Plottables;
+
+/// <summary>
+/// A tooltip displays a text bubble pointing to a specific location in X/Y space.
+/// The position of the bubble moves according to the axis limits to best display the text in the data area.
+/// </summary>
+public class Tooltip : Text, IPlottable, IHasLine, IHasFill
```
bclehmann, line 7: I'm a little conflicted on inheriting from `Text` here, since `Text` inherits from `LabelStyleProperties` and I don't think we respect things like `LabelBackgroundColor`, and if they differ I think we get the label drawing overtop of us with a different fill colour.

I think this would be alleviated if we didn't define our own FillStyle, but then we wouldn't support things like hatched fills (which honestly is probably fine). I think it's definitely simpler if we have the tooltip be separate and then it will only have fields that behave in a consistent manner with eachother

bclehmann (commented):
This looks good, I just have a couple of thoughts but a lot of them are minor and nitpicky

swharden:
Hi @CoderPM2011, thanks so much for this PR! I'm going to put some time into this now and will refactor it a bit, merge, and publish a new package today 🚀 

I suspect I'll disentangle the inheritance structure, and also address Ben's concern about text customization along the way

swharden:
Note that the "cutting" isn't pixel perfect, so I'll probably remove that feature...

image

image

swharden:
cookbook recipes

image

swharden:
> publish a new package today

publish a new package ~today~ _very soon_ 😅

PR #4841, opened 2025-03-06
LogDecadeMinorTickGenerator

macaba:
This adds a new minor tick generator for logarithmic charts.

I have assumed the behaviour of the existing LogMinorTickGenerator is correct, hence why I've gone with creating a new "LogDecade" tick generator. This should also prevent regressions in user apps as this new tick generator is opt-in.
It might well be the case that the code in this PR should actually be the new behaviour of the existing LogMinorTickGenerator, I leave that decision up to the reviewer.

What follows is a quick visual tour of the notable difference between LogMinorTickGenerator (the existing implementation) and LogDecadeMinorTickGenerator

## LogMinorTickGenerator

Code:
```
NumericAutomatic yTickGenerator = new()
{
    LabelFormatter = logTickLabels,
    MinorTickGenerator = new LogMinorTickGenerator() { Divisions = 10 },
    IntegerTicksOnly = true,
    TargetTickCount = 10
};
```

Behaviour when chart has 2 decades (this is correct & this should be the same with both generators)
image

Behaviour when chart has 1 decade (minor ticks are missing)
image

Behaviour when chart has less than 1 decade [0.9 decade] (minor ticks are missing)
image

## LogDecadeMinorTickGenerator

Code:
```
NumericAutomatic yTickGenerator = new()
{
    LabelFormatter = logTickLabels,
    MinorTickGenerator = new LogDecadeMinorTickGenerator() { TicksPerDecade = 10 },
    IntegerTicksOnly = true,
    TargetTickCount = 10
};
```

Behaviour when chart has 2 decades (this is correct & this should be the same with both generators)
image

Behaviour when chart has 1 decade (8 minor ticks are present and correct)
image

Behaviour when chart has less than 1 decade [0.9 decade] (7 minor ticks are present and correct)
image

## Notes

* On most of the charts, the "0.1" and/or "1" labels are now missing, that is probably a bug outside the scope of this PR. The workaround in my application is to add & subtract a tiny amount to the max/min axis limit.
* It might be good to have a feature along the lines of "Always show tick label at axis extents" so that there will always be a tick label at the axis extents (notably, this would show "0.1" and "0.9" for the previous screenshot).


macaba:
Thinking about it more, I think it makes sense for this to be a separate "LogDecade" tick generator forevermore, as it specifically encodes the behaviour of "ticks per log decade" with the `TicksPerDecade` property even if the display is a fractional decade or many decades, which is probably a different intent to the original Log tick generator which has a `Divisions` property.

swharden:
Hi @macaba, thanks for this PR! I'll add a cookbook recipe for this, merge in a few minutes, and it will be in the next release 🚀 

> It might well be the case that the code in this PR should actually be the new behaviour of the existing LogMinorTickGenerator

I'm happy to defer to the giants in the space! Maybe we can survey what matlab and matplotlib do and mimic them. I don't mind changing the default as long as the cookbook demonstrates how to go back to the original behavior in case someone wants to keep their new plots super consistent with ones from previous to versions.

> defer to the giants ... matlab and matplotlif

I probably won't do this research myself considering I'm catching up on PRs and other open issues, but if you or someone in the future wants to check this out and recommend a better default, let me know and we can reference this issue and get the new default merged in quick!

Thanks again 😄 
Scott

PR #4869, opened 2025-03-25
Kernel Density Estimation

bclehmann:
This adds functions for KDE, which can be used to create density plots:
image

There are 4 kernels supported: Epanechnikov (default), Gaussian, Uniform (note this is on [-1, 1], not the standard uniform distribution), and Triangular. There is currently only 1 bandwidth rule, which is Scott's rule. The API allows passing in arbitrary kernels and bandwidths, should the user wish to.

```cs
var ys = SampleData.Faithful;

var hist = Histogram.WithBinCount(80, ys);

var histPlot = myPlot.Add.Histogram(hist);
histPlot.BarWidthFraction = 0.8;
foreach(var bar in histPlot.Bars)
{
    bar.FillColor = Colors.LightBlue;
}

foreach(var kernel in Enum.GetValues<KdeKernel>())
{
    var densityEstimate = hist.Bins.Select((x, i) => KernelDensity.Estimate(x, ys, kernel)).ToArray();
    double scale = ys.Max() / densityEstimate.Sum();

    var rescaledDensityEstimate = densityEstimate.Select(x => x * scale).ToArray();

    var scat = myPlot.Add.Scatter(hist.Bins, rescaledDensityEstimate);
    scat.MarkerSize = 0;
    scat.LegendText = kernel.ToString();
}
```

There is some awkwardness in the API, as there's also a `ProbabilityDensity` class, which assumes normally-distributed data. I don't know how you want to clean up the naming ambiguity.

## Other thoughts

The impetus for this was https://github.com/dotnet/BenchmarkDotNet/pull/2711, where they want to replace some R scripts with C#. On that list is density plots, which is what pushed me to implement it after having wanted it for a while.

This can also be used to create violin plots, but that might be something I defer for another day.

Lastly, there are 2D variations of this, which could potentially be useful for heatmaps (ScottPlot 4 had a slightly quirky implementation of this on binary heatmaps for Gaussian and box/uniform kernels), but I don't know if they're necessary. I would probably not go with the ScottPlot 4 API, since supporting only binary data was an odd choice on my part.


swharden:
Hey @bclehmann, thanks for this PR! The delay in merging is a lot longer than typical (#4975) but I'm happy to be getting to it now 🚀

swharden:
lol, thank you for teaching me about https://en.wikipedia.org/wiki/Scott%27s_rule this is a gem

PR #4873, opened 2025-03-26
Font: add support for weight, slant, and width

Christoph-Wagner:
Instead of just specifying if a font should be bold and/or italic, this now allows to choose exact styles everywhere. The old interfaces have been kept for backwards compatibility. 

```cs
myPlot.Font.SetExact("Myriad Pro", SKFontStyleWeight.Normal, SKFontStyleSlant.Upright, SKFontStyleWidth.Condensed);
```


aespitia:
would there be a way to have it figure out the weight/style just based on the font itself?  i think we've run into some issues where a font's default was that it was Condensed or Book or some other non-normal setting by default and the font wasn't detected at all.  I think most people won't know that they have to use those other settings, they just know the font name they want to use.  

This would be a great addition, I've definitely run into some font issues before, I hope this gets merged in!

swharden:
Hi @Christoph-Wagner, thanks so much for this PR! It took me longer to get to it than typical (#4975) but I'm happy to be reviewing it now.

I definitely want to get these new features implemented, but I hesitate because I think I would prefer to keep SkiaSharp types out of ScottPlot's primitives (FontStyle and LabelStyle).

If you want to create ScottPlot primitives for weight, slant, and width that have extension methods to convert them to SkiaSharp primitives, that would be a good option.

If someone gets to this before I do I will be happy to review it soon, otherwise I'll mark this as "defer" and I'll close some of the other open PRs and loop back around to this and make that change myself.

Thanks again for this new feature! 🚀

swharden:
Hey @aespitia, your idea could work! If you create a new PR building on these changes, here's a sequence of events leading to this getting into the next NuGet package
* You could copy these changes into a new PR and make the requested modifications
* You can add your own features too
* I'm out of town until Monday 8/4
* When I get back I'll review your PR and merge it early in the week
* The merge will tag both of you as authors
* I'll then publish a NuGet package probably sometime next week

Does that work for you? I really like the direction this PR is going, and I'd love to have these features in the next version! 🚀

swharden:
@Christoph-Wagner, thank you SO MUCH for this PR! The commits have been made in #5013 and extended, and when I merge that PR I'll credit you as a co-author. I'll close this PR since the changes are being tracked over there 👍

PR #4897, opened 2025-04-03
PolarPlot: Supported filled areas

### LINKED ISSUE #4885: PolarPlot: Supported filled areas
It would be nice if the unit circle could be filled with a color.

A useful situation is producing a figure like this where the figure and data area backgrounds are the control color, but the unit circle is white...

Image

I did this by adding a white circle shape below the axis, which works, but it's not elegant...

CoderPM2011:
I think the need for #4885 is the background fill function of `PolarAxis`, which is to fill the circle (data range) of `PolarAxis` with a specified color.  (If I misunderstood, please tell me 😅)

- The circle (data range) is the maximum of the radius of all `Circles` and the length of `Spokes`.

For this reason I added a new process `RenderBackgroundColor` to `Render` to fill the circle (data range) with the specified color, and implemented `IHasFill` to support the fill color setting.

### Related changes
1. `PolarAxis` implements `IHasFill`
1. Added `RenderBackgroundColor` method and used it in `Render`
1. Added optional spoke length parameter to `PolarAxis` in `PlottableAdder`

### Demo Image
image

### Demo Code
```cs
formsPlot1.Plot.DataBackground.Color = Colors.LightGray;
// Specifies the default length of the spokes.
var polarAxis = formsPlot1.Plot.Add.PolarAxis(radius: 100, spokeLength: 100);
polarAxis.FillColor = Colors.Pink;
```

resolve #4885

### DIFF
--- a/src/ScottPlot5/ScottPlot5/PlottableAdder.cs
+++ b/src/ScottPlot5/ScottPlot5/PlottableAdder.cs
@@ -915,11 +915,11 @@ public class PlottableAdder(Plot plot)
         return plottable;
     }
 
-    public PolarAxis PolarAxis(double radius = 1.0)
+    public PolarAxis PolarAxis(double radius = 1.0, double? spokeLength = null)
     {
         PolarAxis polarAxis = new() { };
         polarAxis.SetCircles(radius, 5);
-        polarAxis.SetSpokes(12, radius * 1.1);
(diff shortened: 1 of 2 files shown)


swharden:
This is awesome @CoderPM2011, thanks so much! 😊🚀

It took me a while to get to this one (#4975) but I'm catching up.

Thanks again!
Scott

Issue #4903, opened 2025-04-07
Plot.GetImageBytes can hang


NeilMacMullen:
**Issue:** (Describe the bug here)

Plot.GetImageBytes can hang when attempting to retrieve the image bytes.   The problem _may_ be related to using smaller values for width/height although rendering the same chart to a Plot that is backing a small WPFPlot does not cause issues.  I am trying to narrow the problem down to provide a more specific repro case but one clue may be that I've noticed that calling plot.ShowLegend(edge) as well as setting plot.Legend.Alignment/Orientation can sometimes cause WpfPlot to hang.  Maybe a layout calculation loop ?

**ScottPlot Version:** (What NuGet package are you using?)

5.0.55

**Code Sample:** See http://scottplot.net/faq/repro/ for tips about creating reproducible code samples

```csharp
  using var plot = new Plot();
 ... generate chart 
  var bytes = plot.GetImageBytes(width, height, ImageFormat.Png);
```

NeilMacMullen:
Notes...

It appears that XAxisBase is being called with size of 0.  This causes an infinite loop in DateTimeAutomatic.GenerateTicks.

dateTime4 is {01/01/0100 00:00:00}
increment is 2147483647 (0x7fffffff) which causes unit.Next to constantly return the start value

NeilMacMullen:
It looks like RenderManager.RenderOnce is generating a RenderPack with left and right set to the same value.  Maybe the legend is (which is set to edge.right) is squashing the rendering area down  to zero-width section.

NeilMacMullen:
I'm not sure if it's sufficient to catch all cases but maybe an appropriate fix is to modify DateTimeAutomatic.Regenerate...

```cs
public void Regenerate(CoordinateRange range, Edge edge, PixelLength size, SKPaint paint, LabelStyle labelStyle)
{
	if (range.Span >= (double)TimeSpan.MaxValue.Days || double.IsNaN(range.Span) || double.IsInfinity(range.Span)
       || (size.Length ==0)
)
	{
		Ticks = Array.Empty<Tick>();
		return;
	}

```

NeilMacMullen:
Trivial test case to illustrate the problem.  Can be added to AxisTickTests.cs

```csharp
 [Test]
 public void Test_DateTimeAutomaticDoesNotHang()
 {
     Plot plt = new();
     DateTime dt1 = new(2023, 01, 01);
     DateTime dt2 = new(2024, 01, 01);
[...]

PR #4911, opened 2025-04-09
DateTimeAutomatic: fix infinite loop for zero-sized axes

NeilMacMullen:
PR to address #4903 

DateTimeAutomatic can easily enter an infinite loop when generating ticks for a zero-sized axis.  

The code changes take a "belt and braces" approach to this by checking for a number of suspicious conditions including:
- zero/negative sized pixel span
- zero/negative increment
- failure for unit.Next to advance (at least on the very first datetime)

Note this may conflict with https://github.com/ScottPlot/ScottPlot/pull/4900 although resolution should be simple.


mccabe93:
Just want to chime in that I've used this fix for an Avalonia case and that it works. Without it, my slider for showing/hiding the chart causes the application to hang indefinitely. I can provide a test case if necessary, but imo it's pretty obvious what this PR does.

swharden:
I'm merging this now, but note that there's a change from https://github.com/ScottPlot/ScottPlot/commit/c4a35014f73792fdb1b5b1f158671d45569c6e4a causing a merge conflict I'll resolve before merging. 

Thanks for both your help on this! I wish it didn't take me so long to get to it, but better late than never 😅

PR #4956, opened 2025-06-04
Drawing: improve support for invisible and zero line width rectangles

Fruchtzwerg94:
**Rectangles can't be drawn without a line around them**
Drawing a rectangle with fill only and no line around is currently not possible.

```cs
rectangle.LineStyle.IsVisible = false;
rectangle.LineStyle.Width = 0;
```

A line always is drawn. The reason is that `CanBeRendered` is not checked like done for other forms. SkipSharp itself draws the smallest possible line if 0 is set as width.

Fix: Removed the missleading function and use one where `CanBeRendered` is checked accordingly.


swharden:
Thanks @Fruchtzwerg94, this was some really surprising default behavior but I'm happy you were able to figure out and offer a fix!

I'll merge this now and plan to publish a new release on NuGet within the next few weeks.

Thanks again! 🚀

PR #4970, opened 2025-06-11
WinUI: Add keyboard modifier detection (#4967)

### LINKED ISSUE #4967: WinUI ScottPlot: How to properly detect keyboard modifiers in custom UserActionResponse (Ctrl+MouseWheel zoom)
**ScottPlot Version**: 5.0.55
**Platform**: WinUI 3
**Environment**: .NET 8

Hi! I'm working on a WinUI application using `ScottPlot` and have encountered issues that I'd like to address properly using ScottPlot's built-in systems:

I'm trying to create a custom mouse wheel behavior where:

- Normal mouse wheel: No plot interaction (let parent ScrollViewer handle it)
- Ctrl + mouse wheel: Plot zoom only

I've created a custom `IUserActionResponse` but can't get keyboard modifier detection to work properly. The control doesn't receive `KeyDown`/`KeyUp` events, so `KeyboardState.IsPressed()` always returns false.

````csharp
public class CtrlOnlyMouseWheelZoom : IUserActionResponse
{
    public ResponseInfo Execute(IPlotControl plotControl, IUserAction userInput, KeyboardState keys)
    {
        // This always returns false, even when Ctrl is pressed
        bool isCtrlPressed = keys.IsPressed(StandardKeys.Control);
        
        if (userInput is MouseWheelUp && isCtrlPressed)
        {
            // Handle zoom in
        }
        else if (userInput is MouseWheelDown && isCtrlPressed)
        {
            // Handle zoom out
        }
        
        return ResponseInfo.NoActionRequired;
    }
}
````

**Questions:**
1. Is `KeyboardState.IsPressed()` the correct way to detect modifiers in UserActionResponse?
1. Are there any platform-specific issues with keyboard state detection in WinUI?
1. How does the built-in `KeyboardPanAndZoom` class detect `ZoomModifierKey` - should I follow that pattern?

**Additional observation:** 
The built-in `KeyboardPanAndZoom` also seems to have the same issue with `Ctrl` key detection - this suggests the keyboard modifier detection might have broader issues in WinUI.

**Current workaround:** 
Using WinUI's `PointerWheelChanged` with `e.KeyModifiers.HasFlag(VirtualKeyModifiers.Control)`, but I'd prefer to use ScottPlot's interaction system.

Any guidance on the proper approach would be greatly appreciated!

diluculo:
Implements `UpdateKeyStateFromModifiers` helper to detect Ctrl/Shift/Alt states from mouse events, enabling enhanced interactions like Ctrl+MouseWheel zoom.

Also upgrades Windows App SDK to 1.7 and lowers target framework to net8.0-windows10.0.19041.0 due to v1.5 runtime issues in my environment (can be reverted if needed).

Fixes #4967

### Additional Notes

For future `KeyDown`/`KeyUp` event handling from `SKXamlCanvas` in `WinUIPlot.cs`, the `CreateRenderTarget()` method would need focus-related properties:

````csharp
private static SKXamlCanvas CreateRenderTarget()
{
   return new SKXamlCanvas
   {
       // ... (existing VerticalAlignment, HorizontalAlignment, Background) ...
       
       IsTabStop = true,                    // Enables the control to receive focus
       AllowFocusOnInteraction = true       // Allows focus on mouse interaction
   };
}
````

This PR focuses on mouse-based modifier detection using `PointerRoutedEventArgs`, which doesn't require focus.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@ _Not yet on NuGet..._
 * Horizontal and Vertical Line: Improve support for inverted axes (#4920) @manaruto
 * Rendering: Improve support for invisible and zero line width rectangles (#4956) @Fruchtzwerg94
 * Font: Fixed Font.Set() issue associated with premature typeface disposal (#4910, #4958) @sproott @237779932
+* WinUI: Improve support for modifier keys when combining mouse and keyboard inputs (#4967, #4970) @diluculo
 
 ## ScottPlot 5.0.55
 _Published on [NuGet](https://www.nuget.org/profiles/ScottPlot) on 2025-03-22_


swharden:
Thanks @diluculo! I don't spend a lot of time working with WinUI so I _really_ appreciate input from developers like you who are much more familiar with the platform. I'm going to merge this now and I intend to publish a new set of NuGet packages within the next few weeks.

Thanks again! 🚀
Scott

PR #4985, opened 2025-06-29
Add Clamp overloads for all common value types to avoid boxing and improve performance

kevin100702:
Add Clamp overloads for all common value types to avoid boxing and improve performance

- Added explicit Clamp overloads for sbyte, byte, short, ushort, int, uint, long, ulong, float, double, and decimal.
- Ensures value type calls use the most efficient implementation and do not fall back to the generic version.
- Retained the generic Clamp<T> for compatibility with other types.


heartacker:
Inumber

swharden:
Thanks @kevin100702, this looks great! Micro-optimizations like this can really add up over time, and I appreciate you identifying this potential performance improvement and addressing it in a PR.

I'm merging now, and will publish an updated NuGet package within the next few weeks.

Thanks again! 🚀
Scott

PR #4988, opened 2025-07-02
ScottPlot.Maui: Target .NET 8

### LINKED ISSUE #4943: ScottPlot.Maui package does not support net8.0 target, blocking unit testing in .NET MAUI projects
**Issue:**
When following the official Microsoft guidance for unit testing .NET MAUI app projects (see: https://learn.microsoft.com/en-us/dotnet/maui/deployment/unit-testing?view=net-maui-8.0#configure-a-net-maui-app-project-for-unit-testing), it is recommended to add net8.0 to the TargetFrameworks property in the MAUI app's .csproj. However, adding a reference to the ScottPlot.Maui NuGet package causes build failures on the net8.0 target because the ScottPlot.Maui package does not provide a net8.0 target. This prevents MAUI projects that use ScottPlot from being unit tested as recommended.

**ScottPlot Version:** 5.0.55 (ScottPlot.Maui NuGet package)

**Code Sample:**
The easiest way to reproduce this is to change 
 https://github.com/ScottPlot/ScottPlot/blob/461efbdb139fb7c7d7a4bae185cde26e885b74bd/src/ScottPlot5/ScottPlot5%20Sandbox/Sandbox.Maui/Sandbox.Maui.csproj#L7

to 

```xml
<TargetFrameworks>net8.0-windows10.0.19041.0;net8.0</TargetFrameworks>
```

which adds `net8.0` to the target frameworks and attempt to build the Sandbox.Maui project. The build will fail but note that deploying should still work as long as the appropriate target framework is selected.

Adam--:
Updated ScottPlot.Maui to include `net8.0` as a target framework alongside existing platforms. This allows for unit testing in MAUI apps see: https://learn.microsoft.com/en-us/dotnet/maui/deployment/unit-testing?view=net-maui-8.0#configure-a-net-maui-app-project-for-unit-testing. With this missing, the project cannot compile. This is common practice for MAUI. I don't know if this is necessary for any other platforms.

Resolves #4943

### DIFF
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -1,0 +1,13 @@
+{
+        "name": "ScottPlot",
+        "image": "mcr.microsoft.com/devcontainers/dotnet:9.0",
+        "features": {},
+        "customizations": {
+                "vscode": {
+                        "extensions": [
+                                "ms-dotnettools.csdevkit",
+                                "ms-dotnettools.dotnet-interactive-vscode"
+                        ]
+                }
+        }
+}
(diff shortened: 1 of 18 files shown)


swharden:
Thanks @Adam--! I'll merge this now and deploy a new set of NuGet packages within the next few weeks 👍

swharden:
@Adam--, I'm curious about this bit from the docs you linked... can we ignore it because this is library code?

> It's also necessary to modify your .NET MAUI app project so that it doesn't output an executable for the target framework used by the xUnit test project. This can be achieved by adding a condition to the $(OutputType) build property in the .NET MAUI app project file:
```xml
<OutputType Condition="'$(TargetFramework)' != 'net8.0'">Exe</OutputType>
```

Issue #5005, opened 2025-07-21
Legend Items seem to ignore the marker style setting


aespitia:
**Issue:** It looks like looping through legend items to update the marker style from the default to another markerstyle doesn't work.  From looking at the code it also looks like some plot types set a default marker to a filled circle, but the current implementation ignores it.  

**ScottPlot Version:** 5.0.55

**Code Sample:** 

```cs
ScottPlot.Plot myPlot = new();

List<PieSlice> slices =
[
	new PieSlice() { Value = 5, FillColor = Colors.Red, Label = "Red", LegendText = "R" },
	new PieSlice() { Value = 2, FillColor = Colors.Orange, Label = "Orange" },
	new PieSlice() { Value = 8, FillColor = Colors.Gold, Label = "Yellow" },
	new PieSlice() { Value = 4, FillColor = Colors.Green, Label = "Green", LegendText = "G" },
	new PieSlice() { Value = 8, FillColor = Colors.Blue, Label = "Blue", LegendText = "B" },
];

var pie = myPlot.Add.Pie(slices);
pie.ExplodeFraction = .1;
pie.SliceLabelDistance = 1.4;

var legitems = myPlot.Legend.GetItems();

foreach (var item in legitems)
{
	item.LabelText = item.LabelText + "-modified";
	item.MarkerStyle.Shape = MarkerShape.FilledCircle;
	item.MarkerShape = MarkerShape.FilledCircle;
	item.MarkerColor = item.FillColor;
	item.MarkerSize = 20;

	/*
	myPlot.Legend.ManualItems.Add(new LegendItem()
	{
		LabelText = item.LabelText,
		LabelFontName = item.LabelFontName,
		LabelFontSize = item.LabelFontSize,
		LabelFontColor = item.LabelFontColor,
		MarkerShape = MarkerShape.FilledCircle,
		MarkerColor = item.LineColor,
		MarkerSize = 10
	});
	*/
}

myPlot.ShowLegend();

myPlot.Axes.Frameless();
myPlot.HideGrid();

var bitmap = myPlot.GetImageBytes(400, 300);

Util.Image(bitmap).Dump();
```

PR #5006, opened 2025-07-21
Legend: default marker style

aespitia:
This is to hopefully implement a default marker type, in case you want to keep all of legend defaults, but change the marker style.  Looking for feedback from @swharden or anyone else familiar with the Legend/LegendItems, but hopefully fixes issue #5005

I included a new cookbook item for Legend > Default Marker style

```cs
var sig1 = myPlot.Add.Signal(Generate.Sin(51));
sig1.LegendText = "Sin";

var sig2 = myPlot.Add.Signal(Generate.Cos(51));
sig2.LegendText = "Cos";

myPlot.Legend.MarkerShapeDefault = MarkerShape.FilledCircle;


LegendItem item1 = new()
{
    MarkerColor = Colors.Red,
    MarkerShape = MarkerShape.Cross,
    LabelText = "Alpha"
};

LegendItem item2 = new()
{
    MarkerColor = Colors.Green,
    MarkerShape = MarkerShape.FilledSquare,
    LabelText = "Beta"
};

myPlot.Legend.ManualItems.Add(item1);
myPlot.Legend.ManualItems.Add(item2);

myPlot.ShowLegend();
```

image

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,7 @@ _Not yet on NuGet..._
 * Maui: Improve behavior for interactive plots when their user input processor is disabled (#4990, #4989) @Adam--
 * Fonts: Added new styling options for weight, slant, density, etc. (#5013, #4873) @aespitia @Christoph-Wagner
 * Labels: Added styling options for underline with customizable thickness and offset (#4893) @manaruto
+* Legend: Added the ability to customize default marker shape for legend items (#5005, #5006) @aespitia
 
 ## ScottPlot 5.0.55
 _Published on [NuGet](https://www.nuget.org/profiles/ScottPlot) on 2025-03-22_
--- a/src/ScottPlot5/ScottPlot5 Cookbook/Recipes/General/Legend.cs
+++ b/src/ScottPlot5/ScottPlot5 Cookbook/Recipes/General/Legend.cs
@@ -138,10 +138,10 @@ public class Legend : ICategory
 
             LegendItem item2 = new()
             {
-                MarkerColor = Colors.Green,    
+                MarkerColor = Colors.Green,
                 MarkerShape = MarkerShape.FilledSquare,
                 LabelText = "Beta"
-            };          
+            };
 
             myPlot.Legend.ManualItems.Add(item1);
             myPlot.Legend.ManualItems.Add(item2);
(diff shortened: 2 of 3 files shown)


aespitia:
> wow, that was fast @aespitia, thanks so much!
> 
> Do you think we should roll changes from #4893 into this PR too and merge it all at once? I haven't done a thorough review yet, but adding underline support seems closely related and I'm not sure whether these branches would cause merge conflicts or perhaps if you'd be a good person to consolidate these different PRs into this one because you have the finger on the pulse of what our emerging API is for font styling.
> 
> I'm about to head out camping for several days, but I'll be back Monday and can take a closer look after that! It's at Anastasia State Park in Florida with 100F/38C heat so it'll be quite an experience lol
> 
> Thanks again for all your help on this! I really appreciate all the heavy lifting you do with these complex PRs and nuanced topics 🚀

I think this particular PR is a little different, this one is about legend markers, I had this PR in before the Font Resolver PR, however, I think the #4893 underline changes can probably be merged into the font one, I can make another local branch and do a test merge to see if I run into any conflicts, but it might go in clean.

manaruto:
Hi @aespitia

I have recently opened a new [pull request](https://github.com/ScottPlot/ScottPlot/pull/5031) that builds upon the changes made here.
My PR addresses a regression introduced after this one, MarkerStyle.Size stopped taking effect when MarkerShapeDefault was set and also fixes the marker clipping issue.

It would be great if you could take a look at my changes to ensure they don’t conflict with the intent of your original fix. I tried to preserve the functionality you introduced while resolving the size and clipping issues.

Thanks in advance

aespitia:
> Hi @aespitia
> 
> I have recently opened a new [pull request](https://github.com/ScottPlot/ScottPlot/pull/5031) that builds upon the changes made here. My PR addresses a regression introduced after this one, MarkerStyle.Size stopped taking effect when MarkerShapeDefault was set and also fixes the marker clipping issue.
> 
> It would be great if you could take a look at my changes to ensure they don’t conflict with the intent of your original fix. I tried to preserve the functionality you introduced while resolving the size and clipping issues.
> 
> Thanks in advance

I pulled your branch and the recipe i added still works, I like it, thanks for taking a look.  I wasn't sure what was going on in there, which is why I left that note in there, but it had fixed an existing bug anyway where some of the legend items were using the wrong marker anyway.  thanks again!

swharden:
Quick update, signal plots now show large circle markers

image

I think the issue is here. I'll try to get a quick fix in by editing https://github.com/ScottPlot/ScottPlot/pull/5048

https://github.com/ScottPlot/ScottPlot/blob/a69387cb4fcfa5be240658f5b5b09a7e8419451d/src/ScottPlot5/ScottPlot5/Primitives/Legend.cs#L321-L335

PR #5013, opened 2025-07-30
Fonts: Extend styling options (weight, slant, density, etc.)

aespitia:
This is a branch from #4873, where @Christoph-Wagner added support for more specialized fonts.  Historically, the font resolver would only default to "Normal" or "Regular" width fonts, and would not appropriately find Condensed and other font widths. This PR updates that PR with some notes/feedback from @swharden in order to make primitive types

Sample usage with a system installed font
```cs
myPlot.Font.SetExact("Arial Narrow", FontWeight.Normal, FontSlant.Upright, FontWidth.Condensed);
myPlot.SavePng("bug.png");
```
using Arial Narrow

image

### DIFF
--- a/src/ScottPlot5/ScottPlot5 Cookbook/Recipes/General/Styling.cs
+++ b/src/ScottPlot5/ScottPlot5 Cookbook/Recipes/General/Styling.cs
@@ -354,22 +354,71 @@ public class Styling : ICategory
         }
     }
 
-    public class SetSpecificSystemFont : RecipeBase
+    public class SetFontName : RecipeBase
     {
-        public override string Name => "Set Specific System Font";
-        public override string Description => "This allows you to set a specific font, but must be installed on the system. This font resolver lets you control the weight, slant, and width.";
+        public override string Name => "Set Font by Name";
+        public override string Description => "Set font by its name to apply it to common plot components.";
 
         [Test]
         public override void Execute()
         {
-            myPlot.Font.SetExact("Arial Narrow", FontWeight.Normal, FontSlant.Upright, FontWidth.Condensed);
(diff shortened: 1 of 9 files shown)


swharden:
FYI I'm updating the demos to use Calibri because Arial Narrow isn't working with bold and italic on my system. Weird!

swharden:
I'm also thinking `SKFontStyleWidth` may not be supported... it's not doing anything on my system, and there's not much on google images demonstrating it

swharden:
Hey @aespitia, a quick comment about this solution... I don't think we should store much plot-specific state in the static `Fonts` class because processes with generate tons of plots (e.g., the cookbook) get broken when one plot changes the default options, then all the subsequent new plots have different styles. 

I haven't figured it out exactly yet, but I'm going to try to move that default style state into the Plot... or add a ResetDefaults() and call it before every cookbook recipe. It's probably a niche case, but it would be nice to solve it elegantly.

I welcome any suggestions you may have about the best way to go about this! You don't need to come up with a PR - I'll make a quick fix so I can publish new packages tonight and we can refine this behavior if/as needed.

Thanks!
Scott

swharden:
Hey @aespitia, I implemented the `Reset()` approach in #5048 but now I see the same issue affects default Legend style 😅

I'll squeeze in a `Reset()` there too, then later we can figure out a way to tie these defaults to the `Plot` instead of at the static `ScottPlot` level 👍

swharden:
Typical me, I commented before thorough investigation and got my assessment wrong. Sorry for the noise about legends! There is an issue (see MarkerLegend cookbook recipe below) but I don't think it's related to static storage of defaults

image

Issue #4884, opened 2025-03-30
Polar plot: add option to allow clockwise orientation


swharden:
Polar plots are typically counter-clockwise, but not always.

![Image](https://github.com/user-attachments/assets/4af176b5-df64-428e-b1f6-61fc85e16aa9)

![Image](https://github.com/user-attachments/assets/7a87e285-7228-4ab4-ac34-eecdb86a1b76)

![Image](https://github.com/user-attachments/assets/dea8a90f-94c1-4353-b1b8-b00f8374bb09)

![Image](https://github.com/user-attachments/assets/88e4e09a-ef5e-4719-95e7-592b818ff2c7)

CoderPM2011:
I have found that this can be accomplished simply by reversing the Y-axis.

Test based on https://scottplot.net/cookbook/5.0/PolarAxis/PolarQuickStart/
After adding `formsPlot1.Plot.Axes.InvertY();`, it looks like the following:
![Image](https://github.com/user-attachments/assets/b46277af-2408-429a-a039-a503660b422f)

horlegsimpSon:
I may I found a solution to your problem. Check it out here [click](https://javierlopez.eu/captha/)

mattwelch2000:
Inverting the Y-axis only works when when you leave the rotation at zero. Once you rotate the axis coordinates get placed opposite where they should be.

To my knowledge the only real difference between a clockwise or counter-clockwise polar chart is the direction of the angle. So I just modified the classes to reverse the angle if the axis is clockwise. This works great for me.

Updated PolarAxis.cs by adding an `IsClockwise` property and changing the base `SetSpokes` and `GetCoordinates` methods.
```cs
public bool IsClockwise { get; set; } = false;

public void SetSpokes(Angle[] angles, double length, string[] labels)
{
    if (angles.Length != labels.Length)
        throw new ArgumentException($"{nameof(angles)} and {nameof(labels)} must have equal length");

    Spokes.Clear();
    for (int i = 0; i < angles.Length; i++)
    {
        Angle angle = angles[i];
        if (IsClockwise)
        {
            // reverse the angle when clockwise.
            angle = -angle;
        }
        PolarAxisSpoke spoke = new(angle, length) { LabelText = labels[i] };
        Spokes.Add(spoke);
    }
}

public Coordinates GetCoordinates(PolarCoordinates point)
{
    if (IsClockwise)
    {
        // subtract the rotation and reverse the angle when clockwise.
[...]

PR #5028, opened 2025-08-07
PolarAxis: Fix incorrect rotation direction when axes are inverted

CoderPM2011:
This update fixes the rotation direction based on the inversion state of the X and Y axes.

Maybe it can solve the issue #4884

### demo result
A 30-degree rotation was deliberately added in the demo to test the behavior.

The rotation direction depends on the inversion states of the X and Y axes.
If one axis is inverted and the other is not, the rotation is clockwise.
If both axes share the same inversion state, the rotation is counterclockwise.

image

### demo code
```cs
var multiplot = formsPlot1.Multiplot;

multiplot.AddPlots(4);
for (int i = 0; i < 4; i++)
{
    // add a polar axis to the plot
    var plot = multiplot.Subplots.GetPlot(i);
    var polarAxis = plot.Add.PolarAxis(radius: 100);
    polarAxis.Rotation = Angle.FromDegrees(30);

    var colormap = new ScottPlot.Colormaps.Turbo();
    foreach (double fraction in ScottPlot.Generate.Range(0, 1, 0.02))
    {
        // use the polar axis to get X/Y coordinates given a position in polar space
        double radius = 100 * fraction;
        double degrees = 360 * fraction;
        Coordinates pt = polarAxis.GetCoordinates(radius, degrees);

        // place markers or other plot types using X/Y coordinates like normal
        var marker = plot.Add.Marker(pt);
        marker.Color = colormap.GetColor(fraction);
    }
}

// plot0 No invert axis
multiplot.Subplots.GetPlot(0).Title("+30° No invert axis");

// plot1 Invert X axis
multiplot.Subplots.GetPlot(1).Title("+30° Invert X axis");
multiplot.Subplots.GetPlot(1).Axes.InvertX();

// plot2 Invert Y axis
multiplot.Subplots.GetPlot(2).Title("+30° Invert Y axis");
multiplot.Subplots.GetPlot(2).Axes.InvertY();

// plot3 Invert X & Y axis
multiplot.Subplots.GetPlot(3).Title("+30° Invert X & Y axis");
multiplot.Subplots.GetPlot(3).Axes.InvertX();
multiplot.Subplots.GetPlot(3).Axes.InvertY();

// configure the multiplot to use a grid layout
multiplot.Layout = new ScottPlot.MultiplotLayouts.Grid(rows: 2, columns: 2);
```

### DIFF
--- a/src/ScottPlot5/ScottPlot5/Plottables/PolarAxis.cs
+++ b/src/ScottPlot5/ScottPlot5/Plottables/PolarAxis.cs
@@ -232,7 +232,10 @@ public class PolarAxis : IPlottable, IManagesAxisLimits, IHasFill
         using SKAutoCanvasRestore _ = new(rp.Canvas);
         Pixel origin = Axes.GetPixel(Coordinates.Origin);
         rp.Canvas.Translate(origin.X, origin.Y);
-        rp.Canvas.RotateDegrees(-(float)Rotation.Degrees);
+
+        // Gets the actual rotation angle in Cartesian coordinates, adjusted for axis inversion.
+        Angle effectiveRotation = Axes.XAxis.IsInverted() ^ Axes.YAxis.IsInverted() ? -Rotation : Rotation;
+        rp.Canvas.RotateDegrees(-(float)effectiveRotation.Degrees);
 
         foreach (var spoke in Spokes)
         {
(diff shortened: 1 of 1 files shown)


swharden:
... also @CoderPM2011, it's not totally clear to me whether the purpose of this is an attempt to solve #4884, or whether it's to improve support for polar plots on inverted axes (a use case which seems odd to me, but I'm willing to consider).

After #4884 I'm happy to reconsider this feature if you think it would be useful!

Issue #4950, opened 2025-05-23
Exporting a heatmap as SVG outputs a smoothed svg.


MRC-Karel:
**Issue:** Exporting as a SVG smoothes the output, while the image format does not. 

**ScottPlot Version:** Nuget Scottplot version 5.0.55

**Code Sample:** 
```cs
            ScottPlot.Plot myPlot = new ScottPlot.Plot();

            double[,] data = {
                { 1, 2, 3 },
                { 4, 5, 6 },
                { 7, 8, 9 },
            };

            // add a heatmap to the plot

            myPlot.Add.Heatmap(data);

            myPlot.SavePng("demo.png", 400, 300);
            myPlot.SaveSvg("demo.svg", 400, 300);
```

![Image](https://github.com/user-attachments/assets/b6ce6bf3-f738-4a06-8712-ea84b43b8ea7)
![Image](https://github.com/user-attachments/assets/df1d91bb-1379-47d7-9976-116e1f82df69)

timmer98:
Hi @MRC-Karel and @swharden,

**ScottPlot Version:** 5.1.57 (I checked the issue through a test in the ScottPlot project directly)

I am also trying to output a heatmap to svg and noticed the same issue. But the `heatmap.RenderStrategy = new ScottPlot.Plottables.Heatmap.RenderStrategies.Rectangles();` did not solve it for me. It seems to be some kind of scaling issue here. Changing the `hm.Rectangle = new CoordinateRect(20, 40, 60 ,80);` shows a different picture, but still does not show the complete heatmap. As far as I understood, the setting of hm.Rectangle should not have a sideeffect on what part of the heatmap is shown in the data area or? 

```
// create a heatmap of any size
double[,] data =
{
    { 1, 2, 3 },
    { 4, 5, 6 },
    { 7, 8, 9 },
    { 1, 2, 3 },
    { 3, 3, 3 },
    { 7, 8, 9 },
    { 5, 1, 3 },
};

// add the heatmap to the plot
var hm = myPlot.Add.Heatmap(data);
hm.RenderStrategy = new ScottPlot.Plottables.Heatmap.RenderStrategies.Rectangles(); // For png export, this is deactivated.

var svg = myPlot.SaveSvg("heatmap-rectangle.svg", 500, 500);
svg.LaunchFile();
Thread.Sleep(1000);
```

This leads to 

|  svg | png  |
|---|---|
[...]

PR #5035, opened 2025-08-10
Heatmap: Add independent rendering cell flags (#4950)

CoderPM2011:
The issue described in #4950 is similar to that in #5020.
SVG does not fully support pixel-based images, but the `Heatmap` rendering currently relies on `Drawing.BitmapFromArgbs()`.

To address this, a new optional flag `IndependentCellRendering` has been added.
When set to `true`, each cell is rendered independently using `Drawing.FillRectangle()`.

⚠️ **_Note that enabling this option can lead to performance degradation and significantly larger SVG file sizes when rendering many cells._**
However, when the number of cells is large, the difference is hardly noticeable in the SVG output, so it is generally unnecessary to use `IndependentCellRendering` in such cases. 🤔

For this reason, the feature is disabled by default.
🧐 **_Users should carefully evaluate their needs before enabling it._**

### demo result (SVG)
demo

### demo code
```cs
var plot = formsPlot1.Plot;

double[,] data = {
        { 1, 2, 3 },
        { 4, 5, 6 },
        { 7, 8, 9 },
    };

Heatmap heatmap = plot.Add.Heatmap(data);

heatmap.IndependentCellRendering = true;
formsPlot1.Plot.SaveSvg("demo.svg", 400, 300);
heatmap.IndependentCellRendering = false;
```

### DIFF
--- a/src/ScottPlot5/ScottPlot5/Plottables/Heatmap.cs
+++ b/src/ScottPlot5/ScottPlot5/Plottables/Heatmap.cs
@@ -133,6 +133,15 @@ public class Heatmap(double[,] intensities) : IPlottable, IHasColorAxis
     }
 
     /// <summary>
+    /// Renders each cell independently.
+    /// </summary>
+    /// <remarks>
+    /// <para>Set this property to <see langword="true"/> to improve clarity when cells cannot be distinctly rendered in SVG output.</para>
+    /// <para><b>⚠ Warning:</b> Enabling this option may reduce performance and significantly increase file size. Evaluate its impact before use.</para>
+    /// </remarks>
+    public bool IndependentCellRendering { get; set; } = false;
+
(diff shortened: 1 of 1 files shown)


bclehmann (commented):
I'm of two minds about this, on one hand this render method may well be a good feature, it likely is more efficient (if we avoid building the cached bitmap) for small heatmaps, especially if they're updated more often then they're panned/zoomed.

But also I'd really rather just find a way to set `image-rendering: pixelated` on the generated SVGs and let this be handled by the user's image viewer. It will be drastically better on file size and it should look pixel-identical between renderers.
Unfortunately I don't believe this is supported in Skia for whatever reason, though it would be a relatively easy add. So our only option would be to manually edit the generated XML, which is not an idea that sparks joy for me.

PR #5067, opened 2025-08-26
Barebones Avalonia 3D plot

bclehmann:
Barebones Avalonia 3D plot


swharden:
Let's keep the branch open at https://github.com/bclehmann/ScottPlot/tree/feature/avalonia-3d-plot until we figure out for sure the direction we're going on 3D plots

For now I'll close this PR since it's experimental, Avalonia only, and unlikely to get merged soon. Users who want to experiment with this can clone/build your branch.

PR #5075, opened 2025-08-28
Use real turbo colourmap

bclehmann:
Per our discussions in discord this seems like the right move.

### Before
image

### After
image

I know you're colourblind, but I promise there is a difference 🙃, it's most noticeable in the red splotches in the top right of frame.

### Remaining concerns

You'll notice from the diff that the regenerated license files are not deterministic. It looks like you're deduping licenses with the same authors and license. I don't know if doing it this way is legally kosher (since for some projects which are under a given license we're not actually acknowledging that, even though we have that license included under a different file).

I think if you want to dedupe licenses this way I'd create a file called `THIRD-PARTY-LICENSES.txt` which looks like this (this is what Skiasharp does):

```
Project A, Project B
================
<license text>
================
Project C
================
<license text>
================
Project D, Project E
.
.
.
```

In any case, I would definitely sort the file names or something so that the generated licenses are at least consistent from build to build. Otherwise we'll get a very annoying git history.

Deduping licenses is also probably not that necessary, I'd hope zip files handle duplicated files pretty well even if the package size difference was relevant.


bclehmann:
Also worth noting that while Apache 2 (what Turbo is licensed under) is a perfectly fine license, it's also a big sucker at ~200 lines. I personally don't care, since a) I'm on Rider right now, which collapses the license text in source code and b) when was the last time we ever needed to modify a colormap class.

But if this annoys you, we could parse a string like this and generate the text from elsewhere. Or just check in the generated files and remove the comment from the Turbo.cs file, whatever floats your boat.
```
// Copyright 2019 Google LLC.
// SPDX-License-Identifier: Apache-2.0
```

bclehmann:
> I don't think there's a requirement that the license filename has to be named according to the project, so using the md5 hash of the license seems like a way to make it deterministic again

No, I don't think it matters what the filename is, many programs comply by just having a CLI argument which dumps the licenses. But I do think people reading the notices should be able to tell which projects were written by whom, whether they're reading the source or binary distribution. Even if that isn't a legal requirement (i.e. if we could get away with saying there's code that Steve wrote \~\~somewhere\~\~) I think it's morally what we should do.

To me the single license file seems like the right move, I'll probably do it this evening.

swharden:
Wow, this is _really_ nice! Thanks for your work on this 😄

PR #5078, opened 2025-08-31
Allow SignalXY data source to return IList<Pixel>

bclehmann:
## Rationale

Jimmacle in the discord raised some concerns about the heap allocations in `SignalXYSourceGenericList` and found that reusing a buffer rather than reallocating Pixel arrays in every call was much more performant.

I am interested in implementing that (I got most of the way there on this branch just to prove I did it right: https://github.com/ScottPlot/ScottPlot/compare/main...bclehmann:ScottPlot:feature/reuse-buffer-signalxy-source-generic-list), but for now I think moving the interface to specify an `IList<Pixel>` and letting people handle the dirty work themselves is a fine compromise.

## Details

If simply changing the return type of `ISignalXYSource.GetPixelsToDraw` is acceptable than there's a somewhat simpler PR. But I didn't really want to break user data sources, especially as this change is only really useful (so far at least) to the kinds of people who write those.

So instead I renamed the interface to `ISignalXYSourceGeneric`, and the problematic function to `GetPixelsToDrawGeneric`, and created a new interface with the old name `ISignalXYSource`, which requires implementing `GetPixelsToDraw` returning a `Pixel[]`, and created a default implementation of `ISignalXYSourceGeneric.GetPixelsToDrawGeneric` which wraps it. I believe this makes this a fully source-compatible change, barring any sketchy reflection code that I don't think we have much interest in supporting.

Provided this is a workable approach I think this should be a big improvement with no downside.

## Remaining Concerns

I don't know if `IList` is the right type, it's just what I told the user in the discord I would do. I think `IReadOnlyList` is almost certainly a better choice, and I could be swayed towards `IEnumerable` (which has the benefit of being simpler to implement, though I don't know about its performance characteristics).

### DIFF
--- a/src/ScottPlot5/ScottPlot5/Interfaces/ISignalXYSource.cs
+++ b/src/ScottPlot5/ScottPlot5/Interfaces/ISignalXYSource.cs
@@ -1,6 +1,6 @@
 namespace ScottPlot;
 
-public interface ISignalXYSource
+public interface ISignalXYSourceGeneric
 {
     /// <summary>
     /// Number of values in the data source
@@ -51,7 +51,7 @@ public interface ISignalXYSource
     /// Return pixels to render to display this signal.
     /// May return one extra point on each side of the plot outside the data area.
     /// </summary>
-    Pixel[] GetPixelsToDraw(RenderPack rp, IAxes axes, ConnectStyle connectStyle);
+    IList<Pixel> GetPixelsToDrawGeneric(RenderPack rp, IAxes axes, ConnectStyle connectStyle);
 
     /// <summary>
     /// Return the point nearest a specific location given the X/Y pixel scaling information from a previous render.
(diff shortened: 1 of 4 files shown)


bclehmann:
Oof, CI tells me that this isn't supported on older NET versions (I think we need .NET Standard 2.1).

Which means this approach isn't going to work on .NET Framework, which is pretty rough. You'd think you could do this with abstract classes, but overriding properties (even abstract ones) is illegal, since it will end up hiding the original.

I have a possible implementation where we make `ISignalSourceGeneric` generic over the collection, but this causes more incompatibilities with existing user code (and is more confusing) since the incompatibilities aren't constrained to custom data sources. And it (probably unavoidably?) will break decorators such as the `DraggablePlottableDecorator` as we can't switch on a generic type.

bclehmann:
I'm still wondering if we should abandon backwards compatibility here, since I'd imagine there's very few users who would be affected, and it's an easy change which will probably only touch 1 class in their whole codebase.

swharden:
> CI tells met [default implementations on interfaces] isn't supported on older NET versions

> You'd think you could do this with abstract classes

I may be missing something basic here, but 3 classes inherit `ISignalXYSourceGeneric` so could I just copy that default implementation into each of them? I can even make it virtual. commit incoming...

bclehmann:
> I may be missing something basic here, but 3 classes inherit ISignalXYSourceGeneric so could I just copy that default implementation into each of them? I can even make it virtual. commit incoming...

Yes, but then this is a breaking change for users which have data sources that implement the current interface.

That might be ok, but if it is I think we might as well just change the signature to `IReadOnlyList` and our API is just simpler.

swharden:
> breaking change for users

I'm actually okay with this more than typical, because we're bumping version `5.0` to `5.1` in the next release. If there's ever a good time for a breaking change, this it it!

Also, consider whether there are any other breaking changes you may want to roll into the next release.

PR #5091, opened 2025-09-09
ISignalXYSource: Allow IReadOnlyList implementations

bclehmann:
Discussed in #5078, a breaking change seems fine here given that discussion.


bclehmann:
FWIW I decided against using IEnumerable since there are still a lot of performance problems there (and the purpose of this change is 99% around performance). This does make implementing a collection that allocates once and never again notably harder. I still don't think it's bad, you just can't `yield return` the implementation away.

They are being largely addressed by .NET 10 apparently.

PR #5121, opened 2025-10-16
Fix SignalXY plot for integers

### LINKED ISSUE #5108: SignalXY with long[] shifts points by +0.5 compared to double[]
**Issue:** (Describe the bug here)
When plotting with SignalXY, using long[] for X/Y produces a half-unit shift compared to double[].
Both plots should look identical, but the long[] version appears offset by about +0.5 along both axes.
This can be reproduced with a minimal WPF project as shown below.

**ScottPlot Version:** (What NuGet package are you using?)
ScottPlot.WPF 5.0.26, .NET 6

**Code Sample:** See http://scottplot.net/faq/repro/ for tips about creating reproducible code samples
```xml
<Window x:Class="ScottPlotSignalXYLongBug.MainWindow"
        xmlns="http://schemas.microsoft.com/winfx/2006/xaml/presentation"
        xmlns:x="http://schemas.microsoft.com/winfx/2006/xaml"
        xmlns:d="http://schemas.microsoft.com/expression/blend/2008"
        xmlns:mc="http://schemas.openxmlformats.org/markup-compatibility/2006"
        xmlns:ScottPlot="clr-namespace:ScottPlot.WPF;assembly=ScottPlot.WPF"
        xmlns:local="clr-namespace:ScottPlotSignalXYLongBug"
        mc:Ignorable="d"
        Title="MainWindow" Height="800" Width="400">
    <Grid>
        <Grid.RowDefinitions>
            <RowDefinition Height="*" />
            <RowDefinition Height="*" />
        </Grid.RowDefinitions>
        <ScottPlot:WpfPlot Grid.Row="0" x:Name="WpfPlot1" Width="300" Height="300"/>
        <ScottPlot:WpfPlot Grid.Row="1" x:Name="WpfPlot2" Width="300" Height="300"/>
    </Grid>
</Window>

```

```csharp
using ScottPlot;
using ScottPlot.Plottables;
using System.Windows;

namespace ScottPlotSignalXYLongBug
{
    {
        private SignalXY? _line1;
        private SignalXY? _line2;

        public MainWindow()
        {
            InitializeComponent();

            Loaded += (s, e) =>
            {
                double[] dataX1 = { 0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11 };
                double[] dataY1 = { 0, 0, 0, 1, 1, 1, 2, 2, 2, 3, 3, 3 };
                _line1 = WpfPlot1.Plot.Add.SignalXY(dataX1, dataY1);
                WpfPlot1.Refresh();

                long[] dataX2 = { 0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11 };
                long[] dataY2 = { 0, 0, 0, 1, 1, 1, 2, 2, 2, 3, 3, 3 };
                _line2 = WpfPlot2.Plot.Add.SignalXY(dataX2, dataY2);
                WpfPlot2.Refresh();

                _line1.ConnectStyle = ConnectStyle.Straight;
                _line1.MarkerSize = 3;
                _line1.LineWidth = 0;

                _line2.ConnectStyle = ConnectStyle.Straight;
                _line2.MarkerSize = 3;
                _line2.LineWidth = 0;
            };
        }
    }
}

```

Image

nbeck-SMT:
This PR fixes #5108.

The problem is caused by the fact that the pixel column position is cast to the type of the displayed data. In the case of integers, a data point will be displayed when the current pixel column position rounds down to the value of the data point and the next column position rounds up. This will only happen at half integer positions.

One solution would be to round up the column position for integer types. But this means checking for all integral types, which would break if new types are introduced.

```C#
private (int SearchedPosition, int LimitedIndex) SearchIndex(double x, IndexRange indexRange)
{
    var valueToSearch = (x - XOffset) / XScale;
    if (typeof(TX) == typeof(long) || typeof(TX) == typeof(ulong)
        || typeof(TX) == typeof(int) || typeof(TX) == typeof(uint)
        || ... )
        valueToSearch = Math.Ceiling(valueToSearch);
    NumericConversion.DoubleToGeneric(valueToSearch, out TX x2);
    int index = Array.BinarySearch(Xs, indexRange.Min, indexRange.Length, x2);
    ....
```

Another solution is to use a custom comparer for the binary search that compares the data points with the double pixel position. But this would rely on the fact that the implementation of binary search always uses the left slot of the comparer for array values.

```C#
private (int SearchedPosition, int LimitedIndex) SearchIndex(double x, IndexRange indexRange)
{
    TX dummyValue = NumericConversion.DoubleToGeneric<TX>(0);
    var comparer = Comparer<TX>.Create((v, _) => NumericConversion.GenericToDouble(ref v).CompareTo((x - XOffset) / XScale));
    int index = Array.BinarySearch(Xs, indexRange.Min, indexRange.Length, dummyValue, comparer);
    ...
```

I believe the cleanest solution is to implement a binary search that looks for the index to insert a double into the given array.

Let me know what you think.


swharden:
Hi @bclehmann, I'll let you make the final call on this one. 

On one hand there are multiple good options here, but on the other hand these are brutally complex to test (especially for edge cases) and it only affects half integer positions, so I want to acknowledge that there is a real and significant negative cost to "doing something" vs "doing nothing". We can merge what's here or defer this for review later.

@nbeck-SMT I really appreciate this contribution, but my bandwidth is pretty tight right now #5214 and I'm hesitant to make changes that are hard to test at this time.

nbeck-SMT:
@swharden I read about your medical issues and wish you all the best!

I agree that this is a tricky issue and all of the suggested solutions may introduce other problems. However, I think a bug that leads to incorrect data being displayed is worse than nothing being shown at all. So if you don't want to implement complicated fixes, I would suggest to limit this class to float types only for now.

swharden:
Finally getting around to working on this one, merging it today, and publishing it with the next release within the next few days.

I'm still undergoing medical treatment but things are going well and I hope to be out in about a month https://swharden.github.io

PR #5140, opened 2025-11-06
Make rectangle dragging behavior more natural

fissssssh:
This PR improves the user experience when dragging a rectangle by changing from an absolute "center-to-cursor" snapping model to a relative drag model.

Before:

On every mouse move, the rectangle’s center was repositioned exactly under the cursor.
This caused the rectangle to "jump" unexpectedly, especially if the user clicked near an edge or corner.
After:

When dragging starts (on mouse down), we record:
The original rectangle position (dragBaseRect)
The initial mouse position (dragStartPoint)
During drag, the rectangle moves by the same delta as the mouse, preserving the initial offset between the cursor and the rectangle.
Result: smooth, predictable, and natural-feeling dragging—just like in most design or drawing tools.
Changes:

Store drag anchor state on mouse down
Compute movement delta relative to initial press point
Update rectangle position using original state + delta (avoids cumulative drift)
Fixes unnatural jumping during rectangle manipulation.

### DIFF
--- a/src/ScottPlot5/ScottPlot5/PlottableAdder.cs
+++ b/src/ScottPlot5/ScottPlot5/PlottableAdder.cs
@@ -751,20 +751,20 @@ public class PlottableAdder(Plot plot)
         return plottable;
     }
 
-    public InteractiveRectangle InteractiveRectangle(CoordinateRect rect)
-    {
-        InteractiveRectangle plottable = new() { Rect = rect };
-        plottable.LineStyle.Color = GetNextColor();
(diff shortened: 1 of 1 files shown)


fissssssh:
## New: Interactive Polygon Plottable

In addition to the improved rectangle dragging behavior, this PR now also introduces a fully interactive PolygonPlottable:

Users can click and drag individual vertices to reshape the polygon.
The polygon updates in real time during dragging, providing immediate visual feedback.

fissssssh:
@swharden Can you approve this PR? I really need the starting point coordinates when interacting!

PR #5154, opened 2025-11-15
Target .NET 10

swharden:
* Update all projects (and NuGet packages) to target .NET 10
* Resolve all warnings
* Update CICD to support .NET 10 builds

PR #5163, opened 2025-11-23
Add global.json to enforce .NET 9 SDK

### LINKED ISSUE #5161: ScottPlot.WinUI build fails in VS 2026 with .NET SDK 10.0.100
### Problem
Cannot build `ScottPlot.WinUI` in Visual Studio 2026 with .NET SDK 10.0.100.

**Build Error:**
```
Error NETSDK1140: 1.0 is not a valid TargetPlatformVersion for maccatalyst
```

**Current TargetFrameworks:**
```xml

    net8.0;
    net8.0-windows10.0.19041.0;
    net8.0-ios;
    net8.0-maccatalyst;  
    net8.0-android

```

## Root Cause

* `net8.0-maccatalyst` is incompatible with .NET SDK 10.0.100

## Suggested Fix

Add `global.json` to enforce .NET 8 SDK
```json
{
  "sdk": {
    "version": "8.0.404",
    "rollForward": "latestFeature"
  }
}
```

diluculo:
Fixes #5161

Added `global.json` to specify minimum .NET SDK version 9.0.100.

### DIFF
--- a/src/ScottPlot5/global.json
+++ b/src/ScottPlot5/global.json
@@ -1,0 +1,6 @@
+{
+    "sdk": {
+        "version": "9.0.100",
+        "rollForward": "feature"
+    }
+}


diluculo:
Note: PR #5154 is upgrading all target frameworks to `net10.0-*`, which will fundamentally solve this issue.

This PR provides an immediate fix by enforcing .NET 9 SDK until #5154 is merged. 

However, I believe `global.json` is still valuable for ensuring consistent SDK versions across all development environments. Once #5154 is merged, the SDK version in `global.json` should be updated to `10.0.100`.

bclehmann:
What are the semantics of this? In particular, would it prevent building things on previous versions? We've been following the EOL dates of .NET, and if we continue to do that we would have to support NET 8 for another year (and in practice we rarely actively drop support unless it causes issues) and we have to support .NET Framework versions basically until Microsoft says otherwise (the length of .NET Framework support requirements is probably the only thing that would make us reevaluate this choice).

swharden:
Hi @diluculo, thanks for this PR! However, I decided against this to keep things simple and also allow me to target slightly different .NET versions in different packages which is often helpful when piecewise upgrading this large solution from one .NET version to another

PR #5167, opened 2025-11-29
First-pass Avalonia copy-to-clipboard

bclehmann:
Now that https://github.com/AvaloniaUI/Avalonia/pull/19930 has been merged I can tick a feature off my list (5 years later 🙃). We'll have to think about how we ought to disable this behaviour on earlier Avalonia versions, or just require a version that has this.

This implementation is hacky and I haven't cleaned it up yet, I mostly wanted to see what the capabilities were and what we would have to do to support it. It's mostly the same as other implementations (call `GetImageBytes`, turn it into a platform-specific image, and then pass it to a platform-specific function).

The problem is that Avalonia requires we pass them a bitmap, which even with the device-independent-bitmap (DIB) that we have implemented is only a pseudo-standardized format. That wouldn't be a problem if Avalonia weren't cross-platform, where the preferred byte order of bitmaps can differ. Plus Avalonia seems to be ignoring the pixelFormat parameter, at least on Mac where I'm currently testing, so switching byte orders must be done by hand.

On an ARM Mac it seems that all we need to do is strip out the DIB header and it will have the correct colours. But I would want to test on Windows and Linux as well to make sure it does the right thing everywhere. I'm hoping that x86 vs ARM doesn't change the byte order for any OS, because I can't test Windows on ARM, and testing Mac on x86 or Linux on ARM is inconvenient but doable for me.

Once I've done that I'd like to think about how we ought to change this code, I would probably like to make the `BitmapHeader` struct `internal` again and just expose a method to get only the pixel array for a bitmap. And then come to a decision about how we release this and if we require a newer Avalonia version.


swharden:
This may be relevant
https://github.com/AvaloniaUI/Avalonia/issues/20273#issuecomment-3650332703

swharden:
updating branch and merging as per https://github.com/ScottPlot/ScottPlot/pull/5233#issuecomment-4376226436

PR #5178, opened 2025-12-24
Fix missing "else" in MouseWheelZoom (MouseAxisManipulation.cs)

jurihock:
I suppose an "else" is missing in the MouseWheelZoom function:

https://github.com/ScottPlot/ScottPlot/blob/main/src/ScottPlot5/ScottPlot5/Interactivity/MouseAxisManipulation.cs

```cs
if (axisUnderMouse is not null)
{
	if (ChangeOpposingAxesTogether && axisUnderMouse.IsHorizontal())
	{
		plot.Axes.XAxes.ForEach(xAxis => xAxis.Range.ZoomFrac(fracX, xAxis.GetCoordinate(pixel.X, dataRect)));
	}
	else if (ChangeOpposingAxesTogether && axisUnderMouse.IsVertical()) // <-- else if, not just if
	{
		plot.Axes.YAxes.ForEach(yAxis => yAxis.Range.ZoomFrac(fracY, yAxis.GetCoordinate(pixel.Y, dataRect)));
	}
	else
	{
		double frac = axisUnderMouse.IsHorizontal() ? fracX : fracY;
		float scaledCoord = (axisUnderMouse.IsHorizontal() ? pixel.X : pixel.Y);
		axisUnderMouse.Range.ZoomFrac(frac, axisUnderMouse.GetCoordinate(scaledCoord, dataRect));
	}
}
else
{
	...
}
```


bclehmann (approved):
I thought this was a dead change, but because of the else block on line 32 this does actually fix a (very minor) inconsistency. Good catch

PR #5183, opened 2026-01-01
Make ContextMenuItems_get public

btarb24:
Allows users to Insert items anywhere in list, selectively remove default items, as well as rearrange them as desired.

_set is still hidden in the interface.


bclehmann (approved):
lgtm

PR #5185, opened 2026-01-01
Tag property on IPlottable

btarb24:
Added a new property to IPlottable:
`public object Tag {get; set;}`

This Tag property is consistently available across WPF UI objects to allow consumers to add additional data objects to the UI.
[wpf documentation](https://learn.microsoft.com/en-us/dotnet/api/system.windows.frameworkelement.tag?view=windowsdesktop-10.0)

My specific use case was i allow the user to create trend lines on a candlestick chart and i save them to a database.  The Tag property lets me add my model to the UI object for direct association. The user may have many different charts for the same symbol (1min, 2min, 5min, 15min, etc) and the trend lines need to be synchronized across all charts of that symbol.  The common Tag property saves me from having to create more tedious mapping tables for each chart.

### DIFF
--- a/src/ScottPlot5/ScottPlot5 Demos/Avalonia Demo/Demos/CustomPlotTypeWindow.cs
+++ b/src/ScottPlot5/ScottPlot5 Demos/Avalonia Demo/Demos/CustomPlotTypeWindow.cs
@@ -1,13 +1,7 @@
 using Avalonia.Controls;
-using Avalonia.Input;
 using ScottPlot;
-using ScottPlot.Plottables;
-using SkiaSharp;
-using System;
 using System.Collections.Generic;
 using System.Linq;
-using System.Text;
-using System.Threading.Tasks;
 
 namespace Avalonia_Demo.Demos;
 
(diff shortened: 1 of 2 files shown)


bclehmann:
I think most users would get better results by handling this themselves, it's pretty easy to create a `TaggedPlottable` class which implements `IPlottable` and exposes a tag.

And that way you could implement things like pub-sub or whatever your application model wants more easily. You could also get real typing, a `TaggedPlottable<T>` could expose a tag of type `T` rather than `object?`. I think that would be a big improvement.

My main issue though, is if ScottPlot isn't supposed to do anything with the tag, then I don't think ScottPlot should be responsible for handling it. As an example of how this could get dicey, suppose we had a `Plot.Clone` API. Should tags be preserved when a plot is cloned? And then what does that do to our users if their tags exist on 2 different plots? I think the correct solution there is to make our users handle it in whatever way makes sense for their app.

More generally, I think IPlottable is the wrong abstraction for users to develop appmodels around, and I think the WPF way of tags is also dubious in general.

bclehmann:
It also looks like you have some extra changes in here, I'm seeing a new InteractiveTrendline plottable?

PR #5194, opened 2026-01-16
Marker shapes

btarb24:
Added a couple sets of markers that are particularly useful on candlestick charts.  The primary goals of the markers were:
* encode in whether it is a buy or sell marker
* don't obscure the vertical center of the candle (which hides a wick)
* vertically symmetrical (to avoid perspective weighting that would skew the ability to read delta at a glance - up/down triangles are common offenders of this).

Circles:
image image

Triangles:
image image 

Circle/triangle uses fillStyle & outlineStyle
Bar uses LineStyle
This allows for a thinner border on the circle to avoid overpowering the fill color while keeping from becoming too thin to be easily visible.


src/ScottPlot5/ScottPlot5/PlottableAdder.cs
bclehmann: I think this is a separate PR, you can do an interactive rebase so these changes don't have to be merged in a specific order.

Especially since this PR looks pretty uncontroversial and could be merged with little debate, while a new plottable would inevitably require a closer review.

bclehmann (commented):
LGTM, just a note on the trendline plottable that's included in this PR as well.

I personally don't have much experience with trading, but your rationalization makes a lot of sense to me. I can imagine them being quite useful if I were to have some realtime charting projects in the future.

PR #5208, opened 2026-02-23
InteractiveVerticalLine: Add properties to customize label style

Guillaume-Deville:
Hello,

For my project, I wanted to add a vertical line that displays the X value, similar to an oscilloscope.

InteractiveVerticalLine is easy to use and handles mouse interaction, but it doesn’t provide a LabelStyle property. Because of that, you can’t easily display the X value as text like you can with VerticalLine.

So I extended InteractiveVerticalLine to add support for LabelStyle.

Is this the right way to do it, can you provide feedback ?

Harmony_o6VR2Q9GR9


Guillaume-Deville:
Hello Scott,

Thank you for merging my PR.
I hope your bone marrow transplant goes well. Your dedication is very inspiring.

PR #5209, opened 2026-02-23
Controls: improve pressed key tracking when modifier key state changes

btarb24:
The KeyboardState class only receives key information if the application has focus.  This causes some issues with modifier keys.
* if you focus your application > use ALT-Tab to switch out > click back on the chart to attempt to drag pan it = ALT is still applied even though you had already let it go, which causes a rectangle zoom instead of the desired drag pan.
* you are in another app and you want to use a modifier key with your mouse press back onto the chart, the modifier key is ignored since the app didn't have focus when you received it.

Refactoring the keyboard system entirely may be the best option, but i didn't want to dive that deeply.  Instead, i just put in a low-risk workaround specific to modifier keys.

if .net standard is ever dropped then we could remove the pinvoke and simply go with System.Windows.Input's Keyboard.Modifiers.HasFlag


src/ScottPlot5/ScottPlot5/Interactivity/KeyboardState.cs
```diff
@@ -53,4 +66,10 @@ public override string ToString()
             ? "KeyState with 0 pressed key"
             : $"KeyState with {PressedKeyNames.Count} pressed keys: " + string.Join(", ", PressedKeyNames.Select(x => x.ToString()));
     }
+
+    [DllImport("user32.dll")]
```
bclehmann: This is in cross-platform code, so we can't pull in a Windows library.

One option would be to requery key state when the control is focused and then have the control update the key state itself. That would presumably be fine everywhere. Another option would be to have this class take an interface which can call whichever platform-specific code is necessary to check key state.

bclehmann (changes requested):

src/ScottPlot5/ScottPlot5/Interactivity/KeyboardState.cs
btarb24: That's fair.  I reverted the changes to seek a new solution.

I saw WinUIPlotExtensions has a UpdateKeyStateFromModifiers method that looks at the framework's keyboard state system and updates the KeyState to match which modifiers the framework says are pressed.  I used the same pattern for WpfUIPlotExtensions (and used Claude for the other frameworks I'm less familiar with).

The result should be that this now works with all of the supported frameworks and is crossPlatform on the frameworks that support it.  Tested only with WPF on Windows.  Both of my bullet points in initial comment pass.

btarb24: FWIW, this has been working great.  The 'Author Action Required' tag can seemingly be removed.

swharden: > FWIW, this has been working great

I'll review it from here and merge it shortly

swharden:
There is a problem with this PR preventing merging. This code works in the main branch but not this branch. I'm not totally sure why. I suspect calling `processor.KeyState.Reset();` is problematic when users clear or re-map the default keys associated with mouse responses...

```cs
using ScottPlot;

namespace Sandbox.WinForms;

public partial class Form1 : Form
{
    public Form1()
    {
        InitializeComponent();

        formsPlot1.Plot.Add.Signal(Generate.Sin());
        formsPlot1.Plot.Add.Signal(Generate.Cos());

        var panResponse = formsPlot1.UserInputProcessor.UserActionResponses
            .OfType<ScottPlot.Interactivity.UserActionResponses.MouseDragPan>()
            .Single();

        panResponse.KeysThatLockX.Clear();
        panResponse.KeysThatLockY.Clear();

        panResponse.KeysThatLockX.Add(new ScottPlot.Interactivity.Key("x"));
        panResponse.KeysThatLockY.Add(new ScottPlot.Interactivity.Key("y"));
    }
}
```

Related: https://scottplot.net/demo/5/#custom-mouse-actions

swharden:
I think the fix is to add an optional argument or overload to `processor.KeyState.Reset();` that _doesn't_ reset the _changed_ keys from the platform-specific modifier keys. I'll give this a quick try and report back.

swharden:
> There is a problem with this PR preventing merging

I resolved this issue by not calling `Reset()` but instead only removing the modifier keys from the pressed list. I'll merge now and deploy in the next package soon. If this isn't the ideal fix, we can open up a new PR and pick it up there.

Thanks again for your help with this feature!

PR #5215, opened 2026-03-04
Invisible lines should not be interactable

btarb24:
when an Interactable is marked as IsVisible=false we would still have mouse cursor changes and you could drag the line location even though it wasn't visible.

Updated each of the interactive controls so that GetHandle bails if it's not visible.


btarb24:
Removed InteractiveTrendLine from this PR and added the invisible line fix directly to the InteractiveTrendLine PR.

bclehmann (approved):

swharden (approved):

PR #5234, opened 2026-04-28
Fix AutoScaleAll crash/empty rendering when data range span is zero

AscendLiu:
### Problem

When all Y values are identical or all X values are identical in a dataset, the `AutoScaleAll` method in `FractionalAutoScaler` fails to calculate a valid axis range.

Specifically:
- `yAxis.Range.Span = Max - Min = 0` (or similarly for X axis)
- The padding calculation results in `bottom == top`, both equal to the constant data value
- The axis ends up with zero range, causing the chart to render nothing or become unviewable

This affects scenarios such as:
- Plotting horizontal or vertical lines
- Datasets where all points share the same X coordinate
- Datasets where all Y values are identical (e.g., constant sensor readings)

### Solution

Added fallback logic in `FractionalAutoScaler.AutoScaleAll`:

- **X-axis:** When `left == right` (zero span), manually set range to `left - 1` and `right + 1`
- **Y-axis:** When `bottom == top` (zero span), manually set range to `bottom - 1` and `top + 1`

This ensures the axis always has a non-zero range, allowing the chart to render correctly even with constant data.

### Testing

- [x] Tested with dataset where all Y values are equal
- [x] Tested with dataset where all X values are equal
- [x] Verified normal data (non-zero span) is unaffected by this change

### Screenshots (if applicable)

| Before | After |
|--------|-------|
| (empty chart rendering) | (chart shows data properly) |

### Related Issues

Fixes #[Issue Number] — *create one first if it doesn't exist*

### Checklist

- [x] My code follows the code style of this project
- [x] I have tested my changes locally
- [x] I have considered edge cases (zero-span data)
- [x] This PR focuses on a single bug fix (no unrelated changes)

---

Let me know if you want me to adjust the tone or add more technical details specific to the `FractionalAutoScaler` class.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ _Not yet published..._
 * Scatter: Reduce memory allocations when rendering scatter plots (#5129) @Ben-D-Anderson
 * Interactivity: Add trend lines that extend infinitely in both directions (#5181) @btarb24
 * Controls: Improve modifier key handling when plots regain focus (#5209) @btarb24
+* Autoscaling: Fix zero-span axis ranges when data has identical X and Y values (#5234) @AscendLiu
 
 ## ScottPlot 5.1.58
 _Published on [NuGet](https://www.nuget.org/profiles/ScottPlot) on 2026-03-29_
--- a/src/ScottPlot5/ScottPlot5/AutoScalers/FractionalAutoScaler.cs
+++ b/src/ScottPlot5/ScottPlot5/AutoScalers/FractionalAutoScaler.cs
@@ -70,6 +70,12 @@ public class FractionalAutoScaler : IAutoScaler
             double right = xAxis.Range.Max + (xAxis.Range.Span * RightFraction);
             if (NumericConversion.IsReal(left) && NumericConversion.IsReal(right))
             {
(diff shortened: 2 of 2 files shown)


swharden:
I think this may have broken something. The histogram cookbook example now renders poorly, with axes +/- = which lines up with this PR...

Before:
image

After:
image

I'll track this issue in a separate PR and tag this one.

swharden:
The issue with histograms is:
* the fractional autoscaler is setup BEFORE data is added to the histogram when there is no data
* the new code sets axis limits to value values, signaling they have been set already
* on first render, the axis limits are not changed, so they always display -1 to 1

swharden:
Fixed in #5251

AscendLiu:
I tested versions 5.1.59 and Avalonia 12.0.4, and it seems that this issue has been resolved. I reviewed the changes in 5.1.59 but didn't quite understand why. At the time, this pull request (PR) could have solved my issue, but I might have made a mistake somewhere. Below is a demo of how I reproduced the issue at that time
[TestScottPlot.zip](https://github.com/user-attachments/files/29232700/TestScottPlot.zip)

PixPin_2026-06-23_12-02-35

PR #5241, opened 2026-05-07
Avalonia: Customizable wheel event handling.

### LINKED ISSUE #5240: Avalonia. Can't scroll a page with a plot inside if the mouse is over the plot.
A fix that prevents event bubbling in AvaPlot when nested in ScrollViewer was previously introduced in #5223.
But now in case when you've got a Plot with UserInputProcessor.Disable() inside ScrollViewer, you can't scroll the page while the mouse is over the Plot. Therefore, I believe that this behavior should be configurable at the plot control level.

oktrue:
fix #5240

### DIFF
--- a/src/ScottPlot5/ScottPlot5 Controls/ScottPlot.Avalonia/AvaPlot.cs
+++ b/src/ScottPlot5/ScottPlot5 Controls/ScottPlot.Avalonia/AvaPlot.cs
@@ -21,6 +21,8 @@ public class AvaPlot : Controls.Control, IPlotControl
 
     public GRContext? GRContext => null;
 
+    public bool HandleMouseWheelEvent { get; set; }
+
     public float DisplayScale { get; set; }
 
     public AvaPlot()
@@ -131,8 +133,8 @@ public class AvaPlot : Controls.Control, IPlotControl
         {
             UserInputProcessor.ProcessMouseWheel(pixel, delta);
         }
-		
-        e.Handled = true; // Prevent the event from bubbling up to parent controls (e.g., ScrollViewer)
+
+        e.Handled = HandleMouseWheelEvent; // Prevent the event from bubbling up to parent controls (e.g., ScrollViewer)
     }
 
     protected override void OnKeyDown(KeyEventArgs e)
--- a/src/ScottPlot5/ScottPlot5 Controls/ScottPlot.Avalonia/AvaPlot.cs
+++ b/src/ScottPlot5/ScottPlot5 Controls/ScottPlot.Avalonia/AvaPlot.cs
@@ -21,7 +21,7 @@ public class AvaPlot : Controls.Control, IPlotControl
 
     public GRContext? GRContext => null;
 
-    public bool HandleMouseWheelEvent { get; set; }
+    public bool HandleMouseWheelEvent { get; set; } = true;
 
     public float DisplayScale { get; set; }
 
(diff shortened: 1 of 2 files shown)


src/ScottPlot5/ScottPlot5 Controls/ScottPlot.Avalonia/AvaPlot.cs
```diff
@@ -131,8 +133,8 @@ protected override void OnPointerWheelChanged(PointerWheelEventArgs e)
         {
             UserInputProcessor.ProcessMouseWheel(pixel, delta);
         }
-		
-        e.Handled = true; // Prevent the event from bubbling up to parent controls (e.g., ScrollViewer)
+
+        e.Handled = HandleMouseWheelEvent; // Prevent the event from bubbling up to parent controls (e.g., ScrollViewer)
```
bclehmann: Would it not make more sense to read UserInputProcessor.IsEnabled?

I'm assuming that you would never want to scroll when user input is enabled, and you would never want to block scroll when it's disabled, which means we should be using the same boolean everywhere

oktrue: There's another scenario where the user wants to remove only the mouse wheel processing from UserInputProcessor and leave other actions (benchmark, dragging, etc.) enabled. In this case, it won't work.

radekbocon:
I have a workaround for this, but scrolling feels off, so it would be nice to have it merged. ;) 

In case someone needs worarkound: 

```
    ChartElement.AddHandler(PointerWheelChangedEvent, PointerWheelHandler, RoutingStrategies.Tunnel);

    private void PointerWheelHandler(object? sender, PointerWheelEventArgs e)
    {
        var scrollViewer = ChartElement.FindAncestorOfType<ScrollViewer>();
        scrollViewer?.Offset += e.Delta * -40;
        e.Handled = true;
    }
```

src/ScottPlot5/ScottPlot5 Controls/ScottPlot.Avalonia/AvaPlot.cs
swharden: @bclehmann I'm merging this now and will release a package shortly. If you object to this change we can open a new PR to improve it 👍

swharden:
This PR reversed a change implemented earlier in https://github.com/ScottPlot/ScottPlot/pull/5223/changes which set the default value of `e.Handled` to `true`.

I see this PR now sources that value from `HandleMouseWheelEvent`, but to maintain the current behavior I'll change this PR to set the default value of `HandleMouseWheelEvent` to `true` before merging.

mike-kilo:
I was struggling with this issue just a couple of days ago, after updating to 5.1.58 caused our custom mouse wheel actions to stop working.

The root cause analysis I could do, is the following - you want to swallow the mouse wheel events to prevent them from bubbling up while being handled by `UserInputProcessor`. Setting `e.Handled` to true unconditionally does it.

However - there are some cases that you want to add your own event handler (my case). You also want the default actions not to take place. 
If you attach an event handler to `PointerWheelChanged`, this event won't fire up because of the earlier `e.Handled` being set to true.

The solution is quite simple, in my opinion.
In the `protected override void OnPointerWheelChanged(PointerWheelEventArgs e)` method, the `e.Handled` should be set to `UserInputProcessor.IsEnabled`. 
This way, if you use `UserInputProcessor`, the event is swallowed and handled internally by ScottPlot. Otherwise it bubbles up and requires the user to attach custom event handler, where they decide whether to handle it completely (`e.Handled = true`) or allow for further bubbling up (`e.Handled = false`).

PR #5266, opened 2026-08-26
Fix some minor issues with using DateTime on the x-axis

pjt33:
I expected

```cs
plot.Axes.DateTimeTicksBottom();

var ticks = new DateTimeManual();
foreach (var date in xs) ticks.AddMajor(date, date.ToString("M"));
plot.Axes.Bottom.TickGenerator = ticks;
```

to work, but it fails at runtime with a confusing message. This PR, which could be split into two (one for each commit) improves the error message and fixes the error.

1. The error message complained that `Date axis must have a ITickGenerator generator`, which didn't make sense because `DateTimeManual` *is* an `ITickGenerator`. But the actual test that was failing is against `IDateTimeTickGenerator`, so that should be the type named in the error message.
2. Logically it seems that `DateTimeManual` should be an `IDateTimeTickGenerator`, and I can't see any reason why it shouldn't be.


pjt33:
NB I can work around this for the time being, so there's no urgency to pull it, but since it's not much more effort to provide the fix than to open an issue I hope I'm being helpful by making it a PR.

